PR #47, opened 2024-12-03
chore: cleanup redis

ChaoticTempest:
Added `StoreResult` and `StoreError` for cleaner handling of redis underlying storage

### DIFF
--- a/chain-signatures/node/src/protocol/signature.rs
+++ b/chain-signatures/node/src/protocol/signature.rs
@@ -486,7 +486,7 @@ impl SignatureManager {
                 ) {
                     Ok(generator) => generator,
                     Err((presignature, err @ InitializationError::BadParameters(_))) => {
-                        presignature_manager.insert_mine(presignature).await;
+                        presignature_manager.insert(presignature, true).await;
                         tracing::warn!(sign_request = ?sign_request_identifier, presignature_id, ?err, "failed to start signature generation");
                         return Err(GenerationError::CaitSithInitializationError(err));
                     }
@@ -703,7 +703,7 @@ impl SignatureManager {
         // add back the failed presignatures that were incompatible to be made into
         // signatures due to failures or lack of participants.
         for presignature in failed_presigs {
-            presignature_manager.insert_mine(presignature).await;
+            presignature_manager.insert(presignature, true).await;
         }
     }
 
--- a/chain-signatures/node/src/protocol/triple.rs
+++ b/chain-signatures/node/src/protocol/triple.rs
@@ -145,10 +145,12 @@ impl TripleManager {
     }
 
     pub async fn insert(&mut self, triple: Triple, mine: bool) {
-        tracing::debug!(id = triple.id, mine, "inserting triple");
-        self.gc.remove(&triple.id);
+        let id = triple.id;
+        tracing::debug!(id, mine, "inserting triple");
         if let Err(e) = self.triple_storage.insert(triple, mine).await {
             tracing::warn!(?e, mine, "failed to insert triple");
+        } else {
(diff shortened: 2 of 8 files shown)


chain-signatures/node/src/protocol/signature.rs
```diff
@@ -703,7 +703,7 @@ impl SignatureManager {
         // add back the failed presignatures that were incompatible to be made into
         // signatures due to failures or lack of participants.
         for presignature in failed_presigs {
-            presignature_manager.insert_mine(presignature).await;
+            presignature_manager.insert(presignature, true).await;
```
volovyks, line 706: This interface was separated on porpouse. `insert_mine(presignature)` is more readable then `insert(presignature, true)`.

chain-signatures/node/src/protocol/triple.rs
```diff
@@ -144,19 +144,11 @@ impl TripleManager {
         }
     }
 
-    pub async fn insert(&mut self, triple: Triple) {
-        tracing::debug!(id = triple.id, "inserting triple");
+    pub async fn insert(&mut self, triple: Triple, mine: bool) {
+        tracing::debug!(id = triple.id, mine, "inserting triple");
         self.gc.remove(&triple.id);
```
volovyks, line 149: Let's follow the same pattern when removing things from gc. It is not consistent with how it's done in presignatures.

volovyks (commented):
LGTM, tests were not executed for some reason

chain-signatures/node/src/protocol/triple.rs
```diff
@@ -176,6 +168,21 @@ impl TripleManager {
...
+                tracing::warn!(id, ?store_err, "triple is generating");
+                GenerationError::TripleIsGenerating(*id)
+            } else if self.gc.contains_key(id) {
+                tracing::warn!(id, ?store_err, "triple is garbage collected");
+                GenerationError::TripleIsGarbageCollected(*id)
+            } else {
+                tracing::warn!(id, ?store_err, "triple is missing");
+                GenerationError::TripleIsMissing(*id)
+            }
+        })
```
ppca, line 183: need to insert this triple into gc here?

chain-signatures/node/src/protocol/triple.rs
```diff
@@ -185,56 +192,18 @@ impl TripleManager {
...
-                tracing::warn!(id1, ?e, "failed to take triple");
-                if let Err(e) = triples.insert(triple_0).await {
-                    tracing::warn!(id0, ?e, "failed to insert triple back");
-                }
-                return Err(GenerationError::TripleIsMissing(id1));
+                return Err(err);
             }
         };
-
         self.gc.insert(id0, Instant::now());
```
ppca, line 205: Let's not do gc.insert here, let's do it in fn take() so that it's all done in take(), and when the 2nd triple take fails in take_two(), the insert() will revert the gc.insert by a gc.remove

chain-signatures/node/src/protocol/signature.rs
ppca: I personally like insert(x, bool), since if so we don't need to remember to do insert_mine() rather than insert() when the presig or triple is mine.

chain-signatures/node/src/protocol/triple.rs
ppca: this is still unchanged tho

ChaoticTempest: I wonder if you still had an older view of the code, but I changed it already

ppca (dismissed):

ppca (dismissed):

ppca (approved):

PR #50, opened 2024-12-03
feat: separate contract update and participant state from main loop

ppca:
Changes included:
1. get contract state/config and participant state in separate handles than the main loop.
contract state, config and mesh_state are being passed around as Arc<RwLock<...>> so different tasks can read/write from it. Contract state and contract config are updated in ContractUpdater.run(), while mesh_state is updated in Mesh.run(). 

2. in fn ping(), fetch_participant_state are run in parallel 

3. within each iteration of the loop in MpcSignProtocol.run(), the contract state, config and mesh_state are read only at the beginning of the iteration, and will be consistent within each iteration. cfg and mesh are removed from ctx.

### DIFF
--- a/chain-signatures/node/Cargo.toml
+++ b/chain-signatures/node/Cargo.toml
@@ -21,6 +21,7 @@ cait-sith = { git = "https://github.com/LIT-Protocol/cait-sith.git", features = 
 ], rev = "8ad2316" }
 clap = { version = "4.2", features = ["derive", "env"] }
 chrono = "0.4.24"
+futures = "0.3"
 google-datastore1 = "=5.0.4"
 google-secretmanager1 = "5"
 hex = "0.4.3"
--- a/chain-signatures/node/Cargo.toml
+++ b/chain-signatures/node/Cargo.toml
@@ -21,7 +21,6 @@ cait-sith = { git = "https://github.com/LIT-Protocol/cait-sith.git", features = 
 ], rev = "8ad2316" }
 clap = { version = "4.2", features = ["derive", "env"] }
 chrono = "0.4.24"
-futures = "0.3"
 google-datastore1 = "=5.0.4"
 google-secretmanager1 = "5"
 hex = "0.4.3"
--- a/chain-signatures/node/src/contract_updater.rs
+++ b/chain-signatures/node/src/contract_updater.rs
@@ -1,0 +1,50 @@
+use crate::config::Config;
+use crate::protocol::ProtocolState;
+use crate::rpc_client;
+use near_account_id::AccountId;
+use std::sync::Arc;
+use std::time::{Duration, Instant};
+use tokio::sync::RwLock;
+
+pub struct ContractUpdater {
+    rpc_client: near_fetch::Client,
+    mpc_contract_id: AccountId,
+}
+
+impl ContractUpdater {
+    pub fn init(rpc_client: near_fetch::Client, mpc_contract_id: AccountId) -> Self {
+        Self {
+            rpc_client,
+            mpc_contract_id: mpc_contract_id.clone(),
+        }
+    }
+
+    pub async fn run(
+        &self,
+        contract_state: Arc<RwLock<Option<ProtocolState>>>,
+        config: Arc<RwLock<Config>>,
+    ) -> anyhow::Result<()> {
+        let mut last_state_update = Instant::now();
+        let mut last_config_update = Instant::now();
+        loop {
+            if last_state_update.elapsed() > Duration::from_millis(1000) {
+                let mut contract_state = contract_state.write().await;
+                *contract_state =
+                    rpc_client::fetch_mpc_contract_state(&self.rpc_client, &self.mpc_contract_id)
+                        .await
+                        .ok();
+                last_state_update = Instant::now();
+            }
+            if last_config_update.elapsed() > Duration::from_millis(300) {
+                let mut config = config.write().await;
+                if let Err(error) = config
+                    .fetch_inplace(&self.rpc_client, &self.mpc_contract_id)
+                    .await
+                {
+                    tracing::error!("could not fetch contract's config: {error:?}");
+                }
+                last_config_update = Instant::now();
+            }
+        }
+    }
+}
--- a/chain-signatures/node/src/contract_updater.rs
+++ b/chain-signatures/node/src/contract_updater.rs
@@ -3,7 +3,7 @@ use crate::protocol::ProtocolState;
 use crate::rpc_client;
 use near_account_id::AccountId;
 use std::sync::Arc;
-use std::time::{Duration, Instant};
+use std::time::Duration;
 use tokio::sync::RwLock;
 
 pub struct ContractUpdater {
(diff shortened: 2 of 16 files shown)


chain-signatures/node/src/mesh/mod.rs
```diff
@@ -29,38 +31,59 @@ impl Options {
     }
 }
 
+#[derive(Clone)]
+pub struct MeshState {
+    /// Participants that are active at the beginning of each protocol loop.
```
volovyks, line 36: It should not be tied to the beginning of the loop anymore

chain-signatures/node/src/mesh/mod.rs
```diff
@@ -29,38 +31,59 @@ impl Options {
...
+}
+
 pub struct Mesh {
     /// Pool of connections to participants. Used to check who is alive in the network.
-    pub connections: connection::Pool,
+    connections: connection::Pool,
 
     /// Participants that are active at the beginning of each protocol loop.
-    pub active_participants: Participants,
+    active_participants: Participants,
```
volovyks, line 53: Don't we have this and other fields in MeshState?

chain-signatures/node/src/mesh/mod.rs
```diff
@@ -110,8 +121,31 @@ impl Mesh {
...
     }
+
+    pub async fn run(
+        mut self,
+        contract_state: Arc<RwLock<Option<ProtocolState>>>,
+        mesh_state: Arc<RwLock<MeshState>>,
+    ) -> anyhow::Result<()> {
+        let mut last_pinged = Instant::now();
+        loop {
+            if last_pinged.elapsed() > Duration::from_millis(300) {
```
volovyks, line 136: Is 300ms what we have now? That can create an additional load on RPC. My expectation was ~1s

chain-signatures/node/src/mesh/mod.rs
```diff
@@ -110,8 +121,31 @@ impl Mesh {
...
+                    let mut mesh_state = mesh_state.write().await;
+                    *mesh_state = MeshState {
+                        active_participants: self.active_participants().clone(),
+                        active_potential_participants: self.active_potential_participants().clone(),
+                        potential_participants: self.potential_participants().await.clone(),
+                        stable_participants: self.stable_participants().await.clone(),
+                    };
+                    last_pinged = Instant::now();
+                }
+            }
```
volovyks, line 148: Should we wait for ~100ms here? Otherwise, these loops will burn CPU resources.

chain-signatures/node/src/mesh/mod.rs
```diff
@@ -110,8 +121,31 @@ impl Mesh {
...
+        contract_state: Arc<RwLock<Option<ProtocolState>>>,
+        mesh_state: Arc<RwLock<MeshState>>,
+    ) -> anyhow::Result<()> {
+        let mut last_pinged = Instant::now();
+        loop {
+            if last_pinged.elapsed() > Duration::from_millis(300) {
+                if let Some(state) = contract_state.read().await.clone() {
+                    self.establish_participants(&state).await;
+                    let mut mesh_state = mesh_state.write().await;
+                    *mesh_state = MeshState {
```
volovyks, line 140: Why can't we fetch and set the updated info? Store all the information in one place and update it. Get rid of `establich_participants()`.

chain-signatures/node/src/mesh/mod.rs
```diff
@@ -29,38 +31,59 @@ impl Options {
     }
 }
 
+#[derive(Clone)]
+pub struct MeshState {
+    /// Participants that are active at the beginning of each protocol loop.
+    pub active_participants: Participants,
+
+    /// Potential participants that are active at the beginning of each protocol loop. This
```
volovyks, line 39: Same here.

chain-signatures/node/src/contract_updater.rs
```diff
@@ -0,0 +1,50 @@
...
+
+    pub async fn run(
+        &self,
+        contract_state: Arc<RwLock<Option<ProtocolState>>>,
+        config: Arc<RwLock<Config>>,
+    ) -> anyhow::Result<()> {
+        let mut last_state_update = Instant::now();
+        let mut last_config_update = Instant::now();
+        loop {
+            if last_state_update.elapsed() > Duration::from_millis(1000) {
```
volovyks, line 30: Is this logic duplicated?

chain-signatures/node/src/mesh/mod.rs
ppca: yep

chain-signatures/node/src/mesh/mod.rs
ppca: it only runs every 300 ms

chain-signatures/node/src/mesh/mod.rs
ppca: I'll update comments

chain-signatures/node/src/mesh/mod.rs
volovyks: Yes, but how many elapsed checks will you do during this 300ms?

ppca: ah makes sense

chain-signatures/node/src/mesh/connection.rs
```diff
@@ -68,24 +69,47 @@ impl Pool {
 
         let connections = self.connections.read().await;
 
-        let mut status = self.status.write().await;
         let mut participants = Participants::default();
+
+        // FuturesUnordered for parallel execution
+        let mut tasks = FuturesUnordered::new();
```
ChaoticTempest, line 75: There are issues with using `FuturesUnordered` outlined in github.com/rust-lang/futures-rs/issues/2387. Better to use `tokio::task::JoinSet` for this

chain-signatures/node/src/protocol/consensus.rs
```diff
@@ -81,15 +81,18 @@ pub trait ConsensusProtocol {
...
 }
 
 #[async_trait]
 impl ConsensusProtocol for StartedState {
+    #[allow(unused_variables)]
     async fn advance<C: ConsensusCtx + Send + Sync>(
         self,
         ctx: C,
         contract_state: ProtocolState,
+        cfg: Config,
```
ChaoticTempest, line 95: just use `_` for the unused variable, that way we don't miss anything if we change this later

ppca:
fetching participant state in parallel will come in another PR

chain-signatures/node/Cargo.toml
```diff
@@ -21,6 +21,7 @@ cait-sith = { git = "https://github.com/LIT-Protocol/cait-sith.git", features =
 ], rev = "8ad2316" }
 clap = { version = "4.2", features = ["derive", "env"] }
 chrono = "0.4.24"
+futures = "0.3"
```
ChaoticTempest, line 24: no longer need this since you got rid of `FuturesUnordered`

chain-signatures/node/src/mesh/mod.rs
```diff
@@ -29,89 +31,61 @@ impl Options {
...
+        let mut last_pinged = Instant::now();
+        loop {
+            if last_pinged.elapsed() > Duration::from_millis(300) {
+                if let Some(state) = contract_state.read().await.clone() {
+                    self.connections.establish_participants(&state).await;
+                    self.ping().await;
+                    last_pinged = Instant::now();
+                }
             }
+            tokio::time::sleep(Duration::from_millis(100)).await;
```
ChaoticTempest, line 88: no need to check every 300ms, rather just have sleep and we can get rid of maintaining `last_pinged` as well. Also no need for state clone.
```suggestion
            {
                let state = contract_state.read().await;
                if let Some(state) =  {
                    self.connections.establish_participants(&*state).await;
                }
            }
            tokio::time::sleep(Duration::from_millis(300)).await;
```

chain-signatures/node/src/contract_updater.rs
```diff
@@ -24,28 +24,29 @@ impl ContractUpdater {
...
                 if let Err(error) = config
                     .fetch_inplace(&self.rpc_client, &self.mpc_contract_id)
                     .await
                 {
                     tracing::error!("could not fetch contract's config: {error:?}");
                 }
-                last_config_update = Instant::now();
             }
-            tokio::time::sleep(Duration::from_millis(100)).await;
+            tokio::time::sleep(Duration::from_millis(1000)).await;
```
ChaoticTempest, line 49: I guess it doesn't matter that we're doing config update now every 1s, but we should eventually optimize the polling to get both config and state in one RPC call, but that can be done in another PR

ChaoticTempest (approved):

PR #55, opened 2024-12-05
chore(refactor): better testing interface

ChaoticTempest:
Newly improved testing interface for better test iterability.

Everything should now live under `Cluster`  and it's equivalent actions `WaitAction` and `SignAction`.

Example:
```rs

#[test(tokio::test)]
async fn test_signature_basic() -> anyhow::Result<()> {
    let nodes = cluster::spawn().wait_for_running().await?;
    nodes.wait().signable().await?;
    let outcome = nodes.sign()
        .rogue_responder()
        .deposit(NearToken::from_yoctonear(10000))
        .await?;

    let state = nodes.wait().running_on_epoch(1).await?;
    Ok(())
}

```


integration-tests/chain-signatures/tests/cases/mod.rs
```diff
@@ -28,191 +27,156 @@ pub mod nightly;
...
-    })
-    .await
+    let mut nodes = cluster::spawn().wait_for_running().await?;
+    nodes.wait().signable().await?;
+    let _ = nodes.sign().await?;
+
+    tracing::info!("!!! Add participant 3");
+    nodes.add_participant(None).await.unwrap();
+    let state = nodes.wait().running().signable().await.unwrap();
+    let _ = nodes.sign().await.unwrap();
```
ailisp, line 37: what is the reason for unwrap() instead of `?`

ailisp (approved):
Nice! A lot simpler and easy to use. Do you also test integration-tests/chain-signatures/src/main.rs still work?

integration-tests/chain-signatures/tests/cluster/mod.rs
```diff
@@ -0,0 +1,136 @@
...
+impl Cluster {
+    pub fn len(&self) -> usize {
+        self.nodes.len()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.nodes.is_empty()
+    }
+
+    pub fn url(&self, id: usize) -> Url {
```
volovyks, line 50: Should we use the account id to identify the node?

integration-tests/chain-signatures/tests/cases/mod.rs
```diff
@@ -399,162 +363,123 @@ fn dummy_triple(id: u64) -> Triple {
...
-    with_multichain_nodes(MultichainConfig::default(), |ctx| {
-        Box::pin(async move {
-            let state_0 = wait_for::running_mpc(&ctx, Some(0)).await?;
-            assert_eq!(state_0.participants.len(), 3);
-            actions::batch_duplicate_signature_production(&ctx, &state_0).await?;
-            Ok(())
-        })
-    })
-    .await
+    let nodes = cluster::spawn().wait_for_running().await?;
```
volovyks, line 482: Are we getting nodes or cluster here?

integration-tests/chain-signatures/tests/actions/wait.rs
```diff
@@ -0,0 +1,255 @@
...
+
+type Epoch = u64;
+
+enum WaitActions {
+    Running(Epoch),
+    MinTriples(usize),
+    MinMineTriples(usize),
+    MinPresignatures(usize),
+    MinMinePresignatures(usize),
+    Signable(usize),
```
volovyks, line 18: It is not clear what Signable means. How is it different from Running?

volovyks (approved):
Looks good!
Have the build time or time of execution improved?

integration-tests/chain-signatures/tests/cases/mod.rs
ChaoticTempest: unwrap will give you the backtrace as it panics meanwhile `?` will give you the anyhow error message only once its forward past `main`

integration-tests/chain-signatures/tests/cluster/mod.rs
ChaoticTempest: yeah more utilities can be added later, I'm just doing it based on what we have. We can have something like `url_by_id` eventually

integration-tests/chain-signatures/tests/cases/mod.rs
ChaoticTempest: `nodes` is type of ` Cluster` here since they're both equivalent

integration-tests/chain-signatures/tests/actions/wait.rs
volovyks: How is it different from having 1  presignature?

ChaoticTempest: So right now, it requires X number of presignatures but we can have further conditions down the line for it. It's just from a usability standpoint, when you write `signable`, you don't care about what is going on underneath and only care that we can sign something

ppca (dismissed):
Nice! Much simpler! signable() is great!

volovyks (approved):
Looks good!

PR #62, opened 2024-12-09
fetch participant state parallel

ppca:
Using tokio JoinSet.
`pub async fn ping(self: Arc<Self>) -> Participants` takes in Arc<Self> so that it can be passed among different tasks when calling `join_set.spawn()`

### DIFF
--- a/chain-signatures/node/src/mesh/connection.rs
+++ b/chain-signatures/node/src/mesh/connection.rs
@@ -10,6 +10,8 @@ use crate::protocol::ParticipantInfo;
 use crate::protocol::ProtocolState;
 use crate::web::StateView;
 use mpc_keys::hpke::Ciphered;
+use std::sync::Arc;
+use tokio::task::JoinSet;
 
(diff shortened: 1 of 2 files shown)


chain-signatures/node/src/mesh/connection.rs
```diff
@@ -60,41 +62,77 @@ impl Pool {
...
+        // Update the active participants
         let mut active = self.current_active.write().await;
         *active = Some((participants.clone(), Instant::now()));
+
         participants
     }
 
-    pub async fn ping_potential(&self) -> Participants {
+    // self typed Arc<Self> so it can be passed between tokio tasks
+    pub async fn ping_potential(self: Arc<Self>) -> Participants {
```
ailisp, line 135: looks like some logics can be reused with `ping`, but we can do in a future refactor PR

ailisp (approved):
Great optimization!

volovyks (approved):
LGTM

PR #63, opened 2024-12-12
feat: added faster linker for macos

ChaoticTempest:
**NOTE**: requires at least Xcode 15 to be installed on your mac machine.

Goes from 11m to 5m fresh integration test build with this.


volovyks (approved):

ppca (approved):

volovyks:
@ChaoticTempest, I have not noticed any improvements since this change. Do you know if I need to make any additional steps? I've installed the latest Xcode.

ChaoticTempest:
hmm, I guess this change isn't really needed if you have the latest xcode installed. rustc seems to be smart enough to check this, so this isn't really needed, more of a note to update to the latest xcode for faster compilation if that wasn't the case before

PR #64, opened 2024-12-12
Take two triples in atomic way

volovyks:
Making all operations atomic prevents race conditions, especially in multithreaded and multi-node design. It also decreases the number of Redis calls and simplifies the code, although we lose some error types.

I plan to adopt this approach in other places if we stick with it.

### DIFF
--- a/chain-signatures/node/src/protocol/triple.rs
+++ b/chain-signatures/node/src/protocol/triple.rs
@@ -179,23 +179,23 @@ impl TripleManager {
         id0: &TripleId,
         id1: &TripleId,
     ) -> Result<(Triple, Triple), GenerationError> {
-        if self.generators.contains_key(&id0) {
+        if self.generators.contains_key(id0) {
             tracing::warn!(id0, "triple is generating");
             return Err(GenerationError::TripleIsGenerating(*id0));
-        } else if self.gc.contains_key(&id0) {
+        } else if self.gc.contains_key(id0) {
             tracing::warn!(id0, "triple is garbage collected");
             return Err(GenerationError::TripleIsGarbageCollected(*id0));
-        } else if self.generators.contains_key(&id1) {
+        } else if self.generators.contains_key(id1) {
             tracing::warn!(id1, "triple is generating");
             return Err(GenerationError::TripleIsGenerating(*id1));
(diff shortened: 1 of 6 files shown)


ChaoticTempest:
Why not use `redis::transaction` or `redis::pipeline` for this? A lua script would imply running a lua runtime and all the overhead that brings

volovyks:
@ChaoticTempest atomic and transaction do not allow us to have conditional logic, return errors, etc.

volovyks:
From what I see, Lua scripts are the only thing we can use to make all these operations in a completely isolated way.
Redis should be able to execute tens of thousands of such Lua scripts per second. We are very far from it to be a bottleneck.

ChaoticTempest:
I took a look at benchmarks and Lua scripts actually don't have a huge performance impact (around +4-5%) but that's fine in the grand scheme of things

chain-signatures/node/src/protocol/triple.rs
```diff
@@ -170,82 +170,59 @@ impl TripleManager {
...
     /// Take two unspent triple by theirs id with no way to return it. Only takes
     /// if both of them are present.
     /// It is very important to NOT reuse the same triple twice for two different
     /// protocols.
     pub async fn take_two(
         &mut self,
-        id0: TripleId,
-        id1: TripleId,
+        id0: &TripleId,
+        id1: &TripleId,
```
ChaoticTempest, line 180: `TripleId` implements `Copy` so no need to pass by reference. It's also cheaper to pass it by value than reference here too since pointers are fat (i.e. 8 bytes in 64bit systems)

chain-signatures/node/src/storage/triple_storage.rs
```diff
@@ -54,20 +54,88 @@ impl TripleStorage {
         Ok(result)
     }
 
-    pub async fn take(&self, id: &TripleId) -> StoreResult<Triple> {
+    pub async fn take_two(&self, id1: &TripleId, id2: &TripleId) -> StoreResult<(Triple, Triple)> {
```
ChaoticTempest, line 57: same with this one

ChaoticTempest (dismissed):

ailisp (approved):
Looks good!

PR #67, opened 2024-12-16
Remove mpc recovery

volovyks:
Since we are not planning to work on this code base - let's remove it. It should make future refactoring, building, and testing speed optimizations easier.
This PR does not include any optimizations in the project structure, let's do that in a separate PR.


mpc-recovery/DEPLOY.md
```diff
@@ -1,86 +0,0 @@
-# Manually Deploying mpc-recovery to GCP
-
-## Requirements
```
volovyks, line 3: @auto-mausx not sure if it is useful for us now

auto-mausx:
Can we remove the running services? If so let me destroy the TF resources for MPC recovery first, then we can merge this and remove the terraform code.

ppca:
> Can we remove the running services? If so let me destroy the TF resources for MPC recovery first, then we can merge this and remove the terraform code.

That is a good idea

ppca (approved):

ChaoticTempest (changes requested):
since you moved the `config.toml` file to the parent dir, you need to remove one `..` so it's pointing to the right target folder

volovyks:
@ChaoticTempest can you please point me at what path you are talking about?

ChaoticTempest:
@volovyks actually nevermind, the git diff looked funky because it showed moving one file to another

ChaoticTempest (approved):

PR #68, opened 2024-12-16
Atomic insertion of presignatures and triples

volovyks:
Insert and take operations must be atomic. Lua may seem an overhead when we have 2 simple operations, but I want to have a unified approach for storage and flexibility. Already planning to extend those scripts with additional checks.
Also, scripts can be preloaded to Redis, so there is no need to transfer and build them on each call. We may add that optimization when required.

### DIFF
--- a/chain-signatures/node/src/storage/triple_storage.rs
+++ b/chain-signatures/node/src/storage/triple_storage.rs
@@ -33,12 +33,32 @@ impl TripleStorage {
 
     pub async fn insert(&self, triple: Triple, mine: bool) -> StoreResult<()> {
         let mut conn = self.connect().await?;
-        if mine {
-            conn.sadd::<&str, TripleId, ()>(&self.mine_key(), triple.id)
-                .await?;
-        }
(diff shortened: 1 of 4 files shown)


volovyks:
@ChaoticTempest I've fixed the tests

chain-signatures/node/src/storage/triple_storage.rs
```diff
@@ -33,12 +33,32 @@ impl TripleStorage {
...
+
+            return "OK"
+        "#;
+
+        let _: String = redis::Script::new(script)
+            .key(self.mine_key())
+            .key(self.triple_key())
+            .arg(triple.id)
+            .arg(triple)
+            .arg(mine.to_string())
```
ChaoticTempest, line 58: can't we just send it as a bool instead of a string?

volovyks: By default, it is serialized as o or 1, so our script will look like `if mine == "1"`. True/false is more readable both in code and when debugging redis storage directly.

ChaoticTempest (approved):

PR #69, opened 2024-12-17
feat: parallelize triple, presignatures and signatures as tokio tasks

ChaoticTempest:
This makes each triple it's own tokio task that gets spunned in the background. Additionally, the whole of triple tasks is allocated a default budget of 100ms before needing to yield back to the main loop. This is to circumvent the blocking nature of how long it takes.

Take a look at nightly, on my machine, it's able to generate 100 signatures in ~75 seconds with an 8 node setup:
```
RUST_LOG=integration_tests_chain_signatures=info,mpc_node=info cargo test -- test_nightly_signature_production --show-output --ignored
```


volovyks:
Let’s make a fair comparison. What were the results of the same test before these changes?

I don’t expect a major improvement in the number of signatures generated per minute, as it’s constrained by the availability of triples and presignatures (we start with none when the node launches). Additionally, triple generation is bottlenecked by the messaging layer.

What I do expect is a low (around 3 seconds) and stable response time for signature requests, even during active triple generation. To verify this, we should deploy the changes on our dev network and observe how they perform.

Another good signal is a drop in error rate (~15% now during triple generation). Since it is mostly caused by high latency.

chain-signatures/node/src/protocol/triple.rs
```diff
@@ -325,10 +674,11 @@ impl TripleManager {
             if self.len_potential().await >= cfg.triple.max_triples as usize {
                 false
             } else {
+                let tasks = self.tasks.read().await;
                 // We will always try to generate a new triple if we have less than the minimum
                 self.len_mine().await < cfg.triple.min_triples as usize
-                    && self.introduced.len() < cfg.max_concurrent_introduction as usize
-                    && self.generators.len() < cfg.max_concurrent_generation as usize
+                    && tasks.introduced.len() < cfg.max_concurrent_introduction as usize
```
ppca, line 680: is this same as self.len_introduced()?

ppca (commented):
overall looks nice! Just have a few questions/comments

chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -650,6 +651,54 @@ impl PresignatureManager {
...
+        let triple_manager = state.triple_manager.clone();
+        let presignature_manager = state.presignature_manager.clone();
+        let active = active.clone();
+        let protocol_cfg = protocol_cfg.clone();
+        let pk = state.public_key.clone();
+        let sk_share = state.private_share.clone();
+        let messages = state.messages.clone();
+
+        tokio::task::spawn(async move {
+            let mut presignature_manager = presignature_manager.write().await;
```
jakmeier, line 669: The benefits of spawning multiple tasks for concurrent execution might be reduced if the first thing we do is to acquire a write lock on a shared object across all tasks. Especially if we keep the lock until it is all done, which I think is the case here.

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -811,4 +812,69 @@ impl SignatureManager {
...
+        let signer = ctx.signer().clone();
+        let mpc_contract_id = ctx.mpc_contract_id().clone();
+
+        // NOTE: signatures should only use stable and not active participants. The difference here is that
+        // stable participants utilizes more than the online status of a node, such as whether or not their
+        // block height is up to date, such that they too can process signature requests. If they cannot
+        // then they are considered unstable and should not be a part of signature generation this round.
+
+        tokio::task::spawn(tokio::task::unconstrained(async move {
+            let mut sign_queue = sign_queue.write().await;
```
jakmeier, line 841: Again, we acquire a write lock for shared state, hence only one task at the time will get passed this line.

chain-signatures/node/src/protocol/triple.rs
```diff
@@ -35,52 +44,204 @@ pub struct Triple {
...
+        me: Participant,
+        my_account_id: &AccountId,
+        epoch: u64,
+    ) -> JoinHandle<GeneratorOutcome> {
+        tokio::task::spawn({
+            let mut generator = self.clone();
+            let my_account_id = my_account_id.clone();
+            async move { generator.execute(me, &my_account_id, epoch).await }
+        })
+    }
```
jakmeier, line 100: This looks better, at first glance what we do in `TripleGenerator::execute`, I think we get real parallelism here.

chain-signatures/node/src/protocol/triple.rs
```diff
@@ -346,190 +696,91 @@ impl TripleManager {
...
+        let active = active.clone();
+        let protocol_cfg = protocol_cfg.clone();
+
+        tokio::task::spawn(async move {
+            if let Err(err) = self.stockpile(&active, &protocol_cfg).await {
+                tracing::warn!(?err, "running: failed to stockpile triples");
+            }
+
+            {
+                let mut messages = messages.write().await;
```
jakmeier, line 763: This looks like a potential lock-contention again on `messages` if this queue is used by many tasks.

jakmeier (commented):
The overall idea seems to be to transform some form of `for future in futures { future.await } ` into `join!(futures).await`, is that about right? Essentially, doing all work items of a main loop iteration in parallel rather than in sequence?

This generally makes sense, if we can run things that take significant time to execute in parallel, then we should. The questions are (1) is the time per task significant and (2) can we actually run it in parallel.

I don't know the answer to either. Probably (1) is true. For (2), I'm really not sure.

Often, how much parallelism  we really get is dictated by how much shared state each work item requires. Ideally, state is organized such that each task has its own private state and requires no shared (mutable) state with other tasks. Otherwise, they will  be queueing behind each other to acquire the same lock, serializing the work again.

I left a few comments in places where it looks like this could be an issue.

(Oh and just to let you know, I didn't look at all changes in detail, I'm relying on others to do a full review. Since they are familiar with the code base, they are better suited to do a proper review.)

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -811,4 +812,69 @@ impl SignatureManager {
...
+        let rpc_client = ctx.rpc_client().clone();
+        let signer = ctx.signer().clone();
+        let mpc_contract_id = ctx.mpc_contract_id().clone();
+
+        // NOTE: signatures should only use stable and not active participants. The difference here is that
+        // stable participants utilizes more than the online status of a node, such as whether or not their
+        // block height is up to date, such that they too can process signature requests. If they cannot
+        // then they are considered unstable and should not be a part of signature generation this round.
+
+        tokio::task::spawn(tokio::task::unconstrained(async move {
```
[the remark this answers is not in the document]
ChaoticTempest: makes the signature task be prioritized such that the tokio runtime does not steal work from it and give it to other tasks

chain-signatures/node/src/protocol/triple.rs
```diff
@@ -93,12 +254,181 @@ pub struct TripleManager {
...
+
+        let mut interval = tokio::time::interval(Duration::from_millis(5));
+        let started = Instant::now();
+
+        // Go through each running task and see if it's done. This will apply a protocol_budget which will
+        // yield back control to the main loop if the time is up. If it is done, remove it from the ongoing_tasks.
+        // If the TripleGenerator is not done after this, a new task will be spawned in the next iteration
+        // in the case that the TripleGenerator is waiting.
+        while let Some((id, task)) = self.ongoing_tasks.pop_front() {
+            interval.tick().await;
```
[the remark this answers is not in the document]
ChaoticTempest: tokio's flavor of `sleep` but more nuanced. It allows you to sleep with respect to how much time you've taken already to do the work. So if you want to do something like a cycle constrained to 100ms exactly, but somewhere in the middle you do 10ms of work, then `interval.tick` here will only sleep for 90ms

chain-signatures/node/src/protocol/triple.rs
```diff
@@ -93,12 +254,181 @@ pub struct TripleManager {
...
+            if started.elapsed() > self.protocol_budget {
+                self.ongoing_tasks.push_back((id, task));
+                break;
```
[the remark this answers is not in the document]
ChaoticTempest: because we used up all our protocol budget. The whole of triple tasks is only given 100ms by default. If it goes beyond that, yield back control to the main loop. This does not stop all the running tasks from completing but just that we're no longer waiting for them to complete

chain-signatures/node/src/protocol/triple.rs
```diff
@@ -93,12 +254,181 @@ pub struct TripleManager {
...
+                    continue;
+                }
+            };
+            match outcome {
+                Ok((mut msgs, triple)) => {
+                    if let Some((triple, mine)) = triple {
+                        self.remove(id);
+                        triples.push((triple, mine));
+                    }
+                    messages.append(&mut msgs);
```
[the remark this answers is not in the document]
ChaoticTempest: these are just the normal protocol messages to be sent out to other nodes

chain-signatures/node/src/protocol/triple.rs
ChaoticTempest: yes, but the reason why we don't use it is because we'd be acquiring another read lock which we've already acquired a write lock so no point in calling the function directly

chain-signatures/node/src/protocol/presignature.rs
ChaoticTempest: good find! This will definitely contend with the signature task. I left this in for now, but I meant to separate out the contention of the presignature_manager in this PR as well similar to how `TripleManager` had internalized it's locks. I will make it less contentious in another PR or maybe in the same PR

chain-signatures/node/src/protocol/signature.rs
ChaoticTempest: `sign_queue` should actually just be turned into async channel here because it's only used by the indexer and the `SignatureManager` thus eliminating the need for it to be a shared state

chain-signatures/node/src/protocol/triple.rs
ChaoticTempest: good point, the idea is that acquiring the write lock for messages is short lived so actually better if it were an MPSC channel for this purpose, because otherwise we run into this issue if somebody were to try to acquire the message lock somewhere later

chain-signatures/node/src/protocol/signature.rs
volovyks: I was also thinking about moving signature requests to Redis. This is part of preparations for horizontal scaling, where multiple servers will need to access the same pool of signature requests. Also, it is a step forward to a design, where we have one or more separate indexers that are adding requests to the queue.

jakmeier: Making it an async channel sounds like a great next step to decoupling shared state of different tokio tasks.

I suspect moving signature requests to Redis would also benefit from this decoupling.

chain-signatures/node/src/protocol/triple.rs
jakmeier: > the idea is that acquiring the write lock for messages is short lived

It looks like right now the lock is held for the duration of the `self.poke` call, which seems to be all the relevant work this task does.

> so actually better if it were an MPSC channel for this purpose, because otherwise we run into this issue if somebody were to try to acquire the message lock somewhere later

I think an MPSC channel could make a lot of sense here. If you change it such that the task only needs a Sender<MpcMessage> (or something similar) rather than a full lock on the message queue, suddenly all poke calls can execute in parallel.

ChaoticTempest:
I'll make further changes in a separate PR for message channels, so this PR is ready to be merged in whenever

volovyks (approved):
Great work! Let's merge this and iterate with improvements.
And tests improvements are cool too!

ppca (approved):
sounds good to use mpsc channel and other separations in a separate PR.

PR #70, opened 2024-12-19
Restrict triples and presignatures reusage

volovyks:
We are having issues with the pre-signature usage. We should be 100% sure that is never the case. The same goes for triples.

In this PR, I've made the system strict. We should not allow the reinsertion of used triples and pre-signatures.
I've removed the reinsertion logic, it is seldom, but adds complexity to the logic.

Should fix https://github.com/sig-net/mpc/issues/40

### DIFF
--- a/chain-signatures/node/src/protocol/signature.rs
+++ b/chain-signatures/node/src/protocol/signature.rs
@@ -675,7 +675,7 @@ impl SignatureManager {
             }
 
             let Some(my_request) = my_requests.pop_front() else {
-                // TODO: we are loosing presignature now with no good reason
+                tracing::warn!("Unexpected state, no more requests to handle");
                 continue;
             };
 
--- a/chain-signatures/node/src/protocol/signature.rs
+++ b/chain-signatures/node/src/protocol/signature.rs
@@ -639,7 +639,7 @@ impl SignatureManager {
             if sig_participants.len() < threshold {
                 tracing::warn!(
                     participants = ?sig_participants.keys_vec(),
-                    "intersection of stable participants and presignature participants is less than threshold, treshing presignature"
+                    "intersection of stable participants and presignature participants is less than threshold, trashing presignature"
                 );
                 continue;
             }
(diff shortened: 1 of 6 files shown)


chain-signatures/node/src/protocol/signature.rs
```diff
@@ -680,7 +675,7 @@ impl SignatureManager {
             }
 
             let Some(my_request) = my_requests.pop_front() else {
-                failed_presigs.push(presignature);
+                // TODO: we are loosing presignature now with no good reason
```
volovyks, line 678: @ChaoticTempest I'm not fixing this TODO, since this part if rewritten in your PR.

volovyks:
@ChaoticTempest  I will replace GC with this solution in my next PR.

chain-signatures/node/src/storage/triple_storage.rs
```diff
@@ -190,6 +210,13 @@ impl TripleStorage {
             TRIPLE_STORAGE_VERSION, self.node_account_id
         )
     }
+
+    fn used_key(&self) -> String {
+        format!(
+            "used_triples:{}:{}",
+            TRIPLE_STORAGE_VERSION, self.node_account_id
+        )
+    }
```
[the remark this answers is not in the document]
volovyks: Do you want to move the `format!()` logic to construction to prevent executing it often?

chain-signatures/node/src/storage/presignature_storage.rs
```diff
@@ -7,18 +8,36 @@ use crate::storage::error::{StoreError, StoreResult};
 
 // Can be used to "clear" redis storage in case of a breaking change
 const PRESIGNATURE_STORAGE_VERSION: &str = "v2";
+const USED_EXPIRE_TIME: Duration = Duration::hours(24);
 
 pub fn init(pool: &Pool, node_account_id: &AccountId) -> PresignatureStorage {
```
ChaoticTempest, line 13: I don’t know why we kept using this pattern but this is normally reserved for functions that return more than one value such as a tuple. It’s better in the future for us to have a `PresignatureStorage::new` instead but this isn’t blocking anything

ChaoticTempest (approved):

chain-signatures/node/src/storage/presignature_storage.rs
volovyks: I'm trying to keep everything consistent. Ok, I will take a look.

PR #80, opened 2025-01-02
fastNear rate limit: change fetch contract interval from 1000ms to 3000ms

ppca:


volovyks (approved):
When the data become outdated? Should we change that variable?

volovyks:
How often must we update the data? When the node becomes "unstable"?

ppca:
> How often must we update the data? When the node becomes "unstable"?

this is contract state and config, it updates at its own frequency, whether the node becomes unstable has nothing to do with contract.

PR #81, opened 2025-01-03
Update GC logic and purpose

volovyks:
We should not replace in-memory GC with Redis, since it will create a lot of traffic (at least one call for each message in the system).
I've tried to optimize how GC works, now it is all about protocols. Each protocol can either finish or fail, and that is when we add its id to GC.

The only job for GC now is to prevent the recreation of finished or failed protocols. cc @ChaoticTempest

### DIFF
--- a/chain-signatures/node/src/protocol/triple.rs
+++ b/chain-signatures/node/src/protocol/triple.rs
@@ -96,8 +96,8 @@ pub struct TripleManager {
     /// The set of triples that were introduced to the system by the current node.
     pub introduced: HashSet<TripleId>,
 
-    /// The set of triple ids that were already taken or failed. This will be maintained for at most
-    /// triple timeout period just so messages are cycled through the system.
+    /// The set of triple ids that were already generated or failed. This will be maintained for at most
+    /// triple timeout period to prevent protocol restarts.
     pub gc: HashMap<TripleId, Instant>,
 
     pub me: Participant,
@@ -149,8 +149,6 @@ impl TripleManager {
         tracing::debug!(id, mine, "inserting triple");
         if let Err(e) = self.triple_storage.insert(triple, mine).await {
             tracing::warn!(?e, mine, "failed to insert triple");
-        } else {
-            self.gc.remove(&id);
         }
     }
 
@@ -213,8 +211,6 @@ impl TripleManager {
                     ))
                 })?;
 
-        self.gc.insert(id0, Instant::now());
-        self.gc.insert(id1, Instant::now());
         tracing::debug!(id0, id1, "took two triples");
         Ok((triple_0, triple_1))
     }
@@ -231,8 +227,6 @@ impl TripleManager {
                 tracing::warn!(?store_error, "failed to take two mine triples");
             })
             .ok()?;
-        self.gc.insert(triple_0.id, Instant::now());
-        self.gc.insert(triple_1.id, Instant::now());
 
         tracing::debug!(triple_0.id, triple_1.id, "took two mine triples");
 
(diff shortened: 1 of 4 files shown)


chain-signatures/node/src/protocol/triple.rs
```diff
@@ -290,16 +284,6 @@ impl TripleManager {
         timeout: u64,
     ) -> Result<(), InitializationError> {
         let id = rand::random();
-
-        // Check if the `id` is already in the system. Error out and have the next cycle try again.
```
volovyks, line 294: This call is redundant since we check our in-memory and Redis storage for an id randomly created one line above. This check adds an unnecessary call to Redis (or 2, if we want to check "used triples").

chain-signatures/node/src/protocol/triple.rs
```diff
@@ -359,39 +343,35 @@ impl TripleManager {
         participants: &Participants,
         cfg: &ProtocolConfig,
     ) -> Result<Option<&mut TripleProtocol>, CryptographicError> {
-        if self.contains(id).await || self.gc.contains_key(&id) {
```
volovyks, line 362: The same logic here. We call `get_or_start_generation` after all messages are filtered against GC. That should be enough.

volovyks:
From what I see we do not have a GC mechanism for the Signature protocols. There is only `completed`.
That is because we have retries. But that also means we can not distinguish between old and new messages (after retry). Not sure if that is causing issues now.

Another thing. In case the signature generator fails, we follow this logic:
```rust
if generator.proposer == self.me {
    if generator.sign_request_timestamp.elapsed() < generator.timeout_total {
         // ...
    } else {
        self.completed.insert(sign_request_identifier.clone(), Instant::now());
        // ...
    }
}
break false;
```
It means we are adding a signature to the completed map only if it's `mine`. So it can not be a replacement for GC.
But the main thing is that we are not checking incoming messages against `completed` requests.
I want to discuss our strategy here, but this is not directly related to this PR.

volovyks:
The tests are failing because we are checking if Triple was garbage collected before taking it. And with this new logic, it is. I will need to refactor a bit more.

chain-signatures/node/src/protocol/triple.rs
ChaoticTempest: this check is not redundant because creating an id that collides with one of the ids already in our network is unlikely but still a possibility. So on the chance that we do, let's error out first and have it try again later. Otherwise, this id will actually make it all the way to triple generation phase, complete the triple generation, and then try to be inserted into storage which it will fail at that stage, taking up all the time with this process. It's better to check this first then to have this case

chain-signatures/node/src/protocol/triple.rs
```diff
@@ -523,6 +503,7 @@ impl TripleManager {
                         // Protocol done, remove it from the ongoing pool.
                         self.ongoing.remove(id);
                         self.introduced.remove(id);
+                        self.gc.insert(*id, Instant::now());
```
ChaoticTempest, line 506: the issue with inserting it once it's done generating is that the GC expires it before it's ever gets used

volovyks: GC means we will not accept messages with this ID and will not restart a completed or failed protocol.
But now I'm reconsidering the whole approach. Maybe we need to deal with that on the messaging layer.

chain-signatures/node/src/protocol/triple.rs
volovyks: What is the probability of that? My guess is ~ a chance to guess someone's private key. But calls to Redis are very real and happen often.

ChaoticTempest: It's very very unlikely it happens, but if it does, then it could break some things we did not intend for it to. Which is what I am trying to avoid. I would rather have us directly call redis in this case because the cost is not that bad. This is per triple generation anyways, so it doesn't matter. The biggest cost there is already the triple generation process

chain-signatures/node/src/protocol/triple.rs
ChaoticTempest: Currently GC is more than just a filter for the messages though. It's used for more accurate error reporting when we take triples as well.

I'm in favor of making a dedicated message filter where we alert the message channel that a particular id has been completed

volovyks:
@ChaoticTempest if there is anything you want to do with GC - please, create a PR.

PR #91, opened 2025-01-07
feat: make sign queue use less shared state

ChaoticTempest:
Makes the sign queue less contentious with indexer. This allows signature manager to progress without indexer taking up the write lock for too long.

This makes it so that sign queue only gets created in the `SignatureManager`, where a channel `(sign_tx, sign_rx)` is create between the indexer and the protocol loop. The `sign_tx` is used by the indexer to send messages over the channel to `sign_rx` which is acquired by the `SignatureManager` on creation and organizing of message requests


chain-signatures/node/src/protocol/signature.rs
```diff
@@ -36,81 +42,54 @@ pub struct SignRequest {
...
+    me: Participant,
+    sign_rx: Arc<RwLock<mpsc::Receiver<SignRequest>>>,
+    requests: HashMap<Participant, VecDeque<SignRequest>>,
 }
 
 impl SignQueue {
-    pub fn new() -> Self {
-        Self::default()
+    pub fn channel() -> (mpsc::Sender<SignRequest>, mpsc::Receiver<SignRequest>) {
+        mpsc::channel(MAX_SIGN_REQUESTS)
```
volovyks, line 53: What would the behavior be if we got more than 1024 requests? Will they be dropped or wait to be added?

volovyks (dismissed):
I'm unsure if this is a narrow place, but it makes sense overall. We will need to implement such a design for messages.

I have a feeling that we will keep sign requests in Redis, but it's a good solution for now.

ppca (dismissed):

chain-signatures/node/src/protocol/signature.rs
ChaoticTempest: yeah, they'll be dropped but we'll have a whole entire sweep about adding these configs to the contract when we have more demand for it

ppca (approved):

ppca (approved):

PR #92, opened 2025-01-08
fix: added triples and presigs back to storage and fix EXPIRE on used

ChaoticTempest:
- Adds back triple and presignatures to storage if they do not meet the threshold criteria. A new parameter `back` is added for cases where we are sure we want to add these back to storage. Mainly added back to circumvent the cases of inconsistency when one node decides to drop a triple and presignature but others do not or were not aware of it. In the current code, they would never drop these as they do not own them and will purely sit in storage. It's better to have the chance later to use them otherwise we'd still be at the same problem of taking up maximum amount of triples/presignatures we can have per the config.
- Updates to redis 7.4.2 in tests for `HEXPIRE` where `EXPIRE` was deleting the whole set before, which isn't ideal for when we want to expire per triple id or presignature id. Especially when a new triple or presignature gets used, the time to live on this set would get updated and effectively never get deleted.

### DIFF
--- a/chain-signatures/node/src/protocol/presignature.rs
+++ b/chain-signatures/node/src/protocol/presignature.rs
@@ -190,10 +190,14 @@ impl PresignatureManager {
         }
     }
 
-    pub async fn insert(&mut self, presignature: Presignature, mine: bool) {
+    pub async fn insert(&mut self, presignature: Presignature, mine: bool, back: bool) {
         let id = presignature.id;
         tracing::debug!(id, mine, "inserting presignature");
-        if let Err(store_err) = self.presignature_storage.insert(presignature, mine).await {
+        if let Err(store_err) = self
+            .presignature_storage
+            .insert(presignature, mine, back)
+            .await
+        {
             tracing::error!(?store_err, mine, "failed to insert presignature");
         } else {
             // Remove from taken list if it was there
@@ -437,9 +441,10 @@ impl PresignatureManager {
                         participants = ?presig_participants.keys_vec(),
                         "running: the intersection of participants is less than the threshold"
                     );
-                    // We are not inserting triples back to
-                    //    - prevent triple reusage
-                    //    - prevent a situation where we have a full stockpile of triples that can not be used
+                    // TODO: do not insert back triples when we have a clear model for data consistency
+                    // between nodes and utilizing only triples that meet threshold requirements.
+                    triple_manager.insert(triple0, true, true).await;
+                    triple_manager.insert(triple1, true, true).await;
                 } else {
                     self.generate(
                         &presig_participants,
@@ -640,7 +645,7 @@ impl PresignatureManager {
         });
 
         for (presignature, mine) in presignatures {
-            self.insert(presignature, mine).await;
+            self.insert(presignature, mine, false).await;
         }
 
         if !errors.is_empty() {
--- a/chain-signatures/node/src/protocol/signature.rs
+++ b/chain-signatures/node/src/protocol/signature.rs
@@ -642,6 +642,9 @@ impl SignatureManager {
                     participants = ?sig_participants.keys_vec(),
                     "intersection of stable participants and presignature participants is less than threshold, trashing presignature"
                 );
+                // TODO: do not insert back presignature when we have a clear model for data consistency
+                // between nodes and utilizing only presignatures that meet threshold requirements.
+                presignature_manager.insert(presignature, true, true).await;
                 continue;
             }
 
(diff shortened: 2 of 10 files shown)


chain-signatures/node/src/storage/triple_storage.rs
```diff
@@ -150,33 +155,34 @@ impl TripleStorage {
...
             redis.call("HDEL", KEYS[2], id1, id2)
 
-            -- Add the triples to the used set and set expiration time
-            redis.call("SADD", KEYS[3], id1, id2)
-            redis.call("EXPIRE", KEYS[3], ARGV[1])
-    
+            -- Add the triples to the used set and set expiration time. Note, HSET is used so
+            -- we can expire on each field instead of the whole hash set.
+            redis.call("HSET", KEYS[3], id1, "1", id2, "1")
+            redis.call("HEXPIRE", KEYS[3], ARGV[1], "FIELDS", 2, id1, id2)
```
volovyks, line 184: nice catch

integration-tests/README.md
```diff
@@ -5,14 +5,7 @@
...
 
 ```BASH
-docker pull ghcr.io/near/sandbox
-docker pull redis:7.0.15
-```
-
-For M1 you may want to pull the following image instead:
-
-```BASH
-docker pull ghcr.io/near/sandbox:latest-aarch64
```
volovyks, line 15: Isn't this still relevant?

volovyks (approved):
Let's make sure we revert when we have the solution for consistency.

chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -437,9 +441,10 @@ impl PresignatureManager {
                         participants = ?presig_participants.keys_vec(),
                         "running: the intersection of participants is less than the threshold"
                     );
-                    // We are not inserting triples back to
-                    //    - prevent triple reusage
-                    //    - prevent a situation where we have a full stockpile of triples that can not be used
+                    // TODO: do not insert back triples when we have a clear model for data consistency
```
ppca, line 444: why don't we check threshold first then do `triple_manager.take_two_mine()`? Then we don't need to take then insert back

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -642,6 +642,9 @@ impl SignatureManager {
                     participants = ?sig_participants.keys_vec(),
                     "intersection of stable participants and presignature participants is less than threshold, trashing presignature"
                 );
+                // TODO: do not insert back presignature when we have a clear model for data consistency
+                // between nodes and utilizing only presignatures that meet threshold requirements.
+                presignature_manager.insert(presignature, true, true).await;
```
ppca, line 647: same here. we can check threshold first then take presignature

ppca (commented):
It seems like if we can check threshold first, then do take_mine(), then we do not need to insert anything back.

integration-tests/README.md
ChaoticTempest: nope, we don't need sandbox at all

chain-signatures/node/src/protocol/presignature.rs
ChaoticTempest: how would you compare threshold first?

ppca: ah, my bad.

chain-signatures/node/src/storage/triple_storage.rs
```diff
@@ -51,8 +51,11 @@ impl TripleStorage {
             local triple_id = ARGV[1]
             local triple_value = ARGV[2]
             local mine = ARGV[3]
+            local back = ARGV[4]
 
-            if redis.call("SISMEMBER", used_key, triple_id) == 1 then
+            if back == "true" then
+                redis.call("HDEL", used_key, triple_id)
+            elseif redis.call("HEXISTS", used_key, triple_id) == 1 then
```
ppca, line 58: what's the difference between HEXISTS and SISMEMBER?

ppca (approved):

ppca:
And just for my own clarification: our current code does not treat back=true/false differently?

chain-signatures/node/src/storage/triple_storage.rs
ChaoticTempest: Theyre methods for different types. One is for set and one is for map

ChaoticTempest:
> And just for my own clarification: our current code does not treat back=true/false differently?

before this PR, there was no `back` at all because we never added back triples/presignatures. The `back` is so we ignore the `if exists in used set` check

Issue #71, opened 2024-12-19
Add more performance metrics


jakmeier:
### Description

Goal: Have Grafana metrics that tell us how much of the protocol invocation delay is due to CPU work, how much is due to network delays, and how much is due to unnecessary overhead.

Measuring each of these directly may be difficult. But we can measure a couple of things and get to the same high-level result.

We can have the following metrics, with tags per protocol type:

- [ ] protocol_time: Measure the wall-clock time between initiating a protocol until it finishes. (Probably already exists as metric in some form.)
- [ ] poke_time: Every time we call `poke`, measure how much CPU time we spend on it and sum it up for all `poke` calls for the protocol from its creation until it is done. Submit the total poke time per protocol invocation to a histogram metric.
- [ ] client_send_delay: Start the clock anytime we process a `Action::SendMany` or `Action::SendPrivate` and stop it once we are done sending the messages. Submit each time individually to a histogram metric.

Each of these data points is collected locally on each node individually but aggregated as a global histogram. Showing the histogram in Grafana directly can already be useful. Additionally, we can use the sum & count aggregations from the collected histogram to show more interesting statistics:

- Relative CPU time: Display ( poke_time_sum / protocol_time_sum ) to show what percentage of the time we are CPU-bound.
- Message rounds per protocol invocation: Display (client_send_delay_count / protocol_time_count) to show how many messages we send (globally) per protocol invocation.
- Relative sending overhead: (client_send_delay_sum /  protocol_time_sum) to show the percentage of the time we are spending on sending messages out.
- Average network delay: (protocol_time_sum - poke_time_sum - client_send_sum) / protocol_time_count

Ideally, CPU time and network delay should make up >90% of the total time, with the sending overhead being small. Also, the network delay should be close to the observed ping delay. Otherwise, we have to investigate.

*edit 2025-01-02, adding the diagram*

```txt
         Poke called    <----------------------------------------
                |                                               ^ 
                |   (poke_delay)                                | 
                V                                               | 
  Poke returns Action::SendPrivate                              | 
                |                                               | 
                |   (client_send_delay)                         | 
                V                                               | 
      HTTP Call starts                                          | repeat until done
                |                                               | protocol_time = the total delay of all rounds
                |   (HTTP delay)                                | 
                V                                               | 
      HTTP response arrived                                     | 
                |                                               | 
                ------------------------------------------------| 

```

jakmeier:
[...]

PR #93, opened 2025-01-10
feat: add MessageChannel for async protocol message receiving and sending

ChaoticTempest:
Lots of code here. The crux of what this PR does is:
- new `MessageChannel` that processes inbound and outbound messages in a background task. 
- makes receiving of messages from `/msg` go into `MessageChannel` to be sorted and finally be received by protocols thorugh the `MessageReceiver` interface.
- protocols now directly call `MessageChannel::send` to send messages outbound so they no longer block when doing this operation.
- some cleanup where we now have a `NodeClient` to call `msg` and `state` endpoints.


volovyks (approved):
Nice design for message amnagement
And good refactoring

chain-signatures/node/src/node_client.rs
```diff
@@ -0,0 +1,117 @@
...
+    }
+
+    pub async fn state(&self, base: impl IntoUrl) -> Result<StateView, SendError> {
+        let mut url = base.into_url()?;
+        url.set_path("state");
+
+        let resp = self
+            .http
+            .get(url)
+            .timeout(Duration::from_millis(self.options.state_timeout))
```
ppca, line 111: I tried this approach before, but doing .timeout() directly used to cause inconsistent timeouts. Like I set timeout to be 100ms, but actually it takes much longer to timeout. That's why I changed to use tokio::timeout(), which was much more consistent. 
Seems to be a common observation, but I forgot the reason why. But I'm not particularly good at multi-task or multi-threading stuff, I'd say better check this issue out online and see if it will continue to be an issue for us now.

chain-signatures/node/src/node_client.rs
```diff
@@ -0,0 +1,117 @@
...
+    async fn post_msg(&self, url: &Url, msg: &[Ciphered]) -> Result<(), SendError> {
+        let resp = self
+            .http
+            .post(url.clone())
+            .header("content-type", "application/json")
+            .json(&msg)
+            .send()
+            .await?;
+
+        let status = resp.status();
```
ppca, line 81: This function would need a timeout too otherwise this might get stuck in unlucky situation like the aurora mainnet incident where their /msg endpoint is unavailable

chain-signatures/node/src/protocol/cryptography.rs
```diff
@@ -94,53 +92,40 @@ impl CryptographicProtocol for GeneratingState {
                 Action::Wait => {
                     drop(protocol);
                     tracing::debug!("generating: waiting");
-                    let failures = self
```
ppca, line 97: why is this not needed any more?

chain-signatures/node/src/protocol/message.rs
```diff
@@ -141,10 +175,123 @@ impl MpcMessageQueue {
...
+    inbox: Arc<RwLock<MessageInbox>>,
+    outbox: MessageOutbox,
+
+    config: Arc<RwLock<Config>>,
+    mesh_state: Arc<RwLock<MeshState>>,
+}
+
+impl MessageExecutor {
+    pub async fn execute(mut self) {
+        let mut interval = tokio::time::interval(Duration::from_millis(100));
```
ppca, line 215: Does this mean execute() will not clear the inbox if the time taken exceeds 100ms?

chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -573,43 +571,55 @@ impl PresignatureManager {
...
-                                    timestamp: Utc::now().timestamp() as u64
+                                    data,
+                                    timestamp: Utc::now().timestamp() as u64,
                                 },
-                            ))
-                        }
+                            )
+                            .await;
                     }
-                    Action::SendPrivate(p, data) => messages.push((
```
ppca, line 601: oh, there's no way the returned action is SendPrivate? That's actually nice to know

chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -638,11 +648,16 @@ impl PresignatureManager {
...
-                        return false;
+                        remove.push(*id);
+                        break;
                     }
                 }
             }
-        });
+        }
+
+        for id in remove {
```
ppca, line 658: so remove tracks the failed generators?

chain-signatures/node/src/protocol/cryptography.rs
ChaoticTempest: because this is now done in the `MessageChannel`

chain-signatures/node/src/node_client.rs
ChaoticTempest: interesting, let me check myself then

chain-signatures/node/src/node_client.rs
ChaoticTempest: this one already has the default timeout when the Client gets built

chain-signatures/node/src/protocol/message.rs
ChaoticTempest: I'm not sure what you mean by clear here. The inbox never gets cleared unless the protocols themselves takes the specific messages associated to that protocol on `state.recv`

For interval, when the amount of work exceeds 100ms, the next `interval.tick()` immediately runs without a sleep, so we don't sleep at all.

chain-signatures/node/src/protocol/presignature.rs
ChaoticTempest: yeah, it's simplified so that all messages coming to `/msg` endpoint is encrypted or private by default

chain-signatures/node/src/protocol/presignature.rs
ChaoticTempest: yup, before it was because we had `generators.retain` which removes the generator, but that did not allow us to call async functions, so instead we have to iterate over the generators, but that means we cannot remove them inplace. So we have this new `remove` vector that keeps track of which ids to remove after we initially iterated over the generators

chain-signatures/node/src/protocol/cryptography.rs
```diff
@@ -191,29 +156,10 @@ impl CryptographicProtocol for GeneratingState {
 impl CryptographicProtocol for WaitingForConsensusState {
     async fn progress<C: CryptographicCtx + Send + Sync>(
         mut self,
-        ctx: C,
-        cfg: Config,
-        mesh_state: MeshState,
+        _ctx: C,
+        _cfg: Config,
+        _mesh_state: MeshState,
```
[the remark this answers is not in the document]
ChaoticTempest: some state actually needs them so we can't just drop them right now. We should instead be simplifying the interface so that we can go back to how we did `ctx.mesh_state()` instead of passing them. This will be left for another PR if anyone wants to take it up

ppca (commented):
also appreciate your answer in https://github.com/sig-net/mpc/issues/71 as I am adding metrics to measure delay on each step per protocol, and this change will affect how my definition and implementation should be. Greatly appreciated! 🙏

chain-signatures/node/src/protocol/message.rs
ppca: I meant exhaust all messages, which is what used to happen. But I get it here because this channel runs its execute() in parallel and there's no definite time that it runs, so having a cap on how much time it consumes is important, otherwise it keeps receiving message so it will always be running

chain-signatures/node/src/node_client.rs
ppca: teach me why after you find out 🙏

chain-signatures/node/src/node_client.rs
ppca: Do you know what the default timeout is? If my memory was correct, I think that was also an inconsistent timeout :( likely similar to the other one I mentioned.

ChaoticTempest: Its in the new function, where the builder is created

chain-signatures/node/src/protocol/message.rs
ChaoticTempest: hmm, good idea. I'll add it in a subsequent PR or maybe this PR if the timeout needs to be adjusted, but this loop is what was in the protocol loop originally so very unlikely it'll block us

chain-signatures/node/src/node_client.rs
ChaoticTempest: So, I made a test for this, and there's barely any inconsistent timeouts that I would notice. Inside one of the `reqwest` `request_timeout` test, adding the elapsed check consistently yields 101ms with 100ms timeout. So this shouldn't be an issue

```rs
#[tokio::test]
async fn request_timeout() {
    let _ = env_logger::try_init();

    let server = server::http(move |_req| {
        async {
            // delay returning the response
            tokio::time::sleep(Duration::from_millis(300)).await;
            http::Response::default()
        }
    });

    let client = reqwest::Client::builder().no_proxy().build().unwrap();

    let url = format!("http://{}/slow", server.addr());

    let time = std::time::Instant::now();
    let res = client
        .get(&url)
        .timeout(Duration::from_millis(100))
        .send()
        .await;
    println!("elapsed: {:?}", time.elapsed());

    let err = res.unwrap_err();

    if cfg!(not(target_arch = "wasm32")) {
        assert!(err.is_timeout() && !err.is_connect());
    } else {
        assert!(err.is_timeout());
    }
    assert_eq!(err.url().map(|u| u.as_str()), Some(url.as_str()));
}
```

ChaoticTempest:
I'll merge this in and see what dev does overnight with our metrics

chain-signatures/node/src/node_client.rs
ppca: This was the PR where I changed to use tokio::timeout: https://github.com/near/mpc/pull/889, 

The description there:

> The previous implementation caused troubles: 1) 1s timeout have already stuck the protocol; 2) 500ms timeout still see many timeouts on /state endpoint. I am guessing the .timeout() that came with reqwest package cannot successfully time out an async call, the counting of time may not have considered task/thread switching.
So I switched to using tokio::time::timeout and now problem solved.

It might be useful to verify on dev on taking one node offline, and see if the 1s timeout here would stuck the protocol. If not, then I guess our recent optimization of code has made this earlier issue disappear.

PR #94, opened 2025-01-10
Triple/Presig/Signature protocol performance metrics

ppca:
This PR adds the following metrics for triple, presig and signature protocol:
- accrued wait time per protocol: add up the wall clock time difference between pokes that return SendMany/SendPrivate/Return
- before poke delay per protocol: time between generator creation and first poke
- count of pokes per protocol: number of pokes that return SendMany/SendPrivate/Return, this is a best-effort estimate of the number of rounds of communication per protocol.

And add the following for messages:
- client send delay: time between message getting into inbox to message being ready to be sent over wire

### DIFF
--- a/chain-signatures/node/src/metrics.rs
+++ b/chain-signatures/node/src/metrics.rs
@@ -483,6 +483,15 @@ pub(crate) static LATEST_BLOCK_NUMBER: LazyLock<IntGaugeVec> = LazyLock::new(|| 
         &["chain", "node_account_id"],
     )
     .unwrap()
+
+pub(crate) static PRESIGNATURE_BEFORE_POKE_DELAY: Lazy<HistogramVec> = Lazy::new(|| {
+    try_create_histogram_vec(
+        "multichain_presignature_before_poke_delay_sec",
+        "per presignature protocol, delay between generator cration and first poke that returns SendMany/SendPrivate",
+        &["node_account_id"],
+        Some(exponential_buckets(1.0, 1.5, 20).unwrap()),
+    )
+    .unwrap()
 });
 
 pub(crate) static NUM_UNIQUE_SIGN_REQUESTS: LazyLock<CounterVec> = LazyLock::new(|| {
@@ -490,6 +499,105 @@ pub(crate) static NUM_UNIQUE_SIGN_REQUESTS: LazyLock<CounterVec> = LazyLock::new
         "multichain_sign_requests_count_unique",
         "number of multichain sign requests, marked by sign requests indexed and deduped",
         &["chain", "node_account_id"],
+    )
+    .unwrap()
+
+pub(crate) static PRESIGNATURE_ACCRUED_WAIT_DELAY: Lazy<HistogramVec> = Lazy::new(|| {
+    try_create_histogram_vec(
+        "multichain_presignature_accrued_wait_delay_sec",
+        "per presignature protocol, total accrued wait time between each poke that returned SendMany/SendPrivate/Return",
+        &["node_account_id"],
+        Some(exponential_buckets(1.0, 1.5, 20).unwrap()),
+    )
+    .unwrap()
+});
+
+pub(crate) static TRIPLE_BEFORE_POKE_DELAY: Lazy<HistogramVec> = Lazy::new(|| {
+    try_create_histogram_vec(
+        "multichain_triple_before_poke_delay_sec",
+        "per triple protocol, delay between generator cration and first poke that returns SendMany/SendPrivate",
+        &["node_account_id"],
+        Some(exponential_buckets(1.0, 1.5, 20).unwrap()),
+    )
+    .unwrap()
+});
+
+pub(crate) static TRIPLE_ACCRUED_WAIT_DELAY: Lazy<HistogramVec> = Lazy::new(|| {
+    try_create_histogram_vec(
+        "multichain_triple_accrued_wait_delay_sec",
+        "per triple protocol, total accrued wait time between each poke that returned SendMany/SendPrivate/Return",
+        &["node_account_id"],
+        Some(exponential_buckets(1.0, 1.5, 20).unwrap()),
+    )
+    .unwrap()
+});
+
+pub(crate) static SIGNATURE_BEFORE_POKE_DELAY: Lazy<HistogramVec> = Lazy::new(|| {
+    try_create_histogram_vec(
+        "multichain_signature_before_poke_delay_sec",
+        "per signature protocol, delay between generator cration and first poke that returns SendMany/SendPrivate",
+        &["node_account_id"],
+        Some(exponential_buckets(1.0, 1.5, 20).unwrap()),
+    )
+    .unwrap()
+});
+
+pub(crate) static SIGNATURE_ACCRUED_WAIT_DELAY: Lazy<HistogramVec> = Lazy::new(|| {
+    try_create_histogram_vec(
+        "multichain_signature_accrued_wait_delay_sec",
+        "per signature protocol, total accrued wait time between each poke that returned SendMany/SendPrivate/Return",
+        &["node_account_id"],
+        Some(exponential_buckets(1.0, 1.5, 20).unwrap()),
+    )
+    .unwrap()
+});
+
+pub(crate) static TRIPLE_LATENCY_TOTAL: Lazy<HistogramVec> = Lazy::new(|| {
+    try_create_histogram_vec(
+        "multichain_triple_latency_total_sec",
+        "Latency of multichain triple generation, start from generator creation, end when triple generation complete.",
+        &["node_account_id"],
+        Some(exponential_buckets(5.0, 1.5, 20).unwrap()),
+    )
+    .unwrap()
+});
+
+pub(crate) static TRIPLE_POKES_CNT: Lazy<HistogramVec> = Lazy::new(|| {
+    try_create_histogram_vec(
+        "multichain_triple_pokes_cnt",
+        "total pokes per triple protocol",
+        &["node_account_id"],
+        None,
+    )
+    .unwrap()
+});
+
+pub(crate) static PRESIGNATURE_POKES_CNT: Lazy<HistogramVec> = Lazy::new(|| {
+    try_create_histogram_vec(
+        "multichain_presignature_pokes_cnt",
+        "total pokes per presignature protocol",
+        &["node_account_id"],
+        None,
+    )
+    .unwrap()
+});
+
+pub(crate) static SIGNATURE_POKES_CNT: Lazy<HistogramVec> = Lazy::new(|| {
+    try_create_histogram_vec(
+        "multichain_signature_pokes_cnt",
+        "total pokes per signature protocol",
+        &["node_account_id"],
+        None,
+    )
+    .unwrap()
+});
+
+pub(crate) static MSG_CLIENT_SEND_DELAY: Lazy<HistogramVec> = Lazy::new(|| {
+    try_create_histogram_vec(
+        "multichain_msg_client_send_delay_ms",
+        "Delay between message creation and sending to the client",
+        &["node_account_id"],
+        Some(exponential_buckets(0.5, 1.5, 20).unwrap()),
     )
     .unwrap()
 });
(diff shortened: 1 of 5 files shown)


volovyks (commented):
Should we collect data for each protocol and then post it once the protocol produces a pre-signature? Doing that for each message can be computationally heavy and harder to process.

ppca:
I'll see why the integration tests are failing. If it is related to metrics taking too much resources, then I'll see how to reduce the load, maybe merging all rounds per protocol together into one metric.

chain-signatures/node/src/http_client.rs
```diff
@@ -177,6 +187,22 @@ impl MessageQueue {
...
+                            .entry(*id)
+                            .and_modify(|ts| {
+                                if *timestamp > *ts {
+                                    *ts = *timestamp;
+                                }
+                            })
+                            .or_insert(*timestamp);
+                    }
+                }
+
```
jakmeier, line 205: I am not sure I'm following the intention here exactly.
I think this calculates the minimum timestamp of messages for each id.

But why? Is this to have a single delay for a batch of messages sent at once? Maybe a short comment could help to make it easier to understand.

chain-signatures/node/src/http_client.rs
```diff
@@ -202,9 +228,23 @@ impl MessageQueue {
                     errors.push(err);
                 } else {
                     compacted += msgs.len();
+                    let send_latency = start.elapsed().as_millis() as f64;
                     crate::metrics::SEND_ENCRYPTED_LATENCY
                         .with_label_values(&[account_id.as_str()])
-                        .observe(start.elapsed().as_millis() as f64);
+                        .observe(send_latency);
+
+                    for (id, msg_timestamp) in presignature_msg_creation_time {
```
jakmeier, line 236: nit: you shadow the `id` variable of the parent loop, which makes it a bit hard to read. Consider renaming to something like participant_id or node_id and protocol_id or msg_id etc.

chain-signatures/node/src/http_client.rs
```diff
@@ -202,9 +228,23 @@ impl MessageQueue {
...
+                        if let Some(client_send_delay) =
+                            crate::util::elapsed_duration_till(msg_timestamp, start)
+                        {
+                            crate::metrics::PRESIGNATURE_CLIENT_SEND_DELAY
+                                .with_label_values(&[account_id.as_str(), &id.to_string()])
+                                .observe(client_send_delay.as_secs_f64());
```
jakmeier, line 242: optional, nit: The result of `with_label_values` can be stored and reused. Looking up a metric is relatively cheap but still expensive enough to have caused significant slowdown at nearcore more than once.

With the id as a label, there is not much you can do. But without it, assuming account_id is our own id and doesn't change, you can look it up once outside the loops and reuse in each iteration.

jakmeier (commented):
A great start! You've captured the relevant metrics exactly, as far as I can tell.

As I explain in comments, I do think this collects excessively many separate time series in Prometheus though. I would also prefer the approach suggested by @volovyks with regards how the data is represented in the metrics database.

ppca:
This PR adds the following metrics for triple, presig and signature protocol: 
- accrued wait time per protocol: add up the wall clock time difference between pokes that return SendMany/SendPrivate/Return
- before poke delay per protocol: time between generator creation and first poke

And add the following for messages:
- client send delay: time between message getting into inbox to message being ready to be sent over wire

chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -604,6 +636,18 @@ impl PresignatureManager {
                         }
                     }
                     Action::SendPrivate(to, data) => {
+                        if !self.poked_latest.contains_key(id) {
+                            presignature_before_poke_delay_metric
+                                .observe(generator.timestamp.elapsed().as_secs_f64());
+                            self.poked_latest
+                                .insert(*id, (Instant::now(), Duration::from_micros(0), 1));
```
ChaoticTempest, line 643: hmm, this looks a bit weird. How come we have `Duration::from_micros(0)`?

chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -164,6 +164,8 @@ pub struct PresignatureManager {
     /// will be maintained for at most presignature timeout period just so messages are
     /// cycled through the system.
     gc: HashMap<PresignatureId, Instant>,
+    /// latest poked time, total acrued wait time and total pokes per presignature protocol
+    poked_latest: HashMap<PresignatureId, (Instant, Duration, u64)>,
```
ChaoticTempest, line 168: why not have this be a part of per `PresignatureGenerator`? Would make it easier to be retrieved vs checking if it exists

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -234,6 +234,8 @@ pub struct SignatureManager {
     failed: VecDeque<(SignRequestIdentifier, GenerationRequest)>,
     /// Set of completed signatures
     completed: HashMap<SignRequestIdentifier, Instant>,
+    /// latest poked time, total acrued wait time and total pokes per signature protocol
+    poked_latest: HashMap<SignRequestIdentifier, (Instant, Duration, u64)>,
```
ChaoticTempest, line 238: same here but can be put inside `SignatureGeneartor` instead

chain-signatures/node/src/protocol/presignature.rs
jakmeier: This is initializing total_wait to 0, right?

I guess `Duration::new(0,0)` or `Duration::from_secs(0)` are slightly nicer alternatives. But I don't think it matters too much tbh.

jakmeier (commented):
Yeah I think this way of counting the metrics makes more sense to me. Thanks for the changes!

chain-signatures/node/src/protocol/presignature.rs
ppca: yes, that's correct.

chain-signatures/node/src/protocol/presignature.rs
ppca: will do

chain-signatures/node/src/protocol/signature.rs
ppca: done

volovyks (dismissed):

chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -558,6 +578,20 @@ impl PresignatureManager {
...
+                                .observe((now - start_time).as_secs_f64());
+                            generator.poked_latest = Some((now, Duration::from_millis(0), 1));
+                        } else {
+                            let (last_poked, total_wait, total_pokes) =
+                                generator.poked_latest.unwrap();
+                            let now = Instant::now();
+                            let elapsed = now - last_poked;
+                            generator.poked_latest =
+                                Some((now, total_wait + elapsed, total_pokes + 1));
+                        }
```
ChaoticTempest, line 594: should probably make this into its own function since its used quite a bit, and also should do a `if let Some(val)` instead of checking on `is_none` especially since we're using the values inside of the option in one of the branches anyways. Also, why are we doing `Instant::now - last_poked()` when we could just do `last_poked.elapsed()`
```suggestion
                        let (total_wait, total_pokes) = if let Some((last_poked, total_wait, total_pokes)) = &generator.poked_latest {
                            (total_wait + last_poked.elapsed(), total_pokes + 1)
                        } else {
                            let start_time = generator.timestamp;
                            presignature_before_poke_delay_metric
                                .observe(start_time.elapsed().as_secs_f64());
                            (Duration::from_millis(0), 1)
                        };
                        let last_poked = Instant::now();
                        generator.poked_latest = Some((last_poked, total_wait, total_pokes));
```

chain-signatures/node/src/metrics.rs
```diff
@@ -494,6 +494,116 @@ pub(crate) static CONFIGURATION_DIGEST: LazyLock<IntGaugeVec> = LazyLock::new(||
     .unwrap()
 });
 
+pub(crate) static PRESIGNATURE_BEFORE_POKE_DELAY: LazyLock<HistogramVec> = LazyLock::new(|| {
+    try_create_histogram_vec(
+        "multichain_presignature_before_poke_delay_sec",
+        "per presignature protocol, delay between generator cration and first poke that returns SendMany/SendPrivate",
+        &["node_account_id"],
+        Some(exponential_buckets(1.0, 1.5, 20).unwrap()),
```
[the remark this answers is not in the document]
ppca: Good catch! I'll change the metric to be in milliseconds, and with exponential_buckets(1.0, 1.5, 25), basically a range from 1  ms to 25 seconds.

chain-signatures/node/src/metrics.rs
```diff
@@ -494,6 +494,116 @@ pub(crate) static CONFIGURATION_DIGEST: LazyLock<IntGaugeVec> = LazyLock::new(||
...
+    )
+    .unwrap()
+});
+
+pub(crate) static PRESIGNATURE_ACCRUED_WAIT_DELAY: LazyLock<HistogramVec> = LazyLock::new(|| {
+    try_create_histogram_vec(
+        "multichain_presignature_accrued_wait_delay_sec",
+        "per presignature protocol, total accrued wait time between each poke that returned SendMany/SendPrivate/Return",
+        &["node_account_id"],
+        Some(exponential_buckets(1.0, 1.5, 20).unwrap()),
```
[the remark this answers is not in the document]
ppca: Sounds good. Changing to milliseconds, and a range of exponential_buckets(10, 1.5, 25), which is 10 ms to ~4 mins.

chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -578,6 +601,23 @@ impl PresignatureManager {
                                 )
                                 .await;
                         }
+                        let (total_wait, total_pokes) =
+                            if let Some((last_poked, total_wait, total_pokes)) =
+                                &generator.poked_latest
+                            {
+                                (
+                                    *total_wait + (generator_poke_time - *last_poked),
```
ChaoticTempest, line 609: this is probably not doing what you're thinking it's doing. `generator_poke_time` is a recorded as `Instant::now` at the beginning of the loop, not how much time is elapsed when you go to do the minus operation. Take a look at this snippet:
```
use std::time::{Duration, Instant};

fn main() {
    let x = Duration::from_millis(200);
    let y = Instant::now();
    let z = y - Duration::from_millis(1000);
    
    println!("{:?}", z);
    std::thread::sleep(x);
    println!("{:?}", y - Duration::from_millis(1000));
}
```

It'll print the same value even though only 200ms has passed.

ppca: yeah so I am doing this because what Jakob suggested was to separate out wait between pokes and the cpu time spent on each poke. That's why I added a SIGNATURE/TRIPLE/PRESIGNATURE_POKE_CPU_TIME metric which measures per poke cpu time, and the accrued wait will just be the total clock time wait between pokes.

chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -580,6 +614,20 @@ impl PresignatureManager {
...
+                            presignature_before_poke_delay_metric
+                                .observe((now - start_time).as_secs_f64());
+                            generator.poked_latest = Some((now, Duration::from_millis(0), 1));
+                        } else {
+                            let (last_poked, total_wait, total_pokes) =
+                                generator.poked_latest.unwrap();
+                            let now = Instant::now();
+                            let elapsed = now - last_poked;
+                            generator.poked_latest =
+                                Some((now, total_wait + elapsed, total_pokes + 1));
```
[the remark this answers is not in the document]
ppca: done

chain-signatures/node/src/protocol/presignature.rs
ChaoticTempest: I'm not sure I get what you're saying. Are you trying to use `generator_poke_time` as a timestamp before poking? Why aren't we just calling `last_poked.elapsed()` before the poke then?

ppca: yep. `generator_poke_time` is the timestamp right before poking. and yeah `generator_poke_time - *last_poked` is essentially the same as calling `last_poked.elapsed()` before the poke. I'm using the `generator_poke_time` so that it can be used to calculate both `delay before first poke = generator_poke_time - generator_created_timestamp` and `wait between this and last poke = generator_poke_time - last_poked`.

ChaoticTempest (approved):

chain-signatures/node/src/protocol/presignature.rs
ppca: yeah that's right! I'll address it in a followup PR

PR #107, opened 2025-01-15
feat: make encryption per partition instead of per message

ChaoticTempest:
- make encryption sent to other nodes be per request/partition now instead of each message, so from `send(Partition = Vec<Ciphered>)` to `send(Partition = Ciphered(Vec<Message>))` is the effective change.
- internalize message decryption to the `MessageInbox` such that when nodes call send to other nodes, there's no wait for the response due to all the decryption that is happening.
- optimize serialization of `SignMessage`. json serialization is very inefficient, especially for byte vectors, so `bincode` is used to get away from all that json overhead. `bincode` is for the most part just raw byte serialization.
- some cleanup of error variants that were no longer useful or being used.


chain-signatures/node/src/protocol/cryptography.rs
```diff
@@ -219,22 +197,20 @@ impl CryptographicProtocol for ResharingState {
...
-                                        data,
-                                    },
-                                )
-                                .await;
-                        }
-                        None => return Err(CryptographicError::UnknownParticipant(to)),
+                    if self.new_participants.get(&to).is_none() {
+                        tracing::error!("resharing: send_private unknown participant {to:?}");
                     }
+                    ctx.channel()
```
ppca, line 203: Does this mean we will send the message even tho it's unknown recipient? Why?

ppca (approved):
LGTM!

chain-signatures/node/src/protocol/cryptography.rs
ChaoticTempest: it doesn't particularly matter because the message channel will just handle it, but would probably save compute to not do send it over so I'll add an else

chain-signatures/node/src/protocol/message.rs
```diff
@@ -666,59 +673,51 @@ impl MessageReceiver for NodeState {
...
+    ) -> Result<T, MessageError> {
+        let msg = cipher_sk
+            .decrypt(&encrypted, Self::ASSOCIATED_DATA)
             .map_err(|err| {
                 tracing::error!(error = ?err, "failed to decrypt message");
-                CryptographicError::Encryption(err.to_string())
+                MessageError::Encryption(err.to_string())
             })?;
-        let SignedMessage::<Vec<u8>> { msg, sig, from } = serde_json::from_slice(&message)?;
+        let Self { msg, sig, from } = bincode::deserialize(&msg).unwrap();
```
jakmeier, line 720: I would be careful with this change.

Unlike JSON, bincode is not self-describing. In other words, it does not encode field information. Instead, you must use the exact same struct for serialization and deserialization. If anything in the serialized type changes, for example you add a new field to a variant of `MpcMessage`, this may break deserialization. This manifests as either an explicit deserialization error or, if it cannot detect a mismatchmaybe, it just silently deserializes invalid data.

For non-distributed systems, this is usually not a problem.
But you have to consider that nodes will not update all at the same time. Hence, there will be times where different versions of a struct like `MpcMessage` are used in the system. You should be able to handle this gracefully.

Typically, one either uses a format that allows dynamic changes, like JSON or protobuf. Then you can add new fields or make old fields optional, without breaking anything. If you use something like bincode, however, it requires manual versioning, which can be a bit painful.

If you are looking for an easy JSON replacement but with binary encoding, take a look at [CBOR](https://en.wikipedia.org/wiki/CBOR), it is supported with serde: https://github.com/enarx/ciborium

Long term, it might make sense to think about using something like protobuf.

jakmeier (commented):
Hey, I noticed you want to introduce bincode on the wire, so I left my thought on it.

(It might also make sense to have the format change in a separate PR from "make encryption per partition instead of per message" .)

chain-signatures/node/src/protocol/message.rs
ChaoticTempest: I see, yeah that's a good point. I'll change it to CBOR in this PR

> (It might also make sense to have the format change in a separate PR from "make encryption per partition instead of per message" .)

the reason why it's a part of this PR is because changing from `Partition = Vec<Encrypted>` to `Partition = Encrypted` breaks some things due to how json serialization works, like how serializing a bundle of messages would more than increase its size up to 10x what it was before. This was easy to check before when we encrypted per message and summed them up before partitioning them, but since partition now happens before encryption, things aren't as well

ChaoticTempest:
Moved to ciborium for cbor encoding. Also added unit tests for serialization changes. So our `Message`s are now future and backwards compatible

chain-signatures/keys/src/hpke.rs
```diff
@@ -37,6 +38,14 @@ pub struct PublicKey(<Kem as hpke::Kem>::PublicKey);
 #[derive(Clone, PartialEq, Eq, Serialize, Deserialize)]
 pub struct SecretKey(<Kem as hpke::Kem>::PrivateKey);
 
+impl fmt::Debug for SecretKey {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("SecretKey")
+            .field("key", &hex::encode(self.to_bytes()))
```
volovyks, line 44: This looks dangerous. Should we take a few characters instead of the whole key?

chain-signatures/node/src/protocol/cryptography.rs
```diff
@@ -136,6 +113,7 @@ impl CryptographicProtocol for GeneratingState {
                         public_key = hex::encode(r.public_key.to_bytes()),
                         "generating: successfully completed key generation"
                     );
+                    // TODO: handle secret storage error
```
volovyks, line 116: What kind of handling you want to add?

chain-signatures/node/src/protocol/message.rs
```diff
@@ -666,175 +721,200 @@ impl MessageReceiver for NodeState {
...
             tracing::warn!(?timeouts, "messages expired");
         }
     }
 
+    /// Compact the messages in the outbox into partitions of at most 256kb.
+    pub fn compact(&mut self) -> HashMap<MessageRoute, Vec<Partition>> {
+        let mut compacted = HashMap::new();
+        for (route, messages) in self.messages.drain() {
+            let entry = compacted.entry(route).or_insert_with(Vec::new);
+            entry.extend(partition_256kb(messages));
```
volovyks, line 874: Have you experimented with the optimal size of the partition? Not sure what is the average size of it in practice, but a quick Google search suggested 1-10Mb.

integration-tests/chain-signatures/tests/cluster/mod.rs
```diff
@@ -1,146 +0,0 @@
...
-use anyhow::Context;
-use integration_tests_chain_signatures::containers::DockerClient;
-use integration_tests_chain_signatures::{utils, NodeConfig, Nodes};
-use near_workspaces::{Account, AccountId, Contract, Worker};
-use url::Url;
-
-use crate::actions::sign::SignAction;
-use crate::actions::wait::WaitAction;
-
-pub fn spawn() -> ClusterSpawner {
```
volovyks, line 23: Are these deletions relevant? Are they intentional?

chain-signatures/keys/src/hpke.rs
ChaoticTempest: We already have a debug for the sign secret key but doesn't hurt to take a few characters off

chain-signatures/node/src/protocol/cryptography.rs
ChaoticTempest: if it doesn't get stored to secret storage, we should store the key in memory and keep retrying

volovyks (approved):

volovyks (approved):

PR #116, opened 2025-01-16
Eth indexer not require start block env variable

ppca:
- this will enable the eth indexer to start from block_number = latest_block from rpc call - 15 if the last_processed_block was empty or too far behind.
- Also added 2 metrics: eth sign request count and eth indexed block height

### DIFF
--- a/chain-signatures/node/src/indexer_eth.rs
+++ b/chain-signatures/node/src/indexer_eth.rs
@@ -330,7 +330,7 @@ pub fn run(
                 let latest_handled_block =
                     context.indexer.last_processed_block().await.unwrap_or(0);
 
-                const MAX_DELAYED_BLOCKS: u64 = 15;
+                const MAX_DELAYED_BLOCKS: u64 = 100;
                 let next_block_to_handle: u64 = {
                     if latest_block - latest_handled_block < MAX_DELAYED_BLOCKS {
                         latest_handled_block + 1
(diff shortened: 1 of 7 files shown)


chain-signatures/node/src/indexer_eth.rs
```diff
@@ -314,26 +319,37 @@ pub fn run(
...
+
+                let latest_handled_block =
+                    context.indexer.last_processed_block().await.unwrap_or(0);
+
+                const MAX_DELAYED_BLOCKS: u64 = 15;
+                let next_block_to_handle: u64 = {
+                    if latest_block - latest_handled_block < MAX_DELAYED_BLOCKS {
+                        latest_handled_block + 1
+                    } else {
+                        latest_block.saturating_sub(MAX_DELAYED_BLOCKS) + 1
```
ailisp, line 338: Great optimization!
Due to current rate limit failure rate, there is a small issue. I notice due to frequent rate limit hit, sometimes it can be 5-10 blocks delayed, so there is a small chance of >15 block delayed. If that happens, some block would be skipped.

Therefore, I suggest doing this check only once, when there is no last handled block.

ppca: I'll increase the delay to 50 blocks. I also want this fix to help the case where a node goes back online after being offline for a while, in that case we don't want the node to start way back.

ailisp: Sounds good!

ailisp (approved):

chain-signatures/node/src/indexer_eth.rs
volovyks: @ppca I think we should apply the same logic (use the same function) that is used in the NEAR Indexer. I do not think there is a need to differentiate,

PR #121, opened 2025-01-17
feat: add NodeId(id) to every log entry for integration-tests

ChaoticTempest:
Makes debugging what is going wrong with a specific node by introducing the `NodeId(id)` appended to every logline in integration tests. These will not appear on the actual prod environments.

For example:
```sh
NodeId(1) 2025-01-17T21:54:40.653693Z  INFO mpc_node::protocol::consensus: started(initializing): starting key generation as a part of the participant set
```

### DIFF
--- a/chain-signatures/node/src/cli.rs
+++ b/chain-signatures/node/src/cli.rs
@@ -5,17 +5,14 @@ use crate::node_client::{self, NodeClient};
 use crate::protocol::message::MessageChannel;
 use crate::protocol::{MpcSignProtocol, SignQueue};
 use crate::storage::app_data_storage;
-use crate::{indexer, indexer_eth, mesh, storage, web};
+use crate::{indexer, indexer_eth, logs, mesh, storage, web};
 use clap::Parser;
 use deadpool_redis::Runtime;
 use local_ip_address::local_ip;
 use near_account_id::AccountId;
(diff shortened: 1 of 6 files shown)


volovyks (commented):
Not sure if it's a good idea ti introduce one more node identifier. We already have accounId and node index from cait-sith library. Should we use one of them instead? Probably accountId, It should be easier to use with tracing since it's available from the start.

ChaoticTempest:
the issue with using `account-id`s is that we're creating dev accounts (`dev-{long-timestamp}.near`). in our integration tests. It's really hard to tell a part which node is which just based on them. The node id is effectively the index from cait-sith as long as there are no weird edge cases when it comes to adding new participants

volovyks:
Yes, in a cryptographic context, it is. I do not mind using it. I am just not sure if there is a convenient way to set it as a parameter in tracing.

ChaoticTempest:
> Yes, in a cryptographic context, it is. I do not mind using it. I am just not sure if there is a convenient way to set it as a parameter in tracing.

hmm, what do you mean? it already gets set by default with this PR in our integration tests. No need for you to write it manually each time

volovyks:
I see. I have not noticed it is optional and can be limited in production. Ok.

volovyks (approved):

PR #129, opened 2025-01-22
eth indexer: subscribe to log events instead of pulling blocks

ppca:
This PR changes the way we get sign requests from ethereum chain. Previously, we pull for latest block every 5 seconds, if there is a request in the block, we process it. The problem with that is that there's always a delay between block becoming available and us pulling for it no matter how frequent we pull.
The approach in this PR is, we subscribe to the logs in the sign request via web socket, and those logs will be pushed to us when they become available. This will minimize delay in previous approach.
There is much less ethereum RPC calls to be made because we essentially have one web socket and that's it.

### DIFF
--- a/chain-signatures/node/src/indexer_eth.rs
+++ b/chain-signatures/node/src/indexer_eth.rs
@@ -5,6 +5,7 @@ use crypto_shared::kdf::derive_epsilon_eth;
 use crypto_shared::ScalarExt;
 use hex::ToHex;
 use k256::Scalar;
+use near_account_id::AccountId;
 use serde::{Deserialize, Serialize};
 use std::str::FromStr;
 use std::time::{Duration, Instant};
@@ -24,7 +25,7 @@ pub struct Options {
     #[clap(long, env("MPC_INDEXER_ETH_RPC_HTTP_URL"))]
     pub eth_rpc_http_url: String,
 
-    /// The contract address to watch
+    /// The contract address to watch without the `0x` prefix
     #[clap(long, env("MPC_INDEXER_ETH_CONTRACT_ADDRESS"))]
     pub eth_contract_address: String,
 }
@@ -146,7 +147,11 @@ fn parse_event(log: &Log) -> anyhow::Result<SignatureRequestedEvent> {
     })
 }
 
-pub async fn run(options: &Options, sign_tx: mpsc::Sender<SignRequest>) -> anyhow::Result<()> {
+pub async fn run(
+    options: &Options,
+    sign_tx: mpsc::Sender<SignRequest>,
+    node_near_account_id: &AccountId,
+) -> anyhow::Result<()> {
     let contract_address = H160::from_str(&options.eth_contract_address)?;
 
     let signature_requested_topic = H256::from_slice(&web3::signing::keccak256(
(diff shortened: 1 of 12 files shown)


integration-tests/src/main.rs
```diff
@@ -22,9 +22,11 @@ enum Cli {
         nodes: usize,
         #[arg(short, long, default_value_t = 2)]
         threshold: usize,
+        #[arg(long, default_value = "ws://localhost:8545")]
+        eth_rpc_ws_url: String,
         #[arg(long, default_value = "http://localhost:8545")]
-        eth_rpc_url: String,
-        #[arg(long, default_value = "0x5FbDB2315678afecb367f032d93F642f64180aa3")]
+        eth_rpc_http_url: String,
+        #[arg(long, default_value = "99bbA657f2BbC93c02D617f8bA121cB8Fc104Acf")]
```
ailisp, line 29: Should we revert this one? Current deploy script is expected to deterministically deploy local contract to 0x5FbDB2315678afecb367f032d93F642f64180aa3

ailisp (approved):
Looks great! A huge enhancement!

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -291,77 +147,71 @@ fn parse_event(log: &Log) -> anyhow::Result<SignatureRequestedEvent> {
...
-                        continue;
+    loop {
+        match web3::transports::WebSocket::new(&options.eth_rpc_ws_url).await {
+            Ok(ws) => {
+                let web3_ws = web3::Web3::new(ws);
+                match web3_ws.eth_subscribe().subscribe_logs(filter.clone()).await {
+                    Ok(mut filtered_logs_sub) => {
+                        tracing::info!("Ethereum indexer connected and listening for logs");
+
+                        let mut heartbeat_interval = tokio::time::interval(Duration::from_secs(30));
```
ChaoticTempest, line 174: maybe every 1min would be better

chain-signatures/node/src/web/mod.rs
```diff
@@ -121,7 +116,6 @@ async fn state(Extension(state): Extension<Arc<AxumState>>) -> Result<Json<State
     tracing::debug!("fetching state");
     // TODO: rename to last_processed_block when making other breaking changes
     let latest_block_height = state.indexer.last_processed_block().await.unwrap_or(0);
-    let latest_eth_block_height = state.eth_indexer.last_processed_block().await.unwrap_or(0);
     let is_stable = state.indexer.is_stable().await;
```
ChaoticTempest, line 119: Also now that we're publishing to multiple chains, the `stable` participants is gonna be funky. We need to refactor our code so that the stable check is checked per chain instead of only near chain. Issue created: https://github.com/sig-net/mpc/issues/137

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -291,77 +147,71 @@ fn parse_event(log: &Log) -> anyhow::Result<SignatureRequestedEvent> {
...
+                                                        tracing::error!(?err, "Failed to send ETH sign request into queue");
+                                                    }
+                                                });
+                                            }
+                                        }
+                                        Err(err) => {
+                                            tracing::warn!("Ethereum log subscription error: {:?}", err);
+                                            break;
+                                        }
+                                    }
```
ChaoticTempest, line 198: this reduces the need for scoping by a bit:
```suggestion
                                    let Ok(log) = log.inspect_err(|err| {
                                        tracing::warn!("Ethereum log subscription error: {:?}", err);
                                    }) else {
                                        break;
                                    };
                                    tracing::info!("Received new Ethereum sign request: {:?}", log);
                                    crate::metrics::NUM_SIGN_REQUESTS_ETH
                                        .with_label_values(&[node_near_account_id.as_str()])
                                        .inc();
                                    if let Ok(sign_request) = sign_request_from_filtered_log(log) {
                                        let sign_tx = sign_tx.clone();
                                        tokio::spawn(async move {
                                            if let Err(err) = sign_tx.send(sign_request).await {
                                                tracing::error!(?err, "Failed to send ETH sign request into queue");
                                            }
                                        });
                                    }
```

ChaoticTempest (approved):
LGTM besides a couple nits

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -291,77 +147,71 @@ fn parse_event(log: &Log) -> anyhow::Result<SignatureRequestedEvent> {
...
 
-                if next_block_to_handle <= latest_block {
-                    if let Err(err) = handle_block(next_block_to_handle, &context).await {
-                        tracing::warn!(%err, "failed to handle eth block");
-                        tokio::time::sleep(Duration::from_secs(5)).await;
-                        continue;
+    loop {
+        match web3::transports::WebSocket::new(&options.eth_rpc_ws_url).await {
+            Ok(ws) => {
+                let web3_ws = web3::Web3::new(ws);
```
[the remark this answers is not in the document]
ppca: yeah since this loop is trying to recreate the socket endlessly if it fails to be created

chain-signatures/node/src/indexer_eth.rs
ppca: this looks much nicer!!

ChaoticTempest (approved):

PR #133, opened 2025-01-23
eth upgradable contract

ailisp:
This version only allows contract owner to upgrade, we may figure out a better permission control such as voting to agree contract upgrade in future. I deployed this proxy contract to 0x6348104D80e1376b59789cA5b65f9a45708A16d6 on sepolia.


chain-signatures/contract-eth/README.md
```diff
@@ -122,3 +104,16 @@ After [this commit](https://github.com/sig-net/mpc/pull/58/commits/f2308fe3c7352
...
+To upgrade the contract, after change to ChainSignatures.sol, you can use the following command:
+```bash
+npx hardhat run scripts/upgradeContract.js --network <network>
+```
+
+It will upgrade the implementation contract to the new version and keep the proxy contract address the same. Make sure you always extends the storage and never override it. For more details, please refer to [Writing Upgradeable Contracts](https://docs.openzeppelin.com/upgrades-plugins/writing-upgradeable).
+
+## Update the public key
+
+To update the public key, you also need to update the contract. Add a `reinitialize` function to the contract that set the public key to new one, and upgrade the contract as above.
```
ppca, line 119: what do you mean by ```Add a `reinitialize` function to the contract that set the public key to new one```?

chain-signatures/contract-eth/scripts/upgradeContract.js
```diff
@@ -0,0 +1,57 @@
...
+const fs = require('fs');
+
+async function main() {
+  // Get network info
+  const network = await ethers.provider.getNetwork();
+  const networkName = network.name;
+  const deploymentPath = `deployments/${networkName}.json`;
+
+  // Check if deployment file exists
+  if (!fs.existsSync(deploymentPath)) {
```
ppca, line 11: This relies on the local deployment file? Does this mean we need to commit that file to git every time we do initialDeploy or updateContract?

chain-signatures/contract-eth/contracts/ChainSignatures.sol
```diff
@@ -48,7 +47,10 @@ contract ChainSignatures {
     event SignatureRequested(bytes32 indexed requestId, address requester, uint256 epsilon, uint256 payloadHash, string path);
     event SignatureResponded(bytes32 indexed requestId, SignatureResponse response);
 
-    constructor(PublicKey memory _publicKey) {
+    /// @custom:oz-upgrades-unsafe-allow constructor
+    constructor() initializer {}
+
+    function initialize(PublicKey memory _publicKey) public initializer {
         publicKey = _publicKey;
     }
```
ppca, line 56: Do we need to add storage gap or is it handled by the openzepplin already? https://blog.openzeppelin.com/validate-smart-contract-storage-gaps-with-openzeppelin-upgrades-plugins

chain-signatures/contract-eth/README.md
```diff
@@ -122,3 +104,16 @@ After [this commit](https://github.com/sig-net/mpc/pull/58/commits/f2308fe3c7352
...
 There are certain small gas optimizations possible to reduce a little more gas, but at this stage I think it is not a priority.
+
+## Upgrade contract
+
+To upgrade the contract, after change to ChainSignatures.sol, you can use the following command:
+```bash
+npx hardhat run scripts/upgradeContract.js --network <network>
+```
+
+It will upgrade the implementation contract to the new version and keep the proxy contract address the same. Make sure you always extends the storage and never override it. For more details, please refer to [Writing Upgradeable Contracts](https://docs.openzeppelin.com/upgrades-plugins/writing-upgradeable).
```
ppca, line 115: I remember seeing that the order of declaring variables also can't change when we modify the contract? Does openzepplin warn against it?

ppca: ah, it does warn!
```
Upgrading implementation...
Error: New storage layout is incompatible

ChainSignatures: Deleted `threshold`
  > Keep the variable even if unused

contracts/ChainSignatures.sol:43: Inserted `threshold`
  > New variables should be placed after all existing inherited variables

contracts/ChainSignatures.sol:44: Inserted `hihi`
  > New variables should be placed after all existing inherited variables
```

chain-signatures/contract-eth/README.md
ailisp: I'll add an example in unit test

chain-signatures/contract-eth/contracts/ChainSignatures.sol
ailisp: Thanks for point out. I didn't know this. After reading this I think we don't need it, we can append new variables as long as we keep existing variables.

chain-signatures/contract-eth/scripts/upgradeContract.js
ailisp: done

chain-signatures/contract-eth/README.md
ailisp: done

volovyks (approved):

PR #139, opened 2025-01-27
Update protocol default configs

volovyks:
Let's set the default number of concurrent protocols to a higher number. This should stress test our multithreading design and messaging layer.
At first, I wanted to separate the T and P parameters, but this made little sense while introducing the breaking change.

The number of signature protocols remains unconstrained. Limits for T and P protocols are separate (each node can introduce 512 T and 512 P protocols and as a result process 1024*12(number of nodes in the system) parallel protocols)
cc @auto-mausx 

@ChaoticTempest ~12,000 Tokyo tasks seem reasonable since they are mostly IO-bound. We can adjust these numbers after we have more metrics. For now higher number will give us better data.

### DIFF
--- a/chain-signatures/contract/src/config/impls.rs
+++ b/chain-signatures/contract/src/config/impls.rs
@@ -29,8 +29,8 @@ impl Default for ProtocolConfig {
         Self {
             message_timeout: min_to_ms(5),
             garbage_timeout: hours_to_ms(2),
-            max_concurrent_introduction: 2,
-            max_concurrent_generation: 2 * MAX_EXPECTED_PARTICIPANTS,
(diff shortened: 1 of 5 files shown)


volovyks:
Limiting on the contract level is more challenging since we will receive requests from many chains. For now, I'm assuming we are getting everything from NEAR.

cc @esaminu @Pessina please, check if such a price function makes sense. I believe it is more transparent.

volovyks:
@ChaoticTempest I've updated the limits, let's merge this. I believe we are hitting the contract limit now. And the system is proven to work fine with 16.

ChaoticTempest (approved):

PR #145, opened 2025-01-28
Fix load tests

volovyks:


volovyks:
@ChaoticTempest we can try integrating these load tests with benchmarking on CI. We can have multiple scenarios with different loads, spikes, etc. Also, it generates an HTML report.

ChaoticTempest (approved):

PR #149, opened 2025-01-29
chore: add separate metrics for sign respond and sign generation

ChaoticTempest:
Add additional metrics for signing:
- signature generation latency: time taken between request received and signature generation completed.
- respond latency: time it takes to publish the signature. 
- logging now has the elapsed times as well for signatures.

We can see how long signature generation takes a bit more precisely with this

### DIFF
--- a/chain-signatures/node/src/metrics.rs
+++ b/chain-signatures/node/src/metrics.rs
@@ -41,7 +41,7 @@ pub(crate) static NUM_SIGN_SUCCESS: Lazy<CounterVec> = Lazy::new(|| {
     .unwrap()
 });
 
-pub(crate) static SIGN_LATENCY: Lazy<HistogramVec> = Lazy::new(|| {
+pub(crate) static SIGN_TOTAL_LATENCY: Lazy<HistogramVec> = Lazy::new(|| {
     try_create_histogram_vec(
         "multichain_sign_latency_sec",
         "Latency of multichain signing, start from indexing sign request, end when publish() called.",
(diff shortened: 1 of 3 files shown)


chain-signatures/node/src/metrics.rs
```diff
@@ -51,6 +51,26 @@ pub(crate) static SIGN_LATENCY: Lazy<HistogramVec> = Lazy::new(|| {
     .unwrap()
 });
 
+pub(crate) static SIGN_GENERATION_LATENCY: Lazy<HistogramVec> = Lazy::new(|| {
+    try_create_histogram_vec(
+        "multichain_sign_gen_latency_sec",
+        "Latency of multichain signing, from received sign request to signature generation complete.",
```
volovyks, line 57: Is it from receiving or from starting the protocol?

chain-signatures/node/src/rpc.rs
```diff
@@ -499,6 +518,7 @@ async fn try_publish_eth(
             tracing::info!(
                 request_id = ?CryptoHash(*request_id),
                 tx_hash = ?tx_hash,
+                elapsed = ?timestamp.elapsed(),
```
volovyks, line 521: Don't we want latency for Ethereum?

chain-signatures/node/src/rpc.rs
```diff
@@ -432,6 +447,9 @@ async fn try_publish_near(
     crate::metrics::SIGN_LATENCY
```
volovyks, line 447: Since we have many sign metrics, can we rename this SIGN_LATENCY_TOTAL or something similar?

chain-signatures/node/src/rpc.rs
ChaoticTempest: that's in a TODO. I'm leaving it up to @ppca when she gets back

chain-signatures/node/src/metrics.rs
ChaoticTempest: start of the protocol

volovyks: Let's modify the description then, these metrics are becoming hard to understand

ChaoticTempest: done

volovyks (approved):

PR #150, opened 2025-01-29
Release 1.0.0

volovyks:
merge after https://github.com/sig-net/mpc/pull/139


CHANGELOG.md
```diff
@@ -1,8 +0,0 @@
-# Changelog
```
volovyks, line 1: We should either use it or remove it. I'm planning to keep the changelog in our release notes.

chain-signatures/test-contracts/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mpc-test-contract"
-version = "0.1.0"
+version = "1.0.0"
```
ChaoticTempest, line 3: test-contract should not be updated for versions since it's never used in the node

chain-signatures/contract/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mpc-contract"
-version = "1.0.0-rc.5"
+version = "1.0.0"
```
ChaoticTempest, line 3: are we planning on locking the contract yet? If not, we should still have it be in RC

chain-signatures/keys/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "mpc-keys"
-version = "0.1.0"
+version = "1.0.0"
```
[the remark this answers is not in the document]
ChaoticTempest: just do it on the workspace level:
```
# chain-signatures/Cargo.toml
[workspace.package]
version = "1.0.0"

# chain-signatures/keys/Cargo.toml
[package]
name = "mpc-keys"
version.workspace = true
```

auto-mausx:
I'm gonna add some infra changes for machine type for this PR

CHANGELOG.md
ChaoticTempest: yeah we don't need the changelog. The releases page can auto generate the diff for us

chain-signatures/contract/Cargo.toml
volovyks: We need to do that someday. We can always change the major version.

ChaoticTempest (dismissed):

auto-mausx (approved):

PR #151, opened 2025-01-29
chore: make eth chain optional

ChaoticTempest:
makes the eth chain optional from the CLI


volovyks (approved):
Nice, thank you!
@auto-mausx it's a breaking change

auto-mausx:
What ENV vars will change for the container? Right now I don't have any ETH variables defined in TF for partner mainnet.

ChaoticTempest:
there's no need to pass anything. What we have for terraform should just work as-is

PR #152, opened 2025-01-30
doc: fix typos in API docs

jakmeier:
The correct spelling of the function is `experimental_signature_deposit`. The code has it spelled correctly. But in the docs it was `experimantal_signature_deposit`.

Also, the link to the source code used the wrong path, with the directory name `chain-signatures/contracts` instead of `chain-signatures/contract`.


volovyks (approved):
Thank you!

PR #154, opened 2025-01-30
doc: add diagrams to architecture overview

jakmeier:
I think adding these two diagrams here would make sense, for anyone who might be peeking at what's going on in the repo.

No rush to merge it in, though.

Let me know if I should change any specific terms for the public docs.


ChaoticTempest (approved):
Makes sense to me, thanks for adding in the diagrams!

volovyks (approved):
Looks good. We can change it as we go.

PR #157, opened 2025-01-31
feat: integration tests prestockpile

ChaoticTempest:
This is a decently big PR for prestockpiling triples/presignatures before a test gets ran. This is mainly needed to have more consistent benchmarks.

The following was changed in this PR:
- moved some library code from integration tests folder to integration lib.
- `cluster::spawn` by default will prestockpile based on the configuration set.
- `disable_prestockpile` can be used to disable it.
- had to move triple storage and presignature storage clear calls to right after resharing is done due to it clearing the storage right after generating keys.


integration-tests/tests/cases/mod.rs
```diff
@@ -257,22 +249,19 @@ async fn test_triple_persistence() -> anyhow::Result<()> {
...
-    docker_client.create_network(docker_network).await?;
-    let redis = containers::Redis::run(&docker_client, docker_network).await;
-    let redis_url = Url::parse(redis.internal_address.as_str())?;
-    let redis_cfg = deadpool_redis::Config::from_url(redis_url);
-    let redis_pool = redis_cfg.create_pool(Some(Runtime::Tokio1)).unwrap();
-    let presignature_storage = storage::presignature_storage::init(
-        &redis_pool,
-        &AccountId::from_str("test.near").unwrap(),
-    );
+    let spawner = ClusterSpawner::default()
```
volovyks, line 252: What is the benefit of spawning a cluster here?

integration-tests/src/containers.rs
```diff
@@ -650,4 +662,86 @@ impl Redis {
...
+
+    pub fn triple_storage(&self, id: &AccountId) -> mpc_node::storage::TripleStorage {
+        mpc_node::storage::triple_storage::init(&self.pool(), id)
+    }
+
+    pub fn presignature_storage(&self, id: &AccountId) -> mpc_node::storage::PresignatureStorage {
+        mpc_node::storage::presignature_storage::init(&self.pool(), id)
+    }
+
+    pub async fn stockpile_triples(&self, cfg: &NodeConfig, participants: &Participants, mul: u32) {
```
volovyks, line 682: I was expecting a storage solution. We can generate N trioles and pre-signatures and reuse them constantly. That will require hardcoding participant set, keys, etc. Not sure what is easier.

volovyks (commented):
Is the intention here to generate N triples and pre-signatures in advance and measure only signature generation?

integration-tests/tests/cases/mod.rs
ChaoticTempest: ClusterSpawner doesn't actually spawn a cluster here. It's used as a way to configure the env. In this case, it initializes the network for Redis. So wherever you `ClusterSpawner::default()`, it is configuration, not spawning vs `cluster::spawn`

integration-tests/src/containers.rs
ChaoticTempest: Not sure what you mean by storage solution here? It's storing N*mul triples with this function per node.

Hardcoding is a whole slew of work there. The reason it's done this way is because presignatures are dependent on the initial key, which would mean we would then need to inject the key into the contract and each node would also need to intake their own share of the key.

ChaoticTempest:
> Is the intention here to generate N triples and pre-signatures in advance and measure only signature generation?

yes, mostly to measure how much signatures we can produce

integration-tests/tests/cases/mod.rs
ChaoticTempest: the spawner itself is the configuration though. It's only a couple places where we use it like this, so not worth it to clean it up right now

volovyks (approved):

PR #167, opened 2025-02-05
OTLP setup

volovyks:
image

Tracing with OTLP lets us collect structured, time-based data and send them to Jaeger or other visualization framework. Works both locally and on the deployed env. It will help us to find narrow places, get more insights about latency of protocols and signature requests.


integration-tests/README.md
```diff
@@ -26,6 +26,35 @@ cargo test -p integration-tests --jobs 1 -- --test-threads 1
...
+
+### FMT
+This is the default logging format used for local development. It outputs logs to the console in a human-readable format.
+
+### OTLP
+OpenTelemetry Protocol is used for exporting logs to an OpenTelemetry collector. This is useful for integrating with various observability backends.
+
+#### Configuring Local OTLP Logging
+1. **Setup OTLP Backend**
+Start OTLP backend of your choiceю For example [Jaeger](https://www.jaegertracing.io/docs/getting-started/).
```
jakmeier, line 40: What's the ю doing here? :smile:

integration-tests/README.md
```diff
@@ -26,6 +26,35 @@ cargo test -p integration-tests --jobs 1 -- --test-threads 1
 cargo test -p integration-tests --features docker-test
 ```
 
+## Logging and Tracing
+We have three types of logging available:
+
```
jakmeier, line 31: Very nice to include docs! Is there a reason why it is in `integration-tests/README.md`? Is this only meant for integration tests?

jakmeier (approved):
Nice, looking forward to gaining more insights into performance bottlenecks from these traces!

integration-tests/README.md
volovyks: Hiding from choice :)

integration-tests/README.md
volovyks: Not only. Our top level README is minimalistic, I did not wanted to add Logging sections there.
But it is a good point. I've added `Contributing` ling to the main Readme and now it points to this doc.

jakmeier (dismissed):

chain-signatures/node/src/logs.rs
```diff
@@ -61,22 +148,95 @@ where
...
+        .with_ansi(atty::is(atty::Stream::Stderr))
+        .with_line_number(true)
+        .with_thread_names(true)
+        .event_format(NodeIdFormatter::new(node_id))
+        .with_filter(EnvFilter::from_default_env());
+
+    let tracer_otlp_provider = init_otlp_traces(env, node_id, options.otlp_endpoint.as_str()).await;
+    let tracer_otlp = tracer_otlp_provider.tracer("mpc");
+
+    if is_running_on_gcp().await {
```
ChaoticTempest, line 210: since debug id is no longer specified from integration tests, can you add in a flag in Options that is specified by integration tests to disable GCP logging. Some of us are using GCP instances for development so integration tests should disable GCP related logging.

ChaoticTempest (commented):
overall looks good to me, but the tests seem to be failing right now on a couple cases. Probably related to what you were seeing with ContractStateRollback, so something isn't right with the changes

Cargo.toml
```diff
@@ -38,8 +38,13 @@ serde_json = "1"
 sha3 = "0.10.8"
 thiserror = "1"
 tokio = { version = "1.28", features = ["full"] }
-tracing = "0.1.41"
-tracing-subscriber = { version = "0.3.19", features = ["env-filter"] }
+tracing = "0.1.35"
+tracing-subscriber = { version = "0.3.0", default-features = false, features = [
```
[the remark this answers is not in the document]
volovyks: It is really hard to find a combination of versions that works.

chain-signatures/node/src/logs.rs
volovyks: It is a layered system. Presence of the GCP layer should not affect your experience.

auto-mausx (dismissed):

auto-mausx:
Great job tackling this, hopefully we can gain some insight and dive deeper into telemetry as we use this.

chain-signatures/node/src/logs.rs
ChaoticTempest: It does because it's a layered. All logs lose their colored formatting and are captured within `{}` when outputted within a GCP environment. GCP log manager will parse it but not so much for our terminals which is a terrible experience for development on top of GCP VMs

volovyks:
@ChaoticTempest the tests were failing because of the static node account_ids. We were reading them from previous test run instead of starting key generation protocol.
BTW, it may be a good idea to hard code keys and skip key generation in all tests except the one that is actually testing key generation.

ChaoticTempest:
> BTW, it may be a good idea to hard code keys and skip key generation in all tests except the one that is actually testing key generation.

There are some issues with hardcoding the key:
- we need to know the number of nodes ahead of time, since each share would have to be known, and we can't just generate more shares without going through resharing/keygen all over again. So this will break if we decide to test more nodes
- some work is needed to do the hardcoding into our persistent DB since that isn't easily exposed in our integration tests right now

PR #168, opened 2025-02-05
feat: add benchmarking e2e sign requests and generation

ChaoticTempest:
This adds in e2e sign request benchmarking for local benchmarking.

- Metrics based benchmarking by pulling in metrics from a node.
- new feature flag `bench` needed to do this so we can get exact measurements
- ~~note integration build script is janky with `cargo bench`, so follow the instructions in the integration-tests/README.md to run benchmarks.~~

integrating into CI will be left for another time in another PR. Someone else can do this part if they want to


chain-signatures/node/src/web/mod.rs
```diff
@@ -47,8 +47,13 @@ pub async fn run(
         )
         .route("/msg", post(msg))
         .route("/state", get(state))
-        .route("/metrics", get(metrics))
-        .layer(Extension(Arc::new(axum_state)));
+        .route("/metrics", get(metrics));
+
+    if cfg!(feature = "bench") {
```
volovyks, line 52: Since we are exposing a new bench URL, should we simplify this and remove the feature flag usage?

integration-tests/benches/sign.rs
```diff
@@ -0,0 +1,239 @@
...
+fn main() {
+    let rt = tokio::runtime::Builder::new_multi_thread()
+        .enable_all()
+        .build()
+        .unwrap();
+    let started = Instant::now();
+    let (nodes, accounts) = rt.block_on(async {
+        let nodes = cluster::spawn()
+            .with_config(|cfg| {
+                cfg.protocol.triple.min_triples = SIGNATURE_AMOUNT as u32 * 4;
```
volovyks, line 81: How do we control triple pre-generation?

volovyks (commented):
Nice work!

Nits:
- `bench` recompile everything each time I run it
- bench process does not stop by itself, I'm stopping it with Ctrl+C (I've waited for ~10m)

The goal for benchmarks is to integrate them into our workflow. All MPC engineers should know how to use, update, and extend them. It would be great to add a reference to the library we use, a bit more comments on what and how we measure in code and in README, etc.

chain-signatures/node/src/cli.rs
```diff
@@ -280,9 +280,9 @@ pub fn run(cmd: Cli) -> anyhow::Result<()> {
                 mesh_handle.await??;
                 protocol_handle.await??;
                 web_handle.await??;
-                eth_indexer_handle.await??;
                 tracing::info!("spinning down");
 
+                eth_indexer_handle.abort();
```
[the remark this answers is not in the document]
ChaoticTempest: awaiting the handle just waits on it forever and doesn't cleanly exit because the indexer will never cease back control, so we have to abort it like the near indexer

integration-tests/benches/sign.rs
ChaoticTempest: just setting the stockpile amount or calling `cluster::spawn().prestockpile(multiplier)` where `multiplier` creates a multiple of the triple stockpile

ChaoticTempest:
> bench recompile everything each time I run it
> bench process does not stop by itself, I'm stopping it with Ctrl+C (I've waited for ~10m)
> The goal for benchmarks is to integrate them into our workflow. All MPC engineers should know how to use, update, and extend them. It would be great to add a reference to the library we use, a bit more comments on what and how we measure in code and in README, etc.

hmm, I guess it doesn't work as well on macos but on linux it does not stall. Haven't had time to work out what's wrong the recompilation step. I'm also adding this as-is just so people can use it until I work out all the issues

volovyks (approved):
LGTM
Please, do the follow-up PR with the docs

PR #169, opened 2025-02-06
audit: disable all initialization for implementation contract in constructor and add keyVersion variable

ppca:
per audit

### DIFF
--- a/chain-signatures/contract-eth/contracts/ChainSignatures.sol
+++ b/chain-signatures/contract-eth/contracts/ChainSignatures.sol
@@ -44,14 +44,24 @@ contract ChainSignatures is Initializable {
 
     mapping(bytes32 => uint256) public depositToRefund;
 
+    uint32 public keyVersion;
+
     event SignatureRequested(bytes32 indexed requestId, address requester, uint256 epsilon, uint256 payloadHash, string path);
     event SignatureResponded(bytes32 indexed requestId, SignatureResponse response);
 
     /// @custom:oz-upgrades-unsafe-allow constructor
-    constructor() initializer {}
+    constructor() {
+        _disableInitializers();
+    }
 
-    function initialize(PublicKey memory _publicKey) public initializer {
+    function initialize(PublicKey memory _publicKey, uint32 _keyVersion) public initializer {
         publicKey = _publicKey;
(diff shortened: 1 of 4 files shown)


chain-signatures/contract-eth/contracts/ChainSignatures.sol
```diff
@@ -44,14 +44,24 @@ contract ChainSignatures is Initializable {
...
     /// @custom:oz-upgrades-unsafe-allow constructor
-    constructor() initializer {}
+    constructor() {
+        _disableInitializers();
+    }
+
+    function initialize(PublicKey memory _publicKey, uint32 _keyVersion) public initializer {
+        publicKey = _publicKey;
+        keyVersion = _keyVersion;
+    }
```
ChaoticTempest, line 60: it's more gas efficient to use calldata since it is immutable and does not get allocated in heap
```suggestion
    function initialize(PublicKey calldata _publicKey, uint32 _keyVersion) public initializer {
        publicKey = _publicKey;
        keyVersion = _keyVersion;
    }
```

chain-signatures/contract-eth/contracts/ChainSignatures.sol
```diff
@@ -44,14 +44,24 @@ contract ChainSignatures is Initializable {
...
 
-    function initialize(PublicKey memory _publicKey) public initializer {
+    function upgradeToV2(PublicKey memory _publicKey, uint32 _keyVersion) public reinitializer(2) {
```
ChaoticTempest, line 62: hmm, why is this `upgradeToV2` and not just `upgrade`? Also `memory` here can be `calldata` too

ppca: was hoping to flag this is the upgrade function that actually initialize the contract to version 2.

volovyks: I see that this is a common practice with such naming, but I do not understand the flow.

volovyks: Why do we need this function in the first place?

volovyks:
@ppca I believe this one is outdated

PR #171, opened 2025-02-06
feat: make MessageOutbox::send packet their own task

ChaoticTempest:
This makes all `send` their own separate tasks such that we don't wait for all of them to complete before moving on. This allows for the message executor to process the next set of messages incoming and outcoming without being blocked


volovyks:
Let's wait for benchmarking PR to be merged and see how this one performs.

chain-signatures/node/src/protocol/message.rs
```diff
@@ -964,70 +937,33 @@ impl MessageOutbox {
...
-                Ok(Err((route, partition, err))) => {
-                    // since we failed, put back all the messages related to this
-                    retry.push_back((route, partition));
-                    errors.push(err);
-                }
-                Err(err) => {
-                    tracing::warn!(?err, "outbox: task failed to send message");
-                }
+                tokio::spawn(timeout(
+                    Duration::from_millis(cfg.message_timeout),
```
jakmeier, line 941: Optional suggestion: Consider logging and/or collecting metrics for messages we drop afer the tokio task gets cancelled after `Duration::from_millis(cfg.message_timeout)`. Something equivalent to the previous `tracing::warn!(?timeouts, "messages expired");` in the deleted `fn expire()`.

At the moment, it seems we only log warnings on each retry iteration but we don't see anywhere if it eventually succeeded or if it failed in the end.

chain-signatures/node/src/protocol/message.rs
```diff
@@ -964,70 +937,33 @@ impl MessageOutbox {
...
-        // Add back the failed attempts for next time.
-        for (route, partition) in retry {
-            let entry = self.messages.entry(route).or_default();
-            entry.extend(
-                partition
-                    .messages
-                    .into_iter()
-                    .zip(partition.timestamps.into_iter()),
-            );
-        }
```
jakmeier, line 1030: nit: It looks like retrying later is no longer a thing, right? Would be nice to mention that in the commit message, if that's indeed the intention of this change.

jakmeier (commented):
Love the simplification here!
Just two small things to point out, see the review comments.

volovyks:
@ChaoticTempest let's get back to this. The only thing (except the conflict) that is blocking is the benchmark result.

ChaoticTempest:
@volovyks the benchmarks won't show us anything new with this one, since it only benches signature generation as a whole. With this it'll only show a light variation since this wasn't a bottleneck, just a nice thing to do to not block the message channel task from continuing

volovyks (dismissed):

ppca (approved):

ChaoticTempest:
closing in favor of https://github.com/sig-net/mpc/pull/458

PR #172, opened 2025-02-06
chore: use workspace dependencies & rename crypto-shared to mpc-crypto

ChaoticTempest:
Moved common dependencies into `[workspace.dependencies]` so we can utilize the same version across the workspace. This will give us some sanity when modifying multiple packages with different versions.

Also, since the `crypto-shared` package's name was not consistent, I renamed it to `mpc-crypto` to align with the other packages.


volovyks (commented):
Good move
I will remove load tests not to waste time on them

Issue #158, opened 2025-02-03
Decouple Node from Integrations


volovyks:
### Description
Currently, we do not have a clear separation between the node code and Indexers. Since we are planning to add support for many more blockchains, such coupling can slow us down. Here’s my proposition:

1. **Introduce the concept of `Integration`**: 
   Instead of adding new indexers and new functions for publishing signatures, a new integration should:
   - Listen for new signature requests and [add them to the queue](https://github.com/sig-net/mpc/issues/79).
   - Listen for a node to publish a response.
   - Publish the result on the target chain. This approach may also make it easier to publish responses on a different chain from where the request originated.

2. **Separate or remove block height from `is_stable`**:
   - This will allow us to completely decouple node code from integration code.
   - We can check this inside the integration itself, but I would rather skip this step and take an "optimistic" approach. This should be an acceptable solution if we implement Step 2 from @jakmeier's [document](https://docs.google.com/document/d/1rU0diFhHGvviKetw9XVnqGlDnX4TlJC5xKXl4OeBhGw/edit?pli=1&tab=t.0#heading=h.viie31rfopdq) (checking if the signature request was indexed on each node before the start of the protocol using the akc-nack approach).

3. **Proposed Project Structure**:
   - Such an approach should allow us to achieve a project structure similar to the one proposed by @ChaoticTempest in [this issue](https://github.com/sig-net/mpc/issues/156).

Let’s discuss this further. cc @jakmeier @ChaoticTempest @ppca

ChaoticTempest:
nit: I wouldn't call them integration because that's a bit too general. I would much rather us just start with calling them `Chain` as an interface. If things evolve to the point where we're integrating something that is not a chain, then we can then have a higher level interface `Integration` or something of that nature.

IMO, the basic parts of what we need from each chain should just encompass the following:
```rs
trait Chain {
    fn publish(&self, signature) -> Result<...>;

    // or on initialization of chain, we can have the channel for us already.
    fn recv(&self) -> SignatureRequest;

    fn is_stable(&self) -> ChainInfo;

    // not sure if this one would be good to have, maybe useful for tracking stable.
    fn state(&self) -> ChainState;
}
```

We don't need to bake in consensus interfaces here like voting, as this should purely be for listening for sign requests and publishing.

volovyks:
OK, the Chain name should be fine.

@ChaoticTempest, the proposed interface requires the Node to know about all supported chains, call publish, is_stable, etc. My proposition is as follows:
[...]

PR #176, opened 2025-02-06
feat: poc of a request queue in redis

jakmeier:
This commit shows how a redis list can be used to decouple the sign request
producers (indexers) from the rest of the MPC node.

This is essentially just ripping out a mpsc::channel() with a redis FIFO queue.
Most code changes are just replacing the corresponding sender / receiver to a
redis pool.

The biggest change in this is that we have to serialize `SignRequest` to queue
it. This caused a conflict with the time_added field on the request. Using
`Instant` only works for relative timing within a single process and it is
therefore not designed to be serialized. Replacing it with `SystemTime` works
opens up problems with system time changes, in which case the `.elapsed()` time
could fail to work or return invalid data. I think this is fine, as we only use
the time for metrics. Thus, I use `unwrap_or_default` to convert failures in
time measuring to 0 ms duration.

### DIFF
--- a/chain-signatures/node/src/indexer.rs
+++ b/chain-signatures/node/src/indexer.rs
@@ -1,6 +1,7 @@
 use crate::protocol::Chain::NEAR;
 use crate::protocol::{Chain, SignRequest};
 use crate::storage::app_data_storage::AppDataStorage;
+use crate::storage::request_queue::push_sign_request;
 use crypto_shared::{derive_epsilon, ScalarExt};
 use k256::Scalar;
 use near_account_id::AccountId;
@@ -13,8 +14,8 @@ use serde::{Deserialize, Serialize};
 use std::ops::Mul;
 use std::sync::Arc;
 use std::thread::JoinHandle;
-use std::time::{Duration, Instant};
-use tokio::sync::{mpsc, RwLock};
+use std::time::{Duration, Instant, SystemTime};
+use tokio::sync::RwLock;
 
 /// Configures indexer.
 #[derive(Debug, Clone, clap::Parser)]
@@ -172,7 +173,7 @@ impl Indexer {
 struct Context {
     mpc_contract_id: AccountId,
     node_account_id: AccountId,
-    sign_tx: mpsc::Sender<SignRequest>,
+    redis_pool: deadpool_redis::Pool,
     indexer: Indexer,
 }
 
@@ -255,7 +256,7 @@ async fn handle_block(
                     epsilon,
                     entropy,
                     // TODO: use indexer timestamp instead.
-                    time_added: Instant::now(),
+                    time_added: SystemTime::now(),
                 });
             }
         }
@@ -278,7 +279,7 @@ async fn handle_block(
             entropy = hex::encode(request.entropy),
             "new sign request"
         );
-        if let Err(err) = ctx.sign_tx.send(request).await {
+        if let Err(err) = push_sign_request(&ctx.redis_pool, request).await {
             tracing::error!(?err, "failed to send the sign request into sign queue");
         }
         crate::metrics::NUM_SIGN_REQUESTS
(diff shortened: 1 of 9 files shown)


jakmeier:
The discussion on the design is still ongoing in https://github.com/sig-net/mpc/issues/158

With this PoC PR I want to show that this kind of design isn't really all that much work. Thanks to the work done by the team previously (putting things in different tasks) it is quite easy to decouple it even further.
I hope this helps to progress the discussion.

But to be clear: This is just a proof of concept. I'm not submitting this to be merged just yet. We should first decide on the decoupled design. If we want to go that way, we will have to go all the way. Which means: 
1. Create a redis queue the signature responses to be published
2. Replace the `is_stable` check to be performed on redis values, rather than accessing the indexer directly

volovyks:
1. agree
2. `is_stable` was turned off during the Ethereum launch since it was checking the NEAR indexer only and we were unsure if we wanted to add that to the Ethereum indexer.

chain-signatures/node/src/indexer.rs
```diff
@@ -255,7 +256,7 @@ async fn handle_block(
                     epsilon,
```
volovyks, line 256: TODO: since the goal is to decouple Indexers from the Node, we can move the epsilon calculation to the node and avoid executing this logic in multiple Indexer implementations.

chain-signatures/node/src/storage/request_queue.rs
```diff
@@ -0,0 +1,52 @@
...
+    // timeout = 0.0 will block indefinitely
+    let maybe_request_cbor: Option<Vec<u8>> = conn.brpop(QUEUE_KEY, 0.0).await?;
+
+    if let Some(request_cbor) = maybe_request_cbor {
+        let cursor = std::io::Cursor::new(request_cbor);
+
+        let request = ciborium::de::from_reader(cursor)?;
+        Ok(request)
+    } else {
+        // this is against the redis documentation
```
volovyks, line 48: TODO: add`tracing:warn` or error if we do not expect this behavior.

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -76,16 +77,14 @@ impl SignQueue {
...
-                    err
-                }
-                other => other,
-            }
-        } {
+        loop {
+            let Ok(request) = pop_sign_request(&self.redis_pool).await.inspect_err(|err| {
+                tracing::error!("Failed to read sign request from queue: {err}")
+            }) else {
+                // We cannot handle this request but we can hope the next request works.
```
volovyks, line 84: TODO: add tracing

chain-signatures/node/src/storage/request_queue.rs
```diff
@@ -0,0 +1,52 @@
...
+
+/// Take a signature request from the head of the FIFO queue.
+///
+/// Blocks until a request is available, asynchronously waiting for a
+/// notification by redis.
+pub async fn pop_sign_request(redis_pool: &deadpool_redis::Pool) -> anyhow::Result<SignRequest> {
+    let mut conn = redis_pool.get().await?;
+    // Using a redis list as a FIFO queue, pushing on the left and popping on the right.
+    // timeout = 0.0 will block indefinitely
+    let maybe_request_cbor: Option<Vec<u8>> = conn.brpop(QUEUE_KEY, 0.0).await?;
```
volovyks, line 40: Am I right that in practice we should not expect it to be infinite? Communication with Redis is done over the wire, if there are no signature requests for ~1H, it will fail.

volovyks (commented):
@jakmeier looks good!
@ChaoticTempest please, take a look, we need to make a design decision and move forward with this.

chain-signatures/node/src/storage/request_queue.rs
jakmeier: This isn't a timeout for the network request but rather on the redis side.

It will block in case of an empty queue, until a request is available to be popped. If there is none for 1h, the task does nothing for 1h. It won't ever fail, unless the connection to redis is lost.

ChaoticTempest: hmm, so this blocks the sign queue entirely then right? In that case, we'd need to split sign queue itself out to be its own separate task, or at least a task that purely just pops the requests and organizes the requests.

I'm already making the signature manager a separate task so not a huge issue to do this.

jakmeier: Yeah, `SignQueue::organize` should run in its own task, if we block here. Or, we can also use non-blocking calls to the redis queue if you prefer that.

Issue #174, opened 2025-02-06
Check if Signature request was indexed before starting a protocol initiatad by other node


volovyks:

volovyks:
We discussed with @ChaoticTempest that this check can affect performance since we need to check if each incoming message belongs to the indexed request. We can avoid that if have an explicit start and end of the protocol as discussed with @jakmeier earlier. Related to Step 2 of [this doc](https://docs.google.com/document/d/1rU0diFhHGvviKetw9XVnqGlDnX4TlJC5xKXl4OeBhGw/edit?pli=1&tab=t.0#heading=h.viie31rfopdq).

jakmeier:
Makes sense. But it's a bit more work than a quick fix.

@ChaoticTempest  Couldn't you for now just make the check only when the protocol is start locally? 
Just split `signature_manager.get_or_start_protocol` into `signature_manager.get_protocol` and `signature_manager.start_protocol`. If `get_protocol` returns `None`, check if the request was indexed. Then, if the check was successful, call `start_protocol` and continue.

ChaoticTempest:
@jakmeier I found a somewhat easier way of doing this but required changes to the sign queue. So no need to split a part `signature_manager.get_or_start_protocol` by just taking the request from the sign queue, if it does not exist, just wait until it does from indexer

Issue #130, opened 2025-01-22
State Sync (T and P)


volovyks:

jakmeier:
I discussed this with @volovyks earlier today. This is what I think we agreed on.

This description and simplification over the original proposal makes sense. A simple two-way sync for every link going up is enough for correctness. Anything on top is an optimization we don't necessarily need to implement.
Also, the other simplification, we don't need to track explicitly the used triples. Sending just the unused list is enough, since everything not on the list can be considered used.

The one thing I want to stress, though, is that each node only keeps a list of their own triples / pre-signatures. No need to track information about foreign data. Doing it this way makes the state management easier, since no information about T and P holders is duplicated. The owner of it has the single source of truth. They are also the only node that needs this information, as only they will initiate it. And on top of that, synchronizing only on the IDs owned by the peer can reduce the overhead.

ChaoticTempest:
in terms of implementation, this is how I can see us implement this.

## Endpoints

One endpoint for both sending a list of owned triples, and receiving back a view of what the node owns.

- `/sync`: SyncUpdate &#8594; SyncView
- SyncUpdate(OwnedTripleIds, OwnedPresignatureIds)
- SyncView(ShareTripleIds, SharePresignatureIds)


Two different endpoints for parallelization (maybe better):

- `/sync/view`: () &#8594; SyncView
  - owner views what the other node owns
- `/sync/update`: SyncUpdate &#8594; ()
  - owner pings a node with a list of owned triples/presignatures


## Task

New task(s) will manage this synchronization. Let's call it `ProtocolSync` for now.

It will do the following on every iteration:
- process `SyncUpdate` request:
    - ping storage to remove entries that are now invalid
- process updates from storage:
    - creates a local cache of what participants each protocol has
- call `/sync` to all other nodes.
- update local cache based on the info it receives from other nodes
- process take triple/presignature requests:
    - ping storage to remove
    - update local cache so that next iteration updates all other nodes if necessary.

How does it know what protocol it owns?
- on startup: storage will alert `ProtocolSync` with all owned protocols.
- on protocol generation completed/stored: storage will alert `ProtocolSync`

How does a generator know which triple/presignature to use based on `ProtocolSync`?
[...]

PR #180, opened 2025-02-08
fix(signature): use indexed epsilon, and participant subset

ChaoticTempest:
Fixing one thing in signatures requires fixing another. So this PR fixes a couple things because they all slightly require each other to be correct:

- https://github.com/sig-net/mpc/issues/174
- no longer using message epsilon when initiating a signature generation proposed from another node
- use the participant subset determined by the `SignQueue`. This was not the case before because we would use whoever is active. This however does leave some nodes to still have leftover presignatures if they are not in this subset. Not a huge deal for now with a sufficiently large stockpile, but this should be fixed in the future with [state sync](https://github.com/sig-net/mpc/issues/130).


ChaoticTempest:
In a future PR, I will also refactor how we do failed requests because how we determine the the subset here is quite fixed due to how sign_queue was before.

So in the case where a node were to hop offline in the middle of signature generation, we should have the sign queue reorganize the request. Therefore the sign queue should be handling failed requests and be the one and only places to get sign requests from instead of SignatureManager also managing failed requests

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -87,42 +89,67 @@ impl SignQueue {
...
+            tracing::info!(
+                ?request_id,
+                ?subset,
+                ?proposer,
+                in_subset,
+                is_mine,
+                "sign queue: organizing request"
+            );
+
+            let sign_id = SignRequestIdentifier::new(
```
volovyks, line 107: It is a SignRequest that has request_id inside. Should we unify that? Not a blocker for this PR, but check my comment in https://github.com/sig-net/mpc/issues/183

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -87,42 +89,67 @@ impl SignQueue {
...
                     "skipping sign request: node is NOT in the signer subset"
                 );
             }
         }
     }
 
-    pub fn take_my_requests(&mut self) -> VecDeque<SignRequest> {
-        self.requests.remove(&self.me).unwrap_or_default()
+    pub fn take_my_requests(&mut self) -> VecDeque<(SignRequest, Participant, Vec<Participant>)> {
+        std::mem::take(&mut self.my_requests)
```
volovyks, line 138: Is it safe to work with memory here?
Does it affect performance? Is it a blocking operation for incoming requests?

volovyks (commented):
Great work @ChaoticTempest 
Just a couple of questions

chain-signatures/node/src/protocol/message.rs
```diff
@@ -647,6 +630,17 @@ impl MessageReceiver for RunningState {
                     // We will revisit this this signature request later when the presignature has been generated.
                     continue;
                 }
+                Err(err @ GenerationError::WaitingForIndexer(_)) => {
+                    // We will revisit this this signature request later when we have the request indexed.
```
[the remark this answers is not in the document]
ChaoticTempest: what about this?

chain-signatures/node/src/protocol/signature.rs
ChaoticTempest: yeah this is a stopgap right now before I do a whole entire refactor of our sign queue and signature id

chain-signatures/node/src/protocol/signature.rs
ChaoticTempest: the only time where anything is unsafe is if the code shows `unsafe`. For the most part, any time you see `std::mem` it's an optimization on accessing the heap. This isn't an IO operation so not blocking and purely just multiple assignment operations

chain-signatures/node/src/protocol/message.rs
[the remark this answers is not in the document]
volovyks: you have two this in a row

volovyks (approved):
@ChaoticTempest I think we can merge it (after fixing CI) and release a mew version.

ChaoticTempest:
the latency reshare test is a bit flaky so merging this now as a hotfix

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -421,74 +452,84 @@ impl SignatureManager {
...
-                Ok(&mut generator.protocol)
-            }
-            Entry::Occupied(entry) => Ok(&mut entry.into_mut().protocol),
+
+        let entry = match self.generators.entry(sign_id.clone()) {
+            Entry::Vacant(entry) => entry,
+            Entry::Occupied(entry) => return Ok(&mut entry.into_mut().protocol),
+        };
+
+        let Some((request, our_proposer, subset)) = self.sign_queue.take(sign_id) else {
```
jakmeier, line 477: By taking it out of the queue, doesn't it mean that if any of the later checks fail (`InvalidProposer`, `BadParameters`, maybe more) will lead to an inconsistent state? We will not have processed the signature but we have removed it from our queue. Future calls will return `WaitingForIndexer`, if I understand it right.

jakmeier: @ChaoticTempest I know the PR is merged and closed, just checking if this might be a problem

ChaoticTempest: good shout, it was indeed an issue. @ppca fixed it in this https://github.com/sig-net/mpc/pull/226

PR #181, opened 2025-02-08
feat: one workspace for all packages

ChaoticTempest:
this unifies all packages under one workspace:

- remove `once_cell::Lazy` since `std::sync::LazyLock` is pretty much the same and would use one less dep.
- removes the caching step from docker builds because we don't rebuild that often anymore. We just build an image once and if it works, we deploy it now vs back then when we need images for tests.


volovyks (commented):
Looks good overall. We still have the Docker option in tests. Not sure if we even need it.

volovyks (approved):

volovyks:
@ChaoticTempest looks like deploy node CI task is failing, probably because of the wrong paths

PR #194, opened 2025-02-11
chore: refactor signature id and failed signatures

ChaoticTempest:
`SignId` is now used everywhere instead of request_id. A couple things were made more efficient like no longer passing the part of sign requests around, instead we wholly pass around `SignRequest` which contains all the info the node needs to do signature generation.

- `SignId` is an id agnostic of any chain, just has raw `request_id`, `payload` and `epsilon`.
- `SignArgs` is all the info that is not present in the `SignId` such as `entropy`, `key_version`, and `path`.
- `IndexedSignRequest` is all the info from the indexer which includes `SignId` and `SignArgs`.
- `SignRequest` has more info that is not present on the chain but only inside the node such as `participants` and `proposer`.
- removed `ContractSignRequest` since `SignArgs` basically replaces it.
- `SignatureMessage` now uses `SignId` and does not send `ContractSignRequest` anymore since it was never used. This info is already indexed by each of the nodes so not necessary to confirm this, but the proposer and presignature id is still something that each node needs to come up with, so that is still being passed around. This is not a breaking serialization change since the messages are forward/backwards compatible as long as the field's names are unique/new which they are with this change.
- Failed requests are now a part of the `SignQueue` for better management of requests overall. All failed requests are put at the front of the queue as to be process immediately on the next try. Note that retrying uses the same participants, which is not ideal. In a separate PR, I'll have it so that the participants are reselected but the proposer will stay the same. This selection is deterministic where the seed for randomization is `entropy + retry_count`.


chain-signatures/node/src/protocol/signature.rs
```diff
@@ -28,34 +27,91 @@ use tokio::sync::mpsc::error::TryRecvError;
...
+        }
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
+pub struct SignArgs {
+    #[serde(with = "serde_bytes")]
     pub entropy: [u8; 32],
-    pub time_added: Instant,
+    pub path: String,
```
volovyks, line 76: Why do we need to pass the path around, if we already have epsilon in SignId? My suggestion is still to remove epsilon and payload from SignId.

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -28,34 +27,91 @@ use tokio::sync::mpsc::error::TryRecvError;
...
 }
 
 pub struct SignQueue {
     me: Participant,
-    sign_rx: Arc<RwLock<mpsc::Receiver<SignRequest>>>,
-    my_requests: VecDeque<(SignRequest, Participant, Vec<Participant>)>,
-    other_requests: HashMap<SignRequestIdentifier, (SignRequest, Participant, Vec<Participant>)>,
+    sign_rx: Arc<RwLock<mpsc::Receiver<IndexedSignRequest>>>,
+    my_requests: VecDeque<SignRequest>,
+    other_requests: HashMap<SignId, SignRequest>,
```
volovyks, line 103: Here we have a map for SignId -> SignRequest, but SignRequest has IndexedSignRequestand it has the same SignId.

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -217,50 +270,11 @@ impl SignatureGenerator {
...
-    pub participants: Vec<Participant>,
-    pub request: ContractSignRequest,
-    pub epsilon: Scalar,
-    pub request_id: [u8; 32],
-    pub entropy: [u8; 32],
-    pub sign_request_timestamp: Instant,
-}
-
-#[derive(Debug, Clone, Eq, PartialEq)]
-pub struct SignRequestIdentifier {
```
volovyks, line 233: This is basically our SignId (note for myself)

volovyks (commented):
Lots of good stuff that I'm ready to merge, but I either do not understand or disagree with the core thing in this PR - SignId.

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -28,34 +27,91 @@ use tokio::sync::mpsc::error::TryRecvError;
...
-use near_primitives::hash::CryptoHash;
-
-pub type ReceiptId = near_primitives::hash::CryptoHash;
 
 /// This is the maximum amount of sign requests that we can accept in the network.
 const MAX_SIGN_REQUESTS: usize = 1024;
 
-pub struct SignRequest {
+#[derive(Clone, Eq, PartialEq, Serialize, Deserialize)]
+pub struct SignId {
```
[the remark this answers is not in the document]
ChaoticTempest: wait, why was it initially introduced as utilizing payload/epsilon as well then? This `SignId` is the same as what it was before with `SignRequestIdentifier` just renamed

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -28,34 +27,91 @@ use tokio::sync::mpsc::error::TryRecvError;
...
+    #[serde(with = "serde_bytes")]
     pub request_id: [u8; 32],
-    pub request: ContractSignRequest,
+    #[serde(with = "cbor_scalar")]
+    pub payload: Scalar,
+    #[serde(with = "cbor_scalar")]
     pub epsilon: Scalar,
+}
+
+impl std::hash::Hash for SignId {
```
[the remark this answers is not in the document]
ChaoticTempest: yes, it's needed for hashmap

chain-signatures/node/src/protocol/signature.rs
ChaoticTempest: we can't recover the path from an epsilon, so better to just keep it in our system for now

chain-signatures/node/src/protocol/signature.rs
ChaoticTempest: it doesn't matter if we store them twice, because it's better if we can easily access the request instead of iterating through the whole list of requests

chain-signatures/node/src/protocol/signature.rs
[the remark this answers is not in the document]
ChaoticTempest: ok, I made this now just `SignId(request_id)`

ChaoticTempest:
note, that since the near contract also uses `SignId` for `respond` and internal tracking of requests, the SignId had to be precomputed in the contract and indexer separately similar to eth way of computing it

integration-tests/src/commands.rs
```diff
@@ -40,32 +39,26 @@ pub fn sing_command(contract_id: &AccountId, caller_id: &AccountId) -> anyhow::R
...
+    let signature = Signature {
+        big_r,
+        s,
         recovery_id: 0,
     };
 
     let request_json = format!(
         "'{}'",
-        serde_json::to_string(&json!({"request": request, "response": response})).unwrap()
+        serde_json::to_string(&json!({"sign_id": sign_id, "signature": signature})).unwrap()
```
volovyks, line 61: Note: We have breaking changes in this PR too.

volovyks (approved):

PR #201, opened 2025-02-11
Determine admin acc

volovyks:


chain-signatures/crypto/Cargo.toml
```diff
@@ -17,3 +17,5 @@ sha3.workspace = true
 getrandom = { version = "0.2.12", features = ["custom"] }
 
 [dev-dependencies]
+web3.workspace = true
+secp256k1.workspace = true
```
ChaoticTempest, line 21: `secp256k1` dep isn't used in the tests

ppca (approved):
logic looks good to me

ChaoticTempest (approved):

PR #209, opened 2025-02-13
feat: use IterableMap for near contract

ChaoticTempest:
In the spirit of breaking changes, use `IterableMap` instead of `LookupMap` just so we don't maintain that pesky request counter anymore.


volovyks (dismissed):
Nice! Why haven't we used this structure before? Probably because it’s not in collections.

From what I see, the only breaking change here is the SignatureResult part. Overall, I do not want to invest time in this API since it is unclear if we need it. Maybe we will rewrite it to work like the one on Ethereum, with a significantly simplified design.

We can afford breaking changes here now or after the release since technically we are not releasing on NEAR. We will need to fix some things in our infra and tooling, let's merge this if you are ok with that.

ChaoticTempest:
SignatureResult isn't a breaking change since it's only used within the contract

volovyks (approved):

PR #216, opened 2025-02-13
Error handling in Ethereum contract

volovyks:
We need a way to communicate errors with users in case they provide bad parameters. Not integrated fully for now. I just want to have it in the contract.

### DIFF
--- a/chain-signatures/contract-eth/contracts/ChainSignatures.sol
+++ b/chain-signatures/contract-eth/contracts/ChainSignatures.sol
@@ -30,7 +30,7 @@ contract ChainSignatures is AccessControl {
 
     struct Response {
         bytes32 requestId;
-        Signature response;
+        Signature signature;
     }
 
     uint256 signatureDeposit;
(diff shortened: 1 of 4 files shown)


chain-signatures/contract-eth/contracts/ChainSignatures.sol
```diff
@@ -117,7 +134,21 @@ contract ChainSignatures is AccessControl {
...
+                _responses[i].signature
+            );
+        }
+    }
+
+    /**
+     * @dev Function to emit signature generation errors.
+     * @param _errors The array of signature generation errors.
+     */
+    function respondError(ErrorResponse[] calldata _errors) external {
```
ChaoticTempest, line 146: where does `respondError` get called?

volovyks: Nowhere for now. But if we want to communicate errors in the future - we should have it.

ppca: I guess this is currently not used anywhere, but will be added to backend code for node later in case respond fails.

ChaoticTempest (approved):

volovyks:
Thanks for the test @ppca :)

chain-signatures/contract-eth/contracts/ChainSignatures.sol
```diff
@@ -30,7 +30,12 @@ contract ChainSignatures is AccessControl {
 
     struct Response {
         bytes32 requestId;
-        Signature response;
+        Signature signature;
+    }
+
+    struct ErrorResponse {
+        bytes32 requestId;
+        string error;
```
ppca, line 38: is error a special word in Solidity? asking because it is shown in red here like indexed.

PR #225, opened 2025-02-13
chore: make generating/resharing log less messages

ChaoticTempest:


ppca (approved):
good change! was gonna do the same lolll

PR #226, opened 2025-02-14
fix: push failed signature request into sign queue for non-proposers

ppca:
discovered that eth requests fail when it did not succeed on 1st generator, and need to retry.
Poking the logs, I find that the non-proposer nodes were warning `waiting for indexer`. This means that the request has been removed from sign_queue, and not added back.
This should affect near requests as well, but likely near block time is fast thus less delay on indexer side, thus most requests finish in first attempt.

### DIFF
--- a/chain-signatures/node/src/protocol/signature.rs
+++ b/chain-signatures/node/src/protocol/signature.rs
@@ -364,6 +364,7 @@ impl SignatureManager {
             return Err(GenerationError::WaitingForIndexer(sign_id.clone()));
         };
         if proposer != request.proposer {
+            self.sign_queue.push_failed(request.clone());
             return Err(GenerationError::InvalidProposer(proposer, request.proposer));
         }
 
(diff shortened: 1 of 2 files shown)


chain-signatures/node/src/protocol/signature.rs
```diff
@@ -228,7 +228,7 @@ pub struct SignatureManager {
     /// Ongoing signature generation protocols.
     generators: HashMap<SignId, SignatureGenerator>,
     /// Set of completed signatures
-    completed: HashMap<SignId, Instant>,
+    completed: HashMap<SignId, (PresignatureId, Instant)>,
```
ChaoticTempest, line 231: shouldn't it be (SignId, PresignatureId) instead? probably cleaner that way

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -413,27 +413,30 @@ impl SignatureManager {
...
                         remove.push(sign_id.clone());
 
-                        if generator.request.proposer == self.me {
-                            if generator.request.indexed.timestamp.elapsed()
-                                < generator.timeout_total
-                            {
-                                failed.push(sign_id.clone());
-                                tracing::warn!(?err, "signature failed to be produced; pushing request back into failed queue");
+                        if generator.request.indexed.timestamp.elapsed() < generator.timeout_total {
+                            failed.push(sign_id.clone());
```
ChaoticTempest, line 417: actually, we should move failed to right after remove. It's weird that it was only on timeout_total. Timeout total is for all retries which isn't correct

chain-signatures/node/src/protocol/signature.rs
ppca: hmmm but yeah, (SignId, PresignatrueId) achieves the purpose of completed better, like we can check which pair has already been done before. Maybe I'll loop the queue for each signId and do the refresh_gc for each (SignId, PresignatrueId) pair.

ChaoticTempest: yeah, completed should be `(SignId, PresignatureId)`, because a singular `SignId` can have multiple attempts

chain-signatures/node/src/protocol/signature.rs
ChaoticTempest: ahh, yeah for some reason I was reading it as `>= generator.timeout_total` which means it only is failed on all attempts are done

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -343,7 +343,7 @@ impl SignatureManager {
         cfg: &ProtocolConfig,
         presignature_manager: &mut PresignatureManager,
     ) -> Result<&mut SignatureProtocol, GenerationError> {
-        if self.completed.contains_key(sign_id) {
+        if self.completed.contains_key(sign_id) && self.completed[sign_id].0 == presignature_id {
             tracing::warn!(
                 ?sign_id,
                 presignature_id,
```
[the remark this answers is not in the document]
ppca: ok sounds reasonable

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -343,7 +343,7 @@ impl SignatureManager {
         cfg: &ProtocolConfig,
         presignature_manager: &mut PresignatureManager,
     ) -> Result<&mut SignatureProtocol, GenerationError> {
-        if self.completed.contains_key(sign_id) {
+        if self.completed.contains_key(sign_id) && self.completed[sign_id].0 == presignature_id {
```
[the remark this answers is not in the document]
ppca: Completed is now modified to be keyed by (SignId, PresignatureId)

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -413,27 +426,30 @@ impl SignatureManager {
...
                                     .with_label_values(&[self.my_account_id.as_str()])
                                     .inc();
-                            } else {
-                                self.completed.insert(sign_id.clone(), Instant::now());
+                            }
+                        } else {
+                            self.completed.insert(
+                                (sign_id.clone(), generator.presignature_id),
+                                Instant::now(),
+                            );
```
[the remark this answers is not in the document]
ppca: ah..so when we retry, it will be a different presig now?

ChaoticTempest (approved):

PR #230, opened 2025-02-18
Use low level `call` func to withdraw

volovyks:
Auditors:
```
Using transfer may fail on chains with higher gas costs (e.g., zkSync Era). Since only the admin withdraws fees, switching to a call is recommended.
```

### DIFF
--- a/chain-signatures/contract-eth/contracts/ChainSignatures.sol
+++ b/chain-signatures/contract-eth/contracts/ChainSignatures.sol
@@ -180,12 +180,15 @@ contract ChainSignatures is AccessControl {
         uint256 _amount,
         address _receiver
     ) external onlyRole(DEFAULT_ADMIN_ROLE) {
+        require(_receiver != address(0), "Invalid receiver address");
         require(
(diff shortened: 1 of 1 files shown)


esaminu (approved):

chain-signatures/contract-eth/contracts/ChainSignatures.sol
```diff
@@ -180,12 +180,15 @@ contract ChainSignatures is AccessControl {
         uint256 _amount,
         address _receiver
     ) external onlyRole(DEFAULT_ADMIN_ROLE) {
+        require(_receiver != address(0), "Invalid receiver address");
```
ppca, line 183: what is address(0)?

volovyks: In Ethereum and Solidity, address(0) refers to the zero address, which is 0x0000000000000000000000000000000000000000. It is commonly used to represent an invalid, non-existent, or null address.

ppca (approved):

PR #231, opened 2025-02-18
Check deposit with `=`

volovyks:
Auditors:
```
There's no refund for overpayments. To enforce an exact deposit, update the check to:
require(msg.value == signatureDeposit, "Insufficient deposit");
```
Overall, it makes sense, but it disallows us to accept flexible payments, such as prioritizing signatures in the queue.

### DIFF
--- a/chain-signatures/contract-eth/contracts/ChainSignatures.sol
+++ b/chain-signatures/contract-eth/contracts/ChainSignatures.sol
@@ -110,7 +110,7 @@ contract ChainSignatures is AccessControl {
      * @param _request The signature request details.
      */
     function sign(SignRequest memory _request) external payable {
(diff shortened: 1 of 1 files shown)


volovyks:
We can add a check, that the `deposit < 10 * required_deposit to` prevent big losses on the client side.

volovyks:
Here is what we can do to remain flexible and prevent users spend money by accident: https://github.com/sig-net/mpc/pull/232

DavidM-D:
I think just keep it simple with > on the first contract, I don't think we want lowering the gas cost to be a potentially breaking change. 

The 10x isn't great either, because you can still overpay by 9x and if gas gets much cheaper we could still end up with a breaking change on lowering the gas fee.

volovyks:
Ok then, let's keep it flexible.

PR #234, opened 2025-02-18
configuration digest proposal

volovyks:


chain-signatures/node/src/metrics.rs
```diff
@@ -488,6 +488,15 @@ pub(crate) static NUM_SIGN_REQUESTS_ETH: LazyLock<CounterVec> = LazyLock::new(||
     .unwrap()
 });
 
+pub(crate) static CONFIGURATION_DIGEST: LazyLock<IntGaugeVec> = LazyLock::new(|| {
+    try_create_int_gauge_vec(
```
ppca, line 492: This may not allow negative values

ppca: Ah should be good. it's i64

ppca (approved):

ChaoticTempest (approved):

PR #235, opened 2025-02-18
hardcode contract address

ppca:


volovyks:
updated on the nodes side

PR #240, opened 2025-02-19
chore: log cipher/sign pk alongside digest

ChaoticTempest:
log the cipher/sign pk with digest to verify what gets set in metrics


auto-mausx (approved):

volovyks:
added this to my changes https://github.com/sig-net/mpc/pull/239

PR #243, opened 2025-02-20
eth contract disclaimer

volovyks:
Hey @Pessina @esaminu,
please make sure we have this in our docs.

### DIFF
--- a/chain-signatures/contract-eth/contracts/ChainSignatures.sol
+++ b/chain-signatures/contract-eth/contracts/ChainSignatures.sol
@@ -66,7 +66,7 @@ contract ChainSignatures is AccessControl {
 
     /**
      * @dev Emitted when a signature response is received.
-     * @notice This event can be emitted by anyone. Clients should always verify the validity of the signature.
+     * @notice Any address can emit this event. Clients should always verify the validity of the signature.
      * @param requestId The ID of the request. Must be calculated off-chain.
      * @param responder The address of the responder.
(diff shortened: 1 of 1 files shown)


chain-signatures/contract-eth/contracts/ChainSignatures.sol
```diff
@@ -78,6 +79,7 @@ contract ChainSignatures is AccessControl {
 
     /**
      * @dev Emitted when a signature error is received.
+     * @notice This event can be emitted by anyone. Clients should not relly any business logic on this event.
```
ppca, line 82: typo: relly -> rely

volovyks, line 82:
```suggestion
     * @notice Any address can emit this event. Do not rely on it for business logic.
```

chain-signatures/contract-eth/contracts/ChainSignatures.sol
```diff
@@ -66,6 +66,7 @@ contract ChainSignatures is AccessControl {
 
     /**
      * @dev Emitted when a signature response is received.
+     * @notice This event can be emitted by anyone. Clients should always verify the validity of the signature.
```
volovyks, line 69:
```suggestion
     * @notice Any address can emit this event. Clients should always verify the validity of the signature.
```

ppca (approved):

PR #244, opened 2025-02-20
Update luganodes acc id

volovyks:
Not yet fixed on their side


volovyks:
Nor relevant, we better remove it

PR #247, opened 2025-02-20
turn off near indexer

volovyks:


ppca (approved):
LGTM as a temporary try

volovyks:
We still need our integration-tests. Check: https://github.com/sig-net/mpc/pull/283

PR #248, opened 2025-02-20
ethereum mainnet contract deployment

ppca:
this is to save the mainnet deployment and also enable using requestSign to directly call mainnet contract


volovyks:
Do we need all these files for it?

ppca:
> Do we need all these files for it?

our script only use the deployed address. But I want to keep these so that the immutable mainnet contract has a record

volovyks (approved):

volovyks:
It is 27000 lines, let me take a closer look later please

ppca:
> It is 27000 lines, let me take a closer look later please

sure. maybe keeping the deployed_address is enough?

ChaoticTempest:
yeah I feel like we shouldn't be keeping these artifacts here. If anything, we should have a separate repo for this if we want to bookkeep this somehow

volovyks:
Removed 27000 lines of build info. Cleaned this up per the thread: dropped the generated Ignition output (artifacts, build-info, journal), kept only `deployed_addresses.json` for chain-1. Added `.gitignore` rules so future deployments commit only the address file. Merged develop. The diff is now 12 lines.

Verified against mainnet: the deploy tx created the contract at 0xf8bd…791A with admin 0x3c0f…a445 and deposit 1200000000000000, matching the module and params here.

volovyks:
@ppca, I've prepared this PR for merging. Why is it closed?

PR #249, opened 2025-02-20
Updated docker image to use a pinned redis version 7.4.2

auto-mausx:


chain-signatures/node/redis.conf
```diff
@@ -260,7 +260,7 @@ dbfilename dump.rdb
 # The Append Only File will also be created inside this directory.
 #
 # Note that you must specify a directory here, not a file name.
-dir /data
+dir /etc/redis
```
ChaoticTempest, line 263: how come we're creating the AOF in the /etc/redis folder?

volovyks (approved):

chain-signatures/node/redis.conf
auto-mausx: Because the service that we used to use, created a /data directory, a redis user, and other things on startup, now we are only using the binary, so we just need to point it to a file that has persistence on the disk, which is /etc

PR #250, opened 2025-02-21
feat: use setup.sh instead of build.rs

ChaoticTempest:
this PR cleans up our a build a bit since I discovered a way to nicely build our node and contract without relying on the `build.rs` script which doesn't work well with benchmarking. The `runner` in config.toml is used for this purpose instead.

- remove `.vscode/settings.json` since removing `build.rs` makes it play nicer with rust analyzer.
- no more need for `rust-toolchain.toml` since contract can just be built directly with `cargo +1.81.0 ...`. This means that our test builds will now be using `stable` which was already the case for our docker image builds.
- use `./build-contract.sh` to build contract from now on.

This is also needed for benchmarks, since the 1.81 rust toolchain really kills our mpc-node compilation when benchmarks are added


volovyks (approved):
Looks good overall.

Removing rust_toochain.toml means that devs will potensially work with different Rust versions. I would keep to for the whole workspace.

Oh, I see, it is pinned in the script. But again, isn't toml file a native way to achieve that?

ChaoticTempest:
@volovyks the toml file can only be one per workspace, so it would mean that we would need to split the workspace into two, with one being solely for contract, which isn't ideal either just to have it target a specific rust version for compilation serialization purposes. It's better to just have it be an extra build step than to make that whole sweeping change

PR #251, opened 2025-02-24
feat: move to Rust edition 2024

ChaoticTempest:
requires updating to latest rustc 1.85


volovyks (approved):
LGTM

ChaoticTempest:
this actually doesn't mix very well with our contract builds since that requires 1.81 and consequently edition 2021, so won't merge this one in until contracts are fixed with latest rust versions

PR #258, opened 2025-02-27
Remove cipher pk from parameters

volovyks:


ChaoticTempest (dismissed):

volovyks:
@auto-mausx please, take a look
Can we merge this?

auto-mausx:
I can actually remove all of our infra from this repo, as it is now in sig-kustomize. We will need to keep the partner prefixed infra, but let me send an update to remove our other stuff.

auto-mausx (approved):

PR #259, opened 2025-02-27
Use HPKE to verify sender

volovyks:
This change allows us to remove the signing key and all related complexity.
I will do all the refactoring in case we approve this approach. We will need to pass the sender around to decrypt the message.

### DIFF
--- a/chain-signatures/keys/src/hpke.rs
+++ b/chain-signatures/keys/src/hpke.rs
@@ -68,13 +68,18 @@ impl PublicKey {
         Self::try_from_bytes(bytes).expect("invalid bytes")
     }
 
-    pub fn encrypt(&self, msg: &[u8], associated_data: &[u8]) -> Result<Ciphered, hpke::HpkeError> {
+    pub fn encrypt(
+        &self,
+        msg: &[u8],
+        sender_sk: &SecretKey,
+        associated_data: &[u8],
+    ) -> Result<Ciphered, hpke::HpkeError> {
         let mut csprng = <rand::rngs::StdRng as rand::SeedableRng>::from_entropy();
 
         // Encapsulate a key and use the resulting shared secret to encrypt a message. The AEAD context
         // is what you use to encrypt.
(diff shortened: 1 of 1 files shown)


chain-signatures/keys/src/hpke.rs
```diff
@@ -154,13 +160,35 @@ pub fn generate() -> (SecretKey, PublicKey) {
...
+
+        // Malicious party should not be able to decrypt the message
+        assert!(malicious_sk
+            .decrypt(&ciphered, &sender_pk, associated_data)
+            .is_err());
+
+        // Recipient should not be able to decrypt the message if the sender is wrong
+        assert!(receiver_sk
+            .decrypt(&ciphered, &malicious_pk, associated_data)
+            .is_err());
```
jakmeier, line 186: :heart:

jakmeier (commented):
LGTM

One question though. What is the deployment plan of this?

I ask because nodes running old code will not work with those that have updated. Usually, either everyone needs to update at the same time, with some system downtime while everyone is updating. Or alternatively, we add versioning to the messages, where we accept the both old and new messages for one release cycle, before starting to reject the old versions.

chain-signatures/keys/src/hpke.rs
scafuro: @volovyks I found that hkpe's authentication guarantee's are not the same as the ones achieved by signing a cipher text. Before finalizing the update, let's  briefly discuss if hkpe's guarantees are more than enough for our application.

ChaoticTempest:
One thing about moving to HPKE for auth, we do lose the ability to mitigate replay attacks, since the signature from our signing key can used to verify that the messages hasn't been accepted in our system yet

volovyks:
@ChaoticTempest can you elaborate? I don't know how this replay attack works.

volovyks:
Since HPKE alone does not provide the same level of security guarantees, it was decided to keep using separate keys for signing and encryption.

PR #261, opened 2025-02-28
fix: move system metrics into spawn_blocking task to prevent sleeping main thread

ChaoticTempest:
Our system metrics hard sleeps our main runtime thread for 200ms, which is why in our grafana we see that it's never below 200ms for our main loop iteration. This is not ideal for the runtime as it cannot progress any work due to one of threads being asleep. So if any work was scheduled on that thread, nothing will be done and the runtime would have to steal the task to another thread, and in the worse case, it just stays in that slept thread


volovyks (approved):

ChaoticTempest:
Merged this one to see if our signature rates improve and overwrite the changes from https://github.com/sig-net/mpc/pull/255

Issue #262, opened 2025-03-01
Filter out GC/completed/failed generators at the Message channel level


ChaoticTempest:
We should filter out messages associated to protocols that have already been either completed or failed by the protocols. This should be done at the message level such that we don't require the state to be mutable at Protocol manager level. Currently this handled respectively by each of the protocol managers, but makes its unwieldy to pass around the managers if we need to use them concurrently.

For example, the signature manager would no longer have a hard dependency on the presignature manager, because now both the presignature manager and the signature manager can have their own reference of the presignature storage. So in this case, they would no longer be blocking each other from continuing.

This would mean getting rid of GC (Garbage collection) at each of the manager level, and making it more of a general protocol ID filter at the message level.

How this should look like is something like this:
- each of the protocol tasks communicate to the messaging level that a specific ID should no longer be accepted.
- then each of the manager/protocol tasks no longer needs to worry about the case of GC.

Issue #273, opened 2025-03-05
Task Architecture


ChaoticTempest:
In our MPC system, we should roughly have a similar model to actors with communications being between one-way or bi-directional channels. This is because we have a lot of things running in parallel with their dependencies also running asynchronously in parallel.

Additionally, we don't have to fully rely on the actor framework (yet), as that might pigeonhole us into using actors for things that aren't necessarily the best as actors; which could lead to friction when ideating and developing. So this is left as pure as possible, relying on only runtime tasks and channels to give us the most freedom.

## Tasks

Going forward, this is what I believe our model/architecture for our system should look like:

- Messages: manages all related activities for messaging including filtering and idempotentcy. Forwards messages to subscribers.
- RPC: manages all related activities for RPC
- Protocols:
    - Triples
        - TripleSpawner: task that introduces or listens for new incoming, and spawns `TripleGenerator`.
        - TripleGenerator: task that actively generates a triple. Subscribes to `Messages`.
    - Presignatures
        - PresignatureSpawner: task that introduces or listens for new incoming, and spawns `PresignatureGenerator`.
        - PresignatureGenerator: task that actively generates a presignature. Subscribes to `Messages`.
    - Signatures
        - SignatureSpawner: listens for incoming requests from `SignQueue` and spawns `SignatureGenerator`.
        - SignatureGenerator: task that actively generates a signature. Subscribes to `Messages`.

- [Protocol State Sync](https://github.com/sig-net/mpc/issues/130): task that syncs owned protocols with other nodes, and also communicating what owned stored protocols we can take to formulate the next protocol.
- Storage (Redis): this can be either a task or not. We can start with it as-is (a storage Arc passed around to a redis instance). The only reason to have this as a task is if (de)serialization has an impact on each protocols' performance; also it'll manage all retrying attempts for us such that we don't handle it ad-hoc.

This is not a complete list of the core tasks, but it paints a picture as to how things should work.

volovyks:
@ChaoticTempest, I believe many of the things on the list are already done. Should we close this? Is there anything important left?

ChaoticTempest:
yeah, this is pretty much good to be closed

PR #275, opened 2025-03-06
feat: remove garbage collection, add filtering and idempotency to MessageChannel

ChaoticTempest:
This resolves https://github.com/sig-net/mpc/issues/262.

- Remove the need for GC in all the managers by adding `MessageFilter` to the `MessageInbox`, such that incoming messages will now be filtered based on the id of the message. `TripleId` for `TripleMessage`, `PresignatureId` for `PresignatureMessage`, and `(SignId, PresignatureId)` for `SignatureMessage`.
- Also made `MessageInbox` idempontent such that if the node already received a batch of message, it will not process that message since it has already seen it. This is done by checking the signature of the message.
- `MessageInbox` is now mostly reachable from tests, so we can now directly test it from unit tests.


chain-signatures/node/src/protocol/message/mod.rs
```diff
@@ -361,20 +331,38 @@ impl MessageChannel {
...
+
+    pub async fn filter_presignature(&self, id: PresignatureId) {
+        if let Err(err) = self.filter.send((Protocols::Presignature, id)).await {
+            tracing::warn!(?err, "failed to send filter message");
+        }
+    }
+
+    pub async fn filter_sign(&self, sign_id: SignId, presign_id: PresignatureId) {
+        self.filter(&(sign_id, presign_id)).await;
+    }
```
jakmeier, line 357: Shouldn't it work like this? Seems like you built an abstraction without actually using it.

```suggestion
    /// Marks this message as filtered. This is used to prevent the same message with the
    /// corresponding MessageId from being processed again.
    async fn filter<M: MessageFilterId>(&self, msg: &M) {
        if let Err(err) = self.filter.send((M::PROTOCOL, msg.id())).await {
            tracing::warn!(?err, "failed to send filter message");
        }
    }

    pub async fn filter_triple(&self, id: TripleId) {
        self.filter(&id).await;
    }

    pub async fn filter_presignature(&self, id: PresignatureId) {
        self.filter(&id).await;
    }

    pub async fn filter_sign(&self, sign_id: SignId, presign_id: PresignatureId) {
        self.filter(&(sign_id, presign_id)).await;
    }
```

chain-signatures/node/src/protocol/message/filter.rs
```diff
@@ -0,0 +1,57 @@
...
+        }
+    }
+
+    pub fn contains<M: MessageFilterId>(&mut self, msg: &M) -> bool {
+        // Check if the message is already in the filter. Doing `get` here will also
+        // update the LRU cache and promote the rank of this id to be most recent.
+        self.filter.get(&(M::PROTOCOL, msg.id())).is_some()
+    }
+
+    pub fn extend(&mut self) {
```
jakmeier, line 34: nit: `extend` seems like an odd choice of a name. In Rust, I would personally only use it to do what [std::iter::Extend](https://doc.rust-lang.org/std/iter/trait.Extend.html#tymethod.extend) does.

I would maybe call it `update`, `refresh`, or if you like the extend word, `extend_from_receiver`. But I think using the word extend in such a way without a paramter is against common practice in Rust.

Obviously a matter of opinion, though, feel free to disagree and keep your naming :)

chain-signatures/node/src/protocol/message/mod.rs
```diff
@@ -819,18 +811,19 @@ pub struct MessageOutbox {
...
             client,
-            messages: HashMap::new(),
             account_id: id.clone(),
+            outbox_rx,
+            messages: HashMap::new(),
         }
     }
 
-    pub fn extend(&mut self, outgoing: &mut mpsc::Receiver<SendMessage>) {
+    pub fn extend(&mut self) {
```
jakmeier, line 823: Again, an `extend` method without an argument looks very un-rusty to my eye :shrug:

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -247,16 +246,17 @@ impl SignatureManager {
         public_key: PublicKey,
         epoch: u64,
         sign_rx: Arc<RwLock<mpsc::Receiver<IndexedSignRequest>>>,
+        msg: &MessageChannel,
```
jakmeier, line 249: nit: Generally speaking, it's better to take ownership of the parameter when it's needed inside, pushing the clone to the caller side. This may allow the caller to avoid a clone.

jakmeier (approved):
Nice, removing the GC is great!

The code looks good, just small nits here and there.

One conceptual question though:

Can't we have messages that have already passed the message filter by the time a protocol finishes? I think the GC approach caught all these cases, as the addition to the so-called GC hashmap was done synchronously in the same task. But in your new code, isn't there a race condition that allows additional messages to slip through?

chain-signatures/node/src/protocol/signature.rs
ChaoticTempest: I think I've heard both sides of this. I guess it's more about what is more expensive to clone. Since `MessageChannel` is more easily cloneable we should have a move instead, but otherwise it should be a reference if it's just read from like a byte vector that can be gigantic if cloned.

ChaoticTempest:
> Can't we have messages that have already passed the message filter by the time a protocol finishes? I think the GC approach caught all these cases, as the addition to the so-called GC hashmap was done synchronously in the same task. But in your new code, isn't there a race condition that allows additional messages to slip through?

Good question, yeah that's right. So what we can add is before receiving the messages, the protocol now needs to filter out its messages there too based on the filter to remedy this. The filter can be sitting in the filter_rx channel so we need to immediately update it. Which is possible since we take a lock on the inbox.

This should now be race condition free since the `MessageReceiver::recv` is always called before a protocol goes to process a message.
 
In the future, if we were to move this to a more per-protocol-as-separate-task kind of design with https://github.com/sig-net/mpc/issues/273, then Messages will still filter the incoming messages, since `MessageReceiver` will no longer exist, and instead `Message` channel would track a list of subscribers based on IDs. Before making it to the subscriber, the messages would have already been filtered based on that model

volovyks (dismissed):
Great! Good replacement for GC. And nice refactoring.

ppca (approved):

PR #278, opened 2025-03-08
feat: more robust polling of connections and disconnections

ChaoticTempest:
This makes our polling of node connections more robust such that there's now a task per node connection. Each task will manage individually that connection where the defualt ping-interval is 1s. This code is easily replaced in the future by substituting all the http calls with the more low level primitives like tokio's TLS stream, so no necessary changes to architecture is needed at that point.

A couple changes were made to the logic but it will work the same:
- If we observe new info from the contract such as `AccountId` or `url`, the connection will be dropped and be replaced by a new one with the updated info.
- Reordered the pinging of endpoints. Originally it was `/state` then `/msg` empty, but `/state` is potentially more expensive, so do `/msg` first. 
- There's a mesh state task that checks on all the statuses of the connection. This is a simple check and does it every `ping_interval / 2`.
- before, after resharing, the old connections were still maintained. This PR makes it so that we drop those old connections.
- no more potential connections: these are not necessary since they were only used in conjunction when the network is in resharing. Now, these are basically just the `active` connections since in resharing, only active ones are communicating with each other (i.e. only the new participants set in the `ResharingState`). This makes it a lot simpler for `MeshState` since it now only has `active` and `stable` values


chain-signatures/node/src/web/mod.rs
```diff
@@ -28,8 +28,8 @@ pub async fn run(
     sender: Sender<Ciphered>,
     protocol_state: Arc<RwLock<NodeState>>,
     indexer: Indexer,
-) -> anyhow::Result<()> {
-    tracing::info!("running a node");
+) {
+    tracing::info!("starting web server");
```
volovyks, line 32: Can we keep "node" here? We may have other web servers that are not nodes. And we already have Redis.

volovyks (approved):
Nice work, I like the design.
`is_stable` is still in focus here, but that is a separate issue

ChaoticTempest:
Merging this one for now as I'd like to see how this looks like on devnet, but feel free to leave anymore reviews here and I'll make subsequent changes in future PRs.

chain-signatures/node/src/mesh/connection.rs
```diff
@@ -1,245 +1,221 @@
...
+            }
+        }
     }
 
-    pub async fn potential_participants(&self) -> Participants {
-        self.potential_connections.read().await.clone()
-    }
+    /// Drop connections that are not in the active connections list. Dropped connections
+    /// are no longer polled for their status.
+    fn drop_connections(&mut self, active_conn: HashSet<Participant>) {
```
ppca, line 193: active_conn is a bit misleading in the sense that I first thought it meant the participants that are active, but it actually means the current participants list, not whether it's active?

chain-signatures/node/src/mesh/connection.rs
```diff
@@ -1,245 +1,221 @@
...
-        tracing::debug!(
-            "Pool.establish_participants set participants to {:?}",
-            self.connections.read().await.clone().keys_vec()
-        );
-    }
 
-    async fn set_participants(&self, participants: &Participants) {
-        *self.connections.write().await = participants.clone();
+        // drop the connections that are not in the seen list
+        self.drop_connections(seen);
```
jakmeier, line 154: Hm, without the comment, I would misunderstand the function signature and what it does. Whenever I need a comment like this, I take it as a sign to change the function name to make the comment superfluous.

How about `retain_connections` instead of `drop_connections`?

chain-signatures/node/src/mesh/connection.rs
```diff
@@ -1,245 +1,221 @@
...
-            .await
-            .get(participant)
-            .is_some_and(|state| match state {
-                StateView::Running { is_stable, .. } => *is_stable,
-                _ => false,
-            })
+        for participant in remove {
+            tracing::info!(target: "net[pool]", node = ?participant, "node connection dropping");
+            self.connections.remove(&participant);
+        }
```
jakmeier, line 204: How about using [retain](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.retain)?

Something like this should work
```
self.connections.retain(|&participant, _connection|{
    active_conn.contains(participant)
})
```

jakmeier: And I believe the logging here is currently be doubled by the `impl Drop for NodeConnection` logging anyway.

chain-signatures/node/src/mesh/mod.rs
```diff
@@ -11,84 +12,64 @@ pub mod connection;
...
+    pub fn state(&self) -> &Arc<RwLock<MeshState>> {
+        &self.state
     }
 
-    pub async fn run(
-        mut self,
-        contract_state: Arc<RwLock<Option<ProtocolState>>>,
-    ) -> anyhow::Result<()> {
+    pub async fn run(mut self, contract_state: Arc<RwLock<Option<ProtocolState>>>) {
+        let mut interval = tokio::time::interval(self.ping_interval / 2);
```
jakmeier, line 64: Why self.ping_interval / 2?

PR #287, opened 2025-03-13
feat: protocol state sync

ChaoticTempest:
This is the initial work done for https://github.com/sig-net/mpc/issues/130. This only works for the cases where the node is proposing a new presignature or signature generation.

Some stuff are not implemented since it requires a whole different slew of work to be done (which will be done in a different PR). For example, doing triple/presignature removal by state sync requires the node to know exactly who the owner of the triple or presignature is so a node does not just outright tell another node to remove a triple/presignature that's not theres. This info is something we do not store, since we only store if we own a triple or presignature.

### Changes

This PR introduces `SyncChannel`, `SyncTask`, `SyncUpdate` and `SyncView` for public usage:
- `SyncUpdate` is a request to a other node to get back a `SyncView`. 
    - `SyncUpdate` has a list of owned triples/presignatures that the sending node has.
    - `SyncView` is the returned list of triple/presignature shares that the receiving node has.
    - Future PR will have the receiving node also remove expired shares if they're not present in the `SyncUpdate` 
- `SyncChannel` forwards requests to `SyncTask` from:
    - axum web server on receiving a `SyncUpdate` request. 
    - protocol managers on calling `take_two_triple` or `take_presignature`.
- `SyncTask` processes all these requests and synchronously replies to them.

Some other niceties were added, namely `NodeStateWatcher` which is a way to block a task from continuing until we have the info that we want. So the `SyncTask` will wait until we're in the running state, and then be able to get info like threshold and what participant is the node. We can reuse this with other tasks in the future instead of just blindly trying to call into the `RwLock<ProtocolState>`

Also, non-proposed protocol taking from `SyncChannel`, for following proposed protocols by other nodes, will be done when each protocol task has the ability to block itself before continuing.

### DIFF
--- a/chain-signatures/node/src/protocol/sync/mod.rs
+++ b/chain-signatures/node/src/protocol/sync/mod.rs
@@ -1,8 +1,11 @@
 use std::collections::{HashMap, HashSet};
 use std::sync::Arc;
+use std::time::Instant;
 
 use cait_sith::protocol::Participant;
+use rand::rngs::StdRng;
 use rand::seq::IteratorRandom;
+use rand::SeedableRng;
 use serde::{Deserialize, Serialize};
 use tokio::sync::{mpsc, oneshot, RwLock};
 use tokio::task::JoinSet;
@@ -11,7 +14,7 @@ use crate::mesh::MeshState;
 use crate::node_client::NodeClient;
 use crate::storage::{PresignatureStorage, TripleStorage};
 
-use super::contract::primitives::{intersect, intersect_vec};
+use super::contract::primitives::{intersect, intersect_vec, Participants};
 use super::presignature::{Presignature, PresignatureId};
 use super::triple::{Triple, TripleId};
 
@@ -20,6 +23,19 @@ pub struct SyncUpdate {
     // from: Participant,
     triples: Vec<TripleId>,
     presignatures: Vec<PresignatureId>,
+}
+
+impl SyncUpdate {
+    pub fn empty() -> Self {
+        Self {
+            triples: Vec::new(),
+            presignatures: Vec::new(),
+        }
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.triples.is_empty() && self.presignatures.is_empty()
+    }
 }
 
 #[derive(Debug, Serialize, Deserialize)]
@@ -70,7 +86,7 @@ impl SyncTask {
         (syncer, task)
     }
 
-    async fn process_sync(&self, update: SyncUpdate) -> anyhow::Result<SyncView> {
+    async fn process_sync(&self, update: SyncUpdate) -> SyncView {
         // TODO: check that `from` actually owns the triples/presignatures.
         // TODO: log the errors from storage.
         // TODO: make each process_sync a separate task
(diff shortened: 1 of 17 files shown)


chain-signatures/node/src/protocol/sync/mod.rs
```diff
@@ -0,0 +1,630 @@
...
+                        let mut active = state.active.clone();
+                        // do not broadcast to me
+                        active.remove(&me);
+                        active
+                    };
+
+                    let start = Instant::now();
+                    let task = tokio::spawn(broadcast_sync(self.client.clone(), update, active));
+                    broadcast = Some((start, task));
+                }
```
jakmeier, line 227: Ok, so this is not what I thought we discussed. My impression was, we only want to run sync when a link goes up. 

In our setup, I was expecting that means:

1. When node A starts, it initiates sync with all other peers (bidirectional)
2. When Node B wasn't able to ping node C for some time and then it works again for the first time, then nodes B and C initiate sync between each other.

I thought these would be the only time we do a sync.

Doing it all the time would be quite a different approach. I suspect it would be too expensive but can't be sure. We can try and measure your approach if you want.

But the original idea was that each node maintains an up-to-date dictionary of who has which of their owned pre-signatures and triples by:

1. Syncing on boot
2. Syncing anytime we lost connection for a while
3. Updating the dictionary whenever we consume a triple or pre-signature

That should keep it up to date with far fewer overhead.

chain-signatures/node/src/protocol/sync/mod.rs
```diff
@@ -0,0 +1,630 @@
...
+        }
+        for id in update.presignatures {
+            let entry = self.owned_presignatures.entry(id).or_default();
+            entry.push(self.me);
+        }
+
+        // Update the cache with the info from other participants:
+        for (p, view) in views {
+            for triple in view.triples {
+                let entry = self.owned_triples.entry(triple).or_default();
```
jakmeier, line 108: (for future PRs, maybe add TODO for now)

When this entry doesn't exist, yet, it means the peer holds a triple that we no longer have. There should be some mechanism to have them delete it.

It could be an explicit message, requesting a deletion of all IDs we find here to be obsolete.

Or it could be that we do nothing here and instead send a list of which IDs we currently hold on each sync and then the peer can delete it by iterating through all foreign IDs and removing what is obsolete then.

jakmeier (approved):
It's a big PR but you kept it rather nice and clean! :)

I like the approach to do it in several PRs, adding functionality in steps. As such, I'm in favour of merging this sooner rather than later and testing/improving it from there.

However, it seems you are syncing in a loop, over and over, rather than event based as described in the first paragraph of #130. 

@ChaoticTempest, @volovyks 
Am I right that we still want to eventually have it event-based? Or are we going with a different approach?

chain-signatures/node/src/protocol/sync/mod.rs
```diff
@@ -0,0 +1,630 @@
...
+use crate::storage::{PresignatureStorage, TripleStorage};
+
+use super::contract::primitives::{intersect, intersect_vec, Participants};
+use super::presignature::{Presignature, PresignatureId};
+use super::triple::{Triple, TripleId};
+
+/// How long for a synchronous [`SyncUpdate`] to get back a [`SyncView`] request before timeout.
+const REQUEST_UPDATE_TIMEOUT: Duration = Duration::from_millis(50);
+/// How long for a synchronous [`ProtocolRequest`] to get back a [`ProtocolResponse`] before timeout.
+const REQUEST_PROTOCOL_TIMEOUT: Duration = Duration::from_millis(50);
```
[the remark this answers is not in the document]
ChaoticTempest: yep, I measured locally on my machine while 3 nodes were running at the same time. The average is 1ms on each request

chain-signatures/node/src/protocol/sync/mod.rs
ChaoticTempest: > Ok, so this is not what I thought we discussed. My impression was, we only want to run sync when a link goes up.

Ahh, I might've missed that part when reading over that issue. But I'd say one advantage of constantly syncing like this is that we have this single source of truth for which we can take triples/presignatures from in place of sending an init message.

I think we can pair that sync on uplink with this too, since one very rare case I see potentially happening is that the other node goes offline and then online, but our node hasn't observed it because we only do poll-based active connections right now. I don't think this will likely happen that often, so not a huge concern with this.

So in a future PR, I can make it so that when we observe a change on the mesh state, we can sync up.

> Doing it all the time would be quite a different approach. I suspect it would be too expensive but can't be sure. We can try and measure your approach if you want.

So far, just based on what I've measured locally on my m4 mac mini, doing the sync every 250ms isn't too bad since it takes roughly less than 25ms (but this is without network delay due to it being on the same machine) on both 5 nodes and 10 nodes setup.

ChaoticTempest:
> Am I right that we still want to eventually have it event-based? Or are we going with a different approach?

I wrote it in the other comment, but I think we can have both since sync operation isn't that expensive to do constantly.

chain-signatures/node/src/protocol/sync/mod.rs
volovyks: @ChaoticTempest Your approach should work in practice and yield good results. It also aligns with our current design, where we have a "verified" source of Ts and Ps. However, in the long run, we may still need to implement Ack/Nack. Synchronizing every N ms can mask underlying system issues. Performing synchronization when it's actually needed (e.g., after a node goes offline) may be more challenging but aligns better with a robust state machine approach.
Please, experiment with it. I'm ok with what we have here if the alternative is too hard to implement.

ChaoticTempest: The thing is, Ack/Nack devolves down to this synchronization anyways, but instead of init and ack/nack messages, it's `SyncUpdate` and `SyncView` respectively. The only difference is that this sync mechanism has a bigger delay on requesting the protocol, so the view of who has what protocol can be construed, but the chances of it happening in 250ms is very low.

ChaoticTempest: But then the participant list can still have inconsistent views since every node then needs to have the same sync view. Then in that case we still do need the init message.

So what I imagine we can do is on top of syncing, the proposer aggregates the participant list for their own presignature/triple, and sends out this info as well. This acts as the init message but passively being done in the background. The node can send back ack for these if the node's view are the same, and nack otherwise. That way we then have a series of own triples/presignatures that are ready to be used at any time with this syncing.

Additionally, if a node were to crash and hop back online, they can send a bail message to other nodes if they were already a part of the participant set, but do not have a sync record of the protocol (since this is all in-memory at the moment).

jakmeier: > So far, just based on what I've measured locally on my m4 mac mini, doing the sync every 250ms isn't too bad since it takes roughly less than 25ms (but this is without network delay due to it being on the same machine) on both 5 nodes and 10 nodes setup.

Thanks for this data! How many triples and pre-signatures were in store for this? How many nodes did you sync up with in 25ms?
I'm not so much concerned about a single sync but about the sum of network bandwidth we are taking up for syncs that is no longer available to other tasks.

jakmeier:
> I have benchmarked this locally and have not seen any degredation of performance, so I believe we can just push this to dev for now and see where the pitfalls are.
> 
> @ppca @volovyks let me know if there's anything you guys spot that should be changed later.

Trying things out on dev seems like a good approach to move forward :+1:

chain-signatures/node/src/protocol/sync/mod.rs
```diff
@@ -0,0 +1,630 @@
...
+impl SyncView {
+    fn empty() -> Self {
+        Self {
+            triples: HashSet::new(),
+            presignatures: HashSet::new(),
+        }
+    }
+}
+
+struct SyncCache {
```
volovyks, line 68: Why do we need to cache data here? It adds complexity and introduces an additional source of truth for triples and presignatures. I would say we should always work with Redis directly since it is already a cashing solution.

volovyks: Is is a tmp solution until we extend data in Redis?

chain-signatures/node/src/protocol/sync/mod.rs
```diff
@@ -0,0 +1,630 @@
...
+        }
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.triples.is_empty() && self.presignatures.is_empty()
+    }
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct SyncView {
```
volovyks, line 54: What is the benefit of separating SyncView and SyncUpdate?

PR #303, opened 2025-03-27
performance metrics: fix triple metrics

ppca:
Problems:
1. triple before poke delay showing constant value all the time
2. triple accrued delay p95 is too low (<10ms)

Changes:
1. modify the buckets for the triple metrics
This is important because these metrics are histogram metrics, and the values in the bucket is used as the upper bound of each value interval. And the metric counts how many observed values fall in each interval.
Previously the triple before poke delay metric's bucket has the largest upper bound set too low, which explains why the metric shows a constant value for p20, p50, p95, because the observed values are mostly greater than the biggest upper bound.

3. Lock the poked_latest field of triple generator.
This is necessary because in the current code, the triple generator can be poked by multiple tasks at the same time. Thus not locking it up will lead to the values being inconsistent. I checked our triple_acccrued_delay and triple_pokes_cnt, they are not set most of the time because the poked_latest field was not locked, and modifications was not saved properly hence in the last poke somehow the field is still None.

### DIFF
--- a/chain-signatures/node/src/protocol/triple.rs
+++ b/chain-signatures/node/src/protocol/triple.rs
@@ -287,7 +287,7 @@ impl TripleGenerator {
                         triple_owner == me
                     };
 
-                    tracing::error!(
+                    tracing::info!(
                         id = self.id,
                         ?me,
                         triple_is_mine,
@@ -312,10 +312,6 @@ impl TripleGenerator {
                             let total_pokes = total_pokes + 1;
                             triple_accrued_wait_delay_metric.observe(total_wait.as_millis() as f64);
                             triple_pokes_cnt_metric.observe(total_pokes as f64);
-                            tracing::error!(
-                                "triple total pokes is {total_pokes}, total wait is {:?}",
-                                total_wait
-                            );
                         }
                     }
                     triple_poke_cpu_time_metric
--- a/chain-signatures/node/src/protocol/triple.rs
+++ b/chain-signatures/node/src/protocol/triple.rs
@@ -46,7 +46,7 @@ pub struct TripleGenerator {
     pub protocol: Arc<TripleProtocol>,
     pub timestamp: Arc<RwLock<Option<Instant>>>,
     pub timeout: Duration,
-    poked_latest: Option<(Instant, Duration, u64)>,
+    poked_latest: Arc<RwLock<Option<(Instant, Duration, u64)>>>,
     generator_created: Instant,
 }
 
@@ -74,7 +74,7 @@ impl TripleGenerator {
             protocol,
             timestamp: Arc::new(RwLock::new(None)),
             timeout: Duration::from_millis(timeout),
-            poked_latest: None,
+            poked_latest: Arc::new(RwLock::new(None)),
             generator_created: Instant::now(),
         })
     }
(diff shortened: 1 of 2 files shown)


chain-signatures/node/src/protocol/triple.rs
```diff
@@ -46,7 +46,7 @@ pub struct TripleGenerator {
     pub protocol: Arc<TripleProtocol>,
     pub timestamp: Arc<RwLock<Option<Instant>>>,
     pub timeout: Duration,
-    poked_latest: Option<(Instant, Duration, u64)>,
+    poked_latest: Arc<RwLock<Option<(Instant, Duration, u64)>>>,
```
jakmeier, line 49: It took me a while to understand this. But I think I understand the problem now and your solution works to solve it.

I don't think it's because of multiple tasks running concurrently with the same triple id. I think the triple manager ensures there is only one task per id ongoing. Although, the way it's implemented right now, we sadly cannot rely on Rust's ownership model to guarantee that there is only one task accessing the data.

But I found that even though there is only one task at the time, when we spawn the task to poke, we clone the full TripleGenerator and make modifications on that clone. Then, the changes made in `async fn execute(`  are done to that clone. But the original generator, stored in the `TripleTasks::generators` hash map, is never updated.

Modifications still work for fields that are `Arc`, since a clone only changes the reference count to the same underlying data, rather than actually copying the underlying data.
Therefore, your solution fixes the issue because it uses an Arc.

However, I think this is a footgun in how the code is written right now. Your solution avoids the footgun for this particular case. But could we instead improve the code to prevent similar mistakes in the future?

I would suggest we put `Arc<RwLock<TripleGenerator>>` inside the `TripleTasks::generators` hash map. And then remove `Arc<RwLock<>>` from all fields (timestamp, protocol and poked_last). That way, the spawned tasks will always modify the data inside the hash map, rather than on a temporary copy.
(cc @ChaoticTempest )

Also, if we do it like this, we get rid of per-field inner mutability. Which means we can rely on Rust's mutability checks again. It will ensure that no two tasks have mutable access to different fields of the same generator. That would make the code much easier to reason about, IMO.

chain-signatures/node/src/protocol/triple.rs
```diff
@@ -307,12 +304,15 @@ impl TripleGenerator {
...
-                        triple_pokes_cnt_metric.observe(total_pokes as f64);
+                    {
+                        let poked_latest = self.poked_latest.read().await;
+                        if let Some((last_poked, total_wait, total_pokes)) = *poked_latest {
+                            let elapsed = generator_poke_time - last_poked;
+                            let total_wait = total_wait + elapsed;
+                            let total_pokes = total_pokes + 1;
+                            triple_accrued_wait_delay_metric.observe(total_wait.as_millis() as f64);
+                            triple_pokes_cnt_metric.observe(total_pokes as f64);
+                        }
```
jakmeier, line 315: Would it make sense to report cases where the first poke already yields a `Return`?
Not sure if it *can* happen. But if it does, I would want to see it reflected in metrics.

```suggestion
                        let poked_latest = self.poked_latest.read().await;
                        if let Some((last_poked, total_wait, total_pokes)) = *poked_latest {
                            let elapsed = generator_poke_time - last_poked;
                            let total_wait = total_wait + elapsed;
                            let total_pokes = total_pokes + 1;
                            triple_accrued_wait_delay_metric.observe(total_wait.as_millis() as f64);
                            triple_pokes_cnt_metric.observe(total_pokes as f64);
                        } else {
                            // This is the first and only poke
                            triple_accrued_wait_delay_metric.observe(0.0);
                            triple_pokes_cnt_metric.observe(1.0);
                        }
```

jakmeier (approved):
Nice debugging! Yeah I think this solves the two problems.

You can merge this as is, if you want. But I would really prefer to cleanup the inner mutability on generators, as I elaborated in a comment inside the code. (Can be done in a follow-up PR, too, if that's easier.)

chain-signatures/node/src/protocol/triple.rs
ChaoticTempest: yep, this is definitely tech debt. In the view of https://github.com/sig-net/mpc/issues/273, each generator would just be its own separate task where this would not be as much of a footgun. The issue is right now, the triple manager is managing the going back and forth between waits and errors which should be managed directly by a singular task

PR #304, opened 2025-03-28
feat: sync remove storage

ChaoticTempest:
This now makes our sync feature removes items from storage such that the owner can ping a node with an update with what shares they own. The receiving node then chooses to delete the set of protocols that are present in the node but not in the owning node.

Storage changes:
- reserved set: initial set of reservations, such that a `TripleSlot` or a `PresignatureSlot` is created at the start, and then the reservation is removed once `insert`ed.
- owner sets: a mapping of an owner to a set of protocols. Use for the removal of protocols if outdated via `remove_outdated`.

What's not implemented:
- we can probably remove the mine set now since owner set is a more generalized version of that. The nice part of mine set is that you don't need to keep track of your own participant info where we can immediately create the storage object from node startup.
- checking whether or not the update message is indeed from the owner node
- syncing on connection changes, and instead currently relying on syncing every 10s.


ChaoticTempest:
heh, this AI review thing is pretty handy

chain-signatures/node/src/storage/presignature_storage.rs
```diff
@@ -49,43 +103,173 @@ impl PresignatureStorage {
...
-            elseif redis.call('HEXISTS', used_key, presig_id) == 1 then
-                return {err = 'Presignature ' .. presig_id .. ' is already used'}
+            -- if the presignature has NOT been reserved, then something went wrong when acquiring the
+            -- reservation for it via presignature slot.
+            if redis.call("SREM", reserved_key, presig_id) == 0 then
+                return {err = "WARN presignature " .. presig_id .. " has NOT been reserved"}
+            end
+
+            if redis.call('HEXISTS', used_key, presig_id) == 1 then
+                return {err = 'WARN presignature ' .. presig_id .. ' is already used'}
```
volovyks, line 268: "... already inserted", "used" can be misunderstood

chain-signatures/node/src/storage/presignature_storage.rs
```diff
@@ -49,43 +103,173 @@ impl PresignatureStorage {
...
+                storage: self.clone(),
+            }),
+            Err(err) => {
+                tracing::warn!(?err, "failed to reserve presignature");
+                None
+            }
+        }
+    }
+
+    pub async fn remove_outdated(
```
volovyks, line 174: looks good

volovyks (dismissed):
Great progress! I like the atomic design for removing Ts and Ps.

Representing current T and P generations as slots in storage is a good decision. Separating generated and in-progress Ts and Ps seems redundant.

I agree that we can remove mine storage info—it’s unnecessary if we store the owner.

Who will initiate the sync in the follow-up PRs? I believe it should always be a two-step process: if node A sends me its update, I should send mine in return.

P.S. I'm ok with merging this since the diff is huge and we should not keep it in a separate branch. Also, dev is not in a good state now. But it will be great if somebody else can take a look as well.

integration-tests/tests/cases/sync.rs
```diff
@@ -86,39 +86,22 @@ async fn test_protocol_sync_take() -> anyhow::Result<()> {
...
     }
     for i in 0..10 {
         let p = dummy_presignature(i);
         presignature_set.insert(p.id);
-        presignatures.insert(p, true, false).await.unwrap();
+        presignatures
+            .reserve(i, me)
+            .await
+            .unwrap()
+            .insert(p, me)
```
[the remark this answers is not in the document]
ChaoticTempest: nope, it returns a bool whether the insert was successful or not

chain-signatures/node/src/web/mod.rs
```diff
@@ -222,7 +222,7 @@ async fn bench_metrics() -> Json<BenchMetrics> {
 async fn sync(
     Extension(state): Extension<Arc<AxumState>>,
     WithRejection(Json(update), _): WithRejection<Json<SyncUpdate>, Error>,
-) -> Result<Json<SyncView>> {
-    let view = state.sync_channel.request_update(update).await;
-    Ok(Json(view))
+) -> Result<()> {
```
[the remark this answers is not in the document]
ChaoticTempest: Processing the sync does not yield any errors, so return an error code at all would be unnecessary here. If you're talking about whether or not the node receives the request, then that is already done on axum side.

chain-signatures/node/src/storage/presignature_storage.rs
```diff
@@ -6,10 +7,51 @@ use redis::{AsyncCommands, FromRedisValue, RedisError, RedisWrite, ToRedisArgs};
...
+
 // Can be used to "clear" redis storage in case of a breaking change
 const PRESIGNATURE_STORAGE_VERSION: &str = "v7";
 const USED_EXPIRE_TIME: Duration = Duration::hours(24);
 
+/// A pre-reserved slot for a presignature that will eventually be inserted.
+#[derive(Clone)]
+pub struct PresignatureSlot {
+    id: PresignatureId,
+    me: Participant,
```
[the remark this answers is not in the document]
ChaoticTempest: the me is used to determine later whether we are the owner or not. It doesn't have to be stored here, just that it's more convenient to do so.

chain-signatures/node/src/storage/presignature_storage.rs
ChaoticTempest: not sure what you mean. Used and inserted are two distinct states of stored items, so I'm not sure how they can be misunderstood, plus we don't log whether something is inserted but `stored`.

integration-tests/tests/cases/sync.rs
```diff
@@ -86,39 +86,22 @@ async fn test_protocol_sync_take() -> anyhow::Result<()> {
...
+            .await;
     }
 
     // Give it some time for sync to process the inserts
     tokio::time::sleep(Duration::from_secs(3)).await;
 
-    let mine = true;
```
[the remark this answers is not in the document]
ChaoticTempest: this test doesn't really do anything at the moment. It's just leftover after I moved the taking of stored protocols out of sync, but will be used in the future

chain-signatures/node/src/storage/presignature_storage.rs
```diff
@@ -6,10 +7,51 @@ use redis::{AsyncCommands, FromRedisValue, RedisError, RedisWrite, ToRedisArgs};
...
+            tracing::warn!(id = self.id, ?err, "failed to insert presignature");
+            false
+        } else {
+            true
+        }
+    }
+
+    // TODO: put inside a tokio task:
+    pub async fn unreserve(&self) {
+        let mut conn = match self.storage.connect().await {
```
ppca, line 41: should we instead make a unreserve() fn in PresignatureStorage?

chain-signatures/node/src/storage/presignature_storage.rs
```diff
@@ -49,43 +103,173 @@ impl PresignatureStorage {
...
+    pub async fn fetch_owned(&self, me: Participant) -> Vec<PresignatureId> {
+        let mut conn = match self.connect().await {
+            Ok(conn) => conn,
+            Err(err) => {
+                tracing::warn!(?err, "failed to connect to redis");
+                return Vec::new();
+            }
+        };
+
+        conn.sunion((&self.reserved_key, owner_key(&self.owner_keys, me)))
```
ppca, line 115: this is fetching either reserved or mine presignatures? What will this function be used for?

chain-signatures/node/src/storage/triple_storage.rs
```diff
@@ -41,37 +85,179 @@ impl TripleStorage {
...
-                return {err = "Triple " .. triple_id .. " has already been used"}
+            -- if the triple has not been reserved, then something went wrong when acquiring the
+            -- reservation for it via triple slot.
+            if redis.call("SREM", reserved_key, triple_id) == 0 then
+                return {err = "WARN triple " .. triple_id .. " has NOT been reserved"}
             end
 
+            if redis.call("HEXISTS", used_key, triple_id) == 1 then
+                return {err = "WARN triple " .. triple_id .. " has already been used"}
+            end
```
ppca, line 257: if triple's triple_key already exists then we should warn too?

chain-signatures/node/src/storage/triple_storage.rs
```diff
@@ -137,8 +327,9 @@ impl TripleStorage {
                 return {err = "Triple " .. ARGV[2] .. " is missing"}
             end
 
-            -- Delete the triples from the hash map
+            -- Delete the triples from the hash map and reserved slots
```
ppca, line 330: why don't we delete the triples from reserved slot when it is being inserted in insert() fn?

ppca (commented):
Is this PR mostly adding functions into Storage and not really using any of them?

chain-signatures/node/src/storage/presignature_storage.rs
ChaoticTempest: Unreserve is only directly associated with a slot. You can't unreserve something that has not been reserved.

chain-signatures/node/src/storage/presignature_storage.rs
ChaoticTempest: it's used in sync when you go to process an update. You take the union of `Mine` and `Reserved`, and send out the update to say these are the list of protocols that we currently own to other nodes. Other nodes will see this list, compare with their own versions and invalidate the ones that the owner no longer tracks (i.e. non-owned nodes will delete their shares that the owners don't have anymore).

chain-signatures/node/src/storage/triple_storage.rs
ChaoticTempest: we do, this call was leftover for some reason, but it doesn't affect anything

chain-signatures/node/src/storage/triple_storage.rs
ChaoticTempest: `reserve` already catches this so no need to add the extra check.

ChaoticTempest:
> Is this PR mostly adding functions into Storage and not really using any of them?

@ppca They're directly used in the managers and sync module in this PR

chain-signatures/node/src/web/mod.rs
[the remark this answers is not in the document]
volovyks: Yes, the script does not return any errors, but Redis can. We can also lose connection with the Redis server.
I'm ok with keeping it as it is for now, but let's change this line to `error`. This is an unexpected behaviour.
```rust
tracing::warn!(?err, "failed to remove outdated presignatures");
```
(for both triples and presignautres)

chain-signatures/node/src/storage/presignature_storage.rs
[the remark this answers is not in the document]
volovyks: It can confuse people. We were using `me` only in the context of "this is the current node PArticipant id". Here, it can be any node.

ppca (approved):

integration-tests/benches/store.rs
```diff
@@ -0,0 +1,263 @@
...
+        );
+        let (sync_channel, sync) = SyncTask::new(
+            &client,
+            triples.clone(),
+            presignatures.clone(),
+            mesh_state.clone(),
+            watcher.clone(),
+        );
+
+        // let sync_handle = tokio::spawn(sync.run());
```
jakmeier, line 123: Why is this code commented out?

integration-tests/tests/cases/sync.rs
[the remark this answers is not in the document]
jakmeier: Maybe you can add a comment? Now this test just looks a bit odd. And who knows how fast the future when we need it again will come ;)

integration-tests/tests/cases/sync.rs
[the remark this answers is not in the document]
jakmeier: Should we assert that it was successful then?

jakmeier (approved):
Looking pretty good, nice to see this come forward!

chain-signatures/node/src/storage/presignature_storage.rs
```diff
@@ -6,10 +7,51 @@ use redis::{AsyncCommands, FromRedisValue, RedisError, RedisWrite, ToRedisArgs};
...
+        {
+            tracing::warn!(id = self.id, ?err, "failed to insert presignature");
+            false
+        } else {
+            true
+        }
+    }
+
+    // TODO: put inside a tokio task:
+    pub async fn unreserve(&self) {
```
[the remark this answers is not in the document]
ChaoticTempest: I was going to add that at first, but it required more refactorings since when we are able to have access to this is only when we have a `&self` reference to it

integration-tests/tests/cases/sync.rs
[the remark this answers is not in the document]
ChaoticTempest: will do in another PR

PR #306, opened 2025-03-31
feat: Helios indexer

ppca:
This PR uses Helios client to subscribe to new block heads and once a new block head is available, the client fetches the block's receipts and filter for the logs matching signature request events.

This does not include implementation to wait for finality. It will come in a followup PR.

Moving web3 to alloy will also come in a separate PR.


chain-signatures/node/Cargo.toml
```diff
@@ -28,6 +28,10 @@ semver = "1.0.23"
 sysinfo = "0.32.0"
 tokio-retry = "0.3"
 tracing-stackdriver = "0.10.0"
+alloy = "0.9.1"
+alloy-sol-types = "0.8.24"
+helios = { git = "https://github.com/a16z/helios", branch = "master" }
```
ChaoticTempest, line 33: we should fork this repo into our org, especially since we're targeting master branch here which will very likely change from underneath us and we'd be beholden to all those breaking changes

integration-tests/src/local.rs
```diff
@@ -175,10 +182,23 @@ impl Node {
...
+                .eth
+                .clone()
+                .map(|eth| eth.execution_rpc_http_url.clone()),
+            eth_contract_address: config
+                .cfg
+                .eth
+                .clone()
+                .map(|eth| eth.contract_address.clone()),
+            eth_network: None,
         };
```
ChaoticTempest, line 202: seems like this is getting redone enough to warrant us doing a function for this conversion from the config object

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -104,10 +136,35 @@ pub struct EthSignRequest {
     pub key_version: u32,
 }
 
-fn sign_request_from_filtered_log(log: web3::types::Log) -> anyhow::Result<IndexedSignRequest> {
+sol! {
```
ChaoticTempest, line 139: is this right that this is compatible with Eth even though it's solana types?

ppca: this is actually solidity code gen loll

ailisp (dismissed):
Looks good!

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -254,138 +322,141 @@ pub async fn run(
...
-                node_near_account_id.clone(),
+async fn process_block(
+    block_number: u64,
+    client: &EthereumClient<FileDB>,
+    eth_contract_addr: Address,
+    sign_tx: mpsc::Sender<IndexedSignRequest>,
+    node_near_account_id: AccountId,
+) -> anyhow::Result<()> {
+    let Some(block_receipts) = client
+        .get_block_receipts(BlockTag::Number(block_number))
```
jakmeier, line 422: side note: I believe this has to do external HTTP requests. This could potentially be slow. We could think about a performance metric to see how much of the indexer delay comes from this step.

since it looks like the indexer delay can be quite significant, we should keep out an eye for which metrics we are missing

jakmeier (approved):
Looks mostly good, please take a look at my questions in comments.

Also, have you already thought about how we can compare before/after performance? Like, are there specific metrics we can look at to find out if the indexer performance is still the same, better, or worse with Helios?

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -254,138 +322,141 @@ pub async fn run(
...
-                    }
-                }
-            }
-            let Ok(unsubscribed) = filtered_logs_sub.unsubscribe().await.inspect_err(|err| {
-                tracing::warn!(
-                    "Failed to unsubscribe from logs: {:?}, will reconnect to websocket",
-                    err
-                );
-            }) else {
-                break;
```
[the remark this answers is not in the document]
ppca: Because I changed the logic to be: subscribe to new heads and once a new head comes in, we process that new block. 
The way helios subscribe to new heads is a receiver rather than a subscription stream, and it's something that comes with the light client progressing on the blockchain, not really fetching from a remote source. So I figure it won't have the same issue of subscription being closed by remote sources like we had before.

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -254,138 +322,141 @@ pub async fn run(
...
     };
 
+    let network = Network::from_str(eth.network.as_str())
+        .map_err(|err| anyhow::anyhow!("Network input incorrect: {:?}", err))?;
+
+    let mut client: EthereumClient<FileDB> = EthereumClientBuilder::new()
+        .network(network)
+        .consensus_rpc(&eth.consensus_rpc_http_url)
+        .execution_rpc(&eth.execution_rpc_http_url)
+        .data_dir(PathBuf::from("/tmp/helios"))
```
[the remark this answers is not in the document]
ppca: This is the place the light client will look for historical data. So this need to be ideally persisted. If wiped, then the light client will start sync from a checkpoint hardcoded in Helios code, and it's gonna be slower to sync up.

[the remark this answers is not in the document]
ppca: And yeah, should make it configurable. I'll do that.

ppca:
> 
> * Is there a way to filter events on Helios/Alloy side? The same way we did that with Web3.
No, they currently don't support that. Does not seem like they plan to support it in near future.
> * Have you tested stability? Can the indexer wait for new blocks for a long (infinite) time?
Tested with Sepolia on my local laptop. Seem to be doing alright.

ppca:
Let's not merge this until https://github.com/sig-net/mpc/pull/312 is merged

chain-signatures/node/src/indexer_eth.rs
ppca: Done.

jakmeier (approved):
I haven't done a full 2nd review, as it's hard to review the changes after a force-push and new changes being spread across many small commits...

But my previous comments seem to be addressed, so I'll approve again :)

PR #308, opened 2025-04-02
use alloy instead of web3 whenever possible

ppca:
except web3::signing::recover


chain-signatures/node/src/rpc.rs
```diff
@@ -4,27 +4,36 @@ use crate::protocol::signature::SignRequest;
...
 use mpc_primitives::Signature;
 
+use alloy::{
+    contract::{ContractInstance, Interface},
+    dyn_abi::DynSolValue,
+    network::EthereumWallet,
+    primitives::U256,
+    providers::ProviderBuilder,
+    transports::http::{Client as ReqwestClient, Http},
+};
```
jakmeier, line 24: nit: I don't see many other cases of more-than-one levels of import group nesting. Not sure if anybody cares, I think we are already quite inconsistent with imports, which is fine by me. But you might want to check if you are using rust analyzer settings in your editor that generates this. I think if you set `imports.granularity.group` to `module`, instead of crate, it will be more in line with existing code.

https://rust-analyzer.github.io/book/features.html?highlight=import#import-granularity

chain-signatures/node/src/rpc.rs
```diff
@@ -366,33 +401,37 @@ impl NearClient {
...
-            contract,
-            account_sk: web3::signing::SecretKey::from_str(&eth.account_sk)
-                .expect("failed to parse eth account sk, should not begin with 0x"),
-        }
+
+        // Get `abi` from the artifact.
+        let abi_value = json.get("abi").expect("Failed to get ABI from artifact");
+        let abi = serde_json::from_str(&abi_value.to_string()).unwrap();
+
+        // Create a new `ContractInstance` of the `Counter` contract from the abi
```
jakmeier, line 428: Counter contract? Is this maybe a copy-paste leftover?

jakmeier (approved):
Nice, I think alloy will be a better dependency for us long-term.

LGTM

chain-signatures/node/src/rpc.rs
ppca: ah yeah. I'll remove it.

ailisp (approved):
lgtm

chain-signatures/crypto/Cargo.toml
```diff
@@ -14,4 +14,4 @@ sha3.workspace = true
 getrandom = { version = "0.2.12", features = ["custom"] }
 
 [dev-dependencies]
-web3.workspace = true
```
volovyks, line 17: Would it be hard to get rid of web3 completely? It's a heavy crate, and our Rust analyzer is already slow.

volovyks (approved):

PR #310, opened 2025-04-03
solana indexer and response

ailisp:


chain-signatures/node/src/indexer_sol.rs
```diff
@@ -0,0 +1,271 @@
...
+        tracing::warn!("solana indexer is disabled");
+        return Ok(());
+    };
+
+    tracing::info!("running solana indexer");
+    let program_id = Pubkey::from_str(&sol.program_address)?;
+    
+    let keypair = Keypair::from_base58_string(&sol.account_sk);
+
+    let cluster = Cluster::Custom(sol.rpc_url.clone(), sol.rpc_url.replace("http", "ws"));
```
volovyks, line 208: Do we need an API key?

chain-signatures/node/src/indexer_sol.rs
```diff
@@ -0,0 +1,271 @@
...
+}
+
+async fn subscribe_to_program_events<C: Deref<Target = Keypair> + Clone>(
+    program: &Program<C>,
+    sign_tx: mpsc::Sender<IndexedSignRequest>,
+    node_near_account_id: AccountId,
+) -> anyhow::Result<()> {
+    tracing::info!("Subscribing to program events");
+    let (sender, mut receiver) = mpsc::unbounded_channel();
+    let event_unsubscriber = program
```
volovyks, line 230: On Ethereum, we are rewriting Web3 code using Alloy crate because it works with proofs solution (Helios). Does this crate support the future integration of proofs?

chain-signatures/node/src/indexer_sol.rs
ailisp: For production helius api key is part of the url. for local testing, shared rpc url is fine

chain-signatures/contract-sol/Anchor.toml
```diff
@@ -0,0 +1,21 @@
+[features]
+seeds = false
+
+[programs.localnet]
+chain_signatures_project = "4kZoBXmUBLveRS3sboGF557tYsR7SzLDsWmP4sz7VQEs"
```
volovyks, line 5: I saw that it is possible to give Solana contracts human readable names. Not a blocker, since we will test it on testnet for now.

volovyks (approved):
Looks fine, we can do the cleanup in a followup PR.

@jakmeier please, take a look. This PR is intended for Testnet release only and does not include any proofs/finality checks. 

@ailisp please, fix the conflicts.

P.S. I've created a couple of issues in solana contract repo, we need to move fixes here once they are available.

chain-signatures/node/src/rpc.rs
```diff
@@ -645,3 +691,76 @@ async fn try_publish_eth(
...
+        },
+        s: signature.s.to_bytes().into(),
+        recovery_id: signature.recovery_id,
+    };
+    
+    let tx = program
+        .request()
+        .signer(sol.payer.clone())
+        .accounts(SolanaRespondAccount {
+            responder: sol.payer.clone().try_pubkey().unwrap(),
```
jakmeier, line 722: Should we return an error rather than unwrap?

chain-signatures/node/src/rpc.rs
```diff
@@ -645,3 +691,76 @@ async fn try_publish_eth(
...
+    crate::metrics::SIGN_TOTAL_LATENCY
+        .with_label_values(&[Chain::Solana.as_str(), "solana"])
+        .observe(action.request.indexed.timestamp.elapsed().as_secs_f64());
+    crate::metrics::SIGN_RESPOND_LATENCY
+        .with_label_values(&[Chain::Solana.as_str(), "solana"])
+        .observe(timestamp.elapsed().as_secs_f64());
+    
+    if action.request.indexed.timestamp.elapsed().as_secs() <= 30 {
+        crate::metrics::NUM_SIGN_SUCCESS_30S
+            .with_label_values(&[Chain::Solana.as_str(), "solana"])
```
jakmeier, line 760: In all 4 calls to `with_label_values`, shouldn't the second label be the node account id?

chain-signatures/contract-sol/programs/chain-signatures-project/src/lib.rs
```diff
@@ -0,0 +1,163 @@
...
+            ),
+            program_state.signature_deposit,
+        )?;
+        
+        emit!(SignatureRequestedEvent {
+            sender: *requester.key,
+            payload,
+            key_version,
+            deposit: program_state.signature_deposit,
+            chain_id: 0,
```
jakmeier, line 57: Will we use this id in the future for routing between chains? Should we maybe pick a number for Solana that's unlikely to collide with any potential future EVM chains, while also being different from uninitialized (0) ?

jakmeier (commented):
Nice and clean implementation!

I left a few questions as comments. Also, I think some metrics use the wrong labels. Otherwise, it looks ready to merge.

ppca (approved):
lgtm. The metric label need to be changed to near node account id.

chain-signatures/node/src/indexer_sol.rs
```diff
@@ -0,0 +1,276 @@
...
+    let Some(sol) = sol else {
+        tracing::warn!("solana indexer is disabled");
+        return Ok(());
+    };
+
+    tracing::info!("running solana indexer");
+    let program_id = Pubkey::from_str(&sol.program_address)?;
+    let keypair = Keypair::from_base58_string(&sol.account_sk);
+
+    let cluster = Cluster::Custom(sol.rpc_url.clone(), sol.rpc_url.replace("http", "ws"));
```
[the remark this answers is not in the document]
ailisp: pattern from their examples, it need both a http url and a websocket url to create this cluster instance

chain-signatures/node/src/indexer_sol.rs
```diff
@@ -0,0 +1,276 @@
...
+    loop {
+        let unsub = subscribe_to_program_events(
+            &program,
+            sign_tx.clone(),
+            node_near_account_id.clone(),
+        ).await;
+        if let Err(err) = unsub {
+            tracing::warn!("Failed to subscribe to solana events: {:?}", err);
+        } else {
+            unsub.unwrap().unsubscribe().await;
```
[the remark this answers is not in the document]
ailisp: it's an imperfect solution. Based on my test and read anchor client code, subscription is keep alive, even when there is no messages it would wait in that tokio thread until there is a new message. But when there is a glitch on network or solana ws server, the subscription is interrupt. In that case, we unsubscribe. At that time, resub will fail for some seconds until network/solana's ws server recovers, but retry without a sleep always fail I guess due to hit the rate limit. Retry with 10s sleep can recover the subscription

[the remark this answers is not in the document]
ailisp: Ideally we can recover the subscription at no time

chain-signatures/node/src/indexer_sol.rs
```diff
@@ -0,0 +1,276 @@
...
+            tracing::info!("unsubscribing to solana events");
+        }
+        tokio::time::sleep(Duration::from_secs(10)).await;
```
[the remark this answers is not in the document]
ailisp: The subscription is kept alive unless network errors, so sleep 10s should happen very rare. The choice of 10 was below 12s of solana finality time with option "finalized", but added processing time it might be 19s as you said and miss a log

chain-signatures/node/src/rpc.rs
ailisp: It is a configuration error (when sol key, that used by node to respond is not provided or incorrect at startup), so I chose to end the node.

chain-signatures/node/src/rpc.rs
ailisp: Good catch

chain-signatures/contract-sol/programs/chain-signatures-project/src/lib.rs
ailisp: it is taken from https://github.com/sig-net/chain-signatures-solana-stub/blob/6ae5c760278756ee84a407d7e19bc12fb9d8e963/chain-signatures-project/programs/chain-signatures-project/src/lib.rs#L98, @esaminu do you think we can use 0x800001f5 here (as you chosen for sig-server.ts)

chain-signatures/node/src/indexer_sol.rs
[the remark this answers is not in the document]
ailisp: Do you have a better pattern in mind?

chain-signatures/node/src/rpc.rs
ChaoticTempest: we should probably make the types more concrete then in the future, such that this unwrap() isn't a code smell later or just in case a node messes up their setup and no SOL related tx makes it in for a while

ChaoticTempest:
We shouldn't be going back to 0.10 borsh versions, but since solana packages only take 0.10, we will use two different borsh versions. One for SOL and one for the rest of MPC. The reason why is because pre 1.0 versions of borsh have security vulnerabilities. For now, this would be constrained to just SOL

volovyks (approved):
@ailisp please, address all other comments and finalize the setup in a follow up PR.

chain-signatures/node/src/indexer_sol.rs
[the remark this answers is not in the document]
jakmeier: Ah, your right. I missed that the while loop only terminates on a disconnect. That makes sense then.

In that case, I think 10s is reasonable. If we see too many disconnects, that's an issue on its own.

PR #313, opened 2025-04-08
feat: sync on node connection change

ChaoticTempest:
Syncing will now happen on connection changes. This means that when a node decides to hop online, we will send out a sync. If it hops offline, we simply do nothing.

Note, this PR also makes `Mesh` listen for changes on node connections to test the reliability of this connection change feature.

This PR bumps up the sync interval to every hour from every 10s for eventual consistency.


jakmeier (changes requested):
Thanks for giving the connection change based approach a first go! We really need this for state sync to make sense long-term.

However, in the current approach, I'm afraid this introduces more instability rather than fixing it. Specifically, this code still considers a peer "stable" as soon as we can ping them. This means we start running state sync because we suspect we are out-of-sync but we also start new protocols in the meantime, based on the out-of-sync data. This will lead to more race conditions like you found in #311.

I'm tempted to merge this in again, just to keep things going. But given how long things have been unstable due to state sync now, I think we should only merge in changes to state sync that improve stability.

That is, I suggest you keep the changes for detecting online/offline nodes back for now. Instead, we should first ensure we keep track of which nodes are in sync and only use them as participants. Afterwards, the changes in this PR can also be beneficial to stability, if integrated properly.

I've also written down what state sync is lacking the most in this comment: https://github.com/sig-net/mpc/issues/130#issuecomment-2792394679

ChaoticTempest:
closing this in favor of https://github.com/sig-net/mpc/pull/461

PR #316, opened 2025-04-10
add artifacts required to compile the project to the repo

jakmeier:
We statically include the artifact

`contract-eth/artifacts/contracts/ChainSignatures.sol/ChainSignatures.json`

at chain-signatures/node/src/rpc.rs:L381.

If we want to keep it that way, I think it would make sense to check it into the source repository, since currently the project simply doesn't compile after a fresh checkout.


ChaoticTempest (commented):
I think it's fine to include this in, since this is purely for tests right now AFAIK. This artifact is generated on the fly when we're building the docker image anyways for deployments, so it shouldn't be too big of a concern. But @ppca @ailisp if I'm missing anything

jakmeier:
@ppca @ailisp Anything to add on your end? Or should we merge this?

ppca:
I think this is good. Our contract is open anyways.

ppca:
Just let me check if this interferes with building our contract. one sec

ppca:
ok it's good. Let's just remember to change this if our contract changes.

ppca (approved):

jakmeier:
Thanks! Finally, checking out a fresh branch will no longer cause `cargo check` and rust-analyzer to scream at me :D

PR #317, opened 2025-04-10
refactor: delete unused `is_stable` code

jakmeier:
`is_stable` has been set to a constant of `true` since #241

The `is_stable` flag was useful when we only looked at the near indexer. It checked of the indexer data is behind.
But now with many chains integrated, we want to move away from chain-specific checks. Hence, is_stable should go away entirely.

This PR removes dead code that was still around from the old is_stable check and  renames some field to make it easier to understand what actually happens.

Specifically, replace `NodeStatus::Active(IsStable=true)` with `NodeStatus::Active`.

For the case where the peer node is online but it still is in a protocol state where it is not ready to participate, we now use `NodeStatus::Inactive`. (Before, we used `NodeStatus::Active(IsStable=false)` for that, although it doesn't really have anything to do with stability.)

These changes are in preparation for state sync to be used for deciding which nodes are indeed ready to participate in a protocol invocation. For that, we want to add a check for a node being in-sync. Which will touch the same code, so it's a good timing to refactor it a bit first.

### DIFF
--- a/chain-signatures/node/src/mesh/connection.rs
+++ b/chain-signatures/node/src/mesh/connection.rs
@@ -14,11 +14,15 @@ use crate::web::StateView;
 
 use super::MeshState;
 
-type IsStable = bool;
-
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 enum NodeStatus {
-    Active(IsStable),
+    /// The connected node responds and is actively participating in the MPC
+    /// network.
+    Active,
+    /// The node responds but is in an inactive NodeState, hence it is not ready
+    /// for participating in any MPC protocols, yet.
+    Inactive,
+    /// The node can't be reached at the moment.
     Offline,
 }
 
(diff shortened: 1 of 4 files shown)


chain-signatures/node/src/mesh/connection.rs
```diff
@@ -207,11 +210,18 @@ impl Pool {
...
+                NodeStatus::Active => {
+                    active.insert(participant, conn.info.clone());
                     stable.push(*participant);
                 }
+                NodeStatus::Inactive => {
+                    // TODO: should we really insert inactive nodes to the active list here?
+                    // For now, in the refactoring PR, I just keep the exact same behavior.
+                    // We can delete this line when in the next PR.
+                    active.insert(participant, conn.info.clone());
+                }
```
jakmeier, line 223: To keep this a pure refactor, I kept the old behavior here. These are all nodes that respond in the `/state` endpoint that they are not in `NodeState::Running` nor in `NodeState::Resharing`. Meaning, they are in one of  { `Starting`, `Started`, `Generating`, `WaitingForConsensus`, `Joining` }.


@ChaoticTempest can you advise what the best treatment would be? For me it would be more intuitive to keep nodes that are currently joining the network outside the "active" participants set. Only once the are in running or resharing state I would add them. But I'm not sure if it will break something.

ChaoticTempest (approved):

chain-signatures/node/src/mesh/connection.rs
ChaoticTempest: Yeah that would make sense to keep them separate. I think we need to do a refactor on what these sets are. When the node is initially doing keygen at the genesis of the network, they use `{ Starting, Started, Generating, WaitingForConsensus, Joining }` as active set. We can definitely be more fine grained with this and take this out of the active set just in case we're interfering with protocol generation side.

But outside of that, I think it's good to add in this now and make keygen use inactive set later

PR #318, opened 2025-04-10
helios: wait for finality

ppca:
Context:
We want to index the requests as soon as possible, but we want to only generate signature and respond when the block that the request belongs to is finalized.

Changes included:
The run() function in indexer_eth.rs now includes the following logic:
1. the main loop indexes requests as they come in, and add the requests into requests_indexed, which is a list ordered by block number;
2. one tokio task refreshes the final block and finalized epoch (a mapping between block number and block hash). The logic is when the latest finalized block fetched by helios changed from the current one we have, we use helios to fetch the finalized block. Helios will give us trusted information on the block: block hash, parent hash. We will then use an untrusted rpc client to fetch all the blocks going in reverse from (latest finalized block number -1) to (cur finalized block number + 1) . While we iterate, checking that the prev block's hash = cur block's parent_hash. This way the block hashes we fetch is trusted because the latest finalized block we fetch from helios is trusted, and we go backwards from its parent hash. 
3. one tokio task takes requests from requests_indexed, and once the request's block number is finalized, we send it to the send queue;
4. one tokio task retries any blocks that failed to be processed and insert the requests into requests_indexed at the position its block number should be ordered at;

@jakmeier let me know if 2 above is reasonable and whether the code for it is doing what it's supposed to do.


ppca:
talked with helios contributor, I need to change the current logic from using hash field of the block from untrusted rpc to hashing the header myself. Will work on the change and then send for review again.

ppca:
I've asked helios main contributor to check the logic in refresh_finalized_epoch, seems like it's doing the right thing.

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -39,8 +39,8 @@ pub struct IndexedSignRequest {
     pub id: SignId,
     pub args: SignArgs,
     pub chain: Chain,
-    pub timestamp: Instant,
-    pub unix_timestamp: u64,
+    pub unix_timestamp_indexed: u64,
+    pub timestamp_sign_queue: Option<Instant>,
```
volovyks, line 43: Why is it an Option?
And `Unwrap`s in multiple places seam dangerous.

integration-tests/src/local.rs
```diff
@@ -141,6 +141,29 @@ impl Node {
...
+        let eth_config = cfg.eth.clone();
+        let cfg = {
+            let new_eth_config = if let Some(eth_config) = eth_config {
+                Some(EthConfig {
+                    account_sk: eth_config.account_sk,
+                    consensus_rpc_http_url: eth_config.consensus_rpc_http_url,
+                    execution_rpc_http_url: eth_config.execution_rpc_http_url,
+                    contract_address: eth_config.contract_address,
+                    network: eth_config.network,
+                    helios_data_path: format!("{}_{}", eth_config.helios_data_path, node_id),
```
volovyks, line 153: Not sure I understand the purpose  of this change. Is it only to add node_id to `helios_data_path`?

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -99,6 +108,13 @@ pub struct EthArgs {
         default_value = "/helios/sepolia"
     )]
     pub eth_helios_data_path: Option<String>,
+    /// refresh finalized block interval in milliseconds
+    #[clap(
+        long,
+        env("MPC_ETH_REFRESH_FINALIZED_INTERVAL"),
+        default_value = "10000"
```
volovyks, line 115: Seams a bit too long. Why 10,000?

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -368,29 +416,79 @@ pub async fn run(
...
-                )
-                .await
-                {
-                    tracing::warn!("Retry failed for block {block_number}: {:?}", err);
-                    blocks_failed.push_back(block_number);
-                } else {
-                    tracing::info!("Successfully retried block: {}", block_number);
-                }
-            }
+    let blocks_failed: Arc<RwLock<VecDeque<(u64, alloy::primitives::B256)>>> =
```
volovyks, line 419: It would be great to track last processed block and keep it in persistent storage.

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -368,29 +416,79 @@ pub async fn run(
...
+
+    let requests_indexed: Arc<RwLock<VecDeque<BlockAndRequests>>> =
+        Arc::new(RwLock::new(VecDeque::new()));
+
+    let finalized_epoch: Arc<RwLock<HashMap<u64, alloy::primitives::B256>>> =
+        Arc::new(RwLock::new(HashMap::new()));
+    let finalized_block: Arc<RwLock<u64>> = Arc::new(RwLock::new(0));
+
+    let client = Arc::new(client);
+    let client_clone = Arc::clone(&client);
```
volovyks, line 430: Does not looks like a good practice. Can we clone where the clone is required and remove `client_clone`? There is two such places.

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -368,29 +416,79 @@ pub async fn run(
...
+    let finalized_block_clone = Arc::clone(&finalized_block);
+    let finalized_epoch_clone = Arc::clone(&finalized_epoch);
```
volovyks, line 432: Same here.

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -368,29 +416,79 @@ pub async fn run(
...
+    tokio::spawn(async move {
```
volovyks, line 433: Let's add docs on top of each task please.

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -368,29 +416,79 @@ pub async fn run(
...
+                &client_clone,
+                &untrusted_rpc_client,
+                &finalized_block_clone,
+                &finalized_epoch_clone,
+            )
+            .await
+            .unwrap_or_else(|err| {
+                tracing::warn!("Failed to refresh finalized epoch: {:?}", err);
+            });
+            tokio::time::sleep(Duration::from_millis(eth.refresh_finalized_interval)).await;
```
volovyks, line 446: We started using `tick` pattern in our code base, let's stick to it here and in other places.

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -418,23 +518,163 @@ pub async fn run(
...
+        {
+            tracing::warn!("Retry failed for block {block_number}: {:?}", err);
+            add_failed_block(&blocks_failed, block_number, block_hash).await;
+        } else {
+            tracing::info!("Successfully retried block: {}", block_number);
+        }
+    }
+}
+
+async fn add_failed_block(
```
volovyks, line 562: Why do we handle ordering of the failed blocks?

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -418,23 +518,163 @@ pub async fn run(
...
+        .iter()
+        .position(|&(num, _)| num >= block_number)
+    {
+        blocks_failed_write.insert(pos, (block_number, block_hash));
+    } else {
+        blocks_failed_write.push_back((block_number, block_hash));
+    }
+}
+
+async fn refresh_finalized_epoch(
```
volovyks, line 578: Please, add docs to this function.

volovyks (commented):
Nice work! Overall looks ok. We will need to add persistence to make sure each block is processed, but that can be done in a separate PR.
I will make a deeper dive into the block verification later today.

integration-tests/src/local.rs
ppca: This is for integration tests. So basically helios light client will keep its chain related data in this helios data path, and I find out that in integration tests, if the 3 nodes use the same path, occasionally weird things happen, likely because it is one same machine, not dockers. So to be on the safe side, I am making this change to make sure each node gets a different helios path

chain-signatures/node/src/indexer_eth.rs
ppca: this is basically 10s. Since ethereum finality is 10+ minutes, 10s wouldn't matter much

chain-signatures/node/src/indexer_eth.rs
ppca: it's not gonna be useful tho. So helios either start from a helios path locally that has historical helios data in it, which means it would be able to directly start from where it left off last time; or helios starts from an outside checkpoint, which in my experience is always quite recent, and we likely won't be able to go back too many blocks. But I can check with helios folk

chain-signatures/node/src/indexer_eth.rs
ppca: I am doing this because this is the only way I find that it can work....I think it's more like if I directly pass in the Arc::new(client), instead of the cloned pointer of the Arc, then the code will have borrow problems. Would appreciate a suggestion on how to make it better tho.

chain-signatures/node/src/indexer_eth.rs
ppca: ok will do

chain-signatures/node/src/indexer_eth.rs
ppca: ah my bad. the one that should be ordered is requests_indexed, not blocks_failed.

volovyks:
@ChaoticTempest @ailisp please, take a look

chain-signatures/node/src/protocol/signature.rs
volovyks: Unwraps are still not ok, since they can lead to `panic`. We should handle Some/None or introduce one more struct without this timestemp,

integration-tests/src/local.rs
volovyks: I guess we can make it a bit shorter:
```rust
let mut cfg = cfg.clone();
if let Some(ref mut eth_config) = cfg.eth {
   eth_config.helios_data_path = format!("{}_{}", eth_config.helios_data_path, node_id);
}
```

chain-signatures/node/src/indexer_eth.rs
volovyks: What we need is a guarantee that each block is processed, even if the node is late by days. It may be OK if Helios has it's own persistence layer, but it would be better if we are in control of it.

ppca: I think Helios just save the data to the path. I think our problem is more in that when we restart a node, all data in the file system are wiped? Or is it not? If it's not wiped, then our node should be able to start where it's left off last time. 
There is also an edge case, which in the case of 15 min finality wait is going to happen for sure, is that we have processed a block and indexed those requests, but since they are not finalized, signatures are not generated. And we currently don't persist these indexed requests also. So we will always lose all requests that have yet been finalized before the node shuts down.
I'm checking with Helios if there's a solution to go back a bit in blocks.

chain-signatures/node/src/protocol/signature.rs
ppca: yeah, I would handle the None case.

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -368,28 +416,85 @@ pub async fn run(
...
+    let blocks_failed: Arc<RwLock<VecDeque<(u64, alloy::primitives::B256)>>> =
+        Arc::new(RwLock::new(VecDeque::new()));
+
+    // A queue of blocks and requests that are indexed but not yet sent to the sign queue
+    let requests_indexed: Arc<RwLock<VecDeque<BlockAndRequests>>> =
+        Arc::new(RwLock::new(VecDeque::new()));
+
+    let finalized_epoch: Arc<RwLock<HashMap<u64, alloy::primitives::B256>>> =
+        Arc::new(RwLock::new(HashMap::new()));
+    let finalized_block: Arc<RwLock<u64>> = Arc::new(RwLock::new(0));
```
ChaoticTempest, line 428: all these looks pretty contentious, just use either a tokio mpsc or broadcast channel instead since you're going to be sending them from one place, and then taking them from another

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -368,28 +416,85 @@ pub async fn run(
...
+    let client = Arc::new(client);
+    let client_clone = Arc::clone(&client);
+    let finalized_block_clone = Arc::clone(&finalized_block);
+    let finalized_epoch_clone = Arc::clone(&finalized_epoch);
+    tokio::spawn(async move {
+        tracing::info!("Spawned task to refresh finalized block and the finalized epoch's blocks");
+        let mut interval =
+            tokio::time::interval(Duration::from_millis(eth.refresh_finalized_interval));
+        loop {
+            interval.tick().await;
```
ChaoticTempest, line 439: you don't need these intervals if you move to using channels. Doing a `channel.recv().await` will block till the next item comes in

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -368,28 +416,85 @@ pub async fn run(
...
+
+    let blocks_failed_clone = Arc::clone(&blocks_failed);
+    let client_clone = Arc::clone(&client);
+    let requests_indexed_clone = Arc::clone(&requests_indexed);
+    let near_account_id_clone = node_near_account_id.clone();
+    tokio::spawn(async move {
+        tracing::info!("Spawned task to retry failed blocks");
+        let mut interval = tokio::time::interval(Duration::from_millis(200));
+        loop {
+            interval.tick().await;
```
ChaoticTempest, line 484: same with this one with using `channel.recv()`

chain-signatures/node/src/indexer_eth.rs
ChaoticTempest: hmm, I would imagine a client should be cloneable by default, unless Helios forgot to implement that

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -463,57 +736,182 @@ async fn process_block(
...
+            &indexed_requests,
+        )
+        .await;
+    }
 
     let block_timestamp = client
         .get_block_by_number(BlockTag::Number(block_number), false)
         .await
         .ok()
         .and_then(|block| block.map(|b| b.header.timestamp()));
```
ChaoticTempest, line 761: is it actually ok for us to continue when there's no timestamp? Like our whole signature generation timeout mechanism relies on it, and without it, it's kind of rough. So if `get_block_by_number` fails, should we truly be able to progress?

chain-signatures/node/src/indexer_eth.rs
ChaoticTempest: finalized_block can be broadcast channel then since we can read from multiple places. Note that each cloned reader will not be fighting each other. They will each have their own separate cloned value when sent.

let me see here for `finalized_epoch`

ChaoticTempest: so, for `finalized_epoch`, we don't have to send entry by entry, we can send over the whole `HashMap` since you're clearing every interval anyways.

ppca: ohhhh one thing tho, the current logic assumes requests_indexed are ordered by block number. This was useful because then we only need to check the first one of requests_indexed against the finalized block, no need to have each request being checked in the background at the same time. With the mspc, such order won't be maintained, and we might have trouble that some retried blocks' requests are received by the requests_indexed later, and the finalized_epoch info is already refreshed and it's past those blocks and those retried blocks cannot be verified since finalized_epoch won't have those blocks anymore....
well, current logic also can't help requests in blocks that are retried successfully later than a finalized block update, so I guess the problem is not worse....
One solution I have is for requests in requests_indexed that is far back in history (to know whether it's far back in history we can compare with the finalized block number we have), we can have another task that handle them. That task would basically walk from the earliest block in the finalized epoch back to that block, verify its validity, and then send it to the sign queue.

ChaoticTempest: so even for `requests_indexed`, one thing you can do is to have a `tokio::watch` channel where only the latest value is stored. So something like a `HashMap` or a `VecDeque` can just be there, and multiple or singular receivers can fetch for this value. You can even modify the watched value in place if you just want to insert something in the queue without cloning the whole thing when you want to update the watched value via `watch_tx.send_if_modified(|vec| ...)`

ppca: trying to understand how `requests_indexed` would work. Is it like we store the whole VecDeque or HashMap of block number and the requests as one element in the tokio::watch channel called `requests_indexed`, and then `send_requests_when_final()` and `retry_failed_blocks` can consume it? But problem is the main process blocks loop and the `retry_failed_blocks()` will both be Producer/Sender in this case, which is against how tokio::watch channel is designed to be used, and `send_requests_when_final()` also need to monitor which requests are already sent....

ppca: I do think it is important not to contend on these, especially `requests_indexed`, because if we later re-use this code for other EVMs that finalize quickly, then locking the whole `requests_indexed` up for update would quickly be a problem.

ppca:
> hmm, I would imagine a client should be cloneable by default, unless Helios forgot to implement that
Did check with Helios, the clone won't point to the same client underneath. So better to Arc it.

chain-signatures/node/src/indexer_eth.rs
ppca: This block timestamp is only used to observe delay between block and indexing now.
For the timestamp in IndexedSignRequest, we are not using the block timestamp at all, we are using the indexed time now, and that is what signature generation timeout is based on. 
I think this makes sense since we are separating delay between block and indexing, and delay between indexing and responding. Better for tracking what is actually going wrong. We might want to observe an overall latency, but we can do as a followup.

chain-signatures/node/src/indexer_eth.rs
jakmeier: Yeah, this is very typical. You use the values in multiple tasks. Each needs a separate owned Arc. That's completely fine and how Arcs should be used.

> Would appreciate a suggestion on how to make it better tho.

I like your implementation as it is. It's a question of style, maybe someone else prefers to put the code in a new block scope with an extra `{}` and then over shadow with something like `let client = Arc::clone(client)`, rather than to create *_clone variable names. But it's all the same in the end.


> hmm, I would imagine a client should be cloneable by default, unless Helios forgot to implement that

The helios client is not clonable as far as I can see. Could be an oversight, as I see the already have Arcs inside. But maybe they intentionally don't want to lock themselves to such an API.

jakmeier: Oh and I also like the explicit use of Arc::clone rather than .clone() here. :)
It makes it clear that this is a cheap operation.

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -428,13 +535,179 @@ pub async fn run(
...
+
+    let latest_finalized_block_number = cur_finalized_block.header.number;
+
+    {
+        let mut epoch_write = finalized_epoch.write().await;
+        epoch_write.clear();
+        epoch_write.insert(
+            latest_finalized_block_number,
+            cur_finalized_block.header.inner.parent_hash,
+        );
```
jakmeier: Is there a race condition here? Could it be that the task running `send_requests_when_final` never sees a finalized epoch because we overwrite it with the next one?

I think this might happen when we are catching up for some reason, or if the other task gets delayed somehow.

ppca:
tested locally on sepolia with our dev contract. Code is working as expected.

chain-signatures/node/src/indexer_eth.rs
jakmeier: Yeah, I generally agree that using channels is better than `Arc<RwLock<VecDeque<T>>>`. It's kind of the same thing anyway. But channels are more optimized.

I have further suggestion on changing from 3 tasks to just 2 tasks to simplify it a bit. But let me start a new comment thread for that.

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -368,28 +416,85 @@ pub async fn run(
...
+                &Arc::clone(&requests_indexed_clone),
+                &finalized_block_clone,
+                &finalized_epoch_clone,
+                sign_tx.clone(),
+                near_account_id_clone.clone(),
+            )
+            .await
+            .unwrap_or_else(|err| {
+                tracing::warn!("Failed to send requests when final: {:?}", err);
+            });
```
jakmeier: Idea / suggestion:

Would it make things simpler to merge this loop with the one calling `refresh_finalized_epoch`?

Then you need to share `finalized_block` and `finalized_epoch` between tasks at all. No Arc, no RwLock. Just a u64 and a HashMap.

Now, we have roughly this:

```
let finalized_block;
let finalized_epoch;

# task 1:
loop
    refresh_finalized_epoch
        update(finalized_block)
        update(finalized_epoch)

# task 2:
loop
    send_requests_when_final
        wait_for_final(requests_indexed.front)
        send_indexed_requests(requests_indexed.front)
```

Instead, it could be a single task doing both:

```
# task 1:
loop
let finalized_block;
let finalized_epoch;

loop
    refresh_finalized_epoch
        update(finalized_block)
        update(finalized_epoch)

    send_requests_when_final(finalized_block, finalized_epoch)
        while final(requests_indexed.front)
            send_indexed_requests(requests_indexed.front)
```

This way, you can just pass in a normal &mut borrow to share the data between both functions.
I don't think these two variables needs to be available anywhere else.

But just as important, it also makes the control flow easier to follow.
Instead of two decoupled tasks where one polls on the output of the other, we do the check right after we updated the relevant data.
This should also avoid a potential race condition I mentioned in another comment.

chain-signatures/node/src/indexer_eth.rs
jakmeier: The suggested architectural change to merge two tasks into one should get rid of this issue.

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -368,28 +416,85 @@ pub async fn run(
...
+    let finalized_epoch_clone = Arc::clone(&finalized_epoch);
+    let near_account_id_clone = node_near_account_id.clone();
+    let requests_indexed_clone = Arc::clone(&requests_indexed);
+    tokio::spawn(async move {
+        tracing::info!("Spawned task to send indexed requests to send queue");
+        let mut interval = tokio::time::interval(Duration::from_millis(200));
+        loop {
+            interval.tick().await;
+            send_requests_when_final(
+                &Arc::clone(&requests_indexed_clone),
```
jakmeier: nit: The clone of the clone is unnecessary ;)

```suggestion
                &requests_indexed_clone,
```

jakmeier (commented):
Great progress!

I think this should mostly work fine as it is. Functionally, I only have one concern about a potential race condition. (Aside from what others have already mentioned.)

I believe things can be simplified a bit, that's probably worth looking into.

Otherwise, I left some nits and style comments. These are all optional, of course, and can be ignored freely. ;)

jakmeier:
> I've made make `requests_indexed` a mpsc channel of `BlockAndRequests`, made `finalized_epoch` a channel of `BlockNumberToHashMap`, and made `failed_blocks` a channel of `BlockNumberAndHash`. The way I implement it now does not require a `finalized_block` to be recorded since latest finalized block number is only used in `refresh_finalized_epoch()`, and not anywhere else. Note this implementation does not enforce ordering of `requests_indexed` any more, and blocks that are retried late can get skipped:
> 
>     1. For all blocks that are processed successfully on the first attempt will be ordered by block number, since we do only process the next block when the prev successful one has all its requests sent to `requests_indexed` sender. We can also spawn a task for sending to `requests_indexed`, won't be a problem since they are still gonna be closely situated in requests_indexed and will get taken care of together when finalized epoch gets updated.
> 
>     2. for all blocks that are retried, if they are retried and success quickly, then they will be situated close to its mate in the same epoch and won't be skipped; but if they are retried and success quite late, i.e. later than the epoch ends, then they will be skipped as they are situated with other requests in some later epochs, and since we are only able to get request from `requests_indexed` one at a time, we can't possibly find that request and process it before another refresh of finalized_epoch.
> 
> 
> It is necessary to implement salvage of those late retried blocks. I'm thinking we can have a dedicated task that just handles them. Basically once we see that they are in history (in `send_requests_when_final()`), we push them into another channel. And that dedicated task will re-use the logic that walks back in history using untrusted rpc, starting from a block fetched from helios (so we are proven), to validate those retried blocks and send them to sign queue if validated.

Sorry, I didn't see this update before submitting my review.

I see you already improved a lot of the code, it's looking much better now!

Still, I think you could simplify the handling of `finalized_epoch` by merging two tasks. But if you like to keep them as separate tasks, that's also fine with me.

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -428,13 +535,179 @@ pub async fn run(
...
+        return Err(anyhow!("Fetching finalized block return None"));
+    };
+
+    let last_finalized_block_number = {
+        let mut finalized_block_write = finalized_block.write().await;
+        if *finalized_block_write == 0 {
+            *finalized_block_write = cur_finalized_block.header.number;
+        }
+        *finalized_block_write
+    };
```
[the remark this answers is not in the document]
ppca: yes exactly. Will change!

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -428,13 +535,179 @@ pub async fn run(
...
+        return Ok(());
+    }
+
+    {
+        *finalized_block.write().await = cur_finalized_block.header.number;
+    }
+
+    let latest_finalized_block_number = cur_finalized_block.header.number;
+
+    {
```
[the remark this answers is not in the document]
ppca: ah this was leftover from when it was a write to the locked values. I'll modify that. 😓

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -428,13 +535,179 @@ pub async fn run(
...
+        epoch_write.clear();
+        epoch_write.insert(
+            latest_finalized_block_number,
+            cur_finalized_block.header.inner.parent_hash,
+        );
+
+        let mut parent_hash = cur_finalized_block.header.inner.parent_hash;
+
+        // go backwards from latest_finalized_block_number - 1, and check that each block's hash == next block's parent hash
+        for i in (last_finalized_block_number + 1..=latest_finalized_block_number - 1).rev() {
```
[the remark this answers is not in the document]
ppca: Sounds good! Nice catch! Will do

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -428,13 +535,179 @@ pub async fn run(
...
+    blocks_failed: Arc<RwLock<VecDeque<(u64, alloy::primitives::B256)>>>,
+    client: Arc<EthereumClient<FileDB>>,
+    eth_contract_addr: Address,
+    node_near_account_id: AccountId,
+    requests_indexed: Arc<RwLock<VecDeque<BlockAndRequests>>>,
+) {
+    if blocks_failed.read().await.is_empty() {
+        return;
+    }
+    while { blocks_failed.read().await.len() } > 0 {
```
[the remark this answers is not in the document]
ppca: This was to explicitly make sure that read lock was released. But now this is a channel also

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -159,11 +183,33 @@ impl EthArgs {
...
         }
     }
 }
 
+#[derive(Clone)]
+pub struct BlockAndRequests {
+    block_number: u64,
+    block_hash: alloy::primitives::B256,
+    indexed_requests: Vec<IndexedSignRequest>,
+}
```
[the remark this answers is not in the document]
ppca: Done!

chain-signatures/node/src/indexer_eth.rs
ppca: Given it a go. Please take a look. Thanks for the suggestion!!

chain-signatures/node/src/indexer.rs
```diff
@@ -252,9 +252,8 @@ async fn handle_block(
                         key_version: arguments.request.key_version,
                     },
                     chain: Chain::NEAR,
-                    // TODO: use indexer timestamp instead.
-                    timestamp: Instant::now(),
-                    unix_timestamp: crate::util::current_unix_timestamp(),
+                    unix_timestamp_indexed: crate::util::current_unix_timestamp(),
+                    timestamp_sign_queue: Some(Instant::now()),
```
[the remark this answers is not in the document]
ppca: these are 2 different timestamps, one is indexed timestamp, one is timestamp it gets sent to the sign queue.

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -463,55 +719,162 @@ async fn process_block(
...
+    let mut final_block_number: Option<u64> = None;
+    let mut finalized_epoch: BlockNumberToHashMap = HashMap::new();
+    while let Some(BlockAndRequests {
+        block_number,
+        block_hash,
+        indexed_requests,
+    }) = requests_indexed.recv().await
+    {
+        // keep looping until the most recent updated finalized epoch contains the block number
+        loop {
```
jakmeier, line 780: Hm, your previous version of the code looked much cleaner than now with the merged tasks :sweat_smile: 

I think if we merge the tasks, we should avoid nesting too many loops.
Now we have a nesting that's literally `while { loop { loop { foo() } } }`. And then inside foo we do loop { interval.tick.await }, adding yet another nested loop.

Also, there is new concern. If we only index a signature request after we have already marked that block as final, I think we will be stuck polling for a finalization refresh and hence be unable to process the request. I believe this timing should not happen on Ethereum but it might happen in faster EVM chains.

What I meant with my previous pseudo code was sort of flipped inside out to what you have now.
This can also be extended to solve bad timing for finality / indexing.
Maybe something like this?

```rust

// This needs to be updated in `process_block`, so it needs to be passed to `retry_failed_blocks` too.
// Probably needs to be in an Arc for that. But no lock, as it's atomic.
let indexed_block: AtomicU64::new(0);

tokio::spawn(async move {
    // this top level code could also be a separate function

    let mut final_block_number: Option<u64> = None;
    let mut finalized_epoch = HashMap::new();

    // this is the only loop in this function
    loop {
        // this top-level sleep is the only one, no intervals in nested functions
        interval.tick().await;

        if final_block_number > indexed_block.load(Ordering::Acquire) {
             // indexing is behind, wait for indexer to catch up
             continue;
        }

        // this is roughly what you have in `send_requests_when_final` now,
        // without the call to `refresh_finalized_epoch`
        // It contains a `while let {}` loop nesting another loop
        send_all_finalized_requests()?;

        // We know we have indexed all requests at least up until final_block_number
        // and all indexed request up to final_block_number are now processed.
        // Hence, we can progress to the next epoch.
        //
        // This essentially what you have in refresh_finalized_epoch now, with the loop removed.
        try_refresh_finalized_epoch(&mut finalized_epoch, &mut final_block_number);
    }
```


I think that would look cleaner.

But feel free to move it back to what you had before (one more tokio task) if you like that better! It's a matter of style.

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -463,55 +719,162 @@ async fn process_block(
...
+                            )
+                            .await
+                        else {
+                            tracing::warn!("Failed to refresh finalized epoch");
+                            continue;
+                        };
+                        if let Some(new_final_block_number) = new_final_block_number {
+                            final_block_number.replace(new_final_block_number);
+                            finalized_epoch.clear();
+                            finalized_epoch.extend(new_finalized_epoch);
```
jakmeier, line 826: This code can be removed if you pass `&mut finalized_epoch` to `refresh_finalized_epoch`. Same for `final_block_number`.

It's a matter of style. Some people really don't like &mut arguments. But IMO, this is a case where the variable is read inside and we do factually update it. It's just that now you do in an extra step that always happens afterwards.

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -418,23 +512,185 @@ pub async fn run(
...
+    helios_client: &Arc<EthereumClient<FileDB>>,
+    untrusted_rpc_client: &RootProvider<Http<AlloyClient>>,
+    last_final_block_number: Option<u64>,
+    refresh_finalized_interval: u64,
+) -> anyhow::Result<(Option<u64>, BlockNumberToHashMap)> {
+    let mut interval = tokio::time::interval(Duration::from_millis(refresh_finalized_interval));
+    let mut finalized_epoch = BlockNumberToHashMap::new();
+    let mut final_block_number: Option<u64> = last_final_block_number;
+    loop {
+        interval.tick().await;
```
jakmeier, line 586: I think we can get rid of the loop. Instead, try once and return afterwards, even if there was no update.

This also moves the interval up one layer, so it's less hidden away.

jakmeier (commented):
Thanks for merging the tasks! I'm not 100% certain this is better than what you had before. See my inline comments

volovyks (approved):

ChaoticTempest:
@volovyks right now, the leader is reelected if the `stable` list of participants changes. Eventually they'll agree on a stable list, and then the true leader is decided based on that

integration-tests/src/main.rs
```diff
@@ -71,14 +73,15 @@ async fn main() -> anyhow::Result<()> {
...
                 threshold,
-                // eth: Some(EthConfig {
-                //     account_sk: eth_account_sk,
-                //     consensus_rpc_http_url: eth_consensus_rpc_http_url,
-                //     execution_rpc_http_url: eth_execution_rpc_http_url,
-                //     contract_address: eth_contract_address,
-                //     network: eth_network,
-                //     helios_data_path: eth_helios_data_path,
-                // }),
+                eth: Some(EthConfig {
```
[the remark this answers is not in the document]
ppca: it works on my machine. trick is to remove the folder you use for helios data path last time.

PR #321, opened 2025-04-17
feat: reorganize sign request participants on failure

ChaoticTempest:
Currently, when a signature generation fails for whatever reason, we do not try to reorganize the participant list. As in the moment the sign request is indexed, the list of participants that will partake in that list will stay static forever. This would easily interfere with signature generation if one node from that static list were to hop offline.

So, this PR makes it so that we do a reorganization of the sign request participants when the request fails, usually on timeout. `SignQueue` will now take the latest view of `stable` and then go through the same process it did to determine the proposer and participants subset. This also helps in the case that multiple nodes do not agree on the same subset/proposer.

- `failed_requests` facilitates this reorganization.
- `expire` will now expire my_requests and failed_requests. `my_requests` already get expired on `SignatureGenerator` error but added here just for clarity and sanity.
- note that since parties can get reorganized into partaking or not, `other_requests` is now no longer just nodes originally in the subset, but nodes that are just not the proposer. This is OK because `expire` will expire these other_requests even if nodes aren't a part of the generation.

This will likely also help with https://github.com/sig-net/mpc/pull/319 where when we have a race condition with choosing the stable set for the first signature just right after a reshare occured.

### DIFF
--- a/chain-signatures/node/src/protocol/signature.rs
+++ b/chain-signatures/node/src/protocol/signature.rs
@@ -51,13 +51,19 @@ pub struct SignRequest {
     pub indexed: IndexedSignRequest,
     pub proposer: Participant,
     pub participants: Vec<Participant>,
+    pub stable: Vec<Participant>,
 }
 
 pub struct SignQueue {
     me: Participant,
     sign_rx: Arc<RwLock<mpsc::Receiver<IndexedSignRequest>>>,
+    /// The requests that belong to us where we will the propose the signature to the chain.
     my_requests: VecDeque<SignRequest>,
+    /// All other requests. Can be requests that we are not the proposer, or one where
+    /// we do not belong to the signature generating subset.
     other_requests: HashMap<SignId, SignRequest>,
+    /// Set of request that failed to be processed during signature generation
+    failed_requests: VecDeque<SignRequest>,
 }
 
 impl SignQueue {
@@ -74,6 +80,7 @@ impl SignQueue {
             sign_rx,
             my_requests: VecDeque::new(),
             other_requests: HashMap::new(),
+            failed_requests: VecDeque::new(),
         }
     }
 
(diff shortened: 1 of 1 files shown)


volovyks (commented):
Logic and the intention make sense. We do not want a protocol to be doomed if one of the nodes goes offline, makes the whole idea of retries useless.
On the other hand, it is a half measure, since we are reelecting participants, but not the proposer.

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -93,6 +100,65 @@ impl SignQueue {
...
+        threshold: usize,
+        stable: &[Participant],
+        indexed: IndexedSignRequest,
+        reorganize: bool,
+    ) -> SignRequest {
+        let sign_id = indexed.id.clone();
+        // NOTE: reorganize, will use the same entropy for reorganizing the participants. The only
+        // thing that would be different is the passed in stable participants.
+        let mut rng = StdRng::from_seed(indexed.args.entropy);
+        let subset = stable.iter().copied().choose_multiple(&mut rng, threshold);
```
jakmeier, line 121: That's reasonable as a iterative improvement.

Long-term, shouldn't we try different subset even of the same stable set?

jakmeier (approved):
Nice small improvement! LGTM

chain-signatures/node/src/protocol/signature.rs
ChaoticTempest: yeah, I think we can move to reorganizing the same stable set later, but requires adding in an attempt counter just to change the entropy

ChaoticTempest:
> On the other hand, it is a half measure, since we are reelecting participants, but not the proposer.

We discussed offline, but this is re-electing the particiapants and the proposer at the same time

PR #334, opened 2025-04-22
chore: remove ctx traits

ChaoticTempest:
This removes the `ConsensusCtx` and `CryptographicCtx` traits and `Ctx` struct since they don't add much value. Instead, we will use `MpcSignProtocol` struct directly. This cuts down the boilerplate and anytime we want to add in new fields, we would have to modify those traits.

This also removes `async_trait` since rust 1.75 stabilized async fn in traits. The only caveat is that these traits can no longer be `pub`lic, but `pub(crate)` due to bounds limitations. We don't really care about that case since we're not writing a public facing library with the `mpc-node` crate anyways.

### DIFF
--- a/chain-signatures/node/src/protocol/cryptography.rs
+++ b/chain-signatures/node/src/protocol/cryptography.rs
@@ -1,4 +1,3 @@
-use super::message::MessageChannel;
 use super::signature::SignatureManager;
 use super::state::{GeneratingState, NodeState, ResharingState, RunningState};
 use super::MpcSignProtocol;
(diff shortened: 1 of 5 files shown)


ChaoticTempest:
also removed som e `tokio::spawn` handles in `cli.rs` since no pointing in explicitly waiting for those. We should only be waiting on protocol/web handles only.

chain-signatures/node/src/protocol/cryptography.rs
```diff
@@ -251,11 +232,10 @@ impl CryptographicProtocol for ResharingState {
...
 }
 
-#[async_trait]
 impl CryptographicProtocol for RunningState {
-    async fn progress<C: CryptographicCtx + Send + Sync>(
-        mut self,
-        ctx: C,
+    async fn progress(
+        self,
+        ctx: &mut MpcSignProtocol,
```
volovyks, line 238: We should rename either field or struct name

volovyks (commented):
The intention behind this structs was to incapsulate data and give devs predefined interface to interact with the protocol. No strong opinion here, since we were exposing almost everything we have anyway.

ppca (approved):

ppca:
looks simpler and easier to add stuff.

ChaoticTempest:
> The intention behind this structs was to incapsulate data and give devs predefined interface to interact with the protocol. No strong opinion here, since we were exposing almost everything we have anyway.

@volovyks data encapsulation only makes sense when you're exposing it as a library for other 3rd parties to use or simplifying an interface which we weren't doing here at all. Instead, it just added more work every time we needed something new. Generally in Rust, using traits is more for generalizing over several cases. Here we only had 1 case

chain-signatures/node/src/protocol/cryptography.rs
ChaoticTempest: hmm, I'll give it another pass another time

Issue #314, opened 2025-04-10
eth publish signature retry fails


ppca:
failed to publish ethereum signature sign_id=SignId("c8c2dcd424cd95cbbdae5ea49e14151c9e9574f2a31f685653d0da3be32358d8") error=Rpc(Error { code: InternalError, message: "replacement transaction underpriced", data: None })

ppca:
Our eth publish retry fails all the time with this error. Seems like this is because we are retrying with the same nonce and same price. 
Will look into a fix. If anyone know a good fix, let me know

ppca:
OK it's not because of retries. This will always happen when we respond() to a lot of sign requests. What happens is in the background, the provider code get_transaction_count(wallet_address) to get the nonce, and use that nonce to send the respond transaction. As you could imagine, when we are concurrently sending out respond txs, many of them will have the same nonce, and those tx will error out. `replacement transaction underpriced` essentially says we are sending another transaction with the same nonce. 

My proposal here is: instead of sending 1 respond() per sign request, we send 1 respond() per batch of sign requests. This is enabled already on our contract design, thanks to @volovyks :
```
/**
     * @dev Function to respond to signature requests.
     * @param _responses The array of signature responses.
     */
    function respond(Response[] calldata _responses) external {
        for (uint256 i = 0; i < _responses.length; i++) {
            emit SignatureResponded(
                _responses[i].requestId,
                msg.sender,
                _responses[i].signature
            );
        }
    }
```
On the backend, we can batch sign requests, and we call 1 respond() for the whole batch. And IMPORTANT, we only send respond() for the next batch when the prev batch's respond() call has completed. The batch at any time will be min(max_batch_size, whatever left in the respond queue). 
I also observed that under our current approach, it takes a long time for a respond() call to return a result. Not sure if it's related to a lot of concurrent tx going out from one wallet address. This new approach will avoid the same problem. And I also plan to add a timeout bound to the respond() call so that we don't wait too long for each tx.

volovyks:
Agree, we need to implement this. Ideally, we should send all available signatures within gas limit. Each blockchain will have it's own limit, we need to find out what it is.

PR #336, opened 2025-04-22
eth respond nonce: batch respond

ppca:


chain-signatures/node/src/rpc.rs
```diff
@@ -39,6 +40,10 @@ const MAX_PUBLISH_RETRY: usize = 6;
 const MAX_CONCURRENT_RPC_REQUESTS: usize = 1024;
 /// The update interval to fetch and update the contract state and config
 const UPDATE_INTERVAL: Duration = Duration::from_secs(3);
+/// The interval to batch send Ethereum responses
+const ETH_RESPOND_BATCH_INTERVAL: Duration = Duration::from_millis(2000);
+/// The batch size for Ethereum responses
+const ETH_RESPOND_BATCH_SIZE: usize = 10;
```
volovyks, line 46: Is it the max we can do? Have you measured it?

chain-signatures/node/src/rpc.rs
```diff
@@ -674,3 +732,154 @@ async fn try_publish_eth(
...
+    near_account_id: &AccountId,
+    start: Instant,
+) {
+    let mut signatures: HashMap<SignId, Signature> = HashMap::new();
+
+    for action in actions.iter() {
+        let expected_public_key =
+            mpc_crypto::derive_key(action.public_key, action.request.indexed.args.epsilon);
+
+        // We do this here, rather than on the client side, so we can use the ecrecover system function on NEAR to validate our signature
```
volovyks, line 841: this one is not about NEAR :)

chain-signatures/node/src/rpc.rs
```diff
@@ -674,3 +732,154 @@ async fn try_publish_eth(
...
+            return;
+        };
+        signatures.insert(action.request.indexed.id.clone(), signature);
+    }
+
+    let mut retry_count = 0;
+    loop {
+        let publish = match client {
+            ChainClient::Near(_) => {
+                unimplemented!("near has no batch publish");
```
volovyks, line 861: unimplemented can lead to a panic and crash of the node as a result
Let's log an error and mention that this is not implemented please.

chain-signatures/node/src/rpc.rs
```diff
@@ -674,3 +732,154 @@ async fn try_publish_eth(
...
+                tracing::warn!(msg, "no client for chain");
+                Ok(())
+            }
+        };
+        if publish.is_ok() {
+            actions.clear();
+            break;
+        }
+
+        retry_count += 1;
```
volovyks, line 876: Not sure if retry logic should be here since we will retry on all signatures, not only on failed ones.

chain-signatures/node/src/rpc.rs
```diff
@@ -533,6 +562,35 @@ async fn execute_publish(
     }
 }
 
+async fn run_batch_respond(
+    client: ChainClient,
+    mut actions_rx: mpsc::Receiver<PublishAction>,
+    batch_interval: Duration,
+    batch_size: usize,
```
volovyks, line 569: It does not look like we are limiting maximum batch size. It can happen that to many requests will be added to a respond call. And then we retry, again and again.

ppca: we are limiting the batch size here: ```if start.elapsed() > batch_interval
            || actions_batch.len() >= batch_size && !actions_batch.is_empty()```

chain-signatures/node/src/rpc.rs
```diff
@@ -533,6 +562,35 @@ async fn execute_publish(
...
+                num_requests = actions_batch.len(),
+                "publishing batch of signatures",
+            );
+            execute_batch_publish(&client, &mut actions_batch, near_account_id, Instant::now())
+                .await;
+            start = Instant::now();
+        }
+        if let Ok(action) = actions_rx.try_recv() {
+            actions_batch.push(action);
+        }
```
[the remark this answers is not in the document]
ppca: but as mentioned in https://github.com/sig-net/mpc/issues/314, we want to avoid sending multiple tx at the same time. That's why I did not put the execute_batch_publish in spawn, because I want to make sure it is done one after another, not at the same time.

chain-signatures/node/src/rpc.rs
```diff
@@ -674,3 +732,154 @@ async fn try_publish_eth(
...
+        .unwrap()
+        .gas(gas)
+        .send()
+        .await
+        .map_err(|err| {
+            tracing::error!(
+                //sign_id = ?action.request.indexed.id,
+                error = ?err,
+                "failed to publish ethereum signature"
+            );
```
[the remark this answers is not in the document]
ppca: yeah will add. It's gonna be a batch of sign ids

chain-signatures/node/src/rpc.rs
ppca: if the publish is successful, the retry loop will break already and no retries will be done.

chain-signatures/node/src/rpc.rs
```diff
@@ -216,20 +221,44 @@ impl RpcExecutor {
...
+        tokio::spawn(async move {
+            run_batch_respond(
+                eth_client,
+                eth_rpc_rx,
+                ETH_RESPOND_BATCH_INTERVAL,
+                ETH_RESPOND_BATCH_SIZE,
+                &near_account_id_clone,
+            )
+            .await;
+        });
```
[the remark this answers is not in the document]
ppca: ah passed it in because I'm thinking we can reuse the logic for other chains too

PR #338, opened 2025-04-23
update solana dev env

volovyks:
Let's use this address as our dev env contract


jakmeier (approved):

volovyks:
@esaminu we are passing Solana dev address now

PR #342, opened 2025-04-24
doc: add mpc node specification

jakmeier:
This documentation formalizes the plan outlined in #130 and the design documents linked there.

I also tried to make it clear how it all relates to the goal of better fault tolerance in the MPC network.


jakmeier:
The rich diff doesn't show the graphs properly. You can look at the file on my branch directly to see it rendered nicely:

https://github.com/jakmeier/mpc/blob/mpc_node_spec/doc/mpc_node_specification.md

doc/mpc_node_specification.md
```diff
@@ -0,0 +1,225 @@
...
+For the signature protocol, all participants must check that requests were
+indexed and finalized, as noted in the graph.
+
+![graph](./graphs/protocol_states.svg)
+
+
+### States
+
+`init` and `done` are implicit states. All other states must be explicitly
+stored and ideally persisted.
```
volovyks, line 58: Would you say we need to save all the information about protocol, including incoming messages? Partial information will not give us much in terms of protocol recovery (for example, after reboot), and all information may be a lot.

doc/mpc_node_specification.md
```diff
@@ -0,0 +1,225 @@
...
+      all messages have been received. (HTTP success response)
+    - Transition to `Prepare` state
+- Once all INIT messages have been answered with a success response
+    - Wait for finality of the request (for signatures only)
+    - Send out a START message all participants
+    - Transition to `Running` state
+    - Start poking own generator and sending out the corresponding messages
+
+Owners must also handle the following error case:
+- Not all participants answered the INIT message with an HTTP success response
```
volovyks, line 106: Here I suggest to send INIT request to all (or at least T+1 nodes) and wait for T Nodes to respond instead of failing or retrying with a new participant set, because the timeout for INIT will slow down the protocol if one of the nodes is slow or not responding at all.

doc/mpc_node_specification.md
```diff
@@ -0,0 +1,225 @@
...
+
+### Node Connection Status
+
+![graph](./graphs/node_connection.svg)
+
+Nodes keep pinging each other, checking their peers status.
+
+Peers that don't respond are marked as `Offline`.
+
+Peers that respond but indicate in their status response that they are not actively participating in the MPC network, are marked as `Inactive`.
```
volovyks, line 172: What are the reasons to be Inactive? Can we skip it and use Syncing only?

doc/mpc_node_specification.md
jakmeier: Sounds good, yeah. Then we have to send the participant list in the START message rather than the INIT message.

doc/mpc_node_specification.md
jakmeier: We already have the inactive state in the code. As I understand it, these are for nodes that are joining the network. I don't think we have anything to sync with them, yet.

doc/mpc_node_specification.md
jakmeier: Good point. Ideally, yes, we store all messages and recover using the [Event-Sourcing](https://dev.to/alisamir/understanding-event-sourcing-a-detailed-guide-4cjp) pattern. I've added this to the document.

That's the long-term goal, at least. In step 1, I would say we just persist which state of the state machine we are in, without the messages. When we reboot on `Prepare`, that's already enough to recover and deal with incoming START messages without issues.

For the `Running` state, we can start by adding a metric that tells us anytime a node boots and has remaining `Running` states. Then at least we see how big the problem is and can prioritize accordingly. Otherwise, we continue doing what we do now, which leads to unstable executions of the protocol  but should recover eventually with a retry after a timeout.

volovyks: Nice pattern. Yes, I would start the implementation only if we see a clear reason for it. Nodes are mostly stable, so the benefit will not be huge. But it adds a lot of traffic and latency to the persistence layer.

jakmeier:
@volovyks and  @ChaoticTempest please take another look at the latest commit with more details on the Prepare phase. And if you roughly agree on the description, let's merge this and make more changes with additional PRs. (Changes on existing PRs are harder to track for everybody involved)

doc/mpc_node_specification.md
```diff
@@ -0,0 +1,370 @@
...
+    - Start poking own generator and sending out the corresponding messages
+
+Owners must also handle the following error case:
+- Not enough participants answered the INIT message with an HTTP success response
+    - Retry, maybe some more participants are online now
+    - After a threshold of retries, switch to a different P or T input that has
+      more active participants.
+
+Other error cases can be handled with a global timeout to abort and retry for
+signatures.
```
ChaoticTempest, line 197: So additionally with https://github.com/sig-net/mpc/pull/352, other participants can send rejections in the case they don't agree for whatever reason like proposing an id that was already owned in the system by somebody else.

It might also be hard to do acknowledgements based on successful HTTP responses since the messaging layer currently handles that, which is why receiving an `Accept` would be better for now. At least its easier to implement that way

volovyks (dismissed):
Good update. Especially the last table.
Let's merge this and add changes as we go.

doc/mpc_node_specification.md
ChaoticTempest: One thing with the implementation in https://github.com/sig-net/mpc/pull/352, only the proposing node keeps track of the prepare state. Other nodes just either accept or reject, where they essentially have an implicit state for prepare. So I don't think we need other nodes to track prepare themselves

doc/mpc_node_specification.md
```diff
@@ -0,0 +1,370 @@
...
+[Event-Sourcing](https://dev.to/alisamir/understanding-event-sourcing-a-detailed-guide-4cjp)
+pattern, while the protocol is running. The actual latest state is only held in
+memory but the stored events allow to recover it from persisted data.
+
+Specifically, this means tha all received messages during the `Running` state
+must be persisted before they are applied. These can be replayed on boot for
+recovery after a crash. After a protocol finishes, events can be discarded.
+
+Note: Unlike fully event-driven architectures, the MPC network does not need to
+have a separate even store component. The Event-Sourcing pattern is only applied
```
ChaoticTempest, line 168:
```suggestion
have a separate event store component. The Event-Sourcing pattern is only applied
```

ChaoticTempest (approved):
I like that we're defining ever case with the protocol cases with the owner vs non-owners, so LGTM with slight modifications on the comments I mentioned before

doc/mpc_node_specification.md
jakmeier: @ChaoticTempest  There are a couple of cells in the final table that rely on the prepare state on all nodes.

Specifically, if `Prepare` is only tracked on the owner, you can have this situations:

- Node A sends `PREPARE` for Presignature X
- Node B accepts, then crashes
- Node A sends `START` for Presignature X
- Node B mutually syncs with node A before it processes the START messages
   - Sync request A->B [Presignature X]
   - Sync response B->A declares missing [Presignature X]
   - Node A will delete Node B from the list of holder of Presignature X
- Node B starts processing messages again and participates in the presignature protocol generation.
- Node B will eventually have the presignature share stored but Node A's directory will be wrong.

As you can see, failing to store the prepare state on the non-owning participant leads to an inconsistent system state.

doc/mpc_node_specification.md
jakmeier: I see. Updated the spec to make it either an HTTP response or a separate message.

The reason why an HTTP response would be preferred is to reduce latency and send overall less messages. HTTP already comes with a response, so I thought we could use that.

But yeah I can see that it's easier to implement with a separate `Accept` message. Let's do that for now and keep the HTTP response as an option in the future, when we have the luxury to micro-optimize :)

jakmeier:
I have just addressed the latest comments. Will merge this when I get a green tick again.

volovyks (approved):

PR #345, opened 2025-04-28
New K6 setup

volovyks:
Not tested yet, but I want to get initial feedback on this.

- This new setup follows new pinger API (with `chain`, `env`, and `check` parameters). cc @esaminu 
- It will be triggered on each merge  with default parameters (Solana high load dev).
- It can be trigered with any other combination of chain, env, check, and strategy parameters.
- Simplified everything to single GithHub Actions job
- Simplified everything to a single JS script

Still need to figure out if K6 saves ENV variables to create a schedule for all this jobs.


auto-mausx:
Great job! I think every time these tests are ran, it's a new environment (i.e. new GHA runner each time). We can sync and discuss a way forward for scheduling via CI. GHA does support cron jobs, so we could probably do something like that on the GHA file itself and select which we would like to run.

.github/workflows/k6-ci-loadtest.yml
```diff
@@ -80,29 +59,11 @@ jobs:
...
-          K6_CLOUD_PROJECT_ID: ${{ secrets.K6_CLOUD_PROJECT_ID }}
-          EVM_CONTRACT: ${{ env.TESTNET_EVM_CONTRACT }}
-          SOL_CONTRACT: ${{ env.TESTNET_SOL_CONTRACT }}
-          ENVIRONMENT: "testnet"
-        with:
-          path: ./infra/loadtests/2hour.js
-          cloud-run-locally: false
\ No newline at end of file
+          cloud-comment-on-pr: true
+          flags: --exit-on-running
```
auto-mausx, line 69: @volovyks I believe this will allow us to not hold runners while the test is executing, see: https://grafana.com/docs/k6/latest/using-k6/k6-options/reference/#exit-on-running

esaminu:
Nice work. The new way the pinger is being called is accurate. I have 2 questions:
- Do we need to keep the old envs simultaneously so that we can transition to the new API? I am thinking we start with failing calls to the new API on k6 and then once we deploy the latest pinger failing calls to the old APIs which we then remove from the dashboard 
- What happens when the event trigger is not `workflow_dispatch` given `github.event.inputs` values are null? will this have to be run manually?

volovyks:
@esaminu 
1. Load tests are not used atm, we can make breaking changes and move faster. I'm ready to merge this PR and test things.
2. If the event is trigged by merge or other event (not workflow_dispatch) `default` values are used (solana dev high load).

auto-mausx (approved):

Issue #339, opened 2025-04-24
Fix setup-env


volovyks:
Ethereum indexer setup is failing all nodes when running `cargo run -- setup-env`.

ppca:
this is fixed. Remember to `rm -rf ` for the helios_data_path you used last time before you start a new run.

volovyks:
@ppca not sure I understand, is that documented somewhere? Can it be cleaned up automatically?

PR #346, opened 2025-04-29
fix solana re subscribe

ailisp:
re subscribe is currently always fail due to: (anchor client) program keeps a reference to the websocket socket that is in a bad state, and resubscribe from there always immediately disconnect. By creating a new anchor client program instance, it also create a new websocket and resub is successful. Tested locally.

### DIFF
--- a/integration-tests/src/lib.rs
+++ b/integration-tests/src/lib.rs
@@ -65,7 +65,7 @@ impl Default for NodeConfig {
                 account_sk: "fS5jS6X5qvaquBV1bg2YWBdYeCiRSUwNAdNpgNkjS72oNxUJcZJZduaq2oCcXJb8erTbtqqq4wxriUmRJk7bMDw"
                     .to_string(),
                 rpc_url: "https://api.devnet.solana.com".to_string(),
-                program_address: "BtGZEs9ZJX3hAQuY5er8iyWrGsrPRZYupEtVSS129XKo".to_string(),
+                program_address: "CMGYAEsqXw5z52R8fmMZwPYQARHPEkGbefJA2FmeHLMh".to_string(),
             }),
         }
     }
--- a/integration-tests/src/main.rs
+++ b/integration-tests/src/main.rs
@@ -70,14 +70,14 @@ async fn main() -> anyhow::Result<()> {
             let config = NodeConfig {
                 nodes,
                 threshold,
-                eth: Some(EthConfig {
-                    account_sk: eth_account_sk,
-                    consensus_rpc_http_url: eth_consensus_rpc_http_url,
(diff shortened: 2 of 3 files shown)


integration-tests/src/lib.rs
```diff
@@ -65,7 +65,7 @@ impl Default for NodeConfig {
                 account_sk: "fS5jS6X5qvaquBV1bg2YWBdYeCiRSUwNAdNpgNkjS72oNxUJcZJZduaq2oCcXJb8erTbtqqq4wxriUmRJk7bMDw"
                     .to_string(),
                 rpc_url: "https://api.devnet.solana.com".to_string(),
-                program_address: "BtGZEs9ZJX3hAQuY5er8iyWrGsrPRZYupEtVSS129XKo".to_string(),
+                program_address: "CMGYAEsqXw5z52R8fmMZwPYQARHPEkGbefJA2FmeHLMh".to_string(),
```
volovyks, line 68: We talked with Osman and agreed to keep that address as our dev env program. Is it necessary to change it?

integration-tests/src/main.rs
```diff
@@ -70,14 +70,14 @@ async fn main() -> anyhow::Result<()> {
...
                 threshold,
-                eth: Some(EthConfig {
-                    account_sk: eth_account_sk,
-                    consensus_rpc_http_url: eth_consensus_rpc_http_url,
-                    execution_rpc_http_url: eth_execution_rpc_http_url,
-                    contract_address: eth_contract_address,
-                    network: eth_network,
-                    helios_data_path: eth_helios_data_path,
-                }),
+                // eth: Some(EthConfig {
```
volovyks, line 73: Is this necessary for this PR?

volovyks (commented):
Looks good overall.

integration-tests/src/lib.rs
ailisp: I can revert this one. It is for local testing. For my local testing, the BtG... address is not working with Osman's script, maybe I need to start from fresh, rebuild the contract from latest commit.

integration-tests/src/main.rs
ailisp: eth in setup-env is broken, eth config should be set to None before it is fixed

integration-tests/src/lib.rs
volovyks: Thhat contract was not set as immutable and controlled by your key, you can update it.

integration-tests/src/main.rs
volovyks: @ppca take a look with @ailisp once you are back
releated: https://github.com/sig-net/mpc/issues/339

volovyks (approved):

PR #350, opened 2025-04-30
Pass k6 parameters as arguments

volovyks:


.github/workflows/k6-ci-loadtest.yml
```diff
@@ -59,12 +59,13 @@ jobs:
...
-          LT_STRATEGY: ${{ github.event.inputs.lt_strategy }}
         with:
           path: ./infra/loadtests/k6-load-test.js
           cloud-run-locally: false
           cloud-comment-on-pr: true
-          flags: --exit-on-running
\ No newline at end of file
+          flags: >
+            --exit-on-running
+            -e LT_STRATEGY=${{ github.event.inputs.lt_strategy }}
```
auto-mausx, line 68: We should set defaults here, or in the .js file. I don't know if the inputs will be there on CI execution.

PR #352, opened 2025-05-01
feat: proposing new protocols through posits

ChaoticTempest:
This PR adds posits where a node must first propose a Posit before an equivalent protocol can be generated. For the most part, the actions that we can taken for new protocol posits are `Propose, Accept, Reject, Abort`. Only the proposer can issue out aborts. Only deliberators can send out `Accept` or `Reject`.

The flow is as follows on the happy path:
- node A proposes a posit.
- node B,C sees the posit.
- node B,C accepts the post & immediately starts the protocol (did the immediate part for now for simplicity).
- node A receives all `Accept`, and start their protocol.

The path where something went wrong:
- node A proposes a posit.
- node B,C sees the posit.
- node B accepts, but node C sees that their view of participants is invalid with node A and thus sends a `Reject`.
- node A sees a singular `Reject`, and immediately broadcasts out `Abort` (due to the immediate starting of the protocol by other nodes).

Note that this PR only adds these `Posits` to presignatures since triple needs to be refactored a bit and signature also has its own deal of work.

So before handling Presignature messages, presignature manager will first see any new posits and handle those, and thus initiating protocols if necessary. Some additional changes were made for presignatures:
- no separate functions for generating an owned presignature vs foreign presignature. They are now all united under `generate`. This means that reserve will only be called after proposition is done and we start. Before, the owner would try to reserve first, then send out the presignature generation request.
- added `FullPresignatureId` to encompass both the `PresignatureId` and the two `TripleId` that it uses.

The thing missing is to wait on all accepts before starting the protocol. I avoided this for now just to get this reviewed. I also imagine us being able to narrow down the list of participants in the case where someone were to reject or not reply. Timeouts also need to be handled eventually for this

### DIFF
--- a/chain-signatures/node/src/protocol/posit.rs
+++ b/chain-signatures/node/src/protocol/posit.rs
@@ -1,0 +1,147 @@
+use cait_sith::protocol::Participant;
+use serde::{Deserialize, Serialize};
+
+use std::collections::{HashMap, HashSet};
+use std::fmt;
+use std::hash::Hash;
+
+/// All actions that can be taken when a new posit is introduced for a protocol.
+#[derive(Debug, PartialEq, Serialize, Deserialize)]
+pub enum PositAction {
+    Propose(Vec<Participant>),
+    Accept,
+    // TODO: Reject can also have a reason
+    Reject,
+    /// Aborts the protocol. Only the proposer can send this.
+    Abort,
+}
+
+/// A counter for a posit. This is used to track the participants that have
+/// accepted the posit alongside storing an intermediary state for the protocol
+/// that the proposer needs to keep track of.
+pub struct PositCounter<Store> {
+    pub participants: HashSet<Participant>,
+    accepts: HashSet<Participant>,
+    store: Store,
+}
+
+/// A collection of posits that are being proposed. This is used to track
+/// the posits that are being proposed and the participants that have
+/// accepted them.
+#[derive(Default)]
+pub struct Posits<Id, Store> {
+    posits: HashMap<Id, PositCounter<Store>>,
+}
+
+impl<T: Hash + Eq + fmt::Debug, Store> Posits<T, Store> {
+    pub fn new() -> Self {
+        Self {
+            posits: HashMap::new(),
+        }
+    }
+
+    pub fn propose(
+        &mut self,
+        me: Participant,
+        id: T,
+        store: Store,
+        participants: &[Participant],
+    ) -> PositAction {
+        let mut accepts = HashSet::new();
+        accepts.insert(me);
+        self.posits.insert(
+            id,
+            PositCounter {
+                participants: participants.iter().copied().collect(),
+                accepts,
+                store,
+            },
+        );
+
+        PositAction::Propose(participants.to_vec())
+    }
+
+    // TODO: make the resp of this synchronous when each of the protocol
+    // managers are their own individual tasks such that they can respond
+    // without the need of the main consensus loop.
+    /// Act on the posit action. This will map the action received to a corresponding
+    /// action to be sent back to the proposer. This will return
+    ///     (ShouldStartProtocol, TemporaryStorageItem, ReplyAction).
+    pub fn act(
+        &mut self,
+        id: T,
+        from: Participant,
+        action: &PositAction,
+        active: &[Participant],
+    ) -> (Option<Vec<Participant>>, Option<Store>, Option<PositAction>) {
+        match action {
+            PositAction::Accept => {
+                let Some(counter) = self.posits.get_mut(&id) else {
+                    tracing::warn!(?id, "received an Accept for a protocol we did NOT propose");
+                    return (None, None, None);
+                };
+
+                counter.accepts.insert(from);
+                let should_start = counter.accepts.len() == counter.participants.len();
+                let should_start =
+                    should_start.then(|| counter.participants.iter().copied().collect());
+
+                let store = if should_start.is_some() {
+                    tracing::info!(?id, "received all Accepts, starting protocol");
+                    self.posits.remove(&id).map(|counter| counter.store)
+                } else {
+                    None
+                };
+
+                (should_start, store, None)
+            }
+            PositAction::Reject => {
+                // TODO: On the first reject, we should abort the protocol for now.
+                // We should be able to narrow down the list of participants
+                // that are rejecting the protocol up until the threshold amount.
+                let reply = if self.posits.contains_key(&id) {
+                    Some(PositAction::Abort)
+                } else {
+                    tracing::warn!(?id, "received a Reject for a protocol we did NOT propose");
+                    None
+                };
+                (None, None, reply)
+            }
+            // There's no action to be done here for Abort. Abort should be handled one level above.
+            PositAction::Abort => (None, None, None),
+            PositAction::Propose(participants) => {
+                if self.posits.contains_key(&id) {
+                    tracing::warn!(
+                        ?id,
+                        ?participants,
+                        "received a protocol posit for an id that we already proposed"
+                    );
+                    return (None, None, Some(PositAction::Reject));
+                }
+
+                // Check that the participants are all active
+                for p in participants.iter() {
+                    if !active.contains(p) {
+                        tracing::warn!(?id, ?active, ?participants, "rejecting protocol posit");
+                        return (None, None, Some(PositAction::Reject));
+                    }
+                }
+
+                // Automatically join the protocol if we're accepting here.
+                (Some(participants.to_vec()), None, Some(PositAction::Accept))
+            }
+        }
+    }
+
+    pub fn len(&self) -> usize {
+        self.posits.len()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.posits.is_empty()
+    }
+
+    pub fn remove(&mut self, id: &T) -> Option<PositCounter<Store>> {
+        self.posits.remove(id)
+    }
+}
--- a/chain-signatures/node/src/protocol/posit.rs
+++ b/chain-signatures/node/src/protocol/posit.rs
@@ -5,6 +5,19 @@ use std::collections::{HashMap, HashSet};
 use std::fmt;
 use std::hash::Hash;
 
+pub type ProposerId = Participant;
+
+pub enum Positor<Store> {
+    Proposer(ProposerId, Store),
+    Deliberator(ProposerId),
+}
+
+impl<T> Positor<T> {
+    pub fn is_proposer(&self) -> bool {
+        matches!(self, Positor::Proposer(_, _))
+    }
+}
+
 /// All actions that can be taken when a new posit is introduced for a protocol.
 #[derive(Debug, PartialEq, Serialize, Deserialize)]
 pub enum PositAction {
@@ -14,6 +27,12 @@ pub enum PositAction {
     Reject,
     /// Aborts the protocol. Only the proposer can send this.
     Abort,
+}
+
+pub enum PositInternalAction<Store> {
+    StartProtocol(Vec<Participant>, Positor<Store>),
+    AbortAllNodes(Vec<Participant>),
+    Reply(PositAction),
 }
 
 /// A counter for a posit. This is used to track the participants that have
@@ -28,14 +47,15 @@ pub struct PositCounter<Store> {
 /// A collection of posits that are being proposed. This is used to track
 /// the posits that are being proposed and the participants that have
 /// accepted them.
-#[derive(Default)]
 pub struct Posits<Id, Store> {
+    me: Participant,
     posits: HashMap<Id, PositCounter<Store>>,
 }
 
 impl<T: Hash + Eq + fmt::Debug, Store> Posits<T, Store> {
-    pub fn new() -> Self {
+    pub fn new(me: Participant) -> Self {
         Self {
+            me,
             posits: HashMap::new(),
         }
     }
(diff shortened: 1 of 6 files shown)


jakmeier (commented):
Nice to see some progress on this. I like the general approach you are taking here @ChaoticTempest !

I feel the chosen code architecture works really well. But maybe it could be applied a bit more consistently, if I'm understanding the intention correctly. (See inline comments.)

However, my main feedback for the PR would be that we should really try to follow a specification-first approach here.

In https://github.com/sig-net/mpc/pull/342, I wrote a first specification draft for you to review, which should ensure we cover all corner cases in the distributed algorithms we implement. Please take another look, I also just updated it with more explanations on the Prepare phase and its intentions.

In this PR, I see you are implementing `posits`. They seem to be similar in concept to what I called the INIT message, or maybe after the input from @volovyks what I called the START message in the specification.

First, can you describe in your words how posits are different from the concept in the specification, if they are different? As I understand it:
- Posits are currently skipping the two-phase protocol to select participants. But you want to add that later by waiting for Accept messages before actually participating.
- Posits can be rejected by peers. We should discuss this and add to the spec how that works and what it achieves.
- Aborting is also introduced in this PR. Is this related to Posits?

Second, could we agree on one single naming scheme? Either we update the specification, or this PR. I have no strong opinion but it's the first time I see `posits` used in distributed algorithm context. I like the name `Prepare` for this phase, since that's what the literature uses. For messages, maybe if INIT and START are too ambiguous, we could use PREPARE and START_PROTOCOL or something like that.

Having said that, I still think these code changes are great! We need to send new message types and handle them just like you implemented anyway. It's just the details that I would like to align on. And then we update either the spec or the PR so that they match.

chain-signatures/node/src/protocol/posit.rs
```diff
@@ -0,0 +1,181 @@
...
+                    return vec![PositInternalAction::Reply(PositAction::Reject)];
+                }
+
+                // Check that the participants are all active
+                for p in participants.iter() {
+                    if !active.contains(p) {
+                        tracing::warn!(?id, ?active, ?participants, "rejecting protocol posit");
+                        return vec![PositInternalAction::Reply(PositAction::Reject)];
+                    }
+                }
```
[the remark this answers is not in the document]
ChaoticTempest: I see, I'll remove this then and just let the messaging layer handle offline participants

chain-signatures/node/src/protocol/posit.rs
```diff
@@ -0,0 +1,181 @@
...
+        match action {
+            PositAction::Propose(participants) => {
+                if self.posits.contains_key(&id) {
+                    tracing::warn!(
+                        ?id,
+                        ?participants,
+                        "received a protocol posit for an id that we already proposed"
+                    );
+                    return vec![PositInternalAction::Reply(PositAction::Reject)];
+                }
```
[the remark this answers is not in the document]
ChaoticTempest: so this one checks on the level of posits that we own, but there is a check one level above that checks if its already generated. I do see that what also needs to be added is a check for what's currently generating.

chain-signatures/node/src/protocol/posit.rs
```diff
@@ -0,0 +1,181 @@
...
+
+                // Check that the participants are all active
+                for p in participants.iter() {
+                    if !active.contains(p) {
+                        tracing::warn!(?id, ?active, ?participants, "rejecting protocol posit");
+                        return vec![PositInternalAction::Reply(PositAction::Reject)];
+                    }
+                }
+
+                // Automatically join the protocol if we're accepting here.
```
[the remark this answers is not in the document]
ChaoticTempest: yep, I have a TODO comment about this in the level above

chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -216,6 +242,73 @@ impl PresignatureManager {
...
+            vec![PositInternalAction::Reply(PositAction::Reject)]
+        } else if self.contains(id.id).await {
+            tracing::warn!(?id, "presignature already generated");
+            Vec::new()
+        } else if matches!(action, PositAction::Abort) {
+            self.abort(from, id.id);
+            Vec::new()
+        } else {
+            self.posits.act(id.id, from, &action, active)
+        };
```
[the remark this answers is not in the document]
ChaoticTempest: they can be in one place but the issue is that `Posits` is generalized to be over triples, presignatures and signatures. Those don't have a general interface right now where we can call into stuff that checks or calls into things that rely on storage

ChaoticTempest:
> First, can you describe in your words how posits are different from the concept in the specification, if they are different? As I understand it:
>
> Posits are currently skipping the two-phase protocol to select participants. But you want to add that later by waiting for Accept messages before actually participating.

Yes, this skips the 2 phase setup for now, since I did it this way to get it working initially. The second phase will be added either in this PR or another PR.

> Posits can be rejected by peers. We should discuss this and add to the spec how that works and what it achieves.

I mentioned this in https://github.com/sig-net/mpc/pull/342/files#r2070563370, but the rejection is for additional checks made by each node just in case a node finds something that is wrong and we would like to not progress this protocol.

> Aborting is also introduced in this PR. Is this related to Posits?

Aborting is currently a stop-gap for the 1 phase setup since protocols are immediately started. This should probably not exist when the 2 phase setup is implemented. But I could see it being maintained to a degree in the 2 phase setup if other nodes were to maintain state about what proposals are in the 1st vs 2nd phase. So the abort would be useful for cleaning up that state if we early abort (but could also see just using a timeout to expire them).

> Second, could we agree on one single naming scheme? Either we update the specification, or this PR. I have no strong opinion but it's the first time I see posits used in distributed algorithm context. I like the name Prepare for this phase, since that's what the literature uses. For messages, maybe if INIT and START are too ambiguous, we could use PREPARE and START_PROTOCOL or something like that.

Yeah no strong opinions on the naming. I just went with `Posit` to see whether it could work or not. The issue is that prepare and start_protocol doesn't have a good categorization as a noun as far as I see. Posit on the other hand acts as both a noun and a verb, and you can attach actions like `accept` and `reject` to it. It's also closely related in the mathematical sense. We could categorize them just purely under `Protocol` just to generalize over all states the protocol can be in:
```
enum Protocol {
    Prepare(...),
    Start(...),
    Generating(...),
    Stored(...),
    Taken(...),
}
```

chain-signatures/node/src/protocol/posit.rs
```diff
@@ -0,0 +1,181 @@
...
+                vec![
+                    PositInternalAction::StartProtocol(
+                        participants.clone(),
+                        Positor::Deliberator(from),
+                    ),
+                    PositInternalAction::Reply(PositAction::Accept),
+                ]
+            }
```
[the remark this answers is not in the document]
ChaoticTempest: removed in the latest code changes

ChaoticTempest:
In the latest changes, I made it a 2 phase setup, such that the proposer will only broadcast out the start action if and only if we have enough participants that accept. So if someone rejects, they will no longer be a part of the protocol. Only the accepts will be a part

ChaoticTempest:
I've also left the naming as `Posit` until we can get to an agreement as to how to name it. The fundamental issue with it is that it starts to gravitate towards being Protocol states entirely which starts to mix with our redis storage and how we can represent state there.

For example, it ends up becoming the full entire protocol lifecycle `Propose => Start => Generating => Available => Taken => Used`. It might be better to fully represent this in storage for consistency and as the one true source of protocol state, but at the same time we require ways to clean it up in cases where node crashes in the midst of one of these states.

chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -245,100 +338,96 @@ impl PresignatureManager {
...
+        let owner = entry.get().owner;
+        if owner != proposer {
+            tracing::warn!(
+                id,
+                ?owner,
+                claimed_proposer = ?proposer,
+                "received abort request for presignature generation where owner does not match"
+            );
+            return;
+        }
```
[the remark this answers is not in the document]
ChaoticTempest: gotcha, but I removed the abort function after removing that action completely

chain-signatures/node/src/protocol/posit.rs
```diff
@@ -0,0 +1,177 @@
...
+            PositAction::Start(participants) => {
+                if self.posits.contains_key(&id) {
+                    tracing::warn!(?id, "received an invalid Start for a protocol we proposed");
+                    return PositInternalAction::Reply(PositAction::Reject);
+                }
+
+                PositInternalAction::StartProtocol(
+                    participants.to_vec(),
+                    Positor::Deliberator(from),
+                )
```
jakmeier, line 127: Please add a check that we are currently in `Prepare` state before we allow starting the protocol. (Which means you have to track that first. We need this anyway, as mentioned in this comment on the spec: https://github.com/sig-net/mpc/pull/342#discussion_r2079145826.)

If we allow going from `init` directly to `Running`, without a check that we are in `Prepare`, we have no way of knowing whether the `INIT` message was sent and received.

I guess technically, this specific case isn't so bad, as long as all nodes behave correctly. But a principle of distributed systems is that we can't assume other nodes to behave correctly. If another node sends a `START` message without sending us `INIT` first, this is a violation of the specification and leads to undefined behaviour. We should check for such a case.

Explicitly checking that we only do defined state transitions makes it much easier to validate that our implementation matches the specification.

chain-signatures/node/src/protocol/posit.rs
```diff
@@ -0,0 +1,177 @@
...
+                    counter.accepts.insert(from);
+                } else {
+                    counter.rejects.insert(from);
+                }
+
+                let enough_votes = counter.accepts.len() >= threshold
+                    && counter.accepts.len() + counter.rejects.len() == counter.participants.len();
+                if !enough_votes {
+                    return PositInternalAction::None;
+                }
```
jakmeier, line 153: Could we also check if we have too many rejections already? Maybe not in this PR but it would be nice to fail early if we know we can't find enough peers with the necessary data still available.

chain-signatures/node/src/protocol/posit.rs
```diff
@@ -0,0 +1,177 @@
...
+
+                tracing::info!(?id, "received all Accepts, starting protocol");
+                let counter = entry.remove();
+                let participants = counter.accepts.iter().copied().collect();
```
jakmeier, line 157: Would it make sense to check if we are in the list of participants and reject otherwise?

jakmeier (commented):
> In the latest changes, I made it a 2 phase setup, such that the proposer will only broadcast out the start action if and only if we have enough participants that accept. So if someone rejects, they will no longer be a part of the protocol. Only the accepts will be a part

Awesome!

> I've also left the naming as Posit until we can get to an agreement as to how to name it.

Fair, let's keep the name for now. Changing names is easy. As long as we agree on the conceptual design, I'm fine with whatever name.

> It might be better to fully represent this in storage for consistency and as the one true source of protocol state, but at the same time we require ways to clean it up in cases where node crashes in the midst of one of these states.

Yes, we should move towards having these states in Redis, in future PRs. Some clean-up logic will be necessary, yes. The point of persisting is to handle cases where we crash. In some states we can recover, in some others we need to clean up. But I expect the clean-up code to be quite minimal.


Overall, we are getting close to where I think we can merge this. But the `Prepare` phase for all nodes is crucial to have, otherwise it's still not a real 2-phase protocol.

chain-signatures/node/src/protocol/posit.rs
```diff
@@ -0,0 +1,177 @@
...
+                store,
+            },
+        );
+
+        PositAction::Propose
+    }
+
+    // TODO: make the resp of this synchronous when each of the protocol managers
+    // are their own individual tasks such that they can respond without the need
+    // of the main consensus loop.
```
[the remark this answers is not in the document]
ChaoticTempest: yeah, I'm not so sure why I wrote that originally anymore. I think originally I had it outside of the messaging layer, so this comment is now outdated. Will remove

chain-signatures/node/src/protocol/posit.rs
ChaoticTempest: Yep, that's possible, but requires adding back the abort I got rid of. Not a bad thing to do, since it would be nice to abort the ongoing protocol, and have the proposer restart it

chain-signatures/node/src/protocol/posit.rs
ChaoticTempest: yeah that would make sense. There could be a case in the future where our changes lead to where we send out a start on a node that's not in the list

chain-signatures/node/src/protocol/posit.rs
```diff
@@ -0,0 +1,177 @@
...
+    ) -> PositInternalAction<S> {
+        match action {
+            PositAction::Propose => {
+                if self.posits.contains_key(&id) {
+                    tracing::warn!(
+                        ?id,
+                        "received a protocol posit for an id that we already proposed"
+                    );
+                    return PositInternalAction::Reply(PositAction::Reject);
+                }
```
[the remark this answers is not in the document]
ChaoticTempest: Yep, will add that check. But I'll also keep this check, because this one is about conflicting propositions. Like other nodes might try to propose a protocol with an ID that conflicts with what we have proposed. This check will keep us from having multiple protocols with the same ID floating around

chain-signatures/node/src/protocol/posit.rs
ChaoticTempest: Added alongside the other ones when checking the correct state that we are in for a particular posit

chain-signatures/node/src/protocol/posit.rs
ChaoticTempest: Added the check but will not abort the protocol for now. Left it as a TODO in the future

chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -150,6 +174,9 @@ pub struct PresignatureManager {
     generators: HashMap<PresignatureId, PresignatureGenerator>,
     /// The set of presignatures that were introduced to the system by the current node.
     introduced: HashSet<PresignatureId>,
+    /// The protocol posits that are currently being proposed by us.
+    posits: Posits<PresignatureId, TriplesTaken>,
```
jakmeier, line 178: Same here: Is this really only for posits proposed by "us"?

chain-signatures/node/src/protocol/posit.rs
```diff
@@ -0,0 +1,270 @@
...
+}
+
+impl<T: Copy + Hash + Eq + fmt::Debug, S> Posits<T, S> {
+    pub fn new(me: Participant) -> Self {
+        Self {
+            me,
+            // 1024 is a good default size for the cache since we don't expect to be have more
+            // than 1024 concurring posits at any time. This should however be configurable in
+            // the future.
+            posits: LruCache::new(NonZeroUsize::new(1024).unwrap()),
```
jakmeier, line 80: Can you explain a bit more about the motivation / benefits for the LRU cache?  (instead of an unlimited set like HashSet)

I guess strictly limited memory usage can be nice. But shouldn't we be deliberate about which posits we abandon?

I think dropping posits as it is coded now can hurt robustness in certain fault scenarios.
For example, a malfunctioning node that sends many `PositAction::Propose` messages will flush out valid posits on all its peers. This can potentially halt the entire system, if all other nodes can't get posits accepted before they are flushed out of the LRU cache.

And if dropping an entry from self.posits is equivalent to going from `Prepare` back to `init`, this also conflicts with the proof table in the specification.

chain-signatures/node/src/protocol/posit.rs
```diff
@@ -0,0 +1,270 @@
...
+                            ?proposer,
+                            "received START on conflicting proposer"
+                        );
+                        self.posits.put(id, positor);
+                        return PositInternalAction::Reply(PositAction::Reject);
+                    }
+                } else {
+                    tracing::warn!(?id, ?from, "received START on protocol we have no info for");
+                    return PositInternalAction::Reply(PositAction::Reject);
+                }
```
jakmeier, line 189: Just checking: Is this handling the case where we get a `Start` message before the `Propose` message?

jakmeier (commented):
This is looking really good now! The only important question is about the LRU cache.

Otherwise, just one more thing: Please consider if you can add unit tests for posits. I think you have written a nice and clean code architecture that should make it quite convenient for testing. The best timing to add such tests is in this PR, second best timing is right afterwards. :)

chain-signatures/node/src/protocol/posit.rs
```diff
@@ -0,0 +1,270 @@
...
+    store: S,
+}
+
+/// A collection of posits that are being proposed. This is used to track
+/// the posits that are being proposed and the participants that have
+/// accepted them.
+pub struct Posits<Id, S> {
+    me: Participant,
+
+    // TODO: probably should be a LruCache to expire items
```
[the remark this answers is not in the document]
ChaoticTempest: ahh yeah, will remove that

chain-signatures/node/src/protocol/posit.rs
```diff
@@ -0,0 +1,270 @@
...
+    /// The posits that our node has proposed.
+    posits: LruCache<Id, Positor<PositCounter<S>>>,
```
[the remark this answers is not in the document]
ChaoticTempest: will update

chain-signatures/node/src/protocol/posit.rs
ChaoticTempest: that's true, it's definitely better to have our own explicit cleanup logic here too. I'll just change it back to hashmap. Initially just had the lru cache just so there weren't many leftover posits that never get cleaned up an "leak" memory but the storage for them is decently small (a `u32` per entry) anyways.

chain-signatures/node/src/protocol/posit.rs
ChaoticTempest: yep, the node does not have any info about this posit, so that would mean no proposer has sent out a propose yet

jakmeier (approved):
Thanks for adding the tests. I'd say this is ready to merge in now.

PR #360, opened 2025-05-06
feat: rely less on managers and more on storage objects for triples and presignatures

ChaoticTempest:
This makes it so that we are going to rely less on the manager objects going forward for things about storage. Instead we should directly call into `*_storage` objects since they provide the concurrency we need.

- renamed len_mine to len_by_owner for clarity since we are passing in the owner/me anyways
- renamed contains_mine/contains_foreign to contains_by_owner since they're one and the same.
- made triple and presignature primitives more concurrent by purely relying on the storage objects instead of manager. 
- note that with this change, the Presignature/Triple is missing or is currently generating error will not bubble up, and solely rely on the timeout mechanism to expire the messages for these protocols. It's effectively the same behavior since we were clearing the queue of messages before when something was missing, but other nodes would be doing the same thing, effectively expiring the protocol.
- note that we still rely on the managers where generating state is still required. I'll have a follow up PR where we handle that case. Storage should be tracking this state too but I'll keep it from this change where it's purely a refactor. This generating state will live in the memory level and not in the storage level. This is because once the node crashes or restarts, we will have no way to recover on these protocols so best if the current generating set also gets reset on start up such that we need no cleanup


chain-signatures/node/src/storage/mod.rs
```diff
@@ -52,6 +52,33 @@ impl Options {
     }
 }
 
+#[derive(Debug, thiserror::Error)]
+pub enum StoreError {
+    #[error("protocol is generating: {0}")]
+    ProtocolIsGenerating(String),
```
volovyks, line 58: Protocol is a bit misleading, we are not accessing the protocol. Maybe `Entry`?

chain-signatures/node/src/protocol/triple.rs
```diff
@@ -589,69 +588,16 @@ impl TripleManager {
...
 
     pub async fn contains_used(&self, id: TripleId) -> bool {
         self.triple_storage.contains_used(id).await
     }
 
-    /// Take two unspent triple by theirs id with no way to return it. Only takes
-    /// if both of them are present.
-    /// It is very important to NOT reuse the same triple twice for two different
-    /// protocols.
-    pub async fn take_two(
```
volovyks, line 603: Managers seam redundant now. Do you want to get rid of them in the future?

volovyks (approved):
Nice refactoring

chain-signatures/node/src/storage/mod.rs
ChaoticTempest: whoops, I actually never use this error type. Was leftover from what I originally envisioned with the storage objects returning this error

chain-signatures/node/src/protocol/triple.rs
ChaoticTempest: so they will just become their own individual tasks in the future. They should solely be responsible for spawning new generators (either through stockpiling or receiving a request to start a new one).

PR #366, opened 2025-05-12
Rename and restructure Solana program stub, add link to solana program repo

volovyks:


chain-signatures/contract-sol/programs/signet-program/Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
-name = "chain-signatures-project"
+name = "signet-program"
 version = "0.1.0"
-description = "Created with Anchor"
+description = "Solana program stub"
```
ChaoticTempest, line 4: probably don't want to call it stub here, maybe something like this instead:
```suggestion
description = "SOL program for Sig Network"
```

ChaoticTempest (approved):

chain-signatures/contract-sol/programs/signet-program/Cargo.toml
volovyks: @ChaoticTempest  The thing is that it is not a real contract (check Readme).

Why SOL?

ChaoticTempest: I see, well nevermind then. SOL is just the equivalent abbreviation with ethereum with ETH and its shorter

volovyks: SOL is a token, I will keep Solana

PR #371, opened 2025-05-16
feat: store & kick participant from triples and presignature on sync view

ChaoticTempest:
This adds in `*_participants` storage key where each triple or presignature will store a list of participant at the surface level of redis such that this list can be queried and modified easily. `SyncView` is now back such that the owning node will receive a list of triple/presignature shares that the other nodes have. The owning node will then use that info to remove participants from their own list for that particular triple or presignature. This will let future usage of taking of owned protocols to effectively only use those who report they have the shares.

This PR however does not implement the part where we utilize this list when taking a protocol from storage, but will be left for another PR instead


chain-signatures/node/src/storage/presignature_storage.rs
```diff
@@ -141,6 +140,19 @@ impl PresignatureStorage {
             .unwrap_or_default()
     }
 
+    pub async fn fetch_participants(&self, id: PresignatureId) -> Vec<Participant> {
```
volovyks, line 143: Keeping holders separately from Ps and Ts can create race conditions.

chain-signatures/node/src/storage/presignature_storage.rs
```diff
@@ -231,31 +243,53 @@ impl PresignatureStorage {
...
 
+    /// Kicks participants from the given presignatures.
+    pub async fn kick_participants(&self, kick: HashMap<PresignatureId, Vec<Participant>>) {
+        if kick.is_empty() {
+            return;
+        }
+        let Some(mut conn) = self.connect().await else {
+            return;
+        };
+        let mut pipe = redis::pipe();
```
volovyks, line 274: We were not using pipelines before, since they are not atomic. We can face race conditions.

chain-signatures/node/src/storage/presignature_storage.rs
ChaoticTempest: im not sure what you mean by race conditions here? Can you give a couple examples of such cases?

chain-signatures/node/src/storage/presignature_storage.rs
ChaoticTempest: look at the line below, calling `pipe.atomic()` makes it atomic

volovyks: You are sending all the actions in one batch, that is good. But during the execution they are not atomic.
It’s not terrible for this type of operation, though.

chain-signatures/node/src/storage/presignature_storage.rs
volovyks: Can it happen that we will store the per-signature,  but not the holders and interpret this as nobody holds it except us?

chain-signatures/node/src/protocol/sync/mod.rs
```diff
@@ -178,25 +184,59 @@ async fn broadcast_sync(
...
+            if !view.triples.contains(&id) {
+                entry.push(*p);
+            }
+        }
+    }
+    triples.kick_participants(triple_kick).await;
+
+    let mut presignature_kick = HashMap::new();
+    for &id in &update.presignatures {
+        let entry = presignature_kick.entry(id).or_insert_with(Vec::new);
```
jakmeier, line 215: nit: could use `or_default`

```suggestion
        let entry = presignature_kick.entry(id).or_default();
```

keeping it as-is is also fine

chain-signatures/node/src/storage/presignature_storage.rs
jakmeier: I'm not sure about this. [`pipe.atomic()`](https://docs.rs/redis/latest/redis/struct.Pipeline.html#method.atomic) wraps the commands in `MULTI` + `EXEC` which creates a [redis transaction](https://redis.io/docs/latest/develop/interact/transactions/). As I understand this, it should be just as atomic as a script. But I'm not super familiar with these operations.

jakmeier (dismissed):
Looks good to me overall.

@ChaoticTempest do you have a strategy for testing the DB timings and impact of it on overall performance? I'm worried this could make things quite unstable if we run sync too often.

chain-signatures/node/src/storage/presignature_storage.rs
volovyks: Oh... I'm quitting AI tools. In fact, MULTI + EXEC is atomic. Works a bit differently. I think for us, the biggest difference is that scripts are blocking all other operations during execution. That is more predictable, but still, this solution should work perfectly.

volovyks (dismissed):

volovyks (dismissed):

PR #372, opened 2025-05-16
chore(refactor): singular storage struct

ChaoticTempest:
This refactors `PresignatureStorage` and `TripleStorage` now to just use `ProtocolStorage` as their one and only storage struct since a lot of functions were duplicated between the two


volovyks (approved):
Nice refactoring

ChaoticTempest:
closing in favor of https://github.com/sig-net/mpc/pull/602

PR #374, opened 2025-05-20
feat: make CryptographicProtocol not return err

ChaoticTempest:
This is a smaller PR in the grander scheme of making our state machine more versatile. We originally had the premise that on any error, we should revert the state back to the previous state. This however created a couple of unforeseen issues along the way such as going into weird states we shouldn't be going to.

This is the start of making our state machine more explicit in the fact that we should only be transitioning into states we want to transition into. No panics and errors are to be returned, instead if we truly cannot recover, we should go into a FailedState where recovery is only feasible if we certain conditions are met and if not, we can either wait for network input or shutdown the node.

The changes in this PR are as follows:
- `CryptographicProtocol` no longer returns an error. We will be handling all cases as possible here.
- If either `Generating` or `Resharing` fails to store the secret share, we will temporarily store it in the state and have it try again in the next iteration. Before, this would just revert into Started/Running respectively for each state, where this is not great to have.
- On protocol error/timeout on either `generating` or `resharing`, we will actually refresh and restart the protocol. Before, this wasn't actually being refreshed. The cait-sith protocol was refreshed, but since we revert back to the previous state, the refresh gets dumped.

Future changes will make `ConsensusProtocol` no longer return error, with each error case being handled properly. Eventually, there will no longer be a clone on the `NodeState` to reflect a state change.

### DIFF
--- a/chain-signatures/node/src/protocol/consensus.rs
+++ b/chain-signatures/node/src/protocol/consensus.rs
@@ -186,6 +186,7 @@ impl ConsensusProtocol for StartedState {
                                 participants,
                                 threshold: contract_state.threshold,
                                 protocol,
+                                failed_store: Default::default(),
                             }))
                         }
                         None => {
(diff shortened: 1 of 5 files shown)


jakmeier (approved):
Make sense to me, I like the direction of the refactor!

PR #380, opened 2025-05-21
FEAT: Added GHA job to run on a schedule

auto-mausx:
This should allow the continual load tests to run on Solana dev on our self hosted runners. I kept the original CI test, but we can update that if needed later.

### DIFF
--- a/.github/workflows/k6-ci-loadtest.yml
+++ b/.github/workflows/k6-ci-loadtest.yml
@@ -10,6 +10,8 @@ on:
       - chain-signatures/**
       - infra/loadtests/**
       - .github/workflows/k6-ci-loadtest.yml
+  schedule:
+    - cron: '0 * * * *'
   workflow_dispatch:
     inputs:
       lt_strategy:
@@ -47,7 +49,7 @@ on:
 jobs:
   load-test:
     if: ${{ github.event_name == 'push' || github.event.pull_request.merged == true || github.event_name == 'workflow_dispatch' }}
-    runs-on: arc-runners-arc-runner-set
+    runs-on: ubuntu-latest
     steps:
       - name: Wait for 5 minutes due to new version deployment
         if: ${{ github.event.pull_request.merged == true }}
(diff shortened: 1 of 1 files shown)


.github/workflows/k6-ci-loadtest.yml
```diff
@@ -94,3 +96,48 @@ jobs:
             --tag strategy=${{ steps.normalized.outputs.lt_strategy }}
             --tag run_id=${{ github.run_id }}
           inspect-flags: --include-system-env-vars
+
+  cron-test:
```
volovyks, line 100: We should be able to reuse the same job and trigger it with either cron, manually, or PR. With different parameters.

.github/workflows/k6-ci-loadtest.yml
```diff
@@ -94,3 +96,48 @@ jobs:
...
+
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Normalize inputs
+        id: normalized
+        run: |
+          echo "lt_chain=${CHAIN:-Solana}" >> "$GITHUB_OUTPUT"
+          echo "lt_env=${ENV:-dev}" >> "$GITHUB_OUTPUT"
+          echo "lt_strategy=${STRATEGY:-constant_low_rate_1h}" >> "$GITHUB_OUTPUT"
```
volovyks, line 112: This will give us 1 request every 5 minutes, we want a bit more for Solana.
`constant_medium_rate_1h` should be fine for now (6 requests per minute)

.github/workflows/k6-ci-loadtest.yml
auto-mausx: Long story short, I don't think there is a good way to do that with Actions. The PR job and manual job will run the first "load-test" job, but the "cron-test" job will run every hour. If we want multiple cron jobs to run on multiple environments, then we can either:
1. Add an additional cron job to this workflow file for each env and network (hardcode default arguments)
2. Add another workflow file that only manages cron jobs with default values (also hard coding default args)

The only way to pass inputs to GHA is with the manually run workflow, or via PR with tag values is another option.

PR #390, opened 2025-05-28
fix clippy generate_keys

ppca:


ppca:
> Why are we excluding generate_keys? Is it hard to fix clippy errors?

No clippy errors was separate, that was fixed by just removing `use hex` from the generate_keys main.rs.

Excluding generate_keys is because I do not want to build generate_keys when building the package when building image using Dockerfile. 
Before this PR, generate_keys use the workspace built at root of mpc, which is not necessary, as it is completely separate and do not use any dependencies built by mpc. The change in this PR will separate generate_keys completely from the rest of the code, i.e. it does not rely on the workspace established at root of mpc, and it will be cargo built with its own config in its .cargo/config.toml, not using the config at root of mpc any more. This also means that changing mpc won't impact generate_keys either.

volovyks (approved):

Issue #382, opened 2025-05-22
helios eth finality: store past finalized_epoch locally


ppca:
We need this so that retried blocks can also be processed in `send_requests_when_final()`.

One way to do this is actually just extend the `finalized_blocks_map` in `send_requests_when_final()` to contain 2 epochs instead of 1.

PR #391, opened 2025-05-28
store past 2 finalized epochs locally so retried blocks get processed

ppca:
https://github.com/sig-net/mpc/issues/382


volovyks:
@ppca there is something wrong with this PR diff

PR #394, opened 2025-06-02
blacksand digest

volovyks:


auto-mausx (dismissed):

auto-mausx (dismissed):

chain-signatures/node/src/cli.rs
```diff
@@ -376,6 +376,30 @@ mod tests {
 
     const ETH_CONTRACT_ADDRESS: &str = "f8bdC0612361a1E49a8E01423d4C0cFc5dF4791A";
 
+    #[test]
+    fn test_digest_blacksand() {
```
ChaoticTempest, line 380: We should eventually move this to a separate crate/repo and not have this here. The purpose of these is pretty much just a check for sanity and it's not directly a test that should be ran constantly on CI

ChaoticTempest (dismissed):

volovyks:
I've separated and extended our digest coverage. It should allow us to better track configuration changes and resolve issues faster.
Also:
- I've made `sign_sk` a required parameter. It was `Option<>` for purely historical reasons.
- Extended starting log parameters list

PR #403, opened 2025-06-05
eth finality: fix stuck refresh_finalized_epoch

ppca:
add retries to fetching blocks and continue instead of  returning Err


chain-signatures/node/src/indexer_eth.rs
```diff
@@ -653,46 +687,47 @@ async fn refresh_finalized_epoch(
...
-                return Err(anyhow::anyhow!(
-                    "Failed to get block {i} from untrusted RPC client"
-                ));
-            };
+
+            let cur_block =
+                match get_block_from_untrusted_rpc_with_retry(untrusted_rpc_client, i, 5).await {
+                    Ok(block) => block,
+                    Err(e) => {
+                        tracing::warn!("Error fetching finalized block {i}: {:?}", e);
```
volovyks, line 707: Lets not use "Error" in the context of a warning.

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -593,6 +593,40 @@ async fn add_failed_block(
...
+                return Err(anyhow::anyhow!(err_msg));
+            }
+            Err(e) => {
+                if retries < max_retries {
+                    retries += 1;
+                    tracing::warn!(
+                        "Failed to get latest finalized block from Helios client: {:?}, retrying",
+                        e
+                    );
+                    tokio::time::sleep(Duration::from_millis(200)).await;
```
volovyks, line 618: Will exponential back-off with limit on time work better?

ppca: Done!

volovyks (approved):
Please, try reducing code duplication once you are back.

PR #404, opened 2025-06-07
feat: add protocol posits to triples

ChaoticTempest:
This adds the posit feature to triples such that they can only be started when threshold participants accept.


volovyks (approved):
looks good

PR #405, opened 2025-06-09
bug: fix infinite loop in send_request_when_final

ppca:


ChaoticTempest:
it looks good to me, but how does this fix inifnite loop? Just looks like a refactoring with early termination instead now

ChaoticTempest (approved):
The initial `loop`, `while` can also be merged into a singular `loop`

PR #407, opened 2025-06-10
feat: presignature as individual tasks

ChaoticTempest:
This makes it so that presignature generation is done through a separate task similar to https://github.com/sig-net/mpc/pull/388 with triples. One extra thing that had to be done with this PR was that the new presignature task will have to check by itself whether the triples are available or not in storage. If they're the owner, they already have the triples, but if they're not, then they will need to fetch and wait until its available in storage.


volovyks (approved):
LGTM!

volovyks (approved):

PR #410, opened 2025-06-12
select signature proposer based on round

jakmeier:
The current proposer selection depends on the stable set. This is non-deterministic.

Long-term, I propose we have a concept of rounds and use timeouts to iterate through rounds. Within each round, proposer selection can be deterministic.

For now, in this commit, I propose we start using such a deterministic algorithm based on rounds. If it does not yield a stable participant, increase the round.

This is still non-deterministic overall.
But every node who sees the first deterministic proposer as stable, will chose it. So it's much more likely that a majority will pick the same proposer, even if they have different stable sets.


ChaoticTempest:
I like the idea for this, I'll try and pull it into what we have

jakmeier:
> I like the idea for this, I'll try and pull it into what we have

Okay, I'll close this then

PR #412, opened 2025-06-16
FIX: increase loadtest response time thresholds to be more realistic

auto-mausx:


volovyks:
What is the average now?

auto-mausx:
Average response time is just above 30s, currently we have a 1500ms threshold.

volovyks:
1500 is not realistic, but 5000 (5s) should be our target.

auto-mausx:
Done, this should give is a better target to aim for.

volovyks:
stale

PR #413, opened 2025-06-16
feat: signature generation as individual tasks

ChaoticTempest:
Added signature tasks for each signature generation. This does not refactor how retrying is done. It is still the responsibility of `SignatureSpawner` to retry signatures once a signature task fails for whatever reason. In such cases, a `SignError::Retry` will be used to retry.

This adds in `PendingRequests` and `PendingPresignature` to wait for the corresponding dependencies before they're available. `SignQueue` no longer pops requests until the request is completely done. We only maintain a queue of `SignId` related to which requests we need to propose ourselves, and a mapping of current requests that may or may not be generating. This makes it easier to manage these requests since we do not need to move around requests anymore. This also is in anticipation of moving towards persistent sign queues.

Note with this change, while in the running state, there's no protocol advancement, progression or messages are processed now since everything is just happening independently through tasks.

### DIFF
--- a/integration-tests/src/lib.rs
+++ b/integration-tests/src/lib.rs
@@ -61,13 +61,7 @@ impl Default for NodeConfig {
             },
             eth: None,
             // TODO solana: remove hardcoded values
-            sol: Some(SolConfig {
-                account_sk: "fS5jS6X5qvaquBV1bg2YWBdYeCiRSUwNAdNpgNkjS72oNxUJcZJZduaq2oCcXJb8erTbtqqq4wxriUmRJk7bMDw"
-                    .to_string(),
-                rpc_http_url: "https://api.devnet.solana.com".to_string(),
-                rpc_ws_url: "wss://api.devnet.solana.com/".to_string(),
-                program_address: "BtGZEs9ZJX3hAQuY5er8iyWrGsrPRZYupEtVSS129XKo".to_string(),
-            }),
+            sol: None,
         }
     }
 }


chain-signatures/node/src/protocol/mod.rs
```diff
@@ -121,7 +121,7 @@ impl MpcSignProtocol {
             let sleep_ms = match node.state {
                 NodeState::Generating(_) => 500,
                 NodeState::Resharing(_) => 500,
-                NodeState::Running(_) => 100,
+                NodeState::Running(_) => 500,
```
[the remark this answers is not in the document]
ChaoticTempest: this is not true. Running state advancement no longer does signature requests nor its messages

chain-signatures/node/src/protocol/message/mod.rs
```diff
@@ -288,13 +289,8 @@ impl MessageInbox {
...
             .retain(|id, _| !self.filter.contains_id(*id, Protocols::Presignature));
-        self.signature.retain(|_epoch, messages| {
-            messages.retain(|_id, messages| {
-                messages.retain(|msg| !self.filter.contains(msg));
-                !messages.is_empty()
-            });
-            !messages.is_empty()
-        });
+        self.signature
+            .retain(|id, _| !self.filter.contains_id(id.id(), Protocols::Signature));
```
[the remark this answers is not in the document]
ChaoticTempest: this is not accurate, as it requires a u64 instead of `SignId` type. The `id()` call returns a u64 which is a hash of both sign id and presignature id

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -59,12 +101,15 @@ pub struct SignQueue {
...
     /// The requests that belong to us where we will the propose the signature to the chain.
-    my_requests: VecDeque<SignRequest>,
-    /// All other requests. Can be requests that we are not the proposer, or one where
-    /// we do not belong to the signature generating subset.
-    other_requests: HashMap<SignId, SignRequest>,
+    my_requests: VecDeque<SignId>,
     /// Set of request that failed to be processed during signature generation
-    failed_requests: VecDeque<SignRequest>,
+    failed_requests: VecDeque<SignId>,
+    /// The pool of rqeuests that we are about to process or are currently processing. Only
```
jakmeier, line 107:
```suggestion
    /// The pool of requests that we are about to process or are currently processing. Only
```

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -80,32 +125,33 @@ impl SignQueue {
...
+            .count()
     }
 
     pub fn is_empty_mine(&self) -> bool {
         self.len_mine() == 0
     }
 
     pub fn len(&self) -> usize {
-        self.my_requests.len() + self.other_requests.len()
+        self.requests.len()
```
jakmeier, line 146: `len()` and how it's used in `is_empt()` is no longer 100% obvious, as we are not counting pending requests. Did you check that callers are okay with that? And should we maybe add comments like "length of running requests, excluding pending requests" to the two methods?

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -194,14 +241,21 @@ impl SignQueue {
...
             }
+            if let Some(pending) = self.pending.remove(&sign_id) {
+                if pending.send(request.clone()).is_err() {
+                    tracing::warn!(
+                        ?sign_id,
+                        "pending sign request channel closed before able to send request"
+                    );
+                }
+            }
+
```
jakmeier, line 257: Is it possible under any circumstances that an incoming request can't be matched? Maybe bad timing or  maybe a duplicated incoming request?

jakmeier (approved):
Left a few questions to consider but overall looks  good.

I like the separation of signature ids in the queue and the requests themselves in a hashmap.

chain-signatures/node/src/protocol/signature.rs
ChaoticTempest: so, this should be sequential right now since only `SignatureSpawner` organizes requests. When requests come in, this is the only point where they would get organized. For duplicate requests, a pending request would never get created since the request is already in the system and `get_or_pending` would return `PendingRequest::Available(request)` here.

Maybe it's less clear right now how the flow of requests goes but that just means we need to do some refactoring to make it clearer

jakmeier (dismissed):

ChaoticTempest:
hmm, for some reason integration tests keep failing with nonce issues or deadline issues. I need to investigate more since it's too consistent for this to be happening. I wonder if it's because now signatures are their own task, it increased CPU time a bit but it shouldn't since there's not as many signature tasks as there are triples and presignature tasks going on

PR #417, opened 2025-06-20
feat: lower message loop sleep interval

ChaoticTempest:
Testing this to see how much better our messaging is. In the future, sending messages out should be a bit smarter and more reactive by reaching the payload limit or timeout.

With this, I've noticed a slight performance increase in our `test_signature_basic` test by about 5-10s

### DIFF
--- a/chain-signatures/node/src/protocol/posit.rs
+++ b/chain-signatures/node/src/protocol/posit.rs
@@ -45,7 +45,7 @@ impl PositAction {
 pub enum PositInternalAction<S> {
     StartProtocol(Vec<Participant>, Positor<S>),
     Reply(PositAction),
-    Rejected,
+    Abort,
     None,
 }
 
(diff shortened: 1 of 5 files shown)


chain-signatures/node/src/protocol/posit.rs
```diff
@@ -45,7 +45,7 @@ impl PositAction {
 pub enum PositInternalAction<S> {
     StartProtocol(Vec<Participant>, Positor<S>),
     Reply(PositAction),
-    Rejected,
+    Abort,
```
jakmeier, line 48: why is this renamed in this PR? :sweat_smile:

jakmeier (approved):

chain-signatures/node/src/protocol/posit.rs
ChaoticTempest: Because I promised to rename it in a different PR from this one: https://github.com/sig-net/mpc/pull/421

PR #420, opened 2025-06-23
Remove fmt layer on GCP

volovyks:
My guess is that this is the reason we have duplicated logs on GCP.


volovyks:
`cases::test_multichain_reshare` is still unstable :/

PR #421, opened 2025-06-23
fix: signature ongoing collisions

ChaoticTempest:
Currently, since posits are checking whether or not a sign_id is ongoing or not, it is likely that there's race conditions between two different attempts for the same sign request. For example, a sign request can be ongoing but then fail for some reason. Some nodes will still have it as ongoing, but others will have moved on and tried to restart it. Best to just maintain two separate sign request attempts at this point (such as a mapping of (SignId, PresignatureId) to SignAttempt).

This also adds in a check for checking the proposer is correct or not on the sign request, since sign request has this knowledge early on based on organization. Before, we were reorganizing requests on every potential failure, but since we added posits, things are a bit different now. So we have to send it back into the sign queue for reordering on every rejection

### DIFF
--- a/chain-signatures/node/src/protocol/posit.rs
+++ b/chain-signatures/node/src/protocol/posit.rs
@@ -45,6 +45,7 @@ impl PositAction {
 pub enum PositInternalAction<S> {
     StartProtocol(Vec<Participant>, Positor<S>),
     Reply(PositAction),
+    Rejected,
     None,
 }
 
@@ -233,7 +234,7 @@ impl<T: Copy + Hash + Eq + fmt::Debug, S> Posits<T, S> {
                 if enough_rejections {
                     tracing::info!(?id, rejects = ?counter.rejects, "received enough REJECTs, aborting protocol");
                     entry.remove();
-                    return PositInternalAction::None;
+                    return PositInternalAction::Rejected;
                 }
 
                 // TODO: have a timeout on waiting for votes. The moment we have enough threshold accepts,
@@ -377,6 +378,6 @@ mod tests {
         let action = posits0.act(id, Participant::from(1), threshold, &PositAction::Reject);
         assert!(matches!(action, PositInternalAction::None));
         let action = posits0.act(id, Participant::from(2), threshold, &PositAction::Reject);
-        assert!(matches!(action, PositInternalAction::None));
+        assert!(matches!(action, PositInternalAction::Rejected));
     }
 }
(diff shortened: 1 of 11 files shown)


ChaoticTempest:
snowballing this a bit by making toxiproxy only enabled when needed in tests

integration-tests/src/cluster/mod.rs
```diff
@@ -166,7 +167,8 @@ impl Cluster {
                 node.account
             }
             None => {
-                let account = self.worker().dev_create_account().await?;
+                let account = utils::dev_gen_indexed(self.worker(), self.account_idx).await?;
+                self.account_idx += 1;
```
volovyks, line 171: Can we be sure this index is not used? We are using `dev_gen_indexed()` and always start from in `create_accounts()`.

chain-signatures/node/src/protocol/posit.rs
```diff
@@ -45,6 +45,7 @@ impl PositAction {
 pub enum PositInternalAction<S> {
     StartProtocol(Vec<Participant>, Positor<S>),
     Reply(PositAction),
+    Rejected,
```
volovyks, line 48: `Rejected` does not sound like an action. Sounds like a status of something.

integration-tests/src/lib.rs
```diff
@@ -48,19 +48,18 @@ impl Default for NodeConfig {
                 max_concurrent_generation: 16,
                 max_concurrent_introduction: 2,
                 triple: TripleConfig {
-                    min_triples: 8,
-                    max_triples: 80,
+                    min_triples: 16,
```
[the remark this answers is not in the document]
ChaoticTempest: It won't slow down anything. Tests don't wait for these conditions to be met. They wait usually only on one presignature to be ready and nothing else

The reason why I set this to be higher is because non-determinism with who gets assigned a triple/presignature is higher when the number is lower. So there's really no reason to set these numbers low at all for tests

integration-tests/src/cluster/mod.rs
ChaoticTempest: what do you mean not used? This is always used in resharing tests

chain-signatures/node/src/protocol/posit.rs
ChaoticTempest: Yeah, I guess Abort would sound better. But if this is the only thing that needs changing, I'll change it in consequent PRs since I'd like this to go in before waiting on integrations to pass again

integration-tests/src/cluster/mod.rs
ChaoticTempest: the account_idx here starts from node.len() after initial create_accounts, so no collision will ever happen

volovyks: Oh, I have not seen that line, should be fine then

volovyks (approved):
Looks good overall, please, address all the TODOs in a followup PR

PR #422, opened 2025-06-23
fix: expire eth and solana requests with different timeouts

ppca:
1. add a field total_timeout to IndexedSignRequest
2. configure different chains to have a different total timeout 
3. expire requests in sign queue according to this total timeout

### DIFF
--- a/chain-signatures/node/src/indexer.rs
+++ b/chain-signatures/node/src/indexer.rs
@@ -233,6 +233,7 @@ async fn handle_block(
                     chain: Chain::NEAR,
                     unix_timestamp_indexed: crate::util::current_unix_timestamp(),
                     timestamp_sign_queue: Some(Instant::now()),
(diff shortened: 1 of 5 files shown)


volovyks (commented):
Make sense overall. From what I understood, we were removing some of the requests from the queue using generic config from the contract, but in reality, each chains should have it's own timeout based on it's finality.

What about the original contract parameter? Do we still need it? It is still used in two other places.

ppca:
> Make sense overall. From what I understood, we were removing some of the requests from the queue using generic config from the contract, but in reality, each chains should have it's own timeout based on it's finality.
> 
> What about the original contract parameter? Do we still need it? It is still used in two other places.

the original paramater in this case is used to bound the timeout of signature generation starting from sign queue time. Not sure if we intend it for other stuff. @ChaoticTempest

PR #423, opened 2025-06-26
feat: use watch::{Receiver, Sender} for Config, MeshState, ContractState

ChaoticTempest:
This removes all the `Arc<RwLock<T>>` for Config, MeshState and ContractState. Instead we will use `watch::{Sender, Receiver}` to be more reactive and listen for changes made to them. This requires less cycles as we no longer need to poll in certain scenarios and merely just listen for changes once woken by the runtime.


jakmeier (approved):
Nice one!

PR #439, opened 2025-07-06
Update indexer_sol subscriber to use emit_cpi!

Pessina:
Currently, the program uses emit!, but Solana/Anchor's Rust crates don't support subscribing to events emitted via Cross Program Invocation (CPI) calls. See the Anchor client library limitation: https://github.com/solana-foundation/anchor/blob/a5df519/client/src/lib.rs#L424

This means our MPC Solana indexer cannot detect signature requests made through CPI calls.

- Replace current Solana Indexer Event Subscriber with new emit_cpi! subscriber


ppca:
What is the status of this PR? Should I review it? @Pessina @volovyks

esaminu (dismissed):
LGTM

ppca:
while testing out bidirectional mpc code, I find this PR not working to extract events out of the logs in a sign respond event. Debugging it. Please don't merge it yet.

ppca:
> @ppca are you using the [new version](https://github.com/sig-net/signet-solana-program/pull/17) that uses `emit_cpi()`?

yeah after Osman updated it to use emit_cpi it now works

chain-signatures/node/src/indexer_sol.rs
```diff
@@ -158,6 +163,7 @@ pub struct SignatureRequestedEvent {
     pub algo: String,
     pub dest: String,
     pub params: String,
+    pub fee_payer: Option<Pubkey>,
```
volovyks, line 166: @Pessina do we need fee payer here? It is never used.

volovyks (approved):
Let's merge it and move forward.

P.S. Before merging, check if Solana indexer is functioning on Dev. We have issues with it now. We should know if this changes are good.

PR #442, opened 2025-07-07
FIX: Added an increase high rate load test for constant LTs

### LINKED ISSUE #437: Increase request frequency on Load Tests
### Description

Currently our load tests are at a relatively low constant rate, the idea of this task is to run ad-hoc tests that increase load to establish a new baseline for our constant SOL load tests

auto-mausx:
This may be a little optimistic for constant rate, 8 rps for 58 mins. We can reduce this amount if it's too crazy, but ideally we can see where we are at with a higher rate like this and adjust as needed. This may be at the upper end of the limit for RPC services, but with Alchemy we just pay for increased usage I believe.

closes #437

### DIFF
--- a/infra/loadtests/k6-load-test.js
+++ b/infra/loadtests/k6-load-test.js
@@ -34,8 +34,8 @@ const strategies = {
         maxVUs: 10,
         stages: [
           { duration: '1m', target: 1 },
-          { duration: '58m', target: 1 },
-          { duration: '1m', target: 0 },
+          { duration: '58m', target: 3 },
+          { duration: '1m', target: 1 },
         ],
       },
     },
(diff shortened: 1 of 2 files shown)


infra/loadtests/k6-load-test.js
```diff
@@ -53,9 +53,9 @@ const strategies = {
         preAllocatedVUs: 5,
         maxVUs: 20,
         stages: [
-          { duration: '1m', target: 1 },
-          { duration: '58m', target: 1 },
-          { duration: '1m', target: 0 },
+          { duration: '1m', target: 2 },
+          { duration: '58m', target: 8 },
```
volovyks, line 57: I would set it to 5 for now and see how it behaves.

Also, if we’re redefining what we consider a high load, let’s update our definitions of medium and low as well — for both constant and ramping flows.

volovyks (approved):

PR #444, opened 2025-07-07
eth indexer: use helios to directly query for finalized blocks

ppca:
Changes involved:
1. the finalized_block_channel only save the latest finalized block number
2. in send_requests_when_final, once the request's block number <= finalized block number, and its block hash == block hash from helios, we send the request to sign queue

### DIFF
--- a/chain-signatures/node/src/indexer_eth.rs
+++ b/chain-signatures/node/src/indexer_eth.rs
@@ -711,40 +711,28 @@ async fn refresh_finalized_block(
             final_block_number
         );
 
-        let Some(last_final_block_number) = final_block_number else {
-            tracing::info!("Last finalized block was None");
+        if final_block_number.is_none() || new_final_block_number > final_block_number.unwrap() {
+            tracing::info!("Found new finalized block!");
             final_block_number.replace(new_final_block_number);
-            send_finalized_block(&finalized_block_send, new_final_block_number).await?;
+            finalized_block_send
+                .send(new_final_block_number)
+                .await
+                .map_err(|err| anyhow!("Failed to send finalized block: {:?}", err))?;
             continue;
-        };
+        }
+
+        let last_final_block_number = final_block_number.unwrap();
 
         if new_final_block_number < last_final_block_number {
             tracing::warn!(
                 "New finalized block number overflowed range of u64 and has wrapped around!"
             );
-            continue;
         }
 
         if last_final_block_number == new_final_block_number {
             tracing::info!("No new finalized block");
-            continue;
         }
-
-        tracing::info!("Found new finalized block!");
-        final_block_number.replace(new_final_block_number);
-        send_finalized_block(&finalized_block_send, new_final_block_number).await?;
     }
-}
-
-// Helper function to send finalized block and handle errors
-async fn send_finalized_block(
-    finalized_block_send: &mpsc::Sender<u64>,
-    block_number: u64,
-) -> anyhow::Result<()> {
-    finalized_block_send
-        .send(block_number)
-        .await
-        .map_err(|err| anyhow!("Failed to send finalized block: {:?}", err))
 }
 
 #[allow(clippy::too_many_arguments)]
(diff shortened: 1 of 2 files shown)


chain-signatures/node/src/indexer_eth.rs
```diff
@@ -638,7 +605,48 @@ async fn add_failed_block(
...
+            .get_block(
+                BlockId::Number(BlockNumberOrTag::Number(block_number)),
+                false,
+            )
+            .await
+        {
+            Ok(Some(block)) => return Ok(block),
+            Ok(None) => {
+                let err_msg = "Block number {block_number} not found from Helios client";
+                return Err(anyhow::anyhow!(err_msg));
```
ChaoticTempest, line 626: just have this function return `Option<Block>` and log inplace. Very easy to end up double logging somewhere this way

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -487,30 +453,31 @@ pub async fn run(
...
-    let (finalized_epoch_send, finalized_epoch_recv) = finalized_blocks_channel();
+    let (finalized_block_send, finalized_block_recv) = finalized_block_channel();
 
     let client = Arc::new(client);
     let client_clone = Arc::clone(&client);
     tokio::spawn(async move {
-        tracing::info!("Spawned task to refresh finalized epoch's blocks");
-        refresh_finalized_epoch(
+        tracing::info!("Spawned task to refresh the latest finalized block");
+        refresh_finalized_block(
```
ChaoticTempest, line 462: why were we refreshing the epoch before, and now we only do block?

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -408,38 +406,11 @@ fn failed_blocks_channel() -> (
...
-#[derive(Debug, Clone)]
-struct FinalizedEpoch {
-    pub start_block: u64,
-    pub end_block: u64,
-    pub blocks: BlockNumberToHashMap,
-}
-
 const MAX_FINALIZED_BLOCKS: usize = 1024;
-fn finalized_blocks_channel() -> (mpsc::Sender<FinalizedEpoch>, mpsc::Receiver<FinalizedEpoch>) {
+fn finalized_block_channel() -> (mpsc::Sender<u64>, mpsc::Receiver<u64>) {
```
ChaoticTempest, line 410: is there no specific type for this u64? It's kind of vague what it is but I'm assuming block height. We should have a type alias if there's nothing provided

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -920,87 +821,54 @@ async fn process_block(
...
+            send_indexed_requests(
+                indexed_requests,
+                sign_tx.clone(),
+                node_near_account_id.clone(),
+            );
+        } else {
+            tracing::error!(
+                "Block {block_number} hash mismatch: expected {block_hash:?}, got {:?}. Chain re-orged.",
+                block.header.hash
+            );
```
ChaoticTempest, line 871: Is there something we should be doing here in the else portion? If something else is doing work after the fact to resolve the reorg, we should have a comment stating what's going to happen

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -920,87 +821,54 @@ async fn process_block(
...
-                if !finalized_epochs.is_empty() {
-                    let oldest_epoch = &finalized_epochs[0];
-                    if block_number < oldest_epoch.start_block {
-                        tracing::error!("Block {block_number} is in history");
-                        //TODO: handle the block in history case which happens when a block is retried
-                        break;
-                    }
-                }
+        // Wait for finalized block if needed
+        while finalized_block_number.is_none() || block_number > finalized_block_number.unwrap() {
```
ChaoticTempest, line 844: the unwrap() makes me question things so probably this is better:
```suggestion
        while finalized_block_number.map_or(true, |n| block_number > n) {
```

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -920,87 +821,54 @@ async fn process_block(
...
 
-                // Check if block is in history (older than our oldest epoch)
-                if !finalized_epochs.is_empty() {
-                    let oldest_epoch = &finalized_epochs[0];
-                    if block_number < oldest_epoch.start_block {
-                        tracing::error!("Block {block_number} is in history");
-                        //TODO: handle the block in history case which happens when a block is retried
-                        break;
-                    }
-                }
```
ChaoticTempest, line 980: is this being handled somewhere else or is it no longer necessary?

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -674,162 +682,55 @@ async fn fetch_finalized_block_from_helio(
...
-            new_finalized_block.header.number,
+            new_final_block_number,
             final_block_number
         );
 
-        let new_final_block_number = new_finalized_block.header.number;
-
-        let Some(last_final_block_number) = final_block_number else {
-            tracing::info!("Last finalized block was None");
+        if final_block_number.is_none() || new_final_block_number > final_block_number.unwrap() {
```
ChaoticTempest, line 714: same with this one:
```suggestion
        if final_block_number.map_or(true, |n| new_final_block_number > n) {
```

chain-signatures/node/src/indexer_eth.rs
ppca: because Helios latest release now enable fetching past 8k blocks from Helios, while previously it only allows for most recent few blocks. 
So previously, to get the blocks in the finalized epoch, which a lot of them are already not saved by helios, we have to query untrusted rpc for the block and walk back from the latest final block fetched from Helios so we can get proven finalized blocks. We save the finalized epoch so we don't have to do this all again each block that is in this epoch in send_requests_when_final.
Now, we can directly fetch blocks from Helios for the blocks in newly finalized epoch, so we can fetch it directly in send_requests_when_final when needed, no walking back or untrusted rpc needed.

chain-signatures/node/src/indexer_eth.rs
ppca: yeah this is no longer necessary because we are not using a locally saved finalized epoch to check whether a block is final.

PR #452, opened 2025-07-16
indexer: remove silent failures and block_heads_rx lagged error

ppca:
Our prev testnet release had 3 nodes silently failing to run eth indexer.
1. This PR changes both eth and solana indexer's run function to not silently fail.
2. this PR changes catchup to run as tokio task instead of before the main loop. This is necessary because with current dev logic, when catchup takes a long time (node offline for long), then if block_heads_rx is initialized before we catchup, then it will be lagged and when we receive from it, we will miss the older blocks when catchup began so another catchup is needed; if block_heads_rx is initialized after we catchup, then there's a gap between the end block of catchup and the current block, and we'd need to catchup again. We might need to catchup a few times until the gap between catchup's end block and current block is well within block_heads_rx's limit so that it's not lagged. If we run the catchup as a tokio task while main loop is running and initialize the block_heads_rx before catchup task is spawned, we won't have the gap problem at all and no more catchup is needed.

### DIFF
--- a/chain-signatures/node/src/indexer_eth.rs
+++ b/chain-signatures/node/src/indexer_eth.rs
@@ -489,6 +489,14 @@ pub async fn run(
 
     let (finalized_block_send, finalized_block_recv) = finalized_block_channel();
 
+    let mut block_heads_rx = match client.subscribe(SubscriptionType::NewHeads).await {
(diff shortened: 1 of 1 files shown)


volovyks (approved):
LGTM

PR #454, opened 2025-07-18
feat: make MessageInbox/Outbox fully async

ChaoticTempest:
This change makes it so that the message inbox and outbox are completely async. Triples, Presignatures, and Signature when subscribing for messages stays the same. Generating and Resharing states now get their messages within their `Action::Wait` step, where there's a global subscriber/receiver for generating and resharing messages in the `MpcSignProtocol`. When we make generating and resharing more async (as in their own task), then we can move these subscribers there. Until then, they'll stay in the global context of the node.

- No more `MessageReceiver` interface.
- No more `MessageExecutor` (replaced by `MessageInbox` and `MessageOutbox` individual tasks).
- PROTOCOL_LATENCY_ITER_MESSAGE is no longer used since message is now no longer a part of state machine.
- MESSAGE_QUEUE_SIZE was also removed since there's no more outgoing message queue.

Overall this PR makes messaging more robust, with no lock on the inbox. The inbox and outbox will not interfere with each other when receiving and sending messages, since there are no dependencies on each other. Our protocols should see a slight speed up

### DIFF
--- a/chain-signatures/node/src/cli.rs
+++ b/chain-signatures/node/src/cli.rs
@@ -293,7 +293,7 @@ pub async fn run(cmd: Cli) -> anyhow::Result<()> {
             let node = Node::new();
             let node_watcher = node.watch();
 
-            let (sender, msg_channel) = MessageChannel::spawn(
+            let msg_channel = MessageChannel::spawn(
                 client,
                 &account_id,
                 config_rx.clone(),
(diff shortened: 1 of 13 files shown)


volovyks (approved):
Looks good — definitely a step in the right direction!

ChaoticTempest:
@jakmeier feel free to review if you think there's anything that looks off, but this is pretty innocuous so gonna just merge for now

PR #458, opened 2025-07-23
feat: make MessageOutbox send on all participants

ChaoticTempest:
This makes it so that the message outbox no longer looks for active participants but looks for all participants. This is because only the protocols should be saying whether they want active/stable or what other metric. The job of the outbox is to just send, so we should use all participants.

This also makes it so that each message send is its own task, so we don't block the outbox from continuing. This supercedes https://github.com/sig-net/mpc/pull/171 since I cannot be bothered to resolve merge conflicts there


volovyks (approved):
I agree, let's not overwhelm this layer with active.

PR #465, opened 2025-07-30
Bidirectional communication diagram

volovyks:
image

Updates/proposals:
- Now we are not moving the deposit transaction on-chain, since it can be created on both the Client and Contract sides
- Now `DEX's read_respond()` only accepts calls from MPC DEX address, no need to sign the payload itself
- DEX.read_respond(`tx_output`) can be DEX.read_respond(`tx_output_hash`), but it will require DEX to run their indexer and provide the output, which will use even more gas. cc @DavidM-D

Questions:
- Why do we store both hash and signature? What is the vector of attack here?

Note:
- @DavidM-D actually, we have 3 options, `DEX.read_responce()` can be called by MPC admin, MPC DEX key, or MPC DEX User key. But I still think that the MPC DEX key is the best option since handling cases where the user is out of funds would be harder.

@Pessina @esaminu , please check these updates. I still need to compare this to our latest POC .

### DIFF
--- a/doc/bidirectional_communication.plantuml
+++ b/doc/bidirectional_communication.plantuml
@@ -1,0 +1,29 @@
+@startuml Sig.Network Bidirectional Communication (Solana <-> Ethereum)
+!theme amiga
+actor Browser
+participant "USDT Contract\non Ethereum" as USDT
+participant "DEX Program\non Solana" as DEX
+participant "Signing Contract\non Solana" as SignContract
+participant "MPC Network" as MPC
+
+Browser -> Browser: Derive user Ethereum DEX USDT \ndeposit address from main address
+Browser -> USDT: Deposit 10 USDT on user address
+
+Browser -> Browser: Build and save Ethereum tx to transfer USDT \nfrom user to main DEX address
+Browser -> DEX: Call **claim_fund(offchain_parameters)**
+
+DEX -> SignContract: Build same tx and call **sign_respond(tx_hash)**
+
+SignContract -> MPC: Request signature
+MPC -> MPC: Sign tx hash with User key \n(sender is DEX, path is user address) \nStore tx ID (hash + signature)
+MPC -> SignContract: Return signature
+SignContract -> Browser: Return signature
+Browser -> USDT: Execute saved transaction
+USDT --> MPC: Notify tx executed (using MPC Ethereum indexer)
+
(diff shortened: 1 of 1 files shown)


esaminu:
Good points @volovyks 
- We leaned towards a `ReadRespondedEvent` instead of a call back to the DEX contract for a few reasons one of them being that in Solana the DEX contract would have to pass the mpc the list of accounts it uses (e.g. to store the requestIds) in advance because they will need to be specified when calling the chain sig contract to call back the DEX contract successfully. There was also another security issue @DavidM-D brought up at the time wrt to the callback related to synchronization. @DavidM-D would be useful if you could detail it here or indicate if it's no longer a blocker. Also if we go this route the read_respond callback would no longer need to be signed since we can enforce the caller has to be the chain sig contract.
- Currently fake net mpc hashes the rlp to get the signature and then concats the signature with the unhashed rlp and hashes that to get the transaction hash to query for the transaction. Only the transaction hash is stored along with other data from the call https://github.com/sig-net/signet-solana-program/blob/bidirectional-communication/clients/response-server/sig-server.ts#L343-L355
- I am open to either of the three options (MPC admin, MPC DEX key, or MPC DEX User key) but ideally would want to do our best to stay away from MPC DEX User key because it would mean the DEX contract would have to run kdf each time

Pessina:
Regarding the issue @DavidM-D mentioned, the keys are already different, no? The first signature will always either be: 

   - path + user_wallet
   - path + contract_address

So we can't trick the system to sign something with the `respond` key, no? We could use the `root` or `root + path` as proposed by @volovyks

------------ 

I think during the calls with @esaminu and @DavidM-D we discussed that MPC directly calling `DEX.read_respond` could be tricky due to gas estimations and transaction building (e.g Solana require the caller to provide all the accounts involved on the tx). 

So I think we agreed that would be better to emit an event (the same way sign works today) with the sign_respond signature + payload signed and the client would be responsible to call the DEX contract. The DEX has the MPC root key (+ path) hardcoded to verify that the signatures it's from the MPC

volovyks:
@Pessina , sorry—I agreed with your argument but didn’t express it clearly in my last message. We don’t need to add "respond" to the path, since we use that key in a single flow. So, the response will be signed with derive(root_mpc_pk, dex_address).

volovyks:
I think it should be a part of documentation, since it involves many parties, not just the PMC, closing for now.

PR #470, opened 2025-08-01
fix: stockpiling not working due introduction

ChaoticTempest:
This should fix the issue with no mine presignatures. This is because when we abort a protocol, the `introduced` counter never gets resetted and so we end up with no presignatures being generated by that particular node if they've failed enough over time.


volovyks (approved):
LGTM, let's test it.

PR #475, opened 2025-08-05
mpc bidirectional

ppca:
This code now passes the complete successful deposit flow (https://github.com/sig-net/solana-contract-examples/blob/main/contract/tests/sign-respond-erc20.ts#L322). 

Next step:
Understand and see how to set up testing all the other cases in https://github.com/sig-net/solana-contract-examples/blob/main/contract/tests/sign-respond-erc20.ts.

I think we could start reviewing.


chain-signatures/node/src/storage/sign_respond_tx_storage.rs
```diff
@@ -0,0 +1,381 @@
...
+            .await
+            .inspect_err(|err| {
+                tracing::warn!(?err, "failed to get number of txs");
+            })
+            .unwrap_or(0)
+    }
+
+    /// Clear all tx storage
+    /// Return true if successful, false otherwise.
+    pub async fn clear(&self) -> bool {
```
volovyks, line 293: Do we need this function?

chain-signatures/node/src/storage/sign_respond_tx_storage.rs
```diff
@@ -0,0 +1,381 @@
...
+        match conn.sismember(&self.success_key, id.clone()).await {
+            Ok(exists) => exists,
+            Err(err) => {
+                tracing::warn!(?id, ?err, "failed to check if tx in completed set");
+                false
+            }
+        }
+    }
+
+    pub async fn contains_failed(&self, id: SignRespondTxId) -> bool {
```
volovyks, line 247: Why are we separating failed and pending?

chain-signatures/node/src/storage/sign_respond_tx_storage.rs
```diff
@@ -0,0 +1,381 @@
...
+
+        if let Err(err) = result {
+            tracing::warn!(?id, ?err, "failed to insert tx into storage");
+            false
+        } else {
+            true
+        }
+    }
+
+    pub async fn mark_tx_success(&self, id: SignRespondTxId) -> bool {
```
volovyks, line 103: What is the logic behind all failed/pending/successful statuses?

chain-signatures/node/src/protocol/mod.rs
```diff
@@ -203,6 +204,13 @@ pub async fn spawn_system_metrics(node_account_id: &str) -> tokio::task::JoinHan
     })
 }
 
+#[derive(Debug, serde::Serialize, serde::Deserialize, Clone, PartialEq, Eq, Copy, Hash)]
+pub enum SignRequestType {
+    Sign,
+    SignRespond,
```
volovyks, line 210: Why do we have SignRepond with and without data? Isn't that the same type?

ppca:
This is still work in progress, will address comments when it's in more stable shape (things are connected to one another)

chain-signatures/node/src/storage/sign_respond_tx_storage.rs
ppca: deleted.

chain-signatures/node/src/protocol/mod.rs
ppca: it's Sign, SignRespond(SignRespondRequestedEvent) and ReadRespond(tx). 
I expect the ReadRespond type to change its name since we will not have read respond soon

ppca:
This is working on top of Felippe's PR that changes Solana indexer to parse CPI events: https://github.com/sig-net/mpc/pull/439.

This PR has implemented:
1. the SignRespondTx
2. the SignRespondTxStorage
3. add field SignRequestType to IndexedSignRequest
4. the logic that process SignRespondRequestedEvent on Solana, create an IndexedSignRequest from it and add the IndexedSignRequest to the indexed requests channel. When the block where this SignRespondRequestedEvent is contained gets finalized, this sign request will be sent to sign queue and a signature will be generated
5. the logic that extract the eth tx to be watched from SignRespondRequestedEvent and the corresponding signature
6. the logic that watch the eth txs: with each block coming in from eth, we already process each block by parsing its logs for SignatureRequestedEvent, we now add the logic that look for eth txs that need to be watched among the block receipts, if such tx is found, it means that tx is completed, with a status fail or success.
7. the logic that process a completed eth tx and create an IndexedSignRequest and add to indexed requests channel in eth indexer. When the block that contains this eth tx gets finalized, this sign request will be sent to sign queue and a signature will be generated.

This PR has not implemented:
1. the checkpoint mechanism we discussed on Aug 7
8. the part that we actually call read_respond(). This is because we are planning a change on API from read_respond() on chain signatures contract to some other function on the dex contract. And i'm not clear about that new api and how the dex contract address will be passed in, so leaving that out for now.

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -879,6 +875,113 @@ async fn process_block(
...
+        sign_respond_tx_map
+            .read()
+            .await
+            .iter()
+            .map(|(id, tx)| (*id, tx.clone()))
+            .filter(|(_, tx)| tx.status == SignRespondTxStatus::Pending)
+            .collect()
+    };
+
+    for (tx_id, mut tx) in remaining_pending_txs {
```
ChaoticTempest, line 942: what's the difference between processing the first set of pending tx then processing remaining ones?

ChaoticTempest (commented):
did an initial review, will come back to do some more later, but looks promising so far

ChaoticTempest:
how are we testing this so far? Feels like this can easily regress if we don't have an integration test

jakmeier (commented):
Great progress!

There are lots of moving parts in this and I believe the first priority is to have something that *just about* works, without breaking things. We will need to iterate on it to make it more general for different chain to chain directions, more robust against error cases, refactor to make it more testable, add test cases, and potentially optimize async flows for performance. But I think we shouldn't wait too much before merging a first version of this to develop.

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -28,6 +34,7 @@ pub(crate) static MAX_SECP256K1_SCALAR: LazyLock<Scalar> = LazyLock::new(|| {
     )
     .unwrap()
 });
+const MAX_CATCHUP_BLOCKS: u64 = 8191;
```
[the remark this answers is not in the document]
volovyks: please, specify that in the comment

chain-signatures/contract-sol/src/lib.rs
```diff
@@ -25,6 +25,29 @@ pub mod signet_program {
 
         Ok(())
     }
+
+    pub fn read_respond(
+        ctx: Context<ReadRespond>,
+        request_id: [u8; 32],
+        serialized_output: Vec<u8>,
+        signature: Signature,
+    ) -> Result<()> {
```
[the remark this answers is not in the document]
volovyks: Yes, we agreed to call the DEX contract directly and pass an ECDSA signature along with the payload.

However, this approach would require updates to other parts of the demo, and the FE team is busy this week and next. I don’t want to block this merge, so let’s proceed with the current design for now and plan to switch to calling the DEX directly as a next step.

cc @esaminu @Pessina

chain-signatures/node/Cargo.toml
```diff
@@ -45,6 +45,13 @@ opentelemetry-appender-tracing = "0.29.1"
 alloy-signer-local = "1.0.9"
 alloy-sol-types = "1.2.0"
 helios = { git = "https://github.com/a16z/helios", branch = "simple-cache" }
+alloy-dyn-abi = "1.2.0"
```
volovyks: Let's follow the existing pattern and add new dependencies at the workspace level.

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -879,6 +875,113 @@ async fn process_block(
         return Ok(());
     };
 
+    let block_receipts_clone = block_receipts.clone();
```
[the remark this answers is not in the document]
ChaoticTempest: you don't need to clone it, you can just borrow it via `for receipt in &block_receipts`

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -879,6 +874,114 @@ async fn process_block(
...
+    let pending_txs: HashMap<SignRespondTxId, SignRespondTx> = {
+        sign_respond_tx_map
+            .read()
+            .await
+            .iter()
+            .map(|(id, tx)| (*id, tx.clone()))
+            .filter(|(_, tx)| tx.status == SignRespondTxStatus::Pending)
```
ChaoticTempest: we should swap the ordering here, so we can filter first, then clone only if we have a pending tx. The first way would be cloning everything

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -879,6 +874,114 @@ async fn process_block(
...
+
+        let read_respond_request: Option<IndexedSignRequest> = completed_tx
+            .create_sign_request_from_completed_tx(&client_clone, 6, total_timeout)
+            .await;
+        if let Some(read_respond_request) = read_respond_request {
+            read_respond_requests.push(read_respond_request);
+            sign_respond_tx_map
+                .write()
+                .await
+                .insert(receipt.transaction_hash.into(), pending_tx.clone());
```
ChaoticTempest: unnecessary clone on `pending_tx`

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -879,6 +874,114 @@ async fn process_block(
...
+            receipt.transaction_hash,
+            if status { "success" } else { "failed" }
+        );
+        let client_clone = client.clone();
+        let mut pending_tx = pending_tx.clone();
+        pending_tx.status = if status {
+            SignRespondTxStatus::Success
+        } else {
+            SignRespondTxStatus::Failed
+        };
```
ChaoticTempest, line 903: how does this status work? It's confusing to me that it returns a bool. I kind of feel like if it's false, the status shouldn't be known so maybe it hasn't been processed yet?

chain-signatures/node/src/indexer_eth.rs
ChaoticTempest: ahh, so this is processing all pending_tx that were not completed in the current block. How does comparing current nonce with tx nonce yield too low as failure? Because I'm wondering why the comparison here is `!=`

chain-signatures/node/src/indexer_sol.rs
```diff
@@ -473,3 +582,60 @@ async fn process_anchor_event(
 
     Ok(())
 }
+
+async fn subscribe_and_process_sign_events<T>(
```
ChaoticTempest, line 653: generic `<T>` here isn't used

chain-signatures/node/src/indexer_eth.rs
volovyks: This status caught my attention too. From what I see, we need Success and Failed only for respond call serialization. But how do they know how to de-serialize it? Ideally we should treat requests as pending and completed. Everything else is up to a DEX that will parse the response. Unless I misunderstood something.

chain-signatures/node/src/sign_respond_tx.rs
```diff
@@ -0,0 +1,608 @@
...
+    Failed,
+    Success,
+}
+
+#[derive(Debug, Clone, Hash, serde::Serialize, serde::Deserialize)]
+pub struct SignRespondTx {
+    pub id: SignRespondTxId,
+    pub sender: Pubkey,
+    pub transaction_data: Vec<u8>,
+    pub slip44_chain_id: u32,
```
volovyks, line 52: We should use CAIP standards, not slip 44. Is it a dependency from FE fakenet?

chain-signatures/node/src/sign_respond_tx.rs
```diff
@@ -0,0 +1,608 @@
...
+    pub key_version: u32,
+    pub deposit: u64,
+    pub path: String,
+    pub algo: String,
+    pub dest: String,
+    pub params: String,
+    pub explorer_deserialization_format: u8,
```
volovyks, line 59: What is explorer in this context?

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -314,28 +320,10 @@ fn sign_request_from_filtered_log(
         unix_timestamp_indexed: crate::util::current_unix_timestamp(),
         timestamp_sign_queue: None,
         total_timeout,
+        sign_request_type: SignRequestType::Sign,
+        participants: None,
```
volovyks, line 324: Do we still need participants in `IndexedSignRequest`?

chain-signatures/node/src/sign_respond_tx.rs
ppca: i followed the naming from fake mpc. @esaminu might know more detail about it, i'm happy to change it

chain-signatures/node/src/indexer_eth.rs
ppca: yeah, we still do. this is for designating the participants for read respond, since only participants that generated the signature for sign_respond() can participate in signature generation for read respond.

chain-signatures/node/src/sign_respond_tx.rs
ppca: yeah this is what fakenet uses.

chain-signatures/node/src/indexer_sol.rs
ppca: It's used for SignRequestedEvent and SignRespondRequestedEvent separately

chain-signatures/node/src/indexer_eth.rs
ppca:
```
impl TransactionReceipt {
    /// Returns the status of the transaction.
    pub const fn status(&self) -> bool {
        match &self.inner {
            ReceiptEnvelope::Eip1559(receipt)
            | ReceiptEnvelope::Eip2930(receipt)
            | ReceiptEnvelope::Eip4844(receipt)
            | ReceiptEnvelope::Eip7702(receipt)
            | ReceiptEnvelope::Legacy(receipt) => receipt.receipt.status.coerce_status(),
        }
    }
```
this is whether top level execution has succeeded: https://eips.ethereum.org/EIPS/eip-658?utm_source=chatgpt.com

ChaoticTempest:
for the most part, it looks good to me, but let's wait on others to see if they notice anything else

chain-signatures/node/src/indexer_eth.rs
volovyks: This is not a blocker, let's continue designing this here: https://github.com/sig-net/mpc/issues/514

chain-signatures/node/src/sign_respond_tx.rs
volovyks: Not a blocker now, let's fix this after the merge https://github.com/sig-net/mpc/issues/515

chain-signatures/contract-sol/src/lib.rs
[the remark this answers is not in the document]
volovyks: let's continue discussion here: https://github.com/sig-net/mpc/issues/516

chain-signatures/node/src/indexer_sol.rs
```diff
@@ -159,80 +173,180 @@ pub struct SignatureRequestedEvent {
...
-    }
+    fn generate_sign_request(
+        &self,
+        tx_sig: Vec<u8>,
+        total_timeout: Duration,
+    ) -> anyhow::Result<IndexedSignRequest> {
+        tracing::info!("found solana event: {:?}", self);
+        if self.deposit == 0 {
+            tracing::warn!("deposit is 0, skipping sign request");
+            return Err(anyhow::anyhow!("deposit is 0"));
```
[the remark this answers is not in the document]
volovyks: @ppca please, address this one

chain-signatures/node/src/indexer_eth.rs
volovyks: https://github.com/sig-net/mpc/issues/517

chain-signatures/node/src/sign_respond_tx.rs
volovyks: @esaminu please, take a look at this question even if we merge this.

volovyks (dismissed):
I've resolved most of our conversations (feel free to reopen), and created a couple of issues for things we should fix or look into in the future. Let's merge this one and iterate.

ChaoticTempest (dismissed):

ChaoticTempest (approved):

PR #476, opened 2025-08-05
continue if no new block head received

ppca:


ChaoticTempest:
why are we having `interval.tick()` at all? Do you just want to log that we haven't received anything for after a certain period of time?

If so, we can just do it in a tokio::select so that we don't need to poll as fast just to check if a block is ready or not since our warning is every 60s. We can probably just do 20s, and leave it up to the runtime to wake the recv when a new block is ready.
```rs
let mut interval = tokio::time::interval(Duration::from_secs(20));

tokio::select! {
    _ = interval.tick() => {
         if update.elapsed() > Duration::from_secs(60) { ... }
    }
    block_heads.recv() => {
        ...
    }
}

```

volovyks:
Closing: targets the pre-rewrite Ethereum indexer, which has been removed.

PR #478, opened 2025-08-12
fix: expire posits and remove introduction count

ChaoticTempest:
This adds expiration to posits. This makes it so that posits can be expired. Currently an interval is used to poll over if any has expired. There's definitely a smarter way of doing this with a queue of expirations since we have an ordering of when timeouts are inserted by default anyways, but the number of posits is relatively low for now so can be something to be done in the future if its an issue.

Only necessary for triples and presignatures for now since they use the introduced count for stockpiling.


volovyks:
@ChaoticTempest, overall looks good, but the build process fails.
@ppca, hopefully this PR will fix our dev, but even with a positive outcome, we will need to give it a good test (24+ hours + load testing).

volovyks:
@ChaoticTempest, I will merge it myself so we can check if it fixes dev and prepare for the Testnet release.

PR #479, opened 2025-08-12
CAIP2 standard

volovyks:
- Added CAIP2 as our standard for network identification (potentially a breaking change)
- Removed our Roadmap from the repo; it is available on our website.
- Min updates


jakmeier (dismissed):
Nice!

What are the breaking changes you had in mind?
Does this require changes to existing smart contracts (MPC contracts and/or user contracts)

volovyks:
@jakmeier `chain_id` is a part of our derivation path. You can check how it works [here](https://github.com/sig-net/mpc/blob/d52cb03a73d3337f51ab692a51308762b6dc4911/chain-signatures/crypto/src/kdf.rs#L34).

Using CAIP 2 for Ethereum would be a breaking change.

Also, we can use [CAIP10](https://github.com/ChainAgnostic/CAIPs/blob/main/CAIPs/caip-10.md). Again, a breaking change. In practice, it would mean going from `{EPSILON_DERIVATION_PREFIX},{CHAIN_ID_ETHEREUM},{requester}` to `{EPSILON_DERIVATION_PREFIX}:{CHAIN_ID_ETHEREUM}:{requester}`.

Note, our actual derivation path is `{EPSILON_DERIVATION_PREFIX},{CHAIN_ID},{requester},{path}`, where `EPSILON_DERIVATION_PREFIX` is `sig.network v1.0.0 epsilon derivation` and path is what the user passed in their call. I do not think that there is a standard that includes all these parameters.

PR #480, opened 2025-08-12
Simplify solana contract stub

volovyks:
There is no way to remove the contract, but I've managed to remove the additional workspace and 70% of the setup. Now it's a simple Rust project with a single dependency.


ChaoticTempest:
how are we building this contract? I don't see any instructions to do this? I'm trying anchor build inside `contract-sol` but getting:
```
thread 'main' panicked at cli/src/lib.rs:1367:47:
Not in workspace.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

ChaoticTempest (approved):

PR #486, opened 2025-08-21
feat: added signature proposer election by rounds

ChaoticTempest:
This adds the signature proposer election based on rounds by @jakmeier previously with https://github.com/sig-net/mpc/pull/410.

This is a bit different since the rounds are stored in the `SignRequest` struct. So the newly organized request will start from there on reorganization of proposer.

### DIFF
--- a/chain-signatures/node/src/protocol/signature.rs
+++ b/chain-signatures/node/src/protocol/signature.rs
@@ -95,7 +95,6 @@ impl PendingRequest {
 pub struct SignRequest {
     pub indexed: IndexedSignRequest,
     pub proposer: Participant,
-    pub participants: Vec<Participant>,
     pub stable: BTreeSet<Participant>,
     pub round: usize,
 }
@@ -160,7 +159,6 @@ impl SignQueue {
 
     fn organize_request(
         &self,
-        threshold: usize,
         stable: &BTreeSet<Participant>,
         participants: &Participants,
         indexed: IndexedSignRequest,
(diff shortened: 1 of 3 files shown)


chain-signatures/node/src/protocol/signature.rs
```diff
@@ -160,16 +162,60 @@ impl SignQueue {
...
+
         // NOTE: reorganize, will use the same entropy for reorganizing the participants. The only
         // thing that would be different is the passed in stable participants.
         let mut rng = StdRng::from_seed(indexed.args.entropy);
-        let subset = stable.iter().copied().choose_multiple(&mut rng, threshold);
+        let mut subset = stable.iter().copied().choose_multiple(&mut rng, threshold);
+        if !subset.contains(&proposer) {
+            // replace the last randomly chosen participant with the proposer
+            subset.pop();
+            subset.push(proposer);
```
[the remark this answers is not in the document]
ChaoticTempest: subset is at least threshold amount, removing 1 and adding 1 will still be at threshold amount

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -160,16 +162,60 @@ impl SignQueue {
...
+        }
+        subset.sort();
```
jakmeier, line 216: If we still select a random subset from `stable` on every node, this is going to be very unstable. Nodes will disagree on the subset quite often if even just a single peer has intermittent disconnects.

This should only run on the proposer. Other nodes must use the set provided by the proposer.

jakmeier (commented):
Thanks for picking this up!

Preserving the round for the reorg is a great addition!

Can we also increase the round by at leats one on a timeout? That way, we won't try with the same proposer again, even if we think it is stable.

chain-signatures/node/src/protocol/signature.rs
ChaoticTempest: sounds good, I will get rid of it. Will be using posit participants intersected with only presignature participants in this case then

PR #489, opened 2025-08-22
feat: make /msg and /sync use cbor instead of json

ChaoticTempest:
This makes it so that both /msg and /sync use cbor serialization instead of json. This should improve the efficiency of sending messages a fair bit.

#### Changes
- updates axum to 0.7.9 due to conflicts with helios
- adds Cbor axum type which is needed for processing cbor serialization


chain-signatures/node/src/web/cbor.rs
```diff
@@ -0,0 +1,120 @@
...
+use axum::body::Bytes;
+use axum::extract::{FromRequest, Request};
+use axum::http::StatusCode;
+use axum::response::{IntoResponse, Response};
+use http::{header, HeaderMap};
+use serde::{de::DeserializeOwned, Serialize};
+
+use crate::protocol::message::cbor_to_bytes;
+
+pub struct Cbor<T>(pub T);
```
volovyks, line 11: We may change or extend our messaging protocol. Should we add an int version to it now so we can handle it easier in the future?

ChaoticTempest: we shouldn't, as versioning should come with the serialization standard. cbor is like borsh where we have to handle it ourselves, but cbor does make it easier for future proofing since we can add in fields without making previous versions invalid. Adding a version number won't do anything in that case. Neither will it add much value for changing standards since we'd just reject on endpoint level anyways if it disagrees

PR #497, opened 2025-08-28
Key versioning support, CAIP-2 Chain Id in key derivation path

### LINKED ISSUE #515: Use CAIP standard in Bidyrectional flow

volovyks:
- Now `key_version` is used and affects the derivation path. The latest key version is now 1. @esaminu we need to mention that in our docs and make changes in signet.js. Since `key_version=0` is still supported, we can do that at our own pace.
- Since we are using `key_version`, the update in the derivation path is not a breaking change.
- Now, with `key_version=1`, we are using CAIP-2 standard as specified in our docs.
- I've simplified KDF logic, removed code duplication
- Keccak256 is used as a hashing algo for all chains. That is breaking change for NEAR, since it was using Sha3, but it is only used in tests.


doc/ACCOUNT_DERIVATION.md
```diff
@@ -13,11 +13,10 @@ At Sig.Network we use the [CAIP-2](https://chainagnostic.org/CAIPs/caip-2) stand
 
 The following examples show how different blockchain networks are identified using CAIP-2:
 - Solana Mainnet: `solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp`
-- Solana Devnet: `solana:EtWTRABZaYq6iMfeYKouRu166VU2xqa1`
 - Ethereum Mainnet: `eip155:1`
-- Ethereum Sepolia: `eip155:11155111`
 - Bitcoin Mainnet: `bip122:000000000019d6689c085ae165831e93`
-- Bitcoin Testnet: `bip122:000000000933ea01ad0ee984209779ba`
+
+For simplicity, we reuse Mainnet derivation paths for all dev and testnet networks. For example, Ethereum Sepolia will have the same derivation path as Ethereum Mainnet.
```
volovyks, line 19: Adding an Env variable appeared to be troublesome for both MPC, Tooling, as as a result - users. Also, it is kind of a duplication, since we already define our environment by the root MPC key.

volovyks:
Merge after https://github.com/sig-net/mpc/pull/509 where I added additional tests.

chain-signatures/contract/src/lib.rs
```diff
@@ -220,9 +221,8 @@ impl VersionedMpcContract {
     /// Key versions refer new versions of the root key that we may choose to generate on cohort changes
     /// Older key versions will always work but newer key versions were never held by older signers
     /// Newer key versions may also add new security features, like only existing within a secure enclave
-    /// Currently only 0 is a valid key version
     pub const fn latest_key_version(&self) -> u32 {
-        0
+        1
```
volovyks: @ppca we will need to update contract version in all env as a part of release process.

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -272,7 +272,7 @@ fn sign_request_from_filtered_log(
         return Err(anyhow::anyhow!("deposit is 0"));
     }
 
-    if event.key_version != 0 {
+    if event.key_version > LATEST_MPC_KEY_VERSION {
```
volovyks, line 283: @ChaoticTempest @jakmeier not related to this PR, but related to guaranteed responses. What is our strategy on invalid requests? We can try processing them anyway. In this case - with latest key version.

chain-signatures/crypto/src/kdf.rs
```diff
@@ -10,44 +10,84 @@ use sha3::{Digest, Keccak256, Sha3_256};
...
+            Chain::Near => "near:mainnet",
+            Chain::Ethereum => "eip155:1",
+            Chain::Solana => "solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp",
+            Chain::Bitcoin => "bip122:000000000019d6689c085ae165831e93",
+        }
+    }
+}
+
+/// Creates a derivation path string using the legacy format
+fn depricated_derivation_path(chain: Chain, sender: &str, path: &str) -> String {
```
ChaoticTempest: nit: typo
```suggestion
fn deprecated_derivation_path(chain: Chain, sender: &str, path: &str) -> String {
```

chain-signatures/crypto/src/kdf.rs
```diff
@@ -10,44 +10,84 @@ use sha3::{Digest, Keccak256, Sha3_256};
...
     hasher.update(derivation_path);
     let hash: [u8; 32] = hasher.finalize().into();
     Scalar::from_non_biased(hash)
 }
 
-const CHAIN_ID_ETHEREUM: &str = "0x1";
-pub fn derive_epsilon_eth(requester: String, path: &str) -> Scalar {
-    let derivation_path =
-        format!("{EPSILON_DERIVATION_PREFIX},{CHAIN_ID_ETHEREUM},{requester},{path}");
+fn hash_derivation_path_keccak(derivation_path: impl AsRef<[u8]>) -> Scalar {
```
ChaoticTempest: nit: we don't need to prefix the function names with hash in the name since keccak and sha3 already hints it

chain-signatures/node/src/indexer_eth.rs
ChaoticTempest: it would be a bit complex to handle invalid requests with older key versions, because then we'd have to keep a mapping of key versions. Even though we're increasing the key version for derivation path, this isn't entirely breaking anything in our MPC because we can just map key version 0 to the previous derivation path. I think we might need to do that if we're bumping the key version to 1 here.

We also need to make sure that everyone is using key version 1 in our client libraries. Or do we just keep it to 0 for now and not break the client libraries

volovyks: Yes, since there is no breaking change for clients, they can upgrade signet.js on their own pace.

In my question, I'm talking about guaranteed responses. Now, if the key version is not supported, we are dropping the request.

chain-signatures/crypto/src/kdf.rs
volovyks: Agree, but in this case we need to rename it to `keccak_derivation_path` and `sha3_derivation_path`, since we are performing an operation on derivation path. In this case it is less clear and functions start with different words, not great for search and auto-completion.

chain-signatures/node/src/indexer_eth.rs
volovyks: Since we now support both 0 and 1, my question is mostly about key versions > 1. Users still can pass them.

chain-signatures/crypto/src/kdf.rs
ChaoticTempest: why can't we just have `derivation_path_keccak` and `derivation_path_sha3`?

chain-signatures/node/src/indexer_eth.rs
ChaoticTempest: Yeah, we just don't return anything. Ideally the contract should match against key versions to see if they're supported or not. If they're not, just throw an error there so the system does not need to index them. It would be a silent error if the contract doesn't do anything

chain-signatures/crypto/src/kdf.rs
volovyks: `hash_derivation_path` means give me a hash of the derivation path
what `derivation_path_sha3` means?

If you want to drop everything redundant, we can call it `sha3()` and `keccak()`

ChaoticTempest: > hash_derivation_path means give me a hash of the derivation path
what derivation_path_sha3 means?

The suffix would signify different flavors the derivation path. It's used quite often this way.

> If you want to drop everything redundant, we can call it sha3() and keccak()

this would then make it sound like purely hash functions, but these function aren't even derivation path related, they're just doing hashing, so it's fine to just call sha3 and keccak

ChaoticTempest (approved):

volovyks:
`test_mesh_update` us failing again

PR #500, opened 2025-08-28
add regression test for presignature timeout

jakmeier:
Adds a test that would fail without the fix in #485

### DIFF
--- a/integration-tests/tests/cases/mpc.rs
+++ b/integration-tests/tests/cases/mpc.rs
@@ -1,4 +1,5 @@
 use deadpool_redis::redis::AsyncCommands;
+use integration_tests::mpc_fixture::fixture_tasks::MessageFilter;
 use integration_tests::mpc_fixture::MpcFixtureBuilder;
 use mpc_node::protocol::presignature::Presignature;
 use mpc_node::protocol::triple::Triple;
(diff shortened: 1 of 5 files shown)


integration-tests/tests/cases/mpc.rs
```diff
@@ -233,3 +234,44 @@ fn sign_arg(seed: u8) -> SignArgs {
...
+        // apply message filter to all nodes
+        .with_outgoing_message_filter(0, create_filter())
+        .with_outgoing_message_filter(1, create_filter())
+        .with_outgoing_message_filter(2, create_filter())
+        // speed up timeout
+        .with_presignature_timeout_ms(2000)
+        .build()
+        .await;
+
+    tokio::time::timeout(Duration::from_secs(10), network.wait_for_presignatures(1))
```
volovyks, line 274: To be sure I understand this test correctly.
- We start 3 nodes
- They try to generate 5 per-signatures
- Some or all (not clear) of this protocols fail, since we drop 20 per-signature messages on each node
- Nodes start new protocols, at least one of them succeeds.

It means that old protocols are not waiting for new messages forever and the counter of ongoing generators represents protocols that are actually alive.

The requirement for such test is: The number of failed protocols must be >= number of concurrent pre-signature generations. This is not guaranteed by the test itself, but it is very nice to have anyways.

volovyks (approved):
Nice!

PR #507, opened 2025-09-05
chore: use at least 4 worker threads if core count is less than 4

ChaoticTempest:
tokio::main by default takes the number of cores as the amount of worker threads. This might cause some deficiencies in our dev env where we have approx 1 core per node.

### DIFF
--- a/chain-signatures/node/src/main.rs
+++ b/chain-signatures/node/src/main.rs
@@ -1,7 +1,18 @@
 use clap::Parser;
 use mpc_node::cli::Cli;
 
-#[tokio::main]
-async fn main() -> anyhow::Result<()> {
-    mpc_node::cli::run(Cli::parse()).await
+fn main() -> anyhow::Result<()> {
+    let num_cpus = std::thread::available_parallelism()
+        .map(|n| n.get())
+        .unwrap_or(1);
+
+    // Ensure at least 4 worker threads if CPU cores < 4
+    let worker_threads = std::cmp::max(num_cpus, 4);
+
+    let rt = tokio::runtime::Builder::new_multi_thread()
+        .worker_threads(worker_threads)
+        .enable_all()
+        .build()?;
+
+    rt.block_on(mpc_node::cli::run(Cli::parse()))
 }


ppca (approved):

jakmeier:
@ChaoticTempest I don't understand the exact motivation behind this change.

What deficiencies are you worried about with a single worker thread? Did you observer deadlocks or something? Or are we blocking the thread too much, e.g. in triple computations or IO?

I'm trying to understand if we have a bigger underlying issue, which might be better to address directly, if we understand the cause.

ChaoticTempest:
@jakmeier I believe we're blocking the thread too much, specifically our logs show that our connection pool is timing out when calling `/state` and this leads to having participants being offline but maybe they're just slow. So generation could be blocking our axum server from serving requests. `/state` could also be taking too long with redis since it queries redis a couple times. We can move over to a `/status` to serve merely as a health ping or get rid of `/state` and only call `/msg (empty)` but `/msg (empty)` also fails from time to time based on logs.

After this PR merged, our dev net has been more stable, but I also added https://github.com/sig-net/mpc/pull/508 which increased timeouts. This didn't work as well for some reason so had to revert for now

jakmeier:
Ok, makes sense. Ideally we should identify where we block for too long and move it off the async executor. Ideally, it's just a few calls to [spawn_blocking](https://docs.rs/tokio/latest/tokio/task/fn.spawn_blocking.html).

Having multiple worker threads only makes it less likely that we have problems.

ChaoticTempest:
do you know any good tools to check what tasks take too long?

jakmeier:
I would give tokio-console a go: https://github.com/tokio-rs/console

I haven't use it in a while and I remember it can be a bit tricky to set up, I had some problems with `tokio_unstable`. But if it works, it's great!

Otherwise, any profiling with e.g. `perf` might also show where we spend most of our CPU time, though that is less suited to the task than tokio-console.

Oh and I'm not saying this is high priority and we should work on this immediately. But if you don't want to look into it now, I would suggest to create an issue. Even if just to have something to go back and read up on what we already tried, next time we run into a problem with (suspected) bad async scheduling.

PR #509, opened 2025-09-08
Add KDF epsilon tests

volovyks:
Added tests ensuring the KDF derivation logic produces consistent results.


volovyks:
@ChaoticTempest this tests failed in this PR without good reason, is that common?
`mesh::tests::test_mesh_contract_update`
`mesh::tests::test_mesh_update`
`mesh::tests::test_pool_update`

ChaoticTempest:
hmm, they may be a bit non determnistic especially since we're running multiple tests in parallel, I'll check about increasing their timeouts later

ChaoticTempest (approved):

jakmeier (commented):
Important addition! Thanks @volovyks

PR #512, opened 2025-09-09
sol indexer: also index non cpi sign event

ppca:
This is just including current production sol code in another tokio task. This will make sure we also are listening for those events


volovyks (approved):

volovyks:
@ppca thank you!

PR #513, opened 2025-09-09
feat: integration tests poll requests by RPC instead of indexer

ChaoticTempest:
This makes it so that requests from near contract come in by RPC instead of indexer. `indexer.rs` now just polls the near contract for pending requests.

## Changes
- removes near lake indexer and lake node; uses `near_workspaces::sandbox` directly now.
- removes localstack since only lake was using it.
- removes AWS dependencies since only localstack and lake was using it.
- removes toxiproxy since it was very intertwined with lake and basically was a requirement.


chain-signatures/contract/src/lib.rs
```diff
@@ -680,6 +680,12 @@ impl VersionedMpcContract {
         }
     }
 
+    pub fn pending_requests_data(&self) -> Vec<(&SignId, &PendingRequest)> {
+        match self {
+            Self::V0(mpc_contract) => mpc_contract.pending_requests.iter().collect(),
```
volovyks, line 685: Not a huge issue, since it is only our testing infrastructure, but what is the limit here?

chain-signatures/contract/src/lib.rs
```diff
@@ -680,6 +680,12 @@ impl VersionedMpcContract {
         }
     }
 
+    pub fn pending_requests_data(&self) -> Vec<(&SignId, &PendingRequest)> {
```
volovyks, line 683: nit, why not 'pending_requests'

chain-signatures/node/src/indexer.rs
```diff
@@ -1,279 +1,212 @@
...
-use std::thread::JoinHandle;
 use std::time::{Duration, Instant};
 use tokio::sync::{mpsc, RwLock};
+use tokio::task::JoinHandle;
 
 /// Configures indexer.
 #[derive(Debug, Clone, clap::Parser)]
 #[group(id = "indexer_options")]
 pub struct Options {
-    /// AWS S3 bucket name for NEAR Lake Indexer
```
volovyks, line 24: @auto-mausx multiple parameters become redundant, please, represent that in our docs.

chain-signatures/node/src/indexer.rs
```diff
@@ -1,279 +1,212 @@
...
+            },
+            chain: Chain::NEAR,
+            unix_timestamp_indexed: crate::util::current_unix_timestamp(),
+            timestamp_sign_queue: Some(Instant::now()),
+            total_timeout: Duration::from_secs(200),
+        }
+    }
+
+    /// Derive entropy deterministically from sign_id
+    fn derive_entropy_from_sign_id(&self, sign_id: &SignId) -> [u8; 32] {
```
volovyks, line 131: I understand that this approach will work for us, but entropy should not be derived from SignId. Why not add it to the request on the contract level?

volovyks (commented):
Happy to remove all these dependencies and related code!
closing https://github.com/sig-net/mpc/pull/447 in favor of this PR

chain-signatures/contract/src/lib.rs
ChaoticTempest: hmm, not sure by what you mean by limit here. There's no caveats besides serialization ones

chain-signatures/contract/src/lib.rs
ChaoticTempest: because `pending_requests` already exists, and I don't want to break anything

chain-signatures/node/src/indexer.rs
ChaoticTempest: same with the last one, it would require changing the `PendingRequest` struct which ends up breaking things if we were to deploy the contract to testnet/mainnet. This is just for testing anyways, so it doesn't matter if we derive the entropy from sign_id. All nodes derive it the same way, and hashing it gives us a uniform entropy anyways so there's no collision

ChaoticTempest: I also don't want to break things just in case we add more things to gov contract for bidirectional communication like checkpoints

chain-signatures/contract/src/lib.rs
volovyks: I wander how many requests can we store and return. But if it's a 1000+ it should be fine. Running a test once would be great.

ChaoticTempest: ahh, the limit is configurable via sandbox so we don't have to worry about it

chain-signatures/node/src/indexer.rs
volovyks: Ok, please, add a disclaimer that this is not a production ready function.

ChaoticTempest: it's already added in the notes where it's used

ppca (approved):

jakmeier:
This is easy to miss but actually really awesome to get cleaned up! It really helps with compilation time as well as the time to run tests.

The time for running integrations tests in CI went down from 41-42 minutes to just 31-33 minutes.

These are the timing for the 3 time consuming steps:


| | [before](https://github.com/sig-net/mpc/actions/runs/17595272182/job/49985926565) | [after](https://github.com/sig-net/mpc/actions/runs/17596506010/job/49989849080) |
| -- | -- | -- |
| Build Chain-Signatures Node | 6m 33s | 4m 50s |
| Build Chain-Signatures Integration Tests | 4m 9s  | 2m 52s |
|  Test | 30m 21s  | 22m 29s |


Thanks @ChaoticTempest for getting this in!

PR #525, opened 2025-09-16
feat: solana sandbox for integration tests

ChaoticTempest:
This adds solana sandbox to integration tests. So far this PR only adds one test with it where we can do a request for a signature from the sandbox.

To run the test, you need the full suite of solana cli tools, including `solana` and `solana-test-validator`. Install from here: https://solana.com/docs/intro/installation. 

The contract is prebuilt from https://github.com/sig-net/solana-signet-program on 0cdfe4069216490cac3a3a4ce508b358503d81d6 with `FR5pWwinRBn35GNhg7bsvw8Q13kRept2pm561DwZCQzT` being the hardcoded program id.


integration-tests/src/cluster/spawner.rs
```diff
@@ -116,6 +120,28 @@ impl ClusterSpawner {
         self
     }
 
+    /// Configures the cluster to spawn with Solana sandbox.
+    /// This method sets up a Solana test validator and configures the SolConfig.
+    pub fn sol(mut self) -> Self {
```
jakmeier: nit: can we try to stick to the full word "solana" rather than "sol"? The full word is only 6 characters and unfortunately sol is also short for solidity.

`SolConfig` for example is ambiguous, and so it `contract-sol`.

integration-tests/tests/cases/mpc.rs
```diff
@@ -18,7 +18,7 @@ const KEY_SHARE_FILE: &str = "tmp/key_shares.json";
 const TRIPLES_FILE: &str = "tmp/triples.json";
 const PRESIGNATURES_FILE: &str = "tmp/presignatures.json";
 
-#[tokio::test(flavor = "multi_thread")]
+#[test_log::test(tokio::test(flavor = "multi_thread"))]
```
jakmeier: Change makes sense! Ideally, for next time, it would be a separate PR. Mixing multiple unrelated changes puts more mental load on reviewers than two separate PRs.

integration-tests/tests/cases/chains.rs
```diff
@@ -0,0 +1,13 @@
...
+async fn test_solana_signature_basic() {
+    let cluster = cluster::spawn().sol().await.unwrap();
+    let _outcome = cluster
+        .sign()
+        .sol()
+        .payload([42u8; 32])
+        .path("test/integration/path")
+        .await
+        .unwrap();
+}
```
jakmeier: By moving all the logic, including assertions, inside the setup, it can be hard to see what the test actually does. Please correct me if I'm getting this wrong.

From my code reading, I understand this submits a sign request to a locally emulated Solana chain and then waits for the response to arrive within 90s. If it finds a matching event, it parses the signature and returns it.

Could we add a check that the signature matches the original request payload and path?

jakmeier (dismissed):
Great progress in making this setup testable! I love to see a setup that self-contained without dependencies on live networks.

@ChaoticTempest what do you think, how maintainable is this approach long-term? (with dependencies on specific tools) How flexible is it to write very specific tests?

Could you see this as the primary way to write integration tests, for anything involving mpc + indexers? (of course, for indexers on other chains, it would rely on the existance and maintenance of other tools)

integration-tests/Cargo.toml
```diff
@@ -28,8 +30,15 @@ tracing.workspace = true
 tracing-subscriber.workspace = true
 thiserror.workspace = true
 url.workspace = true
+uuid = { version = "1.0", features = ["v4"] }
```
volovyks, line 33: Is our goal to move all dependencies versions to the workspace level? If so, let's follow this standard, I can move whatever is left.

chain-signatures/node/src/indexer_sol.rs
```diff
@@ -169,7 +169,7 @@ pub struct SignatureRequestedEvent {
     pub payload: [u8; 32],
     pub key_version: u32,
     pub deposit: u64,
-    pub chain_id: u64,
+    pub chain_id: String,
```
volovyks, line 183: I guess you are using the latest contract from develop, but this is not what @ppca used when adding Bidirectional.
Not a blocker, but we need to understand on what version we are.

volovyks (commented):
Please, add version to Signet Solana program. We will need to migrate it time after time and know what version is currently used.

integration-tests/tests/cases/mpc.rs
ChaoticTempest: noted, but this one was necessary since it collides with other tests once I added the `chains` module which does a global subscriber first and then `mpc` tries to create one itself which fails

integration-tests/Cargo.toml
ChaoticTempest: not necessarily, I think it's better to keep whatever dependency is used for one package just in its own `Cargo.toml` until it's needed by multiple. This way cargo can potentially not compile the whole set of crates in case we only want to build one of our mpc packages

chain-signatures/node/src/indexer_sol.rs
ChaoticTempest: I listed the commit hash in the description of the PR, so that is the version we need to deploy on dev

volovyks: We need to have it in code, for example, in the name of the contract file.

ChaoticTempest: done

volovyks (dismissed):

jakmeier (dismissed):

ChaoticTempest:
This now uses 0.4.0 from our solana signet program. We will need to have 0.4.0 of program deployed to dev before this one can be merged

ppca (approved):

PR #526, opened 2025-09-18
fix: flaky mesh unit tests

ChaoticTempest:
Sometimes the mesh tests might fail, so it was pretty flaky. This adds in what is essentially retrying to them with a max timeout of 1s to reach the end state.


volovyks (approved):

jakmeier:
@ChaoticTempest  do you understand why the test fails, when it does?

ChaoticTempest:
the mockito server might take a bit to startup so that initial PING_INTERVAL that we're sleeping for might not be enough so it's a bit flaky depending if our CI has enough compute to pull through

PR #527, opened 2025-09-18
feat: eth sepolia testnet for integration tests

ChaoticTempest:
This adds in integration testing with ethereum via sepolia testnet.

One change I had to make on the node side was to include `optimistic_requests: bool` such that the eth indexer doesn't wait for finality so tests can pass faster. This option is only enabled for tests and is by default false.


jakmeier (dismissed):
LGTM!

@ChaoticTempest  One question: Would it make sense to move away from Sepolia and instead run a local setup, much like you did for the Solana testing? It seems the Ethereum ecosystem has broad support for such setups.

ChaoticTempest:
@jakmeier I tried a local setup but none supported the light client support that helios requires

jakmeier:
> @jakmeier I tried a local setup but none supported the light client support that helios requires

Ah, I see. Makes sense then, thanks!

ChaoticTempest:
integration tests failing for the logging issue that was fixed in https://github.com/sig-net/mpc/pull/525. So will wait until that one is pushed in then this one should work. Will need to fix merge conflicts anyways for it

jakmeier:
> integration tests failing for the logging issue that was fixed in #525. So will wait until that one is pushed in then this one should work. Will need to fix merge conflicts anyways for it

(another reason why splitting it to its own PR can be beneficial :P)

jakmeier (approved):

PR #530, opened 2025-09-18
chore: remove web3 dependency

jakmeier:
We only used it for re-exported functionality and in tests
for recovering the ETH address.

In the spirit of keeping our dependency tree clean, let's remove it.

Also gets rid of these indirect dependencies:

- headers
- headers-core
- idna (0.4.0)
- web3-async-native-tls


volovyks (approved):
Nice! Thank you for paying attention to this.

PR #531, opened 2025-09-19
Use rust-toolchain config

volovyks:
- Enforce usage of the same Rust version in the project. Now it is 1.86, since it is the latest one supported by NEAR SDK.
- Simplify setup
- Ensure consistency of builds and developer experience


ChaoticTempest:
if the latest nearcore supports 1.86 wasm compiled contracts, then you also need to update our workspaces

volovyks:
@ChaoticTempest please review: https://github.com/ChaoticTempest/near-fetch/pull/30
(and release a new version)

volovyks:
Hm, tests are not starting.
```
Requested labels: warp-ubuntu-latest-x64-4x
Job defined at: sig-net/mpc/.github/workflows/integration.yml@refs/pull/531/merge
Waiting for a runner to pick up this job...
```
@auto-mausx do you know why?

volovyks:
@auto-mausx this is not an issue with this PR. Other jobs are also not starting.
Probably we are hitting limits in warp, but I was not able to check it.
Also, warp is not in the list of our runners.

auto-mausx:
@volovyks Looking at the account, I don't really see an issue from that side, I do notice there's no runners from warp in our runner group. Last successful warp build run was on the 19th PR #533. Going to continue to look into this, and restart the job again and see what happens. I remember us running into this issue before but I think Bo ended up fixing it.

auto-mausx:
I'm tempted to clear the cache on warpbuild and try again, but I don't know if that's going to actually do anything since the runner doesn't even pick up the job 😞

volovyks:
@auto-mausx there is no harm in clearing the cache.

rust-toolchain.toml
```diff
@@ -0,0 +1,4 @@
+[toolchain]
+channel = "1.86.0" # latest version supported by near-sdk-rs
```
ChaoticTempest, line 2: let's not force the whole project to be this version because then we won't be able to keep up with new changes to rust. We already had a way before to just compile the contract with a specific rust version, so let's just do that instead

volovyks: Using a fixed Rust version is a good practice.
Pros:
- Consistency of builds between devs
- Simpler setup locally and on CI
- Consistency of local setup and CI
- Less overhead and rebuilds, fewer versions, less cache, etc.

Cons:
- Yes, we are using a 6-month-old Rust version. But there are zero 1.86+ features in our project.
- We will need to update the version once in a while.

Overall, it is a small price to pay for all the benefits.

volovyks:
Looks like I will need a Linux machine (or a VM) to address the build issues.

ppca:
> Looks like I will need a Linux machine (or a VM) to address the build issues.

you can probably try orbstrack. @auto-mausx has more details on it

auto-mausx:
Unpaid shill:

Older versions · OrbStack Docs https://share.google/FOHWkXhGszM94F3bs

I'd install the older version without the bug that @ppca was talking about

ChaoticTempest:
You can also use GH codespaces that gives you 90 hours of usage for free

ppca:
> Unpaid shill:
> 
> Older versions · OrbStack Docs https://share.google/FOHWkXhGszM94F3bs
> 
> I'd install the older version without the bug that @ppca was talking about

yeah 2.x won't work

rust-toolchain.toml
ChaoticTempest: the thing is that there are constantly new clippy fixes each version, which is not ideal to fix all at once when we decide to upgrade later. 1.90 also introduces a new linker which speeds up build times, so we get much better iteration speed which is why I predominantly want newer versions.

> Consistency of builds between devs
> Simpler setup locally and on CI
> Consistency of local setup and CI

Nice to have, but in general, new rust versions are very stable, so I'm not too concerned on these.

> Less overhead and rebuilds, fewer versions, less cache, etc.

This isn't much of an issue. I'm not sure where you're seeing this.

volovyks: > I'm not sure where you're seeing this.

I constantly see the build cache dropped. Same dependencies are getting rebuilt again and again. One of the reasons, besides jumping between brunches, is different versions of Rust, different build configurations, etc. I'm trying to unify that.

ChaoticTempest: on local or where? Because my local setup never does a rebuild. Like the different versions shouldn't be rebuilding the cache unless you're making changes

volovyks: It can happen in some flows, for example
```bash
time cargo build -p mpc-node
time cargo build -p mpc-crypto
time cargo build -p mpc-node
```
is instant. All good.
But
```bash
time cargo build -p mpc-node
time cargo build -p mpc-contract
time cargo build -p mpc-node
```
will make you wait an extra 6 seconds in my case. Not that bad, but not ideal.

When things get bad, that's when you try to run unit tests:
```bash
cargo test --workspace --exclude integration-tests
```

The project is getting built, then some tests are running, then it is getting built again, and then a bit more.
And the worst is that if you run the command again, it will start the build one more time.

Not saying that it is all caused only because of the Rust version. This is also different targets, debug/release builds, maybe something else. But we need to start somewhere.

I want to experiment with [sccache](https://github.com/marketplace/actions/sccache-action), but it can hide many issues, so first, I want to understand and, if possible, fix the most outstanding ones.

ChaoticTempest: > The project is getting built, then some tests are running, then it is getting built again, and then a bit more.
And the worst is that if you run the command again, it will start the build one more time.
> Not saying that it is all caused only because of the Rust version. This is also different targets, debug/release builds, maybe something else. But we need to start somewhere.

that's actually because of the build script for the first one, since we need to build 2 different things after integration-tests are built: node and contract. Why are you building contract directly anyways? We don't make any changes to contract so it doesn't really matter for the most part in our dev experience.

jakmeier: Let me add my opinion in the mix :)

TLDR: Let's fix the version for each package, either all to 1.86.0 or make an exception to the Near contract.

- Keeping a fixed version makes for a unified dev experience, which in theory should avoid dev setup problems and increase overall productivity.
- Fixing all clippy lints at once, at a moment in time we control, is better than having lints break CI at random moments of Rust releases. It happened recently (see #529) and it will happen again.
- Using new Rust features is nice but not necessary. For projects that other parties depend on, I even see it as good practice to not jump on new features too quickly. Supporting more compiler versions is good. And rarely does a new shiny language feature improve the final product. It's almost like a trade-off between "fun" for internal developers (that's us) and boring but effective productivity. Don't underestimate the positive effect of motivation but also don't fool yourself into thinking it's the most productive use of your time to refactor code into using new Rust features. :P
- Unfortunately, WASM contracts on Near must not use a newer version than 1.86.0 because of LLVM features added in newer versions that are not yet supported by the Near WASM runtime. It's still one or two releases away from being fixed. Until then, we can overwrite the setting for the near contract specifically, or we can hold the entire project back. Either way is fine with me. As long as a fixed Rust version is used for each package.
- I don't think build times are super relevant in this discussion. I can see arguments going both ways. Within a single workspace, it might help to fix the rust version. But I regularly jump between different Rust workspaces all on the same machine. The more restrictive each workspace is on the allowed compiler version requirements are, the fewer build artifacts can be shared. So, for hobby projects and the like, I never restrict the compiler version more than necessary.

jakmeier (approved):

volovyks:
I will get back to this if it becomes a real issue

PR #533, opened 2025-09-20
feat: use /status over /state for checking liveness

ChaoticTempest:
This utilizes a new endpoint called `/status` to check for liveness. `/state` makes use of 4 different calls to redis which can be slow if we have too many calls to redis at the same time. This could be the case when we're generating a ton of triples and presignatures and there's many writes to it, with congestion to the redis connection pool.

`/status` does none of these redis calls and purely checks the current protocol state (i.e. `Running`, `Resharing`, `Generating`, ...).

Should wait until https://github.com/sig-net/mpc/pull/532 gets merged first to see an improvement in our web endpoint metrics

### DIFF
--- a/chain-signatures/node/src/web/mock.rs
+++ b/chain-signatures/node/src/web/mock.rs
@@ -3,7 +3,7 @@ use mockito::ServerGuard;
 
 use crate::{
     node_client::NodeClient,
-    protocol::{contract::primitives::Participants, ParticipantInfo},
+    protocol::{contract::primitives::Participants, state::NodeStatus, ParticipantInfo},
 };
 
 use super::StateView;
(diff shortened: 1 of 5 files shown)


chain-signatures/node/src/web/mod.rs
```diff
@@ -184,6 +185,32 @@ async fn state(Extension(web): Extension<Arc<AxumState>>) -> Result<Json<StateVi
...
+    Resharing,
+    WaitingForConsensus,
+    Running,
+    Joining,
+}
+
+#[tracing::instrument(level = "debug", skip_all)]
+async fn status(Extension(web): Extension<Arc<AxumState>>) -> Json<StateStatus> {
+    let status = match web.node.status() {
+        NodeStatus::Started | NodeStatus::Starting => StateStatus::NotRunning,
```
volovyks: Let's reuse the existing NodeStatus enum type. The difference is only in the naming of a single field.

ppca: +1 to this, otherwise lgtm

ppca (dismissed):

chain-signatures/node/src/web/mock.rs
```diff
@@ -36,6 +36,22 @@ impl MockServer {
             .create_async()
             .await;
 
+        server
+            .mock("GET", "/status")
+            .with_status(201)
+            .with_header("content-type", "text/plain")
```
jakmeier: wouldn't it be better to use application/json here?

```suggestion
            .with_header("content-type", "application/json")
```

jakmeier (dismissed):

chain-signatures/node/src/web/mock.rs
ChaoticTempest: I guess mockito doesn't make use of this at all then

volovyks (approved):

ppca (approved):

PR #536, opened 2025-09-24
feat: add redis latency for triple/presignature storage

ChaoticTempest:
This adds in latency metrics for all the redis operations so we can keep track of how long each operations take. This will be useful for when these operations change and we want to monitor how much of a diff they make

### DIFF
--- a/chain-signatures/node/src/metrics.rs
+++ b/chain-signatures/node/src/metrics.rs
@@ -98,6 +98,17 @@ pub(crate) static SIGN_QUEUE_SIZE: LazyLock<IntGaugeVec> = LazyLock::new(|| {
     .unwrap()
 });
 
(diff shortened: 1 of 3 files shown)


auto-mausx (approved):

ChaoticTempest:
@auto-mausx hmm, is our runners authenticated for docker? This is now popping up `Error response from daemon: unauthorized: authentication required`. Maybe has something to do with clearing the cache

auto-mausx:
I think redis moved to close source, I wonder if they still have open source images

auto-mausx:
nevermind, its a dockerhub issue: https://www.dockerstatus.com/

ChaoticTempest:
alright, waiting it is for this PR then

PR #551, opened 2025-10-02
feat: initial checkpoints 

ChaoticTempest:
This is the initial work towards `Checkpoints` and `PendingRequests` and is merely a refactoring right now of what we have right now.

## Changes
- We use to use `SignRespondTxId` as the id for mapping into the `sign_respond_tx_map`, but this changes that to `SignId` since that will be more generic going forward for all chains.
- `SignatureGenerator` takes in `Checkpoints` to set it for pending publish. We're still using the old channels to send the request for publishing, but these channels will likely be moved to checkpoints in the future.
- `RpcExecutor` takes in `Checkpoints` to mark a request as completed (does nothing for now).

### DIFF
--- a/chain-signatures/node/src/backlog.rs
+++ b/chain-signatures/node/src/backlog.rs
@@ -1,0 +1,416 @@
+use crate::protocol::Chain;
+use crate::sign_bidirectional::{BidirectionalTx, PendingRequestStatus};
+use mpc_primitives::SignId;
+use std::collections::{HashMap, VecDeque};
+use std::sync::Arc;
+use tokio::sync::RwLock;
+
+#[derive(Debug, Clone)]
+pub struct PendingRequests {
+    requests: HashMap<SignId, BidirectionalTx>,
+    /// Queue of transaction IDs waiting to be published
+    pending_publish: VecDeque<SignId>,
+    /// The highest block height that has been processed for this chain
+    processed_block_height: Option<u64>,
+}
+
+impl Default for PendingRequests {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl PendingRequests {
+    /// Creates a new empty PendingRequests container
+    pub fn new() -> Self {
+        Self {
+            requests: HashMap::new(),
+            pending_publish: VecDeque::new(),
+            processed_block_height: None,
+        }
+    }
+
+    /// Inserts a sign-respond transaction into the pending requests map
+    /// Returns Some(old_value) if the key was already present
+    fn insert(&mut self, id: SignId, tx: BidirectionalTx) -> Option<BidirectionalTx> {
+        self.requests.insert(id, tx)
+    }
+
+    /// Removes a sign-respond transaction from the pending requests map
+    /// Returns Some(value) if the key was present
+    fn remove(&mut self, id: &SignId) -> Option<BidirectionalTx> {
+        self.requests.remove(id)
+    }
+
+    /// Gets a clone of a sign-respond transaction from the pending requests map
+    /// Returns Some(value) if the key is present
+    fn get(&self, id: &SignId) -> Option<BidirectionalTx> {
+        self.requests.get(id).cloned()
+    }
+
+    /// Returns the number of pending requests
+    fn len(&self) -> usize {
+        self.requests.len()
+    }
+
+    /// Returns all sign-respond transactions with a specific status
+    pub fn get_by_status(&self, status: PendingRequestStatus) -> HashMap<SignId, BidirectionalTx> {
+        self.requests
+            .iter()
+            .filter(|(_, tx)| tx.status == status)
+            .map(|(id, tx)| (*id, tx.clone()))
+            .collect()
+    }
+
+    /// Add a transaction ID to the pending publish queue
+    pub fn push_pending_publish(&mut self, id: SignId) {
+        self.pending_publish.push_back(id);
+    }
+
+    /// Pop the next transaction ID from the pending publish queue
+    /// Returns the transaction if found
+    pub fn pop_pending_publish(&mut self) -> Option<(SignId, BidirectionalTx)> {
+        let id = self.pending_publish.pop_front()?;
+        let tx = self.get(&id)?;
+        Some((id, tx))
+    }
+
+    /// Get the number of transactions waiting to be published
+    fn pending_publish_count(&self) -> usize {
+        self.pending_publish.len()
+    }
+
+    /// Get all pending transaction IDs and their transactions
+    fn pending_execution(&self) -> Vec<(SignId, BidirectionalTx)> {
+        self.pending_publish
+            .iter()
+            .filter_map(|id| self.get(id).map(|tx| (*id, tx)))
+            .collect()
+    }
+
+    /// Get the processed block height for this chain
+    fn processed_block_height(&self) -> Option<u64> {
+        self.processed_block_height
+    }
+
+    /// Set the processed block height for this chain
+    fn set_processed_block(&mut self, height: u64) {
+        self.processed_block_height = Some(height);
+    }
+}
+
+/// Backlog manages pending sign-respond requests across multiple chains.
+/// Each chain has its own isolated set of pending requests with their own
+/// publish queues.
+#[derive(Debug, Clone)]
+pub struct Backlog {
+    requests: Arc<RwLock<HashMap<Chain, PendingRequests>>>,
+}
+
+impl Default for Backlog {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl Backlog {
+    pub fn new() -> Self {
+        Self {
+            requests: Arc::new(RwLock::new(HashMap::new())),
+        }
+    }
+
+    pub async fn insert(
+        &self,
+        chain: Chain,
+        id: SignId,
+        tx: BidirectionalTx,
+    ) -> Option<BidirectionalTx> {
+        self.requests
+            .write()
+            .await
+            .entry(chain)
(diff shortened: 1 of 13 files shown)


chain-signatures/node/src/checkpoint.rs
```diff
@@ -0,0 +1,432 @@
...
+        Self {
+            requests: HashMap::new(),
+            pending_publish: VecDeque::new(),
+            processed_block_height: None,
+        }
+    }
+
+    /// Inserts a sign-respond transaction into the pending requests map
+    /// Returns Some(old_value) if the key was already present
+    pub fn insert(&mut self, id: SignId, tx: SignRespondTx) -> Option<SignRespondTx> {
```
volovyks: Since these functions are never intended to be accessed directly, we can encapsulate them by removing `pub`. Let's have everything public by default.

chain-signatures/node/src/indexer_sol.rs
```diff
@@ -526,6 +525,8 @@ async fn subscribe_and_process_sign_events(
                     .await
                     {
                         tracing::warn!("Failed to process event: {:?}", err);
+                    } else {
+                        checkpoints.set_processed_block(Chain::Solana, slot).await;
```
volovyks: Here, we update the last processed block once we see a single event from that block. I see that the logic is not finalized yet, but let's not lose other events from the same block.

chain-signatures/node/src/checkpoint.rs
```diff
@@ -0,0 +1,432 @@
+use crate::protocol::Chain;
+use crate::sign_respond_tx::{PendingRequestStatus, SignRespondTx};
+use mpc_primitives::SignId;
+use std::collections::{HashMap, VecDeque};
+use std::sync::Arc;
+use tokio::sync::RwLock;
+
+/// A thread-safe container for managing pending sign-respond transactions for a specific chain
```
volovyks: Let's stop using sign-response terminology. `Bidirectional` is how we call it now :)

volovyks: From what I understand, we will use this structure to store both regular and bidirectional requests in the future.

chain-signatures/node/src/indexer.rs
```diff
@@ -179,7 +184,7 @@ async fn poll_pending_requests(ctx: &Context) -> anyhow::Result<()> {
     // Update metrics
     crate::metrics::LATEST_BLOCK_NUMBER
         .with_label_values(&[Chain::NEAR.as_str(), ctx.node_account_id.as_str()])
-        .set(crate::util::current_unix_timestamp() as i64);
```
volovyks, line 182: Wait, why were we using time before?

chain-signatures/node/src/sign_bidirectional.rs
```diff
@@ -38,8 +34,9 @@ struct AbiField {
 }
 
 #[derive(Debug, Clone, PartialEq, Eq, Hash, serde::Serialize, serde::Deserialize)]
-pub enum SignRespondTxStatus {
-    Pending,
+pub enum PendingRequestStatus {
```
volovyks, line 40: I do not understand these statuses.
My understanding:
- PendingExecution
- PendingPublishSignature
- WaitingForOutcome
- PendingPublishOutcome

Maybe the names are not the best, open for discussion.

volovyks (commented):
Looks good!

chain-signatures/node/src/checkpoint.rs
jakmeier: > Let's have everything public by default.

Did you mean *private* by default?

I personally don't care too much about `pub` unless it's published on crates.io but I think private by default is generally the accepted standard approach.

chain-signatures/node/src/checkpoint.rs
jakmeier: What do you mean by thread-safe? I don't see anything in `PendingRequests` that makes it more suitable to use multi-threaded applications.

chain-signatures/node/src/rpc.rs
```diff
@@ -1395,18 +1408,11 @@ async fn try_publish_sol(
...
 
-    if let SignRequestType::SignRespond(_) = action.request.indexed.sign_request_type {
-        sign_respond_signature_channel.send(
-            action.public_key,
-            action.request.clone(),
-            action.output.clone(),
-            action.participants.clone(),
-        );
-    }
+    if let SignRequestType::SignRespond(_) = action.request.indexed.sign_request_type {}
```
jakmeier: Huh, isn't this a no-op? Can it be removed?

chain-signatures/node/src/checkpoint.rs
```diff
@@ -0,0 +1,432 @@
...
+        self.processed_block_height = Some(height);
+    }
+}
+
+/// Checkpoints manages pending sign-respond requests across multiple chains
+/// Each chain has its own isolated set of pending requests with their own publish queues
+#[derive(Debug, Clone)]
+pub struct Checkpoints {
+    requests: Arc<RwLock<HashMap<Chain, PendingRequests>>>,
+}
```
jakmeier: Some bikeshedding on the naming:

Can we reserve the name `Checkpoint` for when we implement the snapshotting and catch-up from a snapshot? Otherwise I'm afraid we are melding different concepts together that are quite different in reality.

For me, one checkpoint is a snapshot of pending requests per chain. It's a frozen data structure created at a given timestamp.

What we use now in my eyes isn't a checkpoint. This is a frontier of pending requests. A live data structure. The thing we are going to take checkpoint off periodically, but not itself a checkpoint.

To make a valid checkpoint, we will need to align timestamps and reach a consistent state for them. Then we need to sign it by the mpc network itself, to make the checkpoint verifiable. In general, the live-view of pending requests would not be a valid checkpoint suitable for recovery.

I struggle to come up with a good name right now, but something in the direction of `SystemWidePendingRequests` is what I'm thinking. Or maybe just `PendingRequests` and rename the other struct to `ChainPendingRequests` or something.

chain-signatures/node/src/backlog.rs
```diff
@@ -0,0 +1,432 @@
...
+    pub fn set_processed_block_height(&mut self, height: u64) {
+        self.processed_block_height = Some(height);
+    }
+}
+
+/// Checkpoints manages pending sign-respond requests across multiple chains
+/// Each chain has its own isolated set of pending requests with their own publish queues
+#[derive(Debug, Clone)]
+pub struct Checkpoints {
+    requests: Arc<RwLock<HashMap<Chain, PendingRequests>>>,
```
jakmeier, line 107: note: To avoid lock contention, we might want to consider `requests: Arc<HashMap<Chain, RwLock<PendingRequests>>>` instead.

But let's avoid premature optimization. Just something to keep in mind, as we are building this to be a central piece used by many concurrent tasks.

jakmeier (dismissed):
Great incremental change!!

Please address the comments by @volovyks and my own but I'm excited to get this merged soon. :)

chain-signatures/node/src/checkpoint.rs
volovyks: Private, sorry :)

ChaoticTempest: I'll just make them not pub then

chain-signatures/node/src/indexer.rs
ChaoticTempest: it didn't really matter before since the metrics were never used

chain-signatures/node/src/sign_bidirectional.rs
ChaoticTempest: I didn't care enough to change the Failed/Success ones. We can keep debating on what to name each of the status as we add them in on utilization

chain-signatures/node/src/checkpoint.rs
ChaoticTempest: yeah I couldn't come up with a good name either. Maybe the latter with `PendingRequests` and `ChainPendingRequests` would be better

chain-signatures/node/src/rpc.rs
ChaoticTempest: whoops, forgot about this one

chain-signatures/node/src/checkpoint.rs
ChaoticTempest: originally had pending requests be the lock, but eventually moved into checkpoints. I'll remove the comment

ChaoticTempest: I'll change the terminology in a different time

chain-signatures/node/src/sign_bidirectional.rs
volovyks: Sure
cc @ppca

ppca: success/failed is needed to handle different logic to create the message we are sending in read_respond depending on whether the eth tx was success or failure. It is not to represent the status of the entire bidirectional request flow.

chain-signatures/node/src/checkpoint.rs
ChaoticTempest: renamed to `Checkpoints` to `Backlog`, but kept `PendingRequests`

jakmeier (dismissed):

volovyks (dismissed):
The usage of `PendingRequestStatus` still has mixed purposes, but we can deal with it in the followup.

ChaoticTempest:
need another approval for this one. It's now working with the bidirectional test in https://github.com/sig-net/mpc/pull/525. We do not need to wait for it to go in. This one can go in now

jakmeier (approved):

PR #554, opened 2025-10-06
mpc bidirectional: api changes

ppca:
align with our api changes:
1. function name change: sign_respond -> sign_bidirectional, read_respond -> respond_bidirectional
2. api field change: slip44ChainId (number) -> caip2Id (string)
3. api field name change: explorer_deserialization_schema -> output_deserialization_schema, callback_serialization_schema -> respond_serialization_schema
4. removing fields: explorer_deserialization_format and callback_serialization_format, as we decide to use designated formats for eth and solana.

I also changed all the naming in mpc.

Will need to test this change with the latest tests in https://github.com/sig-net/solana-contract-examples before merging


chain-signatures/node/src/respond_bidirectional.rs
```diff
@@ -19,23 +19,27 @@ use tokio::sync::RwLock;
 use tokio::time::Duration;
 
 const MAGIC_ERROR_PREFIX: [u8; 4] = [0xde, 0xad, 0xbe, 0xef];
-const SOLANA_READ_RESPOND_PATH: &str = "solana response key";
+const SOLANA_RESPOND_BIDIRECTIONAL_PATH: &str = "solana response key";
+// hardcode to 0 as this is what we are using for solana
+const RESPOND_SERIALIZATION_FORMAT: u8 = 0;
```
volovyks: It is not clear what 0 and 1 mean. I would say we need to have an enum that includes all supported serialization formats that can be represented as integers.

It may be a good idea to specify that X serialization format is the default one for the Y chain. This way it will be more extendable, as we add more chains and allow initiating bidirectional flows on different chains.

ppca: yeah agree having an enum with all supported formats is better.
I'll test that the flow works first, then address comments.

chain-signatures/contract-sol/src/lib.rs
```diff
@@ -14,9 +14,13 @@ pub mod signet_program {
         request_ids: Vec<[u8; 32]>,
         signatures: Vec<Signature>,
     ) -> Result<()> {
-        // Minimal implementation - just emit the event
+        require!(
```
volovyks: Are these changes required to generate primitives for the Solana indexer?

ppca: not really. I'll remove them.

ppca:
ok the tests in solana-contract-example(https://github.com/sig-net/solana-contract-examples/blob/main/contract/tests/sign-respond-erc20.ts) all passed. I'll address comments

chain-signatures/contract-sol/src/lib.rs
ppca: done.

chain-signatures/contract-sol/src/lib.rs
```diff
@@ -14,9 +14,8 @@ pub mod signet_program {
         request_ids: Vec<[u8; 32]>,
         signatures: Vec<Signature>,
     ) -> Result<()> {
-        // Minimal implementation - just emit the event
```
volovyks, line 17: No reason to remove this.

volovyks (dismissed):
LGTM, let's merge and iterate

auto-mausx (approved):

PR #555, opened 2025-10-06
fix: mesh sync improperly

ChaoticTempest:
There was a point where in the update for `MeshState`, we weren't remove `need_sync` when transitioning to `Active`. Also, the connection wasn't transitioning properly as it would overwrite the syncing on the next ping. 

This fixes those issues, but might put us in the syncing state if sync is going to fail. If it becomes an issue on devnet, will fix later. Just trying to keep this fix small for now. The unit tests for mesh have been updated to accommodate syncing state.

### DIFF
--- a/chain-signatures/node/src/mesh/connection.rs
+++ b/chain-signatures/node/src/mesh/connection.rs
@@ -128,7 +128,8 @@ impl NodeConnection {
                         | OtherNodeStatus::Started
                         | OtherNodeStatus::WaitingForConsensus { .. } => NodeStatus::Inactive,
                     };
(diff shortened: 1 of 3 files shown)


volovyks:
> might put us in the syncing state if sync is going to fail

Isn't that the expected behaviour?
I thought we needed to stay in the "Sync required" state until it is successful.

ChaoticTempest:
yes, but that also means that we're stuck in syncing state forever if sync fails, which it won't yet because we are not responding with anything yet

ppca (approved):

PR #556, opened 2025-10-07
Remove baked docker

volovyks:
Once we have time, let's test how our node behaves and uses memory without baked in Redis.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,3 @@
-FROM redis:7.4.2 AS redis-bin
-
 FROM node:20 AS eth-builder
 WORKDIR /usr/src/app/contract-eth
 COPY chain-signatures/contract-eth/package.json chain-signatures/contract-eth/package-lock.json ./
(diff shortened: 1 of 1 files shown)


volovyks:
@auto-mausx, from my understanding, this should work out of the box, since we are not using baked-in Docker on dev.

auto-mausx:
Correct, however I'm a little worried for when we push what's on develop to main that the redis container might get changed and we end up with an incident.

volovyks:
@auto-mausx, how to deploy to dev without merging the PR? I do not see "run workflow".
Prod action does not have dev option.

auto-mausx:
In order to better keep track of what gets deployed and insure tests get ran beforehand, we removed the option to deploy dev manually at will.

It is also possible to manually build an image and push it with the latest tag for myself, Xyangi and also you I believe.

volovyks:
Ok, let's merge it then and revert after a day or two.

auto-mausx (approved):

volovyks:
Closing this, as it is now clear that Helios caused issues with memory usage.

PR #557, opened 2025-10-07
fix: remove self connection and purely follow contract for status

ChaoticTempest:
Mesh/connection pool no longer maintains a connection to ourselves. This means that we will set ourselves as active when we're in running, and inactive otherwise. This saves us from pinging ourselves and further draining resources.

### DIFF
--- a/chain-signatures/node/src/mesh/mod.rs
+++ b/chain-signatures/node/src/mesh/mod.rs
@@ -5,8 +5,10 @@ use crate::mesh::connection::NodeStatus;
 use crate::node_client::NodeClient;
 use crate::protocol::contract::primitives::Participants;
 use crate::protocol::ParticipantInfo;
+use crate::protocol::ProtocolState;
 use crate::rpc::ContractStateWatcher;
 use cait_sith::protocol::Participant;
+use near_account_id::AccountId;
 use tokio::sync::{mpsc, watch};
 
 pub mod connection;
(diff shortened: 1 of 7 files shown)


chain-signatures/node/src/mesh/mod.rs
```diff
@@ -110,14 +117,65 @@ impl Mesh {
...
+                    if let Some((participant, info)) = my_info {
+                        let new_status = match &contract {
+                            ProtocolState::Initializing(_) | ProtocolState::Resharing(_) => NodeStatus::Inactive,
+                            ProtocolState::Running(_) => NodeStatus::Active,
+                        };
+                        self.connections.connect(contract).await;
+                        self.state_tx.send_modify(|state| {
+                            // if the previous me is different from the current me, remove the
+                            // previous me from the MeshState.
+                            if let Some(previous_me) = previous_me.filter(|old| *old != participant) {
```
volovyks, line 134: This logic seems error-prone. I would say Mesh should not be aware of "me" or "account_id". When creating a mesh, we need to pass a list of connections that do not include the current node. Once we know that the list of participants or something else related has changed, we should recreate the Mesh or pass a list of updated connections.

ChaoticTempest: I don't understand how this would be error prone. Why is what you're suggesting any different than what this is doing right now. Mesh was always passed a list of connections based on the contract and adapts based on the list of participants inside the contract. We always need to be aware of who `me` is, because `me` needs to be set in our active set as well so we can't simply just filter it out when passed to mesh.

volovyks: Sorry for not clarifying what I mean.

The logic is divided between this place and `connect_nodes`. I'm suggesting to have a single function `update_mesh` that contains all the logic. It should make the testing of this component easier and API simpler.

Another think I'm confused about is usage of `account_id` and `me` (participant id) to identify nodes and update connections. It is hard to understand. Can we please document it?

As of the Mesh being aware of who is `me`, I agree that it is ok to pass this information inside to remove self connection.

ChaoticTempest: > The logic is divided between this place and connect_nodes. I'm suggesting to have a single function update_mesh that contains all the logic. It should make the testing of this component easier and API simpler.

I don't see how we can do this though. Updating the mesh state happens very asynchronously. We can't simply just have an `update_mesh` function, since the `Mesh` watches for changes in the connection pool.

We also already have unit tests that spins up this mesh/connection-pool so it's not that hard to do.

volovyks (approved):
Please, add docs in case you think we need them.

jakmeier (approved):
LGTM

PR #565, opened 2025-10-17
fix(audit): RUSTSEC-2025-0073

ChaoticTempest:
This fixes the latest cargo audit.

Moves all our audit exceptions into `.cargo/audit.toml` since that is cleaner instead of adding it to our CI each time. it's also nicer since we can just call `cargo audit` without specifying all our exceptions.

### DIFF
--- a/chain-signatures/node/src/logs.rs
+++ b/chain-signatures/node/src/logs.rs
@@ -1,4 +1,5 @@
 use std::fmt::{self, Display};
+use std::io::IsTerminal;
 use std::sync::OnceLock;
 
 use opentelemetry::trace::TracerProvider as _;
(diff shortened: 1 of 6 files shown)


chain-signatures/node/src/logs.rs
```diff
@@ -212,7 +213,7 @@ pub async fn setup(env: &str, node_id: &str, options: &Options) -> OtlpGuard {
     let log_otlp_layer = OpenTelemetryTracingBridge::new(&log_otlp_provider);
 
     let log_fmt_layer = tracing_subscriber::fmt::layer()
-        .with_ansi(atty::is(atty::Stream::Stderr))
+        .with_ansi(std::io::stderr().is_terminal())
```
volovyks: Is this and other changes (versioning) related to audit.toml?

volovyks (approved):
Nice! Look much cleaner.

PR #566, opened 2025-10-17
feat: Remove total timeout from signature and add test

ChaoticTempest:
Removes total timeout from signature requests, where we have a exponential backoff for old requests that couldn't be fulfilled. 

Each `SignRequest` inside the `SignQueue` now tracks attempts such that we can increase the delay on the next time they need to be processed.

### DIFF
--- a/chain-signatures/node/src/protocol/signature.rs
+++ b/chain-signatures/node/src/protocol/signature.rs
@@ -101,7 +101,6 @@ pub struct SignRequest {
     pub proposer: Participant,
     pub stable: BTreeSet<Participant>,
     pub round: usize,
-    attempts: Attempts,
 }
 
 pub struct SignQueue {
@@ -118,48 +117,6 @@ pub struct SignQueue {
     /// The set of pending request listeners that are waiting for a sign request to be indexed.
     /// They will be notified when a sign request is available.
     pending: HashMap<SignId, oneshot::Sender<SignRequest>>,
-}
-
-#[derive(Debug, Clone, PartialEq)]
-pub struct Attempts {
-    attempts: u32,
-    next_retry_at: Instant,
-}
-
-impl Default for Attempts {
-    fn default() -> Self {
-        Self::new()
-    }
-}
-
-impl Attempts {
-    pub fn new() -> Self {
-        Self {
-            attempts: 0,
-            next_retry_at: Instant::now(),
-        }
-    }
-
-    pub fn schedule_next_attempt(&mut self) {
-        self.attempts = self.attempts.saturating_add(1);
-        self.next_retry_at = Instant::now() + self.compute_retry_delay();
-    }
-
-    pub fn is_ready(&self, now: Instant) -> bool {
-        self.next_retry_at <= now
-    }
-
-    pub fn is_ready_now(&self) -> bool {
-        self.is_ready(Instant::now())
-    }
-
-    pub fn mark_ready(&mut self, now: Instant) {
-        self.next_retry_at = now;
-    }
-
-    fn compute_retry_delay(&self) -> Duration {
-        Duration::ZERO
-    }
 }
 
 impl SignQueue {
@@ -269,7 +226,6 @@ impl SignQueue {
             proposer,
             stable: stable.clone(),
             round,
-            attempts: Attempts::new(),
         }
     }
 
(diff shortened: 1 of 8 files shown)


integration-tests/src/mpc_fixture/fixture_tasks.rs
```diff
@@ -49,8 +49,16 @@ pub(super) fn test_mock_network(
...
+                    let mut should_filter = true;
+                    for filter in filters.iter_mut() {
+                        if !filter(&send_message) {
+                            tracing::info!(?from, ?to, log_msg, "Dropping a message because it didn't pass the test's filter");
+                            should_filter = false;
+                            break;
+                        }
+                    }
+
+                    if !should_filter {
```
jakmeier: Code clarity:

1. Isn't the naming of `should_filter` the wrong way around? (In my mind: To filter = not sending the message. The code only sends the message if `should_filter = true`)
2. Up to you which way you like it better but you could avoid the entire `should_filter` boolean by using a loop label and a direct `continue 'outer` in place of the break.

integration-tests/tests/cases/mpc.rs
```diff
@@ -184,6 +185,179 @@ async fn test_basic_sign() {
...
+            .iter()
+            .any(|action| action.contains("RpcAction::Publish")),
+        "unexpected rpc actions: {actions:?}"
+    );
+
+    let signature_messages = {
+        let log = network.output.msg_log.lock().await;
+        log.iter()
+            .filter(|entry| entry.starts_with("Signature"))
+            .count()
```
jakmeier, line 367: side note: Should we change how logs are represented? I just wanted some dumb way of keeping a log of what happened, without coupling the testing closely with internal implementation types. Curious to hear what you think after you have written your first tests with this approach.

jakmeier (dismissed):
Great stuff!

I see you removed the exponential backoff. Makes sense, I don't think we need, as discussed on Monday. What I thought of later is that we could also make the timeout itself an exponential backoff. That is, wait longer and longer between retries, without actually removing it. Probably not needed either, but I thought I mention it in case we come around and add a backoff mechanism later.

Also, nice use of the component testing setup! Let me know if you have any thoughts after using it. I have some ideas what I want to improve but I'd really like your input before I put too much else into it.

integration-tests/src/mpc_fixture/fixture_tasks.rs
ChaoticTempest: yep, corrected it

integration-tests/tests/cases/mpc.rs
ChaoticTempest: yeah I think we need more robust typing or better convenience helpers for this in the long term. It's going to be hard to follow what this is doing

integration-tests/src/mpc_fixture/fixture_tasks.rs
jakmeier: Ah, yes the code here locally now makes more sense. But you also inverted the meaning of the filter in general, which wasn't what I meant. I'd really prefer staying with the semantics of the standard library [filter](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.filter).

Can we keep it as `filter` returning true means it passes the filter? The local variable `should_filter` here can stay as it is, just needs inverted assignment. Or maybe `should_drop` or `should_send` would be clearer naming.

integration-tests/tests/cases/mpc.rs
jakmeier: Added a bullet point to possibly improve this quarter in #583.

integration-tests/tests/cases/mpc.rs
```diff
@@ -184,6 +185,195 @@ async fn test_basic_sign() {
...
+            .iter()
+            .any(|action| action.contains("RpcAction::Publish")),
+        "unexpected rpc actions: {actions:?}"
+    );
+    assert!(
+        dropped_messages > 0,
+        "expected test filter to drop at least one signature message"
+    );
+    assert!(
+        start.elapsed() >= Duration::from_secs(2),
```
volovyks, line 269: What is the logic behind this check? From what I see, we are making sure that at least 2 seconds passed here:
```rust
if start.elapsed() < Duration::from_secs(2) {
        tokio::time::sleep(Duration::from_secs(2) - start.elapsed()).await;
}
```
so it must be true.

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -258,10 +258,13 @@ impl SignQueue {
                 .with_label_values(&[indexed.chain.as_str(), my_account_id.as_str()])
                 .inc();
 
-            let request = self.organize_request(stable, participants, indexed, 0);
+            let mut request = self.organize_request(stable, participants, indexed, 0);
+            if request.indexed.timestamp_sign_queue.is_none() {
```
volovyks, line 262: Why is this field optional in the first place? Can we make it non optional and ensure timestamp is added at the indexer level?

volovyks (approved):

chain-signatures/node/src/protocol/signature.rs
ChaoticTempest: maybe it had a purpose once upon a time. Maybe @ppca can probably explain why, but I've since removed it in https://github.com/sig-net/mpc/pull/587

integration-tests/tests/cases/mpc.rs
ChaoticTempest: removing this

ChaoticTempest:
closing this one since we removed total_timeout in newer pull requests

Issue #272, opened 2025-03-05
Parallel Signing & Resharing


ChaoticTempest:
On a reshare, the system is currently paused and no new sign requests can be completed. One way to mitigate downtime is to have the system keep on signing with the old set of participants before resharing completes.

We probably want to have two different of resharing that nodes can vote for, either `Resharing(Maintenance)` or `Resharing(Parallel)` where the latter lets us sign while resharing is happening.

Issue #144, opened 2025-01-28
Add the ability to cancel resharing


ChaoticTempest:
Right now, if resharing were to error out in any case, a new resharing protocol would be needed. What happens if this resharing process becomes too long and the new set of participants aren't needed anymore?

In that case, maybe resharing is no longer valid and we should go back into running. The participants should be able to vote to cancel resharing in that case.

volovyks:
closed in https://github.com/sig-net/mpc/pull/568

PR #568, opened 2025-10-21
fix: resharing retrying & timeout

ChaoticTempest:
This adds a fix to the resharing process where we will timeout if it takes too long to progress. The progress is determined by the last action (i.e. receiving a new message, sending messages)

We will default to a timeout of 5 minutes from the last action which should give nodes plenty of time to restart if they are doing so.

Additionally, I made resharing a two phase process. One is awaiting ready, and the other is running the resharing.

The awaiting step is where we will constantly broadcast that we're ready to start every 10 seconds. When all new participants including us is ready, we will go to the next phase which is running the resharing.

So this fixes several cases where:
- node restart and then have to restart the whole resharing again. So here this is resolved by going back to awaiting ready and then running resharing.
- starting the resharing protocol at different times and then timeout at different times. The awaiting step fixes this as now we all start roughly at the same time withing a 10 second interval of the broadcasting of readiness

This also adds a new resharing test so we can see that it works properly when a node is offline and restarts.

### DIFF
--- a/chain-signatures/node/src/protocol/cryptography.rs
+++ b/chain-signatures/node/src/protocol/cryptography.rs
@@ -6,6 +6,7 @@ use super::state::{
     ResharingState, RESHARING_READY_BROADCAST_INTERVAL,
 };
 use super::MpcSignProtocol;
+use crate::protocol::contract::ResharingContractState;
 use crate::protocol::message::{GeneratingMessage, ResharingMessage, ResharingReadyMessage};
 use crate::protocol::state::{PersistentNodeData, WaitingForConsensusState};
 use crate::protocol::MeshState;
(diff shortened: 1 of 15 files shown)


integration-tests/tests/cases/mod.rs
```diff
@@ -217,3 +220,66 @@ async fn test_batch_duplicate_signature() -> anyhow::Result<()> {
...
+
+    // Wait for all online nodes to move to resharing state.
+    nodes.wait().nodes_resharing().await?;
+
+    // Now we should wait to see that we are still in the resharing state even after
+    // a long time, since one participant is offline and cannot complete the resharing.
+    tokio::time::sleep(Duration::from_secs(30)).await;
+    assert!(matches!(
+        nodes.contract_state().await?,
+        mpc_contract::ProtocolContractState::Resharing(_)
```
volovyks, line 263: How much effort would it take to fix these two issues? It is still possible to be "stuck" in the resharing state.
- https://github.com/sig-net/mpc/issues/272
- https://github.com/sig-net/mpc/issues/144

chain-signatures/node/src/protocol/state.rs
```diff
@@ -77,18 +79,47 @@ pub struct RunningState {
...
     /// If the resharing state fails to store data after generating, it gets temporarily
     /// stored here and retried later.
     pub failed_store: Option<SecretKeyShare>,
+    pub started_at: Instant,
+    pub last_activity: Instant,
 }
 
+#[allow(clippy::large_enum_variant)]
+pub enum ResharingPhase {
+    AwaitingReady(ResharingReadyState),
```
volovyks: Would it be possible to rename it to  ResharingAwaitingState? `Ready` in its context sounds like "we are ready for resharing".

chain-signatures/node/src/protocol/message/mod.rs
```diff
@@ -638,6 +653,18 @@ impl MessageChannel {
             }
         }
     }
+
+    pub async fn subscribe_resharing_ready(&self) -> mpsc::Receiver<ResharingReadyMessage> {
+        let Some(subscription) = self.subscribe(SubscribeId::ResharingReady).await else {
+            panic!("failed to subscribe for resharing readiness");
```
volovyks, line 659: panic!()?

volovyks (commented):
Wow, great improvement. And awesome test! Happy to merge it, not sure about those `panic!()`s.

chain-signatures/node/src/protocol/state.rs
ChaoticTempest: yep will rename

chain-signatures/node/src/protocol/message/mod.rs
ChaoticTempest: we also panic for the other subscribe to generation/resharing as well. This is because on startup, we subscribe. If we can't then we can't really do anything. This is because of how subscribing works. If it returns an error, it means something else has been aborted. So this won't really ever be reached unless the whole system goes down

integration-tests/tests/cases/mod.rs
ChaoticTempest: It'll be easy to add in the cancel of resharing, but for parallel signing, we need to consider somethings like if we're kicking somebody, would it be a security concern if we allow them to sign?

ChaoticTempest: added cancelling resharing here: https://github.com/sig-net/mpc/pull/570

ChaoticTempest:
Added another test_resharing_running_participant_restart that does the restart in the middle of reshare is running. This had some other issues that arose like resharing consuming previous reshare messages. 

To fix that, each node during every awaiting step creates their own separate token to be broadcasted out with the ReadyMessage. These tokens are combined later when all new participants are ready to form a new unique token to this very specific reshare running phase such that new reshare running phase would have an entirely different token.

So reshare messages match against these tokens, if the token is not correct, ignore the message.

integration-tests/tests/cases/mod.rs
volovyks: If we are kicking someone because of their misbehavior, the voting process will take several hours in the best-case scenario. The resharing protocol is relatively fast, so it does not matter much (5h vs 5h+1m).

The main reason we wanted parallel signing and resharing was possible downtime. Especially, downtime caused by failed resharing.

volovyks (approved):
Inventing a new approach for message filtering isn’t the best idea 🙂 But adopting a single universal approach for all protocols would be overkill.

chain-signatures/node/src/protocol/cryptography.rs
```diff
@@ -1,18 +1,33 @@
...
 use tokio::sync::mpsc;
 
+pub static RESHARING_RUNNING_TIMEOUT_SECS: AtomicU64 = AtomicU64::new(300);
+
+pub fn resharing_running_timeout() -> Duration {
+    Duration::from_secs(RESHARING_RUNNING_TIMEOUT_SECS.load(Ordering::SeqCst))
+}
+
+pub fn set_resharing_running_timeout(duration: Duration) {
+    RESHARING_RUNNING_TIMEOUT_SECS.swap(duration.as_secs(), Ordering::SeqCst);
```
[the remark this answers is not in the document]
jakmeier: Super nitpicky but Copilot has a point about the semantics fitting better to `store` than `swap`. :wink:

While at it, another detail is that with `swap` + `Ordering::SeqCst`, the compiler has to generate an atomic load with appropriate memory fences to ensure all potentially outstanding stores (or swaps) of other threads to `RESHARING_RUNNING_TIMEOUT_SECS` have been observed by this thread before this thread can perform its own swap. This can lead to quite expensive thread stalling and CPU cache invalidations.

Using `store` instead of `swap` would get rid of the load. I think that would already avoid CPU-level synchronization for x86 targets thanks to their relatively stronger memory model.

If you further change `Ordering::SeqCst` to Ordering::Relaxed`, we should be able to entirely avoid expensive CPU synchronization instructions on all platforms. This is safe since we don't actually use this variable for multi-threaded synchronization. We just need every get and set to be atomic. Relaxed load / store is the best choice for this use case.

But again, this isn't important in the grand scheme of things. Feel free to change it if you want or keep it as is. It won't make a noticeable difference either way.

chain-signatures/node/src/protocol/state.rs
```diff
@@ -77,16 +82,64 @@ pub struct RunningState {
...
+}
+
+impl ResharingPhase {
+    pub fn awaiting(me: Participant) -> Self {
+        let my_token = random::<u64>();
+        Self::Awaiting(ReshareAwaiting {
+            ready_tokens: std::iter::once((me, my_token)).collect(),
+            my_token,
+            // ready to broadcast immediately
+            broadcast_interval: Instant::now() - RESHARING_READY_BROADCAST_INTERVAL,
```
jakmeier, line 133: I think this can technically panic. `Instant` is an opaque type that may start counting at 0 when the program starts. (I think I've observed this with WASM targets.) If we get here in the first 10s of the program, subtraction will negative overflow.

It probably happens to be fine for our use case. But API wise, a `std::time::Instant` should only be used to compare it to another `Instant`, not to some global clock.

jakmeier (commented):
Really nice! This is miles more robust than what we had before!

Long-term, keep in mind that this only solves crashes but not malicious behavior. I believe this approach would not work if, for example, one malicious node sends different `my_token` values to each peer.

We'll have to add it to the list of necessary improvements for byzantine fault-tolerance.

ChaoticTempest:
> Long-term, keep in mind that this only solves crashes but not malicious behavior. I believe this approach would not work if, for example, one malicious node sends different my_token values to each peer.
>
> We'll have to add it to the list of necessary improvements for byzantine fault-tolerance.

AFAIU, we probably cannot add BFT to the resharing step aside from restarting the reshare protocol with a different new participant set.

To visualize this a bit, for resharing, we have old participant set and new participant set, where a subset of the old participants must also be in the new participants. For a participant to be able to get their new key share, they have to communicate. That means they need to be online, which also means all new participants have to be online to get their keyshare. So this would mean that new participants have to be honest. We can maintain up to threshold amount of old participants that must be honest, and can be included in the new participant set.

If we're in resharing running step, and a node behaves maliciously, we really have to just kick them out, and then restart the whole resharing process again. This is due to cait-sith not allowing us to adjust the participants inflight during resharing

jakmeier:
Thanks for the details @ChaoticTempest !

I see, the resharing algorithm itself may not be suitable to be made BFT. But I believe we can still work on:

- Detect when a node behaves dishonest.
- Ensure bad participants can't disrupt the entire process indefinitely.

That said, I don't think making resharing BFT is a useful thing to work on right now. We can always fall back to manual intervention in a process that's only triggered manually anyways. As long as we know the limitations of it, I'm very happy with what we have now with this PR merged in.

PR #570, opened 2025-10-21
feat: add ability to cancel resharing

ChaoticTempest:
This adds in the ability to cancel resharing and go back to the previous running state if something wrong was to happen during the resharing process


volovyks (approved):
Awesome!

PR #572, opened 2025-10-22
feat: disable helios, add regular eth indexing

ChaoticTempest:
Since dev/testnet is broken right now due to helios not working properly. I'm just going to outright disable it as we need to make progress here.

So, this disables helios by default. Adds a new `light_client` feature flag that enables it. If disabled, we regular ethereum RPC. This also adds a local ethereum sandbox to make sure it works properly.

### DIFF
--- a/integration-tests/tests/cases/ethereum.rs
+++ b/integration-tests/tests/cases/ethereum.rs
@@ -71,7 +71,7 @@ async fn test_signature_ethereum() -> Result<()> {
             .query()
             .await?;
         if let Some(event) = events.into_iter().find(|event| {
-            event.request_id == &expected_request_id[..] && event.responder == requester
+            event.request_id == expected_request_id[..] && event.responder == requester
         }) {
             matching_event = Some(event);
             break;
--- a/integration-tests/tests/cases/ethereum.rs
+++ b/integration-tests/tests/cases/ethereum.rs
@@ -21,17 +21,17 @@ async fn test_signature_ethereum() -> Result<()> {
         .as_ref()
         .context("ethereum sandbox not initialized")?;
     let endpoint = eth_ctx.sandbox.external_http_endpoint.clone();
-    let private_key = eth_ctx.sandbox.private_key.clone();
+    let secret_key = eth_ctx.sandbox.secret_key.clone();
     let chain_id = eth_ctx.sandbox.chain_id;
     let contract_address = eth_ctx.contract_address;
 
-    let (client, requester) = eth::client(&endpoint, &private_key, chain_id)?;
+    let (client, requester) = eth::client(&endpoint, &secret_key, chain_id)?;
     let contract = eth::ChainSignaturesContract::new(contract_address, client.clone());
 
     let payload = [7u8; 32];
-    let path = "m/44'/60'/0'/0/0";
+    let path = "test";
     let algo = "secp256k1";
-    let dest = "http://localhost/callback";
+    let dest = "solana:EtWTRABZaYq6iMfeYKouRu166VU2xqa1";
     let params = "{}";
 
     let request = eth::SignRequest {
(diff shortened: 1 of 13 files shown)


integration-tests/tests/cases/ethereum.rs
```diff
@@ -0,0 +1,132 @@
...
+    let chain_id = eth_ctx.sandbox.chain_id;
+    let contract_address = eth_ctx.contract_address;
+
+    let (client, requester) = eth::client(&endpoint, &private_key, chain_id)?;
+    let contract = eth::ChainSignaturesContract::new(contract_address, client.clone());
+
+    let payload = [7u8; 32];
+    let path = "m/44'/60'/0'/0/0";
+    let algo = "secp256k1";
+    let dest = "http://localhost/callback";
```
volovyks: This was added a long time ago and was intended to be a chain ID (now - CAIP 2). WE are not using it yet, but I do not want it to be misleading for people who check how it is used.

volovyks (commented):
Great work, thank you for looking into this!

My only concern is that I’m not sure using a feature flag is the right approach for setting up the light client. Right now, all the setup is centralized in the CLI run() function — including the Indexer parameters. cc @auto-mausx 

Also, using #[cfg(...)] seems a bit heavy. Would it make more sense to separate these two implementations and pick one during node initialization?

volovyks:
BTW, ideally, we should make a branch of 1.8.0, add this fix there (as a hot fix for testnet), then merge it to main and then to develop.

On the other hand... we need to prepare a version that includes all the resharing fixes and add it to mainnet. Doing all that as hot fixes would not be ideal. Let's focus on stabilizing dev env then.

ChaoticTempest:
We might want to gate the usage of light_client behind versions is the thing, which is why feature flags is useful here. It's going to be hard to track down which nodes have light_client enabled or not if it was a CLI option. 

> BTW, ideally, we should make a branch of 1.8.0, add this fix there (as a hot fix for testnet), then merge it to main and then to develop.
>
> On the other hand... we need to prepare a version that includes all the resharing fixes and add it to mainnet. Doing all that as hot fixes would not be ideal. Let's focus on stabilizing dev env then.

Dev also has a couple more changes that testnet doesn't have for 1.8 but I'd rather save us the effort and fully upgrade testnet to what we have on dev. And only if it fails to upgrade on testnet should we do a hotfix.

volovyks:
We can track it in Grafana, not an issue. Also, the usage of the light client should be controlled by a node provider.
And from what I understand, now it can be enabled by the node provider. But with a flag, not a parameter in `run()`.

volovyks (approved):
Not blocking this, since we need a fix. But I would appreciate a third opinion from @jakmeier . Even if this PR is merged.

jakmeier:
> Also, using #[cfg(...)] seems a bit heavy. Would it make more sense to separate these two implementations and pick one during node initialization?

Yeah there are pros and cons for using cfg. Using cfg can be useful here since helios pulls in a serious amount of dependencies. Not compiling and shipping this code when we don't need it can be a good idea.

But in general, I prefer cfg features to not change the logic, only make different options available. And these options should be additive. That is, instead of `light_client`, I would prefer different indexer implementations to each live behind their own cfg. For example `eth_indexer_helios_light_client` and `eth_indexer_direct_rpc_access` as two features. Each with a separate .rs file with a single `#[cfg(...)]` directive right on the `mod` declaration.

We can then enable each indexer adapter in a release build depending on whether we think they are safe and stable to use or not.

Then a CLI option can choose between the options enabled at compile time.

-------

> We might want to gate the usage of light_client behind versions is the thing, which is why feature flags is useful here.

As we are aiming to have non-globally synced updates, nodes need to know what version they run anyway. Probably a own version number like `const VERSION: u64` that's fixed at compile time and then a dynamic field on messages that let's us know what version other nodes are using.

Even without using any cfg, we can enforce rules using this constant `VERSION` to achieve the same compile-time gating that I think you describe here @ChaoticTempest 
But I might have misunderstood your intention.

------

Overall, no strong opinion on the short-term. But long-term I don't like having many `#[cfg(...)]` directives spread across a single file. If we keep the cfg, at least let's pull each implementation into their own file.

PR #575, opened 2025-10-23
Add revoke_join

### LINKED ISSUE #28: Mitigate user error on contract::join call
When a participant goes and tries to `join` the network, they can potentially send in the wrong `sign_pk`. We can either make it so that `join` call validates that the `sign_pk` is correct by having the caller send in a signed message such that we can use `sign_pk` to verify that this is indeed correct. Still some user error there but a bit less now since the user would have to make use of the key, and potentially learn that they're using the wrong key.

A better way of handling this is post-join, by allowing the user/participant to modify their own credentials. There can be some weird side-effects with this such as a node having not picked up the new config of participants sign_pk where they're unable to verify that the messages are indeed coming from this newly updated participant. We will need a new test for this in that case.

volovyks:
Potential participants should have the ability to revoke their `join` call in case they want to update their data.
`revoke_join` is expected to be called manually. After this action, the node will call `join` automatically.

### DIFF
--- a/chain-signatures/contract/src/lib.rs
+++ b/chain-signatures/contract/src/lib.rs
@@ -332,6 +332,34 @@ impl VersionedMpcContract {
     }
 
     #[handle_result]
+    pub fn revoke_join(&mut self) -> Result<(), Error> {
+        log!("revoke_join: signer={}", env::signer_account_id());
+        let protocol_state = self.mutable_state();
+
+        match protocol_state {
+            ProtocolContractState::Running(RunningContractState {
+                candidates,
+                join_votes,
+                ..
+            }) => {
+                let signer_account_id = env::signer_account_id();
+                if candidates.get(&signer_account_id).is_none() {
+                    return Err(JoinError::RevokeNotCandidate.into());
+                }
+
+                // cleanup the existing votes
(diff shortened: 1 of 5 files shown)


chain-signatures/contract/src/lib.rs
```diff
@@ -331,6 +331,34 @@ impl VersionedMpcContract {
...
+                let signer_account_id = env::signer_account_id();
+                if candidates.get(&signer_account_id).is_none() {
+                    return Err(JoinError::RevokeNotCandidate.into());
+                }
+
+                // cleanup the existing votes
+                join_votes.remove(&signer_account_id);
+
+                // remove from candidates
+                candidates.remove(&signer_account_id);
```
ChaoticTempest, line 354: we should not be removing the candidate at all here. This should purely revoke the join vote. Candidates themselves can update their own info through `join`

ChaoticTempest: actually, you're trying to remove the candidacy. It's better to name this to be `remove_candidacy`. Revoke join feels like it would be removing just the votes

ChaoticTempest (approved):

PR #579, opened 2025-10-27
Add blacksand digest test

volovyks:
The digest is not correct now. We need to communicate that with BlackSand.


volovyks:
The wrong digest can be caused by an outdated version of Mainnet. Let's deploy and check again.

ppca:
Do we know when we plan to deploy mainnet? I'm thinking best to first deploy to all old partners including blacksand, once we make sure we have correct digest and fixed all issues if they occur, we invite new partners to join. cc @auto-mausx

volovyks:
closing infavour of https://github.com/sig-net/mpc/pull/632

PR #581, opened 2025-10-28
small: make derive_user_address() public so solana-contract-examples can use it

ppca:
This also includes adding serde feature to alloy-consensus so solana-contract-examples can use it

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -56,7 +56,7 @@ alloy-dyn-abi = "1.4.1"
 ethers-core = "2.0.13"
 secp256k1 = "0.28.2"
 rlp = "0.5"
-alloy-consensus = "1.0.38"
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -56,7 +56,7 @@ alloy-dyn-abi = "1.4.1"
 ethers-core = "2.0.13"
 secp256k1 = "0.28.2"
 rlp = "0.5"
-alloy-consensus = "1.0.38"
+alloy-consensus = { version = "1.0.38", features = ["serde"] }
```
ChaoticTempest, line 59: did we have to add this new feature flag if we're only just making `derive_user_address` public?

ChaoticTempest (approved):

Cargo.toml
ppca: Yes this is needed otherwise solana-contract-examples won't compile

PR #584, opened 2025-10-30
test: Message counting tests

jakmeier:
Add tests checking that T / P / S protocols sent the expected amount of messages.

As noted in the comment, we may not necessarily care about the exact number. But we should understand roughly what they are and run tests to ensure it doesn't change unnoticed.

To implement this test, I also made the testing framework more flexible in how it observed outputs. We can now check specific counts for each protocol invocation.

The printed debug summary per test looks a bit like this:

```
### Participant(2) -> Participant(1)
           1    Forwarded Signature(15446211684585113244)

### Participant(2) -> Participant(0)
           1    Forwarded Posit(0)
           1    Forwarded Signature(15446211684585113244)

### Participant(0) -> Participant(1)
           2    Forwarded Posit(0)
           1    Forwarded Signature(15446211684585113244)

### Participant(0) -> Participant(2)
           1    Forwarded Signature(15446211684585113244)
           2    Forwarded Posit(0)

### Participant(1) -> Participant(0)
           1    Forwarded Posit(0)
           1    Forwarded Signature(15446211684585113244)

### Participant(1) -> Participant(2)
           1    Forwarded Signature(15446211684585113244)
```

### DIFF
--- a/chain-signatures/node/src/protocol/presignature.rs
+++ b/chain-signatures/node/src/protocol/presignature.rs
@@ -39,8 +39,8 @@ pub type PresignatureId = u64;
 /// that was used to generate it.
 #[derive(Copy, Clone, Debug, Eq, PartialEq, Hash, Ord, PartialOrd, Serialize, Deserialize)]
 pub struct FullPresignatureId {
-    id: PresignatureId,
-    pair_id: TripleId,
+    pub id: PresignatureId,
+    pub pair_id: TripleId,
 }
 
 impl FullPresignatureId {
--- a/integration-tests/src/mpc_fixture/builder.rs
+++ b/integration-tests/src/mpc_fixture/builder.rs
@@ -5,6 +5,7 @@ use crate::containers::Redis;
 use crate::mpc_fixture::fixture_interface::SharedOutput;
 use crate::mpc_fixture::fixture_tasks::MessageFilter;
 use crate::mpc_fixture::input::FixtureInput;
+use crate::mpc_fixture::message_collector::CollectMessages;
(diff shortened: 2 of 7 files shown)


jakmeier:
Hm, the test fails for triple 14508508910886555200, it has a different number of messages than the rest.

```
0 -> 1:  10 messages (~as expected)
0 -> 2:  6 messages (~as expected)
1 -> 0:  136 (as expected)
1 -> 2:  6 messages (~as expected)
2 -> 0:  4 messages (instead of ~135)
2 -> 1:  4 messages (instead of ~135)
```

All other triples, as far as I can see in the log, follow the rules that participant 2 sends ~135 to both other participants.
Locally, tests run fine for me, without exception.

So, this isn't even fully deterministic in our implementation. :thinking:

jakmeier:
Ahh, no this is a testing mistake with an unclean stop. I was able to reproduce it locally and it looks like the missing messages are in a buffer to be sent. If I count the messages that are sent after the output is observed, I see there are actually ~135 message for all triples.

I will need to fix that.

jakmeier:
I changed the tests to only check the upper bound on messages. Ultimately, we should care about not sending too many messages, not about specific numbers.

Checking the exact number of messages would only work for finished protocols. To do that, I started implementing something to stop deterministically, finish processing in-flight messages and then collect a consistent snapshot of sent messages after that. But that felt way overengineered for a test like this. Keeping it simple seems much better here.

jakmeier:
Tests are now passing, should be ready for review now. @ChaoticTempest

ChaoticTempest (approved):
nice

PR #585, opened 2025-11-04
FIX: update workflow to tag prod images

auto-mausx:


ChaoticTempest:
hmm, why was Cargo.lock updated? It's killing the tests it seems like

volovyks:
I thought the idea was to expose this info on Grafana as a metric. What is the plan here?

auto-mausx:
intent got moved in with another PR

Issue #499, opened 2025-08-28
Pending request management for bidrectional messaging


jakmeier:
# Goal

We want to ensure every `sign_respond` message will eventually be matched with a `read_respond` call. (See #465)

Any request missing a `read_respond` must be retried stage-by-stage until it succeeds.

Cases that cannot complete will be stuck forever. (Later design iterations can clean those up if it becomes a problem.)

# Limitations of the Current State

In the first implementation of bidirectional messaging #475, the pending requests are stored as follows.

```rust
sign_respond_tx_map: Map<SignRespondTxId, SignRespondTx>
```

In this iteration, messages are added when a new request is observed and then removed again when we finish processing them.

To produce a response, we check every indexed transaction against the pending requests to see if a new outcome is available.

Unfortunately, we would miss requests in the following cases:

 - The in-memory data structure gets wiped because of a reboot.
 - Chain / indexer timing leads to processing of the response before the request.
   - For example, a request on chain A is processed by nodes 1,2,3 while node 4 uses a different indexer which has fallen behind. Node 4 has not seen the request yet. The signed tx gets pushed to chain B. Node 4 has a healthy indexer for chain B, thus sees the result immediately, while it hasn't even seen the request. Node 4 will eventually see the request and put it as pending. But it will never process the response, therefore keep retrying it forever, which won't success because all other nodes have finished processing it already.

# Proposal

- Nodes must track the state of a request based only on the finalized on-chain state. (Ensure all nodes have the same view.)
- Nodes store periodic checkpoints of pending requests at a given block height. (Used to restore missing requests after a reboot.)
    - Checkpoint are per chain
    - Checkpoint are predefined as a function of height and the blockchain's timestamp. E.g. "use the first block after a full hour".
    - When a node reaches a predefined checkpoint height, it dumps all pending requests for that given height of that chain.
    - Checkpoints of chain A also include the height of the last received response of all other chains. Otherwise, it is unclear where to start looking for responses on target chains for requests from chain A.
    - Checkpoints can be requested from other nodes through an API
    - Checkpoints must be verifiable (e.g. signed by the network)
    - (optional) we can also persist checkpoints locally or even put them on a separate storage service
- When processing a request, nodes store them as pending (as done already)
- When processing a request, nodes also have to check if a response has already been published, going through all potential blocks that were already processed but are not ordered "logically after" the currently processed block. Definition of "logically after":
    - Block of chain A at height `h+1` is "logically after" a block of chain A at height `h`
[...]

PR #587, opened 2025-11-06
feat: Add backlog checkpoints w/ simpler bidirectional requests

ChaoticTempest:
This adds in checkpoint creation and recovery from backlog:

- create checkpoints every block interval (hardcoded to ~5~ 20 for eth and ~10~ 120 for sol for now)
- recover from checkpoint on restart (there's a test for this)
- ~checkpoints are rudimentarily uploaded to the contract for now (will be removed in the future once stored in redis since I didn't want to snowball redis into this PR).~ A query of the latest checkpoint for every other node is done on recovery.
- indexers now follow the simpler approach I mentioned in https://github.com/sig-net/mpc/issues/499#issuecomment-3487754525
  - backlog.insert request on indexing Sign or SignBidirectional
  - backlog.watch execution on indexing Respond for SignBidirectional request
  - backlog.remove on Respond for Sign, RespondBidirectional for SignBidirectional requests.

Note the SignRequestType and the BacklogTransaction has a lot of duplicate information and some sporadic clones are about. Will clean up in a future PR.

### DIFF
--- a/chain-signatures/contract/src/lib.rs
+++ b/chain-signatures/contract/src/lib.rs
@@ -24,10 +24,10 @@ use near_sdk::{
     PromiseError, PublicKey,
 };
 use primitives::{
-    CandidateInfo, Candidates, InternalSignRequest, Participants, PendingRequest, PkVotes,
-    SignPoll, SignRequest, StorageKey, Votes, YieldIndex,
+    CandidateInfo, Candidates, Chain, Checkpoint, InternalSignRequest, Participants,
+    PendingRequest, PkVotes, SignPoll, SignRequest, StorageKey, Votes, YieldIndex,
 };
-use std::collections::{BTreeMap, HashSet};
+use std::collections::{BTreeMap, HashMap, HashSet};
 
 use crate::config::Config;
 use crate::errors::Error;
@@ -477,6 +477,8 @@ impl VersionedMpcContract {
                         candidates: Candidates::new(),
                         join_votes: Votes::new(),
                         leave_votes: Votes::new(),
+                        checkpoint_votes: HashMap::new(),
+                        latest_checkpoints: HashMap::new(),
                     });
                     Ok(true)
                 } else {
@@ -520,6 +522,8 @@ impl VersionedMpcContract {
                         candidates: Candidates::new(),
                         join_votes: Votes::new(),
                         leave_votes: Votes::new(),
+                        checkpoint_votes: HashMap::new(),
+                        latest_checkpoints: HashMap::new(),
                     });
                     Ok(true)
                 } else {
@@ -561,6 +565,8 @@ impl VersionedMpcContract {
                         candidates: Candidates::new(),
                         join_votes: Votes::new(),
                         leave_votes: Votes::new(),
+                        checkpoint_votes: HashMap::new(),
+                        latest_checkpoints: HashMap::new(),
                     });
                     Ok(true)
                 } else {
@@ -568,6 +574,58 @@ impl VersionedMpcContract {
                 }
             }
             _ => Err(InvalidState::UnexpectedProtocolState.message(protocol_state.name())),
+        }
+    }
+
+    /// Vote on a checkpoint for a specific chain. When threshold votes are reached
+    /// for the exact same checkpoint (same block_height and pending_transactions),
+    /// it is stored as the latest checkpoint for that chain.
+    ///
+    /// Returns Ok(true) if threshold was reached and checkpoint was stored.
+    /// Returns Ok(false) if threshold was not yet reached.
+    #[handle_result]
+    pub fn vote_checkpoint(&mut self, chain: Chain, checkpoint: Checkpoint) -> Result<bool, Error> {
+        let voter = self.voter()?;
+        let checkpoint_height = checkpoint.block_height;
+        log!("vote_checkpoint: signer={voter}, chain={chain:?}, block_height={checkpoint_height}");
+        let ProtocolContractState::Running(RunningContractState {
+            threshold,
+            checkpoint_votes,
+            latest_checkpoints,
+            ..
+        }) = self.mutable_state()
+        else {
+            return Err(InvalidState::ProtocolStateNotRunning.into());
+        };
+
+        if let Some(latest) = latest_checkpoints.get(&chain) {
+            if checkpoint.block_height <= latest.block_height {
+                log!(
+                    "checkpoint ignored: chain={chain:?} block_height={checkpoint_height} <= latest_height={}",
+                    latest.block_height
+                );
+                return Ok(true);
+            }
+        }
+
+        let chain_votes = checkpoint_votes.entry(chain).or_default();
+
+        // Add this voter's vote for this specific checkpoint
+        let voted = chain_votes.entry(checkpoint.clone());
+        voted.insert(voter);
+        let vote_count = voted.len();
+
+        if vote_count >= *threshold {
+            log!("checkpoint latest updated: chain={chain:?} block_height={checkpoint_height}");
+
+            // Store as latest checkpoint
+            latest_checkpoints.insert(chain, checkpoint);
+            // latest checkpoint reached, clear previous ones:
+            checkpoint_votes.remove(&chain);
+            Ok(true)
+        } else {
+            log!("checkpoint voted: chain={chain:?} block_height={checkpoint_height}, votes={vote_count}/{threshold}");
+            Ok(false)
         }
     }
 
@@ -698,6 +756,8 @@ impl VersionedMpcContract {
                 candidates: Candidates::new(),
                 join_votes: Votes::new(),
                 leave_votes: Votes::new(),
+                checkpoint_votes: HashMap::new(),
+                latest_checkpoints: HashMap::new(),
             }),
             pending_requests: IterableMap::new(StorageKey::PendingRequests),
             proposed_updates: ProposedUpdates::default(),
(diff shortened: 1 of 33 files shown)


chain-signatures/contract/src/lib.rs
```diff
@@ -571,6 +577,58 @@ impl VersionedMpcContract {
...
     }
 
+    /// Vote on a checkpoint for a specific chain. When threshold votes are reached
+    /// for the exact same checkpoint (same block_height and pending_transactions),
+    /// it is stored as the latest checkpoint for that chain.
+    ///
+    /// Returns Ok(true) if threshold was reached and checkpoint was stored.
+    /// Returns Ok(false) if threshold was not yet reached.
+    #[handle_result]
+    pub fn vote_checkpoint(&mut self, chain: Chain, checkpoint: Checkpoint) -> Result<bool, Error> {
```
jakmeier: In the PR description you write

> checkpoints are rudimentarily uploaded to the contract for now (will be removed in the future once stored in redis since I didn't want to snowball redis into this PR).

So, all of these changes to the contract are just temporary, right? No voting on-chain when we have it in redis, right? That how I remember our decision on how we handle checkpoints, self-verifiably stored in redis, nothing on-chain.

It just seems that dumping it to redis would be less invasive than updating the contract, including the migration logic... Am I missing something?

chain-signatures/node/src/backlog.rs
```diff
@@ -1,15 +1,23 @@
...
 use std::sync::Arc;
 use tokio::sync::RwLock;
 
+pub use mpc_primitives::Checkpoint;
+
+/// Checkpoint interval for Ethereum (every 5 blocks)
+const CHECKPOINT_INTERVAL_ETH: u64 = 5;
+
+/// Checkpoint interval for Solana (every 10 blocks)
+const CHECKPOINT_INTERVAL_SOL: u64 = 10;
```
jakmeier: So what is that, around 2 min for Ethereum and about 4 seconds for Solana?

Honestly, I was thinking more like 10 minutes, or even an hour between checkpoints. We need to implement the catch up logic from the checkpoint to the latest block anyway. If that's a couple hundred blocks more, that should still be relatively quick.

On the other hand, creating an aligned snaphot might not be a cheap operation. Doing it too often might use more CPU cycles and DB bandwidth than we want.

jakmeier (commented):
I broadly went through most of the changes. But didn't go too much into details, as I first need to understand the new model that you suggested over at #499 first. But I will submit the comments I have so far.

But yeah, the per-chain checkpoint generation in this PR looks neat and much simpler than I thought it will have to be. Nice!

I'm just confused why we have any changes to the contract in this PR. If a temporary solution is needed, I'd rather not do a temporary on-chain contract migration as part of it.

ChaoticTempest:
> I'm just confused why we have any changes to the contract in this PR. If a temporary solution is needed, I'd rather not do a temporary on-chain contract migration as part of it.

The contract provides `latest_checkpoint` which is what `Checkpoint::recover` uses to restore. Let me just add store to redis then to not have contract migration later. That probably will save us the complication on dev or even worse if it doesn't get done by the next release...

chain-signatures/contract/src/lib.rs
ChaoticTempest: we'd still need to vote for the checkpoint hash on the contract though unless there was something else that we could use for consensus?

chain-signatures/node/src/backlog.rs
ChaoticTempest: yeah I can up it for ethereum then. It was just arbitrary what I had

chain-signatures/contract/src/lib.rs
ChaoticTempest: actually, frequently voting for a checkpoint is a bit bad since that will run into RPC limits we have.

So instead, what we can do is a p2p consensus on the checkpoint when the node restarts. For example, on solana when we start up the node:
- we need to ensure we have at least threshold checkpoints (i.e. wait till at least threshold nodes are online and active). But this makes it so that indexers can only be ran when threshold nodes are active which is not bad either ways since signing can't function without nodes anyways.
- query all active nodes on `/checkpoint?chain=Solana`
- sort these checkpoints by block height from highest to lowest.
- take the threshold lowest checkpoint:
   ```py
   threshold = 4
   checkpoints = [10, 9, 9, 8, 6, 5]
   taken = checkpoints[threshold - 1] 
   taken == 8
   ```
 - I wonder if we have a confirmation step here where we confirm with the rest of the nodes if the `taken` checkpoint is indeed the same across the board.
   - or maybe we don't need a confirmation step if the `/checkpoint` query returns a list of checkpoints in the last ~30mins or so.
   - then we can form a consensus ourselves based on that without much agreement.
   - since the indexers determine modification of pending requests. The checkpoints _should_ all be agreeing but maybe there's some hiccups we're not accounting for in indexers.
   - for this PR at least, I'll ignore this confirmation part because it's a bit complicated

I think that would be the simplest for now without operational expense. 

On a side note, since this requires threshold nodes to be active to start the indexer, this actually makes catchup less of a thing because then we're relatively close to the head of the chain assuming that other nodes are relatively close as well

ChaoticTempest: Implemented this but did not add the confirmation step

chain-signatures/node/src/backlog.rs
ChaoticTempest: bumped to 20 blocks for ethereum and 120 for solana. Tests can change these through environment variables

chain-signatures/primitives/src/lib.rs
```diff
@@ -103,3 +104,139 @@ impl Signature {
...
+
+        Some(interval)
+    }
+
+    pub fn checkpoint_env_vars() -> Vec<(&'static str, &'static str)> {
+        vec![
+            ("CHECKPOINT_INTERVAL_ETHEREUM", "2"),
+            ("CHECKPOINT_INTERVAL_SOLANA", "5"),
+        ]
+    }
```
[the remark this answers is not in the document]
ChaoticTempest: This is purely for testing purposes

chain-signatures/node/src/backlog.rs
```diff
@@ -229,12 +414,177 @@ impl Backlog {
...
+        let interval = chain.checkpoint_interval();
+        tracing::trace!(
+            ?chain,
+            height,
+            ?interval,
+            "backlog updated processed block height"
+        );
+
+        // create a checkpoint on interval
+        if height.is_multiple_of(interval?) {
```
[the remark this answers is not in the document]
ChaoticTempest: this is wrong, `is_multiple_of` is a standard library function and is preferred over doing % num due to num panicking on zero

chain-signatures/node/src/backlog.rs
```diff
@@ -283,23 +704,79 @@ mod tests {
...
+                    caip2_id: "eip155:1".to_string(),
+                    key_version: 0,
+                    deposit: 0,
+                    path: "".to_string(),
+                    algo: "".to_string(),
+                    params: "".to_string(),
+                    program_id,
+                    output_deserialization_schema: vec![],
+                    respond_serialization_schema: vec![],
+                }),
```
jakmeier, line 736: nitpick on style: It would be slightly easier to read (and less code duplication) to use a common constructor and a spread operator, something like:

```suggestion
                SignRequestType::SignBidirectional(SignBidirectionalEvent {
                    dest: "ethereum".to_string(),
                    caip2_id: "eip155:1".to_string(),
                    program_id,
                    ..SignBidirectionalEvent::empty()
                }),
```

Instead of `::empty()`, you could potentially even use `#[derive(Default)]` on `SignBidirectionalEvent` but personally I would give the constructor a different name, since the result isn't a valid default event.

chain-signatures/node/src/checkpoint_consensus.rs
```diff
@@ -0,0 +1,240 @@
...
+    mesh_state: &MeshState,
+    node_client: &NodeClient,
+    threshold: usize,
+    chains: &[Chain],
+) -> HashMap<Chain, Checkpoint> {
+    if threshold == 0 {
+        tracing::warn!("threshold must be greater than 0");
+        return HashMap::new();
+    }
+    if mesh_state.active.participants.is_empty() {
```
jakmeier, line 38: should we extend this check to `participants.len() < threshold`?

chain-signatures/node/src/checkpoint_consensus.rs
```diff
@@ -0,0 +1,240 @@
...
+            Err(err) => {
+                tracing::warn!(
+                    participant = ?participant_id,
+                    url = %info.url,
+                    %err,
+                    "failed to query participant for checkpoints"
+                );
+            }
+        }
+    }
```
jakmeier, line 120: Couldn't this be done in parallel?

chain-signatures/node/src/checkpoint_consensus.rs
```diff
@@ -0,0 +1,240 @@
+//! Node-based checkpoint consensus
+//!
+//! This module implements checkpoint recovery through node consensus rather than
+//! querying the NEAR contract. Nodes query each other's `/checkpoint` endpoints,
+//! collect checkpoints for each chain, and reach consensus on the threshold-lowest
+//! block height checkpoint.
```
jakmeier, line 6: First, I love the use of module level comments for documentation!

But second, I don't think `consensus` is the right term to use here.

The algorithm and its implementation are good. But it doesn't solve the [consensus problem](https://en.wikipedia.org/wiki/Consensus_(computer_science)) in a strict sense. There aren't multiple valid data values, among which the nodes collective decide on one. There is also no guarantee that all nodes reach the same conclusion (a.k.a. consensus) on what the best checkpoint is right now.

Rather, in this code, a node collects a series of valid checkpoints and decides locally which one they want to use for recovery. No shared decision making is involved. So, while I don't care about the specific name, maybe call it `checkpoint_selection` or something like that, please avoid the term `consensus`, as this is a technical term that doesn't fit here.

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -477,12 +492,24 @@ pub async fn run(
...
+    mesh_state: watch::Receiver<MeshState>,
+    node_client: NodeClient,
 ) {
     let Some(eth) = eth else {
         tracing::warn!("ethereum indexer is disabled");
         return;
     };
 
+    // Wait for threshold to be available
+    let threshold = contract_watcher.wait_threshold().await;
```
jakmeier, line 505: At first glance, I thought this means we wait for a threshold of nodes to be available. But it seems we only wait for the config value to be available, right? Why do we need to block on that? And could you make the comment clearer?

chain-signatures/node/src/indexer_eth.rs
```diff
@@ -477,12 +492,24 @@ pub async fn run(
...
+    if threshold > 0 {
+        let mesh_state = mesh_state.borrow().clone();
+        backlog
+            .recover(&mesh_state, &node_client, threshold, &[Chain::Ethereum])
+            .await;
+    }
+
```
jakmeier, line 512: Not necessarily for this PR, but I wonder, does it make sense to unconditionally try and recover from remote-fetched checkpoint?

What if we do a quick reboot, wouldn't we have local data that's more recent? Maybe we can check how far we are behind the latest block height, to see if we even need to catch up with remote checkpoints?

jakmeier (approved):
Looks pretty good! Left some comments to discuss

chain-signatures/contract/src/lib.rs
jakmeier: We don't _need_ consensus on what is the latest checkpoint. All we need is a valid checkpoint from somewhere when we reboot and a way to check it's correct.

So yeah, what you describe makes sense and I like it. It might not be BFT. But it helps us under honest conditions to ensure we don't lose any requests.

To make it BFT later, we still don't need consensus. Since we make checkpoints at fixed intervals and everything is deterministic, nodes are guaranteed to reach the same results. Any node suggesting a different checkpoint is incompatible and maybe malicious. Eitherway, we ignore those and only operate with nodes that reach the same correct result as us.

What we discussed originally was that the MPC network can then self-sign these checkpoints. We may not even need to wait for a threshold of nodes to reach the checkpoint, we can just start signing it and every node only participates after it has reached the same height and has generated the same checkpoint. Finally, rebooting nodes can query a signed checkpoint from a single peer and verify its validity efficiently.

None of this should require any changes to the contract if I'm not mistaken.


> The checkpoints should all be agreeing but maybe there's some hiccups we're not accounting for in indexers.

Well, that would be a bug and may lead to all sorts of problems. At a very fundamental level, we need to ensure all nodes process the same requests from indexers. This is why we need to wait for finality and check for correctness of the provided blocks.

If we really wanted to account for differing inputs to each node, then yeah we would need to do consensus on the indexer outputs. Which is a much more general problem to solve. Bu I don't think it makes sense to catch such cases in checkpoint selection specifically.

chain-signatures/node/src/checkpoint_consensus.rs
ChaoticTempest: I agree, just went with consensus since I didn't have a better name here. Maybe something like checkpoint bootstrap selection would be a better name then?

chain-signatures/node/src/indexer_eth.rs
ChaoticTempest: sounds good, yeah this isn't the clearest. The reason why we block is because when the indexer spins up, the contract is still not necessarily initialized yet, so we need to wait for it to be available (mainly because RPC hasn't polled it's state yet). 

Since there's a bunch of other stuff that wants to wait as well, let me just add something like `wait_available()` then we can do `watcher.threshold()` to make it consistent

chain-signatures/node/src/indexer_eth.rs
ChaoticTempest: yep, I was planning on adding persistence in another PR so we can do something like this

ChaoticTempest:
I'll just implement all these fixes in another PR since this isn't blocking anything and would be nice if all this checkpoint stuff was in so I don't conflict with myself in depending on the changes of this PR. So going to merge this one

chain-signatures/node/src/checkpoint_consensus.rs
jakmeier: Yeah I think calling it checkpoint bootstrapping works well

PR #588, opened 2025-11-06
feat: improve integration test times

ChaoticTempest:
This makes our integration tests more performant.

- reduces the querying wait times since some were pretty big (sign had a 20s interval between queries).
- add in `PregeneratedKeys` which takes @jakmeier work on mpc fixture over for integration tests. This reduces our setup time by about 20s which is pretty good.
- our `test_signature_basic` went from 55-60s to now 30s from my measurement.


ChaoticTempest:
full integration test run went from 38-42mins to now 26mins

jakmeier (approved):
Wow, great improvement! Thanks for connecting the pregenerated data with integration tests.

volovyks:
Impressive!

PR #589, opened 2025-11-06
feat: triple pairs

ChaoticTempest:
cait sith has `generate_triple_many` which allows us to generate multiple triples at once. We switch to generating triple pairs in this PR.

By generating triple pairs, we have the same participants set for both, and thus we do not need to line up participants of similar intersection when generating the presignature.

This ends up simplifying the triple process a bit and leads to less trashed triples.

### DIFF
--- a/chain-signatures/node/src/storage/triple_storage.rs
+++ b/chain-signatures/node/src/storage/triple_storage.rs
@@ -7,6 +7,7 @@ use cait_sith::protocol::Participant;
 use chrono::Duration;
 use deadpool_redis::{Connection, Pool};
 use redis::{AsyncCommands, FromRedisValue, RedisError, RedisWrite, ToRedisArgs};
+use serde::{Deserialize, Serialize};
 
 use near_account_id::AccountId;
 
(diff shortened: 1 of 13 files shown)


chain-signatures/node/src/storage/triple_storage.rs
```diff
@@ -505,81 +523,70 @@ impl TripleStorage {
...
-            -- pop two triples from the self owner set and delete them once successfully fetched
-            local triple_ids = redis.call("SPOP", mine_key, 2)
-            local triples = redis.call("HMGET", triple_key, unpack(triple_ids))
-            if not triples[1] then
-                return {err = "WARN unexpected, triple " .. triple_ids[1] .. " is missing"}
-            end
-            if not triples[2] then
-                return {err = "WARN unexpected, triple " .. triple_ids[2] .. " is missing"}
+            -- pop one pair from the self owner set and delete it once successfully fetched
+            local pair_ids = redis.call("SPOP", mine_key, 1)
```
jakmeier, line 569: Maybe remove the count argument? Then it returns a single value rather than an array, so indexing later (`[1]`) can also be removed.

integration-tests/src/mpc_fixture/builder.rs
```diff
@@ -517,9 +520,23 @@ impl MpcFixtureNodeBuilder {
...
+                            share: pair[0].share.clone(),
+                            public: pair[0].public.clone(),
+                        },
+                        triple1: Triple {
+                            share: pair[1].share.clone(),
+                            public: pair[1].public.clone(),
+                        },
+                    };
+                    let mut slot = triple_storage.reserve(pair_id).await.unwrap();
+                    slot.insert(pair, owner).await;
```
jakmeier, line 539: This works and is ok for this PR.

But probably we should just regenerate the output and change the fixture to store pairs, too. I really want to keep the fixture as close as possible to a dump of the real data as stored in the real DB.

integration-tests/src/mpc_fixture/input.rs
```diff
@@ -1,15 +1,33 @@
...
 use mpc_node::protocol::state::NodeKeyInfo;
 use mpc_node::protocol::triple::Triple;
 use std::collections::BTreeMap;
 
+#[derive(serde::Deserialize, serde::Serialize)]
+pub struct FixtureTriple {
+    pub id: u64,
+    pub share: TripleShare<Secp256k1>,
+    pub public: TriplePub<Secp256k1>,
+}
```
jakmeier, line 14: Similar here, I think we should avoid having a new view of triples just for tests.

jakmeier (approved):
A triple pair, wouldn't that be sextuple? :D

Anyway, it seems like a good idea to keep them stored as pairs of triple, since that's how we always use them anyway.

A thought for future improvements based on this: I wonder if we can simplify our redis calls after this. Taking a pair is inherently atomic, so maybe instead of LUA scripts, we can rely more on single operations such as  `LMOVE` to move atomically from available to used triples.

ChaoticTempest:
I'll take a look at LMOVE and do all the fixes with the fixture in a follow up PR

volovyks (commented):
@ChaoticTempest  Please, fix the metrics in the follow up. Since we are using and storing pairs of triples now, len and len mine are not returning the expected value. We can also start operation git len of pairs, but that requires changes on Grafana.

ChaoticTempest:
showing the metrics be a multiple of 2 isn't very helpful. We should just rename the all the triple stuff on grafana if we want to bother with it otherwise just assume it's a triple pair from now on

PR #590, opened 2025-11-07
feat: compatibility testing against testnet and mainnet

ChaoticTempest:
Adds in compatibility based testing against our mainnet and testnet environments.

I've noticed how easy it was for things to not work after making deployments to our dev environment where old nodes that haven't upgraded yet would then run into problems.

We aren't easily catching this so I added this to our testing as follows. Spin up the latest {testnet, mainnet} node and our latest code node. Set threshold to 2 and have them sign something. If it doesn't work then they're not compatible and that we should be able to transition properly.

Underneath all this, we store the binaries in `artifacts/{mainnet, testnet}/mpc-node.{semver}`. Might be an issue later if the binaries are too big for git but we'll handle that when we get there.

These tests aren't being ran as a part of our CI for now since https://github.com/sig-net/mpc/pull/589 is still breaking some things. In the future, we can enable testnet for CI and get that working. Once that is working as intended, we can enable mainnet.


integration-tests/src/lib.rs
```diff
@@ -525,7 +549,11 @@ pub async fn host(spawner: &mut ClusterSpawner) -> anyhow::Result<Nodes> {
     let node_futures = spawner
         .accounts
         .iter()
-        .map(|account| local::Node::run(&ctx, cfg, account));
+        .zip(std::mem::take(&mut spawner.node_binary_sources).into_iter())
+        .map(|(account, source)| {
+            let binary_path = source.binary_path().unwrap();
```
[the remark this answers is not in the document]
ChaoticTempest: this function doesn't return a anyhow result but a future

volovyks (commented):
Nice, we really need this.

The binary is ~50Mb, and it is in git history forever.
Also, we will need to update it on each release, which is not convenient.

We already have a repository where we push all the releases. We can use it instead. Pull the latest version for testnet and mainnet and test against them.

ChaoticTempest:
@volovyks which repo is that? I don't see it

volovyks:
Probably "container registry" is the right term. @auto-mausx can you please help Phuong here?

ChaoticTempest:
oh the container registry, yeah that's not gonna work for us. That only contains docker images, not the actual binaries this would use. Wouldn't work well on our mac machines either since they're built for linux only

volovyks:
Hm, true
we have not touched them in a while, but we have Dockerised integration tests. Is that an option?

ChaoticTempest:
the reason why we directly run the node binaries is because of performance (especially on CI). And even if we make the nodes run again via docker, we'd still hit issues with linux only builds since that's a whole mess to setup

volovyks:
We can also build the older version based on a Git tag.
- list all tags
- checkout the latest one that follows the release standard (v1.2.3)
- build this version
- checkout latest code
- run integration tests
Another option is to create a registry for binaries.
I really do not want to keep binaries in our git history.

ChaoticTempest:
We'll probably need something like git lfs. It's a lot simpler than just checking out a version and compiling it each time which is a lot of either manually build or a setup script that would be a bit hard to maintain. The manual way no one's going to do

jakmeier:
Really good thinking, exactly the kind of testing we lack!!

Regarding the implementation, +1 from me on not storing binaries in the git repo. Even if we use git LFS to alleviate the size problem, I'm still very much against it.

Binary artifacts should be generated on demand, for the machine using it. Downloading a binary from a git repository and running it locally requires another level of trust into anyone touching these binaries. And it's super fragile if not the entire team uses the same OS. (I'm not on a Mac :P)

Have you considered setting it up using github actions? Building an old version in a separate step shouldn't be drastically more complicated than just building the current one. Throwing this idea at an LLM, I get what seems like a reasonably simple and maintainable solution.

ChaoticTempest:
yeah I can see it becoming an issue, but I just hate the idea of adding another build script to what we have right now. If we can keep it manageable then i'll be fine with it. We can keep it purely for GH actions right now, but testing locally will be a pain

ChaoticTempest:
ok, this now has a GH actions pipeline that does the following with some bash scripts that builds prod versions via git worktrees. So long as your env supports bash, this can be done. Calling the build script won't trigger a rebuild since they check if the binaries exist

scripts/build-compat-binaries.sh
```diff
@@ -0,0 +1,88 @@
...
+
+versions_path = pathlib.Path(sys.argv[1])
+channel = sys.argv[2]
+
+data = json.loads(versions_path.read_text())
+if channel not in data:
+    raise SystemExit(f"Unknown compatibility channel '{channel}'.")
+print(data[channel])
+PY
+}
```
jakmeier, line 38: I don't love running python from bash like this but I guess it works. :shrug: 

I'd usually use `jq` for this, or just write it all in python or maybe even in rust with https://github.com/matklad/xshell/ for bash-like commands.

But IMO it's fine to keep as is.

jakmeier (dismissed):
Looks good. I don't think the newly added bash build script is too bad.

Originally, I was thinking of building the older version in a GitHub action step and then running this locally with `act` (https://github.com/nektos/act).

But I like the bash script and the usage of git worktree!

scripts/build-compat-binaries.sh
ChaoticTempest: yep it's quite dirty. I might just change it over to jq

ChaoticTempest:
we can keep the production compatibility workflow failing for now. It should eventually pass

jakmeier (approved):

PR #592, opened 2025-11-11
fix: parse sol events

ppca:


ChaoticTempest:
hmm, how is this fixing the sol event parsing? Looks like we're only changing the event macro import so far?

ppca:
> hmm, how is this fixing the sol event parsing? Looks like we're only changing the event macro import so far?

Honestly I'm not sure either. But tried out on dev, it's parsing sign events correctly 🤷‍♀️

ChaoticTempest:
hmm, was this with sending some transactions yourself to our devnet and you noticed that for the old contract code, it wouldn't be picked up at all by the indexer?

ppca:
> hmm, was this with sending some transactions yourself to our devnet and you noticed that for the old contract code, it wouldn't be picked up at all by the indexer?

no, this is after we deployed the latest version of solana contract to the new contract address, dev and testnet cannot pick up solana events from the new address. 
Initially it was Signet.js and contract pinger not correctly setup, so I asked Felipe to help correct them. He corrected and sent some events from his local contract-pinger. 
Then I see events coming in from logs on dev but ```Failed to deserialize SignatureRequestedEvent: Unexpected length of input```. However I checked on solscan, the events Felipe sent has the right instruction data ,i.e. starts with the right event tag, and parses to the fields of the event correctly. 
Then in the process of trying to print some logs on dev, I included this change. Not sure if this is related to event macro.

ChaoticTempest:
so is contract pinger working right now? Since I see some events go through in the last day, but it just stops there

ppca:
> so is contract pinger working right now? Since I see some events go through in the last day, but it just stops there

as of today it is working again

ChaoticTempest:
I think I know the issue, it's because the events that was in indexer_sol was different than the events in `signet_program`. So we had them defined multiple times. My latest changes via the checkpoints have them corrected, so we should be good for now

ppca:
I'll close this PR then.

PR #594, opened 2025-11-13
feat: Singular task SignTask for organizing + posit + generation

### LINKED ISSUE #559: Singular task for all protocol tasks
Currently, we have posits that are being done in spawners. This is troublesome to have in spawners, since they conflict with how the spawners need to act immediately, whereas posits require waiting for resources potentially (presignatures deliberators need two triples, signature deliberators need sign request and presignature).

The way we can solve this simply is to move all the posit logic into the generator task which is already being done async. It won't be called generator anymore in this case, maybe `TripleTask`, `PresignatureTask`, `SignatureTask` or we can generalize and call it `ProtocolTask`.

The task flow for this would simply look like:

### proposer
- propose posit
- generation
- store or publish

### deliberator
- deliberate posit:
  - wait if needed (presignature ready, sign request ready)
  - accept if on ready in time, reject otherwise
- generation
- store

## Bidirectional Communication

This consolidation should also be done with the full signature flow with bidirectional communication too.

### proposer
- propose posit
- generation
- publish signature
- wait from indexer: new sign request on tx execution
  - new task can be started immediately from here if restored from checkpoint
- note: new proposer will be elected here based on the sign request
- propose posit
- generation
- publish signature

### deliberator
- deliberate posit:
  - wait for sign request and presignature ready
  - accept if on ready in time, reject otherwise
- generation
- wait from indexer: new sign request on tx execution
  - new task can be started immediately from here if restored from checkpoint
- deliberate posit
- generation

As you can see with the bidirectional flow, we do `posit`, `generate`, `publish` sequence twice. So we can just simply loop over it too for this task. But the idea is to keep this task around so we can track it more easily.

ChaoticTempest:
This resolves https://github.com/sig-net/mpc/issues/559 partially.

This makes signature organizing, positing and generating in signatures become a singular `SignTask` which goes through each of these phases sequentially and backtracks to organizing if something goes wrong.

No more `SignQueue` with this change. `SignSpawner` is now responsible for receiving messages `SignQueue` once processing, while the organizing that `SignQueue` did got moved to the `SignOrganizer` aka the organizing phase of `SignTask`.

Also no more `get_or_pending`. How a `SignTask` gets created is a one to one mapping with an `IndexedSignRequest`. Once we receive a request from indexer, we will immediately create the task. The task will then do all the work necessary for that request.

This is pretty flexible, so that we can later have the `SignTask` be a watchdog on timeout if the request never gets completed. This is one thing that is present in the previous version as well, we never tell the MPC signing that the request got completed via indexer, so it may try to reorganize and posit over and over again. This was and still is an issue, but will be resolved after checkpoints when we can observe that the respond/respond-bidirectional request made it, and thus we can go and cancel the `SignTask` at this point.

### DIFF
--- a/chain-signatures/node/src/protocol/posit.rs
+++ b/chain-signatures/node/src/protocol/posit.rs
@@ -354,6 +354,54 @@ impl<Id: Copy + Hash + Eq + fmt::Debug, S> Posits<Id, S> {
     }
 }
 
+/// A single posit counter that tracks participants accepting/rejecting a proposal.
+/// This is used by individual signature tasks instead of the global Posits mapping.
+pub struct SinglePositCounter {
+    participants: HashSet<Participant>,
+    rejects: HashSet<Participant>,
+    pub accepts: HashSet<Participant>,
+}
+
+impl SinglePositCounter {
+    pub fn new(me: Participant, participants: &[Participant]) -> Self {
+        let mut accepts = HashSet::new();
+        accepts.insert(me);
+        Self {
+            participants: participants.iter().copied().collect(),
+            rejects: HashSet::new(),
+            accepts,
+        }
+    }
+
+    pub fn enough_accepts(&self, threshold: usize) -> bool {
+        self.accepts.len() >= threshold
+    }
+
+    pub fn enough_rejects(&self, threshold: usize) -> bool {
+        self.rejects.len() > self.participants.len() - threshold
+    }
+
+    pub fn meets_totality(&self) -> bool {
+        self.accepts.len() + self.rejects.len() == self.participants.len()
+    }
+
+    pub fn process_action(&mut self, from: Participant, action: &PositAction) -> bool {
+        if !self.participants.contains(&from) {
+            return false;
+        }
+        match action {
+            PositAction::Accept => {
+                self.accepts.insert(from);
+            }
+            PositAction::Reject => {
+                self.rejects.insert(from);
+            }
+            _ => return false,
+        }
+        true
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/chain-signatures/node/src/protocol/signature.rs
+++ b/chain-signatures/node/src/protocol/signature.rs
@@ -336,7 +336,6 @@ impl SignPositor {
 
                     break *presignature_id;
                 } else {
-                    // reject
                     tracing::warn!(
                         ?sign_id,
                         ?from,
@@ -408,9 +407,6 @@ impl SignPositor {
         let is_proposer = proposer == ctx.me;
         let is_deliberator = !is_proposer;
 
-        // GUARANTEE: at least threshold participants from organizing phase.
-        let posit_participants = stable.iter().copied().collect::<Vec<_>>();
-
         // Get the presignature participants - only these nodes participated in generating it
         let presignature_participants = if let Some(ref taken) = presignature {
             taken.presignature.participants.clone()
@@ -441,6 +437,8 @@ impl SignPositor {
             }
         }
 
+        // GUARANTEE: at least threshold participants from organizing phase.
+        let posit_participants = stable.iter().copied().collect::<Vec<_>>();
         let mut counter = SinglePositCounter::new(ctx.me, &posit_participants);
 
         let posit_timeout = Duration::from_secs(60);
@@ -1102,8 +1100,11 @@ impl SignatureSpawner {
             Sign::Request(indexed) => {
                 let sign_id = indexed.id;
 
-                // Skip if we've already seen this request
-                if self.inboxes.contains_key(&sign_id) {
+                // Skip if we already have a task handling this request.
+                // Use tasks instead of inbox map since it may already contain buffered messages
+                // (e.g. a Propose arriving before the indexer notifies us), so we must only look
+                // at the task map to decide whether the request is truly a duplicate.
+                if self.tasks.contains_key(&sign_id) {
                     tracing::info!(?sign_id, "skipping duplicate sign request");
                     return;
                 }
--- a/chain-signatures/node/src/protocol/signature.rs
+++ b/chain-signatures/node/src/protocol/signature.rs
@@ -3,7 +3,9 @@ use crate::backlog::Backlog;
 use crate::config::Config;
 use crate::kdf::derive_delta;
 use crate::mesh::MeshState;
-use crate::protocol::message::{MessageChannel, PositMessage, PositProtocolId, SignatureMessage};
+use crate::protocol::message::{
+    MessageChannel, PositMessage, PositProtocolId, SignatureMessage, Subscriber,
+};
 use crate::protocol::posit::{PositAction, SinglePositCounter};
 use crate::protocol::presignature::PresignatureId;
 use crate::protocol::Chain;
@@ -1013,11 +1015,9 @@ impl SignTask {
                 }
             }
 
-            match phase.advance(&self, &mut state, &mut task_rx).await {
+            phase = match phase.advance(&self, &mut state, &mut task_rx).await {
                 SignPhase::Complete(result) => return result,
-                other => {
-                    phase = other;
-                }
+                other => other,
             }
         }
     }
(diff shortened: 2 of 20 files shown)


chain-signatures/node/src/protocol/posit.rs
```diff
@@ -354,6 +354,54 @@ impl<Id: Copy + Hash + Eq + fmt::Debug, S> Posits<Id, S> {
...
+pub struct SinglePositCounter {
+    participants: HashSet<Participant>,
+    rejects: HashSet<Participant>,
+    pub accepts: HashSet<Participant>,
+}
+
+impl SinglePositCounter {
+    pub fn new(me: Participant, participants: &[Participant]) -> Self {
+        let mut accepts = HashSet::new();
+        accepts.insert(me);
```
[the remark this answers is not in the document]
ChaoticTempest: only use counter in if is_proposer, but definitely needs cleanup

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -787,228 +999,96 @@ impl SignatureSpawner {
...
+                })
+                .await;
+        } else {
+            // No task yet - buffer message for when task is created
+            tracing::trace!(?sign_id, ?from, ?action, "buffering message - no task yet");
+            self.pending_posit_messages
+                .entry(sign_id)
+                .or_default()
+                .push((presignature_id, from, action));
+        }
```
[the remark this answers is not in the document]
ChaoticTempest: need to cleanup later with a TTL for each sign id of signature timeout, since we do a new posit later, but will leave it up for another PR

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -1019,137 +1099,104 @@ impl SignatureSpawner {
...
+                    ?sign_id,
+                    ?elapsed,
+                    "sign request already timed out, skipping"
                 );
-                retry.push(my_request.indexed.id);
                 continue;
             }
 
-            self.propose_posit(&my_request, taken, &participants).await;
+            self.seen_requests.insert(sign_id);
```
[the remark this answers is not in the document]
ChaoticTempest: this is being removed on tasks completion

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -104,323 +58,552 @@ pub struct SignRequest {
...
-                // if we are unable to find the corresponding request, we can remove it.
-                return false;
-            };
-            crate::util::duration_between_unix(
-                request.indexed.unix_timestamp_indexed,
-                crate::util::current_unix_timestamp(),
-            ) < request.indexed.total_timeout
-        });
-    }
+        }
```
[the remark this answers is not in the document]
ChaoticTempest: there's really only ever accept, reject, propose, start cmds for posits. So this won't affect anything since deliberators will only look for propose and start. To receive a start before a propose means something is wrong. The out of order messages isn't an issue here either, because to get a start, the deliberator must have seen a propose first so no out of ordering can happen

chain-signatures/node/src/protocol/triple_serde_test.rs
```diff
@@ -0,0 +1,59 @@
...
+        let share = TripleShare::<Secp256k1> {
+            a: <k256::Scalar as Field>::ZERO,
+            b: <k256::Scalar as Field>::ONE,
+            c: <k256::Scalar as Field>::ZERO,
+        };
+
+        // Serialize it
+        let json = serde_json::to_string_pretty(&share).unwrap();
+        println!("Serialized TripleShare:");
+        println!("{}", json);
```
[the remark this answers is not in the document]
ChaoticTempest: deleting triple_serde_test.rs. That's not supposed to be committed

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -104,323 +58,583 @@ pub struct SignRequest {
...
+        tracing::info!(?sign_id, round = ?state.round, "entering organizing phase");
+        let (stable, proposer) = {
+            let stable = self.wait_stable(ctx, state, threshold).await;
+            if stable.is_empty() {
+                tracing::warn!(?sign_id, round = ?state.round, "no stable participants, reorganizing");
+                state.bump_round();
+                return SignPhase::Organizing(self);
+            }
+
+            let max_rounds = state.round + 512;
```
jakmeier: nit: maybe make 512 a constant defined at the file top

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -1019,137 +1130,90 @@ impl SignatureSpawner {
...
-    ) {
-        // NOTE: signatures should only use stable and not active participants. The difference here is that
-        // stable participants utilizes more than the online status of a node, such as whether or not their
-        // block height is up to date, such that they too can process signature requests. If they cannot
-        // then they are considered unstable and should not be a part of signature generation this round.
+        // Update metrics
+        crate::metrics::SIGN_QUEUE_SIZE
+            .with_label_values(&[self.my_account_id.as_str()])
+            .set(self.seen_requests.len() as i64);
+    }
```
jakmeier, line 1125: nit: the field name `seen_requests` is confusing, without digging deeper I would assume it is a ever growing list of all seen requests ever. Can we call it something like `requests_in_progress`?

integration-tests/src/mpc_fixture/builder.rs
```diff
@@ -416,9 +416,9 @@ impl MpcFixtureNodeBuilder {
         let presignature_storage = storage.presignature_storage.clone();
 
         // prepare all channels for the node
-        let (sign_tx, sign_rx) = SignQueue::channel();
+        let (sign_tx, sign_rx) = mpsc::channel(1024);
         const MAX_CONCURRENT_RPC_REQUESTS: usize = 1024;
-        let (rpc_tx, rpc_rx) = tokio::sync::mpsc::channel(MAX_CONCURRENT_RPC_REQUESTS);
+        let (rpc_tx, rpc_rx) = mpsc::channel(MAX_CONCURRENT_RPC_REQUESTS);
```
jakmeier, line 422: nit: for consistency, maybe make both numbers a constant

jakmeier (dismissed):
I really like the direction of this!

Regarding the number of possible concurrently ongoing signature requests. Did I get this right, we haven't limited it in the past and we still don't limit it now? If so, do you think we should limit it in the future?

chain-signatures/node/src/protocol/signature.rs
ChaoticTempest: Seen request was being removed on completion of the request, but I  removed it in favor of `inboxes: HashMap<SignId, Subscriber>` which buffers the posit messages which can also be used to see if we have the request ongoing or not.

ChaoticTempest:
I also made each indexer send `Sign::Completion(sign_id)` alongside `Sign::Request(request)` so that completions can abort currently running sign tasks that are acting as a watchdogs to constantly retry if the proposer was not successful

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -684,419 +913,267 @@ impl Drop for SignatureGenerator {
...
 
-    async fn handle_requests(
+    fn handle_completion(&mut self, sign_id: SignId) {
+        self.inboxes.remove(&sign_id);
+        if self.tasks.abort(sign_id) {
+            tracing::info!(?sign_id, "aborting signature task due to completion event");
+        } else {
+            tracing::info!(?sign_id, "task already completed or unable to be aborted");
+        }
+    }
```
jakmeier, line 1087: Any peer can send these at any time, right? Doesn't this make it incredibly easy for a single node to censor a signature request? It also breaks the "requests are always answered" guarantee.

I believe we should follow the rule to only trust verified on-chain observations for tracking which requests have been served.

jakmeier (commented):
> I also made each indexer send Sign::Completion(sign_id) alongside Sign::Request(request) so that completions can abort currently running sign tasks that are acting as a watchdogs to constantly retry if the proposer was not successful

I have some doubts on this addition, please see my comment.

chain-signatures/node/src/protocol/signature.rs
ChaoticTempest: no it's the not peer that sends these. It's the indexer with respond/respond-bidirectional events that sends this. So still true to the on-chain observation when requests are completed

jakmeier: Oh I see, my mistake!

jakmeier (approved):

PR #598, opened 2025-11-19
chore: remove unnecessary deps in docker

ChaoticTempest:
for some reason we have protobuf in our docker builds that we never really use


volovyks (approved):
@auto-mausx check your notification setting please

auto-mausx (approved):

PR #600, opened 2025-11-19
eth indexer: refactor helios and direct rpc, fix sign bidirectional flow

### LINKED ISSUE #605: sign respond broken on dev and testnet
I tested it again with solana-contract-examples, it is broken. 
I think the problem is the sender and from_address of BidirectionalTx is wrong: https://github.com/sig-net/mpc/blob/852d7a3e2f4e719aa24f4f2f663693513c0de76d/chain-signatures/node/src/sign_bidirectional.rs#L51

The from_address is set to be the address of mpc signer, not the eth address of the sender.

ppca:
I've removed feature flag, and controlling it by an environment variable which defaults to false. This is easier for development as it makes sure the helios code path compiles whenever changes are made, while with feature flag, the helios path can be error prone as it won't be compiled at all.
both helios and direct rpc eth indexer use almost same code path, so any changes will apply to both.
also solves https://github.com/sig-net/mpc/issues/605


chain-signatures/node/src/indexer_eth/mod.rs
```diff
@@ -0,0 +1,1514 @@
...
+}
+
+#[derive(Clone)]
+pub enum EthereumClient {
+    Helios(indexer_eth_helios::HeliosEthereumClient),
+    DirectRpc(indexer_eth_direct_rpc::RpcEthereumClient),
+}
+
+#[async_trait]
+impl EthereumClientTrait for EthereumClient {
```
ChaoticTempest: We don't need `EthereumClientTrait` at all since `EthereumClient` is an enum. Just implement these functions directly on `EthereumClient` and get rid of the trait. This way we don't need `async_trait` which wraps the future in a Box

chain-signatures/node/src/indexer_eth/indexer_eth_helios.rs
```diff
@@ -0,0 +1,230 @@
...
+            .data_dir(PathBuf::from(&eth.helios_data_path))
+            .with_file_db()
+            .build()
+        {
+            Ok(client) => client,
+            Err(err) => {
+                return Err(anyhow::anyhow!("Failed to build Helios client: {err:?}"));
+            }
+        }
+    };
```
ChaoticTempest: just use anyhow context with these:
```suggestion
    let client = EthereumClientBuilder::new()
        .network(network)
        .consensus_rpc(&eth.consensus_rpc_http_url)
        .context("failed to build consensus rpc")?
        .execution_rpc(&eth.execution_rpc_http_url)
        .context("failed to build execution rpc")?
        .data_dir(PathBuf::from(&eth.helios_data_path))
        .with_file_db()
        .build()
        .context("failed to build helios client")?;
```

chain-signatures/node/src/cli.rs
```diff
@@ -352,7 +352,19 @@ pub async fn run(cmd: Cli) -> anyhow::Result<()> {
...
                 mesh_state.clone(),
                 client.clone(),
-            ));
+            )
+            .await
+            {
+                Ok(eth_indexer) => {
+                    tokio::spawn(async move {
+                        eth_indexer.run().await;
+                    });
```
ChaoticTempest:
```suggestion
                    tokio::spawn(eth_indexer.run());
```

chain-signatures/node/src/indexer_eth/mod.rs
```diff
@@ -0,0 +1,1514 @@
...
+
+    #[allow(clippy::too_many_arguments)]
+    pub async fn run(&self) {
+        let backlog = self.backlog.clone();
+        let contract_watcher = self.contract_watcher.clone();
+        let mesh_state = self.mesh_state.clone();
+        let node_client = self.node_client.clone();
+        let sign_tx = self.sign_tx.clone();
+        let app_data_storage = self.app_data_storage.clone();
+        let client = self.client.clone();
```
ChaoticTempest: - too_many_arguments is not needed
- run should be owning self other we end up with these clones


```suggestion
    pub async fn run(self) {
```

chain-signatures/node/src/indexer_eth/mod.rs
```diff
@@ -0,0 +1,1514 @@
...
+                backlog_clone2,
+                sign_tx_clone,
+            )
+            .await;
+        });
+
+        let blocks_to_process_send_clone = blocks_to_process_send.clone();
+        let client_clone = Arc::clone(&client);
+        if let Some(last_processed_block) = last_processed_block {
+            match Self::catchup_end_block_number(client_clone).await {
```
ChaoticTempest: nit: since this is in an if statement, only clone inside of it.

```suggestion
        if let Some(last_processed_block) = last_processed_block {
            match Self::catchup_end_block_number(Arc::clone(&client)).await {
```

chain-signatures/node/src/indexer_eth/mod.rs
```diff
@@ -0,0 +1,1514 @@
...
+                }
+            }
+        }
+
+        let blocks_to_process_send_clone = blocks_to_process_send.clone();
+        let client_clone = Arc::clone(&client);
+        tokio::spawn(async move {
+            tracing::info!("Spawned task to add new blocks to process");
+            Self::add_new_block_to_process(client_clone, blocks_to_process_send_clone).await
+        });
```
ChaoticTempest: we should put the initial log not inside of an async move block, but inside the functions directly so we can have clean spawn calls like so without polluting our local variable names

```suggestion
        tokio::spawn(Self::add_new_block_to_process(Arc::clone(&client), blocks_to_process_send.clone()));
```

chain-signatures/node/src/indexer_eth/indexer_eth_helios.rs
```diff
@@ -0,0 +1,230 @@
...
+                        e
+                    );
+                        tokio::time::sleep(delay).await;
+                        continue;
+                    }
+                    tracing::warn!(
+                    "Failed to fetch block number {block_id} from Helios client: {:?}, exceeded maximum retry",
+                    e
+                );
+                    return None;
```
ChaoticTempest: the log formatting is weird here

chain-signatures/node/src/indexer_eth/indexer_eth_helios.rs
```diff
@@ -0,0 +1,230 @@
...
+    }
+
+    async fn call(
+        &self,
+        from: Address,
+        to: Address,
+        data: Bytes,
+        block_number: u64,
+    ) -> anyhow::Result<Bytes> {
+        let helios_client = self.client.clone();
```
ChaoticTempest: why do we need a clone here on the client? All clients calls should be by reference

chain-signatures/node/src/indexer_eth/indexer_eth_direct_rpc.rs
```diff
@@ -0,0 +1,226 @@
...
+        block_id: alloy::rpc::types::BlockId,
+    ) -> anyhow::Result<Option<Vec<alloy::rpc::types::TransactionReceipt>>> {
+        self.rpc_call(
+            "eth_getBlockReceipts",
+            vec![json!(to_hex_block_id(block_id))],
+        )
+        .await
+    }
+
+    #[allow(unused)]
```
ChaoticTempest: shouldn't need this

chain-signatures/node/src/indexer_eth/mod.rs
```diff
@@ -0,0 +1,1467 @@
...
+    let event = parse_event(&log)?;
+    tracing::debug!("found eth event: {:?}", event);
+    if event.deposit == U256::ZERO {
+        tracing::warn!("deposit is 0, skipping sign request");
+        anyhow::bail!("deposit is 0");
+    }
+
+    if event.key_version > LATEST_MPC_KEY_VERSION {
+        tracing::warn!("unsupported key version: {}", event.key_version);
+        anyhow::bail!("unsupported key version");
```
ChaoticTempest: Let's make this return an option instead since we're already logging everything here

chain-signatures/node/src/indexer_eth/mod.rs
```diff
@@ -0,0 +1,1467 @@
...
+            }
+            crate::metrics::LATEST_BLOCK_NUMBER
+                .with_label_values(&[Chain::Ethereum.as_str(), node_near_account_id.as_str()])
+                .set(block_number as i64);
+        }
+    }
+}
+
+#[async_trait]
+impl EthereumIndexerTrait for EthereumIndexer {
```
ChaoticTempest: We don't need this trait either. Should be implemented directly on EthereumIndexer

chain-signatures/node/src/indexer_eth/indexer_eth_direct_rpc.rs
ppca: done

chain-signatures/node/src/indexer_eth/mod.rs
ppca: done

chain-signatures/node/src/indexer_eth/mod.rs
ppca: done

chain-signatures/node/src/indexer_eth/mod.rs
```diff
@@ -0,0 +1,1449 @@
...
+        let client = self.client;
+        let eth = self.eth;
+        let sign_tx = self.sign_tx;
+        let node_near_account_id = self.node_near_account_id;
+
+        Self::recover_backlog(&backlog, contract_watcher, mesh_state, node_client).await;
+
+        let last_processed_block = Self::get_last_processed_block(&app_data_storage).await;
+
+        let client = Arc::new(client);
```
ChaoticTempest, line 737: this shouldn't be an Arc. All the internal clients are cloneable. The reqwest client is internally reference counted, and you have an Arc over helios client anyways. So this Arc is useless

ChaoticTempest (dismissed):

chain-signatures/node/src/indexer_eth/indexer_eth_direct_rpc.rs
jakmeier: Looks like `#[allow(unused)]` is still here :thinking: Maybe you forgot to add / push?

jakmeier (dismissed):
Nice change! Feature flags are always annoying, even if justified sometimes.

small complaint on process (for the future): Big refactors and bug fixes really should be separate PRs. Not only is it much easier to review, it also becomes really hard to revert a specific change if it turns out that it didn't fix the problem.

volovyks (dismissed):

Pessina (approved):

PR #602, opened 2025-11-20
feat: protocol storage for triple/presignature storage

ChaoticTempest:
This makes triple and presignature storage a singular reusable struct. This was already done in https://github.com/sig-net/mpc/pull/372 but this one is more up to date especially with TriplePairs now.

This also makes component tests use TriplePairs and persist triple pairs instead of singular triples.


jakmeier (dismissed):
Nice!

ppca (approved):

PR #606, opened 2025-11-21
feat: add protocol versioning

ChaoticTempest:
Adds in protocol versioning to /status. Defaults PROTOCOL_VERSION to 1, since 0 is what we have mainnet.

### DIFF
--- a/chain-signatures/node/src/mesh/mod.rs
+++ b/chain-signatures/node/src/mesh/mod.rs
@@ -202,6 +202,23 @@ mod tests {
 
     const PING_INTERVAL: Duration = Duration::from_millis(10);
 
+    async fn expect_status(
+        watcher: &mut connection::ConnectionWatcher,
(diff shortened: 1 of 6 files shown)


chain-signatures/node/src/mesh/mod.rs
```diff
@@ -440,4 +457,32 @@ mod tests {
 
         mesh_task.abort();
     }
+
+    #[test(tokio::test)]
+    async fn test_protocol_version_mismatch_marks_offline() {
```
jakmeier, line 462: Nice test!

jakmeier (approved):
Makes sense!

I tried writing down the context around, as we discussed it on Slack, in an issue. Please take a look https://github.com/sig-net/mpc/issues/609

volovyks:
@ChaoticTempest thanks for looking into this!

PR #607, opened 2025-11-26
fix: sign task running out of presignatures

ChaoticTempest:
This should fix what is on devnet where the presignatures are being burnt too quickly. We put back the presignature into storage since the only thing that is communicated is what presignature ID to use, not anything inside of it during posit stage. So we should put it back if it doesn't meet the requirements.

Also adding back the intersection of stable & presignature.participants so that we only communicate with the participants that have the presignature.


ppca (dismissed):

chain-signatures/node/src/storage/protocol_storage.rs
```diff
@@ -679,6 +679,78 @@ impl<A: ProtocolArtifact> ProtocolStorage<A> {
         }
     }
 
+    /// Return a taken artifact back to the available pool.
+    pub async fn recycle_mine(&self, me: Participant, taken: ArtifactTaken<A>) -> bool {
```
jakmeier, line 683: Adding this code breaks an important architectural invariant we had so far: Triples and Presignatures only leave storage once

I understand this invariant was introduced before I joined to ensure that Ts and Ps are only used once.
Your change doesn't break that property, which is ultimately what counts.

But if we break the current invariant, what can we replace it with? Making sure the current change doesn't break the thing we are thinking about right now is easy. Making sure no future changes accidentally break things when we are not thinking about it, is hard. That's why being strict about invariants can be really useful.

Personally, I'd prefer a solution where we don't need this recycling mechanism.

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -220,20 +222,35 @@ impl SignOrganizer {
...
                 loop {
                     if let Some(taken) = ctx.presignatures.take_mine(ctx.me).await {
-                        break taken;
+                        let participants = intersect_vec(&[&taken.artifact.participants, &stable]);
+                        if participants.len() < ctx.threshold {
+                            recycle.push(taken);
+                            continue;
+                        }
+
+                        break (taken, participants);
```
jakmeier, line 238: Filtering out for stable participants only makes sense here.

But do we need to take the full P out of storage to make that check?

If we go back to the spec I wrote some time ago (https://github.com/sig-net/mpc/blob/develop/doc/mpc_node_specification.md#participant-selection)

> When a proposer selects a list of participants for a new protocol invocation, it should check these conditions:
>
>    - Participants must be online
>    - Participants must hold the relevant P and T shares

This is exactly what you are adding now, correct?

But I think we still lack the data structure suggested to make the second check.

> To enable the first, nodes keep a connection status for every peer. For the second part, they keep a directory of each owned P and T with a list of share holders. 

The list of share holders would be handy here to avoid taking out the triple. But it would also make the state sync actually useful, since we would use the up-to-date list of participants that still have the P or T, rather than the original list of participants when it was created.

Maybe it's time to finish the state sync implementation?

integration-tests/tests/cases/solana.rs
```diff
@@ -40,3 +40,45 @@ async fn test_solana_signature_basic() -> anyhow::Result<()> {
         anyhow::bail!("signature verification failed");
     }
 }
+
+// Concurrent variant: spawn many sign requests at once against a very small
+// presignature stockpile and assert we make forward progress (no livelock).
+#[test(tokio::test)]
+async fn test_solana_stockpile_depletion() -> anyhow::Result<()> {
```
jakmeier: Does this test actually test the case that was problematic? I would have expected some test that has Ps with a partial participant set, combined with offline nodes that are listed as participant.

ChaoticTempest: it does but it's not that explicit about it. I added tests in components instead and got rid of this test to showcase it further. The idea is to have double the amount of sign requests than we have a stockpile. The stockpile should build up which means that our pending sign requests should complete and not contend with each other. So for cases where the proposer isn't agreed upon or when we burn too many presignatures, this should catch it.

chain-signatures/node/src/storage/protocol_storage.rs
ChaoticTempest: yeah, putting it back is a bit hackish. One of my PRs that I never got back to was having a participant list right at the redis level where we can do intersections in redis instead of in node level. That could fix this but requires more work

chain-signatures/node/src/protocol/signature.rs
ChaoticTempest: yeah, this is just adding back in the intersection check. When the sign related code was consolidated into SignTask, this was deleted, so we now need to do this intersection check again.

It's a good idea to use the latest set based on state sync but it won't avoid this intersection check. If we do want to avoid it, we'll have to add in another priority queue based on participant size

ppca (approved):

chain-signatures/node/src/storage/protocol_storage.rs
volovyks: We will need to get back to this, I've created this issue not to forget about this change: https://github.com/sig-net/mpc/issues/611

chain-signatures/node/src/protocol/signature.rs
volovyks: I think we should trust data from state sync layer when choosing participants, in case something goes wrong - we still have the posits and rejections

ChaoticTempest: I'm saying state sync won't do much in this case, we'll get the latest up to date participants who has the presignatures, but we reject on posits anyways. So state sync participants wouldn't matter much. It's just optimization if we do add it

chain-signatures/node/src/storage/protocol_storage.rs
jakmeier: yeah let's keep this in mind and come back to it when we have a moment

chain-signatures/node/src/protocol/signature.rs
jakmeier: > It's a good idea to use the latest set based on state sync but it won't avoid this intersection check. If we do want to avoid it, we'll have to add in another priority queue based on participant size

@ChaoticTempest  I don't understand this. What is participant size and how does it matter in this context?

What I was saying is that state-sync should already have a more-up-to-date version of `artifact.participants` information, stored separately from the artifact. That's what should be used for participants selection. So yes, there is an intersection check but it shouldn't require loading the artifact.

> I'm saying state sync won't do much in this case, we'll get the latest up to date participants who has the presignatures, but we reject on posits anyways. So state sync participants wouldn't matter much. It's just optimization if we do add it

If state sync is accurate, then what would be a reason for the posit to be rejected?

Being aware of who is a valid participant ahead of time is the main point state sync. If we don't see this as valuable, I wonder why we use state sync at all?

integration-tests/tests/cases/solana.rs
jakmeier: Thanks! The new tests are much better!

But still, does it actually test the case that was problematic? In other words, would the tests have failed before this PR? Sorry if I'm missing something but I didn't find how the test triggers the condition `intersect_vec(&[&taken.artifact.participants, &stable]).len() < ctx.threshold` which seems necessary for the problem to occur.

PR #608, opened 2025-11-27
feat: handle 100% signature guarantee on resharing

ChaoticTempest:
This makes signature be handle on the cases of resharing. Before, we would just terminate the sign tasks on reaching resharing, which is not ideal when we want to guarantee 100% signatures.

So with this PR, there are a couple changes to specifically handle resharing cases when doing signature requests:
- we will now have the `SignatureSpawner` never be aborted since it will always be receiving sign requests and self update it's own relevant information.
- `SignTask` will check if we're in resharing, reset itself to organization and then pause if so. It will then update with new governance info when resharing completes and go back into running.
- component tests can now trigger a psuedo resharing like state to test this case specifically.

NOTE: The only time where we can throw away sign requests should be on key refreshing (changing out the whole master key) or when we no longer support a specific key version. These cases still need to be handled eventually.

I built this on top of the fix for sign requests on devnet, so https://github.com/sig-net/mpc/pull/607 needs to go in first

### DIFF
--- a/chain-signatures/node/src/protocol/signature.rs
+++ b/chain-signatures/node/src/protocol/signature.rs
@@ -231,8 +231,8 @@ enum SignPhase {
 
 impl SignPhase {
     async fn advance(
-        self,
-        ctx: &SignTask,
+        &mut self,
+        ctx: &mut SignTask,
         state: &mut SignState,
         task_rx: &mut mpsc::Receiver<SignTaskMessage>,
     ) -> SignPhase {
@@ -240,7 +240,7 @@ impl SignPhase {
             SignPhase::Organizing(phase) => phase.advance(ctx, state).await,
             SignPhase::Posit(phase) => phase.advance(ctx, state, task_rx).await,
             SignPhase::Generating(phase) => phase.advance(ctx, state).await,
-            SignPhase::Complete(result) => SignPhase::Complete(result),
+            SignPhase::Complete(result) => SignPhase::Complete(*result),
         }
     }
 }
@@ -260,7 +260,7 @@ impl SignOrganizer {
     /// Waits for threshold active participants to be present.
     async fn wait_active(
         &self,
-        ctx: &SignTask,
+        ctx: &mut SignTask,
         state: &mut SignState,
         threshold: usize,
     ) -> Option<BTreeSet<Participant>> {
@@ -293,7 +293,7 @@ impl SignOrganizer {
         }
     }
 
-    async fn advance(self, ctx: &SignTask, state: &mut SignState) -> SignPhase {
+    async fn advance(&mut self, ctx: &mut SignTask, state: &mut SignState) -> SignPhase {
         let sign_id = ctx.sign_id;
         let threshold = ctx.governance.threshold;
         let me = ctx.governance.me;
@@ -312,7 +312,7 @@ impl SignOrganizer {
             let Some(active) = self.wait_active(ctx, state, threshold).await else {
                 tracing::warn!(?sign_id, round = ?state.round, "no active participants, reorganizing");
                 state.bump_round();
-                return SignPhase::Organizing(self);
+                return SignPhase::Organizing(SignOrganizer);
             };
 
             let max_rounds = state.round + ROUND_INTERVAL;
@@ -389,7 +389,7 @@ impl SignOrganizer {
                         "proposer timeout waiting for presignature, reorganizing"
                     );
                     state.bump_round();
-                    return SignPhase::Organizing(self);
+                    return SignPhase::Organizing(SignOrganizer);
                 }
             };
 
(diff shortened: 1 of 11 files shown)


chain-signatures/node/src/rpc.rs
```diff
@@ -297,6 +311,44 @@ impl ContractStateWatcher {
...
+                    me: *me,
+                    threshold: running.threshold,
+                    epoch: running.epoch,
+                    public_key: PublicKey::from(running.public_key),
+                    participants: running.participants.keys().copied().collect(),
+                };
+            }
+            // We're not a participant, wait for next state change
+            tracing::warn!("wait_governance: we are not a participant, waiting for state change");
+            let _ = self.contract_state.changed().await;
```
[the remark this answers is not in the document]
ChaoticTempest: only time this would be happening is when RPC is dropped and the contract state sender gets dropped

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -1224,51 +1340,52 @@ pub struct SignatureSpawnerTask {
...
 }
 
-impl Drop for SignatureSpawnerTask {
-    fn drop(&mut self) {
-        self.abort();
-    }
-}
+// NOTE: We intentionally do NOT implement Drop for SignatureSpawnerTask.
+// The spawner should persist across resharing, so dropping it should not abort the task.
+// The spawner will naturally terminate when all channels close (e.g., during node shutdown).
```
jakmeier: Wait, I don't quite follow, when would `SignatureSpawnerTask` drop anyway? It's stored in `MpcSignProtocol`, right? So the only way it would be dropped is if `MpcSignProtocol` drops. And that's exactly when you call abort anyway.

I don't understand the difference between what you have now and aborting the handle whenever `SignatureSpawnerTask` drops. What am I missing? :confused: 

Also, could you put this comment somewhere closer to the `SignatureSpawnerTask` definition? Or inside `fn abort` just above? Seems a but out of place here and I probably wouldn't read it when I look at the relevant other code.

chain-signatures/node/src/rpc.rs
```diff
@@ -297,6 +311,44 @@ impl ContractStateWatcher {
...
+            let running = self.wait_running().await;
+            if let Some(me) = running.participants.find_participant(&self.account_id) {
+                return GovernanceInfo {
+                    me: *me,
+                    threshold: running.threshold,
+                    epoch: running.epoch,
+                    public_key: PublicKey::from(running.public_key),
+                    participants: running.participants.keys().copied().collect(),
+                };
+            }
```
jakmeier: nit: I think something like this would work, too

```suggestion
            if let Some(gov) = self.governance() {
                return gov;
            }
```

integration-tests/tests/cases/mpc.rs
```diff
@@ -198,6 +198,147 @@ async fn test_basic_sign() {
...
+        .send(Sign::Request(request.clone()))
+        .await
+        .unwrap();
+
+    // Give some time for the sign task to start organizing
+    tokio::time::sleep(Duration::from_millis(50)).await;
+
+    // Trigger resharing while sign task is in progress
+    tracing::info!("triggering resharing");
+    network.trigger_resharing();
```
jakmeier: How do we know it hasn't finished the signature, yet?

jakmeier (dismissed):
Good change and great tests!

Is it really required to merge after #607? Wouldn't it be an independent change that works either way?

chain-signatures/node/src/protocol/signature.rs
ChaoticTempest: ahh that's an old comment. I implemented drop later but forgot to remove this comment. The idea is to have it be dropped when MpcSignProtocol gets dropped but so we can gracefully exit from there but it doesn't have to be there either

integration-tests/tests/cases/mpc.rs
ChaoticTempest: true, I'll add an assert before that the actions hasn't replied yet with the signature

ChaoticTempest:
@jakmeier I just don't want to resolve more merge conflicts so anything that changes the same set of files, I just work off of whatever was changed before. Just makes things simpler in terms of reviews. It'll be rebased no matter what, but at least there's one less friction point

chain-signatures/node/src/protocol/signature.rs
jakmeier: ok so let's just remove the comment then?

volovyks:
> NOTE: The only time where we can throw away sign requests should be on key refreshing (changing out the whole master key) or when we no longer support a specific key version. These cases still need to be handled eventually.

This was discussed in our 1-1, but I want to make sure we don’t forget about it. We should never discard the old root key. All requests must specify a key version, which, in the case of an upgrade, will point to the older key version. As a result, all indexed requests must eventually receive a response.

ChaoticTempest:
@volovyks @jakmeier bumped this to latest of develop

jakmeier (approved):

PR #610, opened 2025-11-27
debug page: display more basic information

jakmeier:
This PR combines has two commits containing two small additions to the debug page.

1. Group tasks on the debug page by name and put them in expandable containers.
2. Add information about node status, participant number of "me", and current T and P stockpile to the debug page view.


jakmeier:
It is still not pretty. But at least the tasks are now grouped and collapsed.

image

ChaoticTempest (approved):

PR #612, opened 2025-12-03
release 1.11.1 merge to main

ppca:


ChaoticTempest (dismissed):
Shouldn't this be 1.11.1 instead of 1.12.0? It's not bad to have 1.12.0 but for the process of this being more of a patch, we should make it a patch release instead.

But I don't care as long as our changelog properly reflects our changes correctly. That would mean that 1.13.0 would exclude all the changes added here for 1.12

ChaoticTempest (dismissed):

ChaoticTempest (approved):

volovyks:
@ppca can we close this one?

PR #613, opened 2025-12-03
fix: deterministic proposer election via scoring

ChaoticTempest:
The sign task issues we're seeing right now on devnet is due to how each sign task could not agree on the correct proposer due to different rounds not being aligned.

So this gets rid of the notion of rounds and instead uses a scoring on each participant `hash(sign_id + participant_id)`, where we then sort all participants in increasing order. We take the first online/stable from this list as our proposer. This is pretty deterministic but comes with the caveat of electing the same leader over and over again if the sign request fails. The only time this should happen is either from timeout where some nodes are not responding which should be pretty rare.

- component tests now have the ability to send sign competions.
- redesigned the depleted stockpile test to first start with how many presignatures we have in the pregenerated file, and then add more not from generation but from us loading in the same set of presignatures (this is fine for testing purposes since we can just set different presignature ids for them)
- adjusted the proposer and deliberator timeouts for the posit phase to be much shorter now. Note that proposer timeout is shorter than deliberator since if we expire on a proposer to issue a START, then the deliberator might end up expiring at the same time and not be able to start. These timeouts aren't ideal either but I will address this in a future PR to make this more robust by not expiring and rather just start on enough accepts.

### DIFF
--- a/chain-signatures/node/src/protocol/signature.rs
+++ b/chain-signatures/node/src/protocol/signature.rs
@@ -36,13 +36,9 @@ use tokio::task::JoinHandle;
 
 use near_account_id::AccountId;
 
-/// The round interval to search for a proposer in the organizing phase.
-const ROUND_INTERVAL: usize = 512;
-
-// If the message comes from someone who isn't the expected proposer,
-// don't immediately reject. Accept proposers who would be selected
-// for any round within +/- ACCEPT_ROUND_RANGE of our current round.
-const ACCEPT_ROUND_RANGE: usize = 3;
+/// No multi-round proposer selection anymore.
+/// We select a deterministic leader per-request using a hash-score
+/// computed from the request id and participant id.
 
 /// Timeout durations for the posit phase. Propose is should be faster than deliberator
 /// for the purpose of starting sooner if we have enough participants.
@@ -76,7 +72,7 @@ enum SignError {
 }
 
 struct SignState {
-    round: usize,
+    attempt: usize,
     indexed: IndexedSignRequest,
     mesh_state: watch::Receiver<MeshState>,
 }
@@ -84,7 +80,7 @@ struct SignState {
 impl SignState {
     fn new(indexed: IndexedSignRequest, mesh_state: watch::Receiver<MeshState>) -> Self {
         Self {
-            round: 0,
+            attempt: 0,
             indexed,
             mesh_state,
         }
@@ -94,8 +90,8 @@ impl SignState {
         &self.indexed
     }
 
-    fn bump_round(&mut self) {
-        self.round += 1;
+    fn bump_attempt(&mut self) {
+        self.attempt += 1;
     }
 }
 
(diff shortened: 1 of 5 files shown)


chain-signatures/node/src/protocol/signature.rs
```diff
@@ -126,13 +131,29 @@ impl SignPhase {
...
+                hasher.update(Into::<u32>::into(p).to_le_bytes());
+                let hash = hasher.finalize();
+                let score = u64::from_le_bytes(hash[0..8].try_into().unwrap());
+                (score, p)
+            })
+            .collect();
+
+        // sort by score ascending (lower scores have priority)
+        scored.sort_by_key(|(score, _)| *score);
+        scored.into_iter().map(|(_, p)| p).collect()
```
jakmeier, line 156: That's a neat trick to establish a shared order, even if the participant set it different!

integration-tests/src/mpc_fixture/fixture_interface.rs
```diff
@@ -47,6 +53,30 @@ pub struct SharedOutput {
     pub rpc_actions: Arc<Mutex<HashSet<String>>>,
 }
 
+/// Broadcast channel for sign completions across all nodes.
+/// When any node publishes a signature, this channel is used to notify
+/// all other nodes so they can abort their tasks for the same SignId.
+pub struct CompletionBroadcaster {
+    pub tx: broadcast::Sender<SignId>,
+}
```
jakmeier, line 61: Why do we need this?

It feels like cheating in the test setup. Real nodes won't know instantly when a signature is generated. We must be robust to handle this case in our distributed system without global knowledge.

jakmeier (commented):
So, if we have problems with aligning rounds, it's important to understand why that happens. Do you have a clear understanding of that? It's not obvious to me. 

Also, could we have a test that creates such a scenario to demonstrate how it happens?
The tests in this PR are valuable for what they test. But they don't seem to reproduce the bug fixed in this PR.

----

Regarding the removal of rounds: If we do it, it is a step backwards regarding the 100% response guarantee. A single faulty node could lead to lost requests if things line up badly.

Of course, we might need to make this step backward anyway, if round alignment as implemented now doesn't work. In that case, I think it should be a temporary solution and we should follow it up by making the posits phase fault tolerant. Which basically means, we need to run consensus to agree on the proposer for a signature.

I'd probably go with the King-Algorithm ([see 17.3 in this document](https://disco.ethz.ch/courses/hs24/distsys/lnotes2/chapter17.pdf)). It should be quite simple to implement and it would even be BFT. (We'll have to start somewhere.)
It requires more messages compared to what we have now but with our number of nodes, it should be ok. And it can always be replaced with other consensus algorithms if needed.

We can predetermine a list of kings using the same deterministic inputs used so far. Then each node can evaluate who they think should be proposer, just like you do it in this PR. In most cases, enough nodes will agree after the first broadcast in line 5. The rest of the algorithm covers all corner cases we would be missing in the current PR.

---

That all written down - This seems to become way over-engineered. I really think we should be able to make a round-based retry approach work for our case, without losing 100% response guarantee.

@ChaoticTempest if you can explain how round alignment fails right now, I may be able to propose standard solutions to that problem.

ChaoticTempest:
> Do you have a clear understanding of that? It's not obvious to me.

I believe it's from getting the sign request at different times, maybe too far a part from one round.
  - our timeouts might be different now than what they were before leading to round increments being different.
  - maybe if we can get the timeouts to be correct again, it would fix this part, but doesn't fix it when multiple nodes decide to be offline for a while and then they join and want to be a part of the sign request.

Another issue could be due to how we wait on the totality of participants before proceeding or expiring, we may end up waiting too long and trying to expire then start on an already expired posit

> Also, could we have a test that creates such a scenario to demonstrate how it happens?
> The tests in this PR are valuable for what they test. But they don't seem to reproduce the bug fixed in this PR.

Yeah I think testing SignOrganizer/SignPositor would probably be better to reproduce this case. In that case, I'll need to do the same thing I did for `CheckpointStorage`: make `ProtocolStorage` an enum over in memory or redis backing. That way we can directly test the SignTask for specific cases.

> I'd probably go with the King-Algorithm ([see 17.3 in this document](https://disco.ethz.ch/courses/hs24/distsys/lnotes2/chapter17.pdf)). It should be quite simple to implement and it would even be BFT. (We'll have to start somewhere.)
> It requires more messages compared to what we have now but with our number of nodes, it should be ok. And it can always be replaced with other consensus algorithms if needed.

I'll take a look at it and see how we can fit it in

integration-tests/src/mpc_fixture/fixture_interface.rs
ChaoticTempest: this was more to mimic the indexer `send(Sign::Completion)` to each node in case it wasn't a part of the sign request since we don't have the indexer here to say a sign request has been completed

ChaoticTempest:
hmm, what about if we do something like on /status, we send our our current time. Other nodes who receive this /status can check to see if there's a significant drift between us and them. If the drift is too large, we just mark the node as not stable.

Then for all stable nodes, we can set the round to be the time modulo over a timeframe higher than the drift

ChaoticTempest:
> Do you have a clear understanding of that? It's not obvious to me.

I just reread our old code, it's because we were actually using the same round over and over again... we weren't doing previous_round+1, but just previous_round, so there was never any incrementation 🫤

jakmeier:
> > Do you have a clear understanding of that? It's not obvious to me.
> 
> I just reread our old code, it's because we were actually using the same round over and over again... we weren't doing previous_round+1, but just previous_round, so there was never any incrementation 🫤

Oh... I see :sweat_smile: 

In that case, rather than adding complexity that might not be needed, why don't we fix the old code? And if we can, let's also add a test where the first attempt is missed and leads to a successful second attempt. (Can be separate PRs)

The ideas with scoring, time shifts, and timeouts are all good. But complexity has a cost on its own and I'm currently not fully convinced it is worth adding.

Ideally, any patch that we add for stability's sake only, should have a test associated with it that proves it is useful in a known and reproducible scenario. Anything else is guess work. And by default, new code has more problems than established code.

ChaoticTempest:
@jakmeier I should be a bit more explicit when saying old code. I mean the previous code before SignTask was a thing. The reason why we never saw any issues with round before SignTask was because we were using the same round and never incrementing. So there's really no easy fix for this

ChaoticTempest:
closing in favor of https://github.com/sig-net/mpc/pull/626

PR #614, opened 2025-12-03
feat: persist latest checkpoint for each chain

ChaoticTempest:
This PR adds persistence of the latest checkpoint for each chain to redis.

How this works with checkpoing bootstrap selection is query all nodes for their checkpoints, compare each chains checkpoint to our local chain checkpoint and take the one with the higher block height (or more recent one).

### DIFF
--- a/chain-signatures/node/src/backlog/mod.rs
+++ b/chain-signatures/node/src/backlog/mod.rs
@@ -5,6 +5,7 @@ use crate::mesh::MeshState;
 use crate::node_client::NodeClient;
 use crate::protocol::{Chain, SignRequestType};
 use crate::sign_bidirectional::{BidirectionalTx, BidirectionalTxId, PendingRequestStatus};
+use crate::storage::checkpoint_storage::CheckpointStorage;
 use anyhow::Context;
 use mpc_primitives::{PendingTx, SignId};
 use std::collections::{hash_map, HashMap};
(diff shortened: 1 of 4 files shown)


chain-signatures/node/src/storage/checkpoint_storage.rs
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    fn checkpoint_key(&self, chain: Chain) -> String {
+        match self {
+            CheckpointStorage::Redis(_, account_id) => {
+                format!("{account_id}:checkpoint:latest:{chain}")
+            }
+            CheckpointStorage::InMemory(_) => format!("checkpoint:latest:{chain}"),
+        }
+    }
```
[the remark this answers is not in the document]
ChaoticTempest: ehh, we can just add this in the future when we change storage versions

chain-signatures/node/src/backlog/mod.rs
```diff
@@ -200,7 +202,12 @@ impl Default for Backlog {
 
 impl Backlog {
     pub fn new() -> Self {
+        Self::persisted(CheckpointStorage::in_memory())
```
volovyks, line 205: Naming and logic around `inMemory`/`persisted` and `new()`/`persisted()` is a bit confusing. It is unclear if you recover old data or not and what is stored in memory. I would say `recover_from_storage` make more sense to me. But maybe I'm a bit opiniated here.

volovyks (approved):
LGTM

chain-signatures/node/src/backlog/mod.rs
ChaoticTempest: So how these functions are supposed to be read is `Backlog::new()` which you don't care about the storage at all. And `Backlog::persisted(CheckpointStorage::Redis(...))` where the backlog's persistence is in redis. 

`recover_from_storage` is more of an action than what you're setting the backing storage to be

jakmeier (approved):

PR #618, opened 2025-12-11
feat: add timeout budget for sign posit+organizer

ChaoticTempest:
relatively small change that might improve some things. This makes it so that we have a budget for doing organizing + positing. This is so it's consistent between proposers and deliberators and that our rounds can align a little bit better in the happy path. 

So with this PR, each sign task starts with a budget of 60s and then they do work until generation. Anything that tries to timeout will rely on this budget.


jakmeier (approved):
LGTM

To confirm: This fixes the missing increase of the "round" or "attempt" counter, right?

ChaoticTempest:
@jakmeier it helps but it doesn't solve it completely. There can still be cases where we'll be misaligned. Like for example, somebody's indexer is off by 60s, so they'll get the sign request at round 0, while everyone else is at round 1 (if they haven't completed the request by then).

PR #619, opened 2025-12-11
feat: in memory used and reserved for ProtocolStorage

ChaoticTempest:
This moves our protocol storage reserved/used into memory land instead of having it be in redis. There's really no reason to have these be persisted. This allows for less round trips to redis and redis operations overall. 

Reserved is really only for at the moment sort of reservations which is not really useful when we restart a node. Best to just have it be cleared from memory instead in cases where we just reserve but then don't do anything with the reservation.

Used is similar. Only necessary for when we do insertions and allowing us to short circuit when it's already used. Doesn't matter much in guarding against trying to use the same artifact over again since this is just an ID.


chain-signatures/node/src/storage/protocol_storage.rs
```diff
@@ -469,6 +454,7 @@ impl<A: ProtocolArtifact> ProtocolStorage<A> {
 
         match result {
             Ok(artifact) => {
+                self.used.write().await.insert(id);
```
jakmeier: Shouldn't you check the result? 

If this returns false, the artifact was already marked as used by another task. I guess it shouldn't happen if the reservation logic works out. But to be locally consistent, wouldn't it be better to return None if insertion failed?

jakmeier (dismissed):
Lovely change! The fewer code we run on redis, the better IMO.

What is the plan for migration? Should we run a clean up script to remove the used/reserved data on existing nodes? Or are we keeping it around as dead data until a full db reset?

ChaoticTempest:
> What is the plan for migration? Should we run a clean up script to remove the used/reserved data on existing nodes? Or are we keeping it around as dead data until a full db reset?

Probably the latter. It's such a small amount of data anyways that's it not worth the hassle cleaning up. The bulk of data is in the artifacts being stored not the IDs which reserved and used are storing.

chain-signatures/node/src/storage/protocol_storage.rs
ChaoticTempest: I see, instead of insert here after we've successfully taken the artifact, we should insert it at the beginning before we take and then remove after the fact if we were unable to take

jakmeier (dismissed):

volovyks (dismissed):
Nice simplification!

volovyks (approved):

PR #620, opened 2025-12-11
Include all protocol failures in the metric

volovyks:


chain-signatures/node/src/metrics.rs
```diff
@@ -375,14 +375,15 @@ pub(crate) static SIGNATURE_GENERATOR_FAILURES: LazyLock<CounterVec> = LazyLock:
...
 
-pub(crate) static SIGNATURE_GENERATOR_SUCCESS: LazyLock<CounterVec> = LazyLock::new(|| {
-    try_create_counter_vec(
-        "multichain_num_total_historical_signature_generators_success",
-        "total signature generator success",
-        &["node_account_id"],
-    )
-    .unwrap()
-});
+pub(crate) static NUM_TOTAL_HISTORICAL_SIGNATURE_GENERATORS_SUCCESS: LazyLock<CounterVec> =
```
volovyks, line 378: Renaming this for consistency with presignatures and triples

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -835,9 +838,7 @@ impl SignGenerator {
                 Action::Wait => {
                     // Wait for the next set of messages to arrive.
                     let msg = self.recv().await.inspect_err(|_| {
-                        if self.proposer == me {
-                            signature_generator_failures_metric.inc();
-                        }
+                        signature_generator_failures_metric.inc();
```
jakmeier, line 841: What is the reason for this change? The commit says that it want to include all failures in the metric now but now why.

volovyks: @jakmeier  There is two metrics, `multichain_num_total_historical_signature_generators_success` and `multichain_signature_generator_failures`, the first one represents all protocols, the second represents only the failures on mine protocols. I want to change that, so the second metric also represent all protocol failures and I can represent that as a success % metric.

```
multichain_num_total_historical_signature_generators_success / (multichain_num_total_historical_signature_generators_success + multichain_signature_generator_failures)
```

We can add mine metrics for all protocols separately.

jakmeier: Ok, thanks for explaining.

So with this change, we have `signature_generator_failures_metric` with the account id as label. This looks redundant to me. Every node will only ever insert value with their own name as label.

I think we should either remove the label on the metric, or make the label the proposer.

If the proposer is used as label, we can read "mine" in Graphana by filtering by node_id = proposer, or we can access the sum by not filtering by the proposer label.

volovyks: To be sure I understand you correctly. Instead of having 2 metrics for `mine` and `total` protocol success and failure we can have a single metric that will pass the `node_id` and report only mine success or failures. So the `node_id` parameter becomes `proposer` parameter. We can represent them separately for each node or sum to get the data about all.

Overall I think this is a good strategy. I wanted to reduce the number of metrics as a next step, but we can make it this way now.

I will create a separate PR then, for all protocols.

jakmeier: > Instead of having 2 metrics for mine and total protocol success and failure we can have a single metric...

yes

> ...that will pass the node_id...

No. I was thinking of the proposer_id, not the node_id.

Although, looking at it again, maybe it should be both? It seems we rely on `node_account_id` a lot in our Grafana boards, even if it's functionally equivalent to `instance`. This seems a bit wasteful, though. Every data set is stored per instance where the data was collected from anyway.

> ...and report only mine success or failures.

No, I would still report all failures on every node.

So the way this works with Prometheus, it collects and stores data sets from each node. If a metric has labels, its data set becomes a vector of numbers rather than a single number. 

Let's assume there are 10 failures across all sign protocols, (Node 1 owned 3, node 2 owned 5 and node 3 owned 2. Today, Prometheus stores data like this:

  | instance=Node1 | instance=Node2 | instance=Node3
-- | -- | -- | --
node_id=Node1 | 3 | 0 | 0
node_id=Node2 | 0 | 5 | 0
node_id=Node3 | 0 | 0 | 2
node_id=* | 3 | 5 | 2




What you had in this PR would result in something like this. (This assume node 3 did not participate in one of the other failures, thus it has observed one less over all failures.)


  | instance=Node1 | instance=Node2 | instance=Node3
-- | -- | -- | --
node_id=Node1 | 10 | 0 | 0
node_id=Node2 | 0 | 10 | 0
node_id=Node3 | 0 | 0 | 9
node_id=* | 10 | 10 | 9


Instead, I suggest we label by proposer_id rather than by node_id.


  | instance=Node1 | instance=Node2 | instance=Node3
-- | -- | -- | --
proposer_id=Node1 | 3 | 3 | 3
proposer_id=Node2 | 5 | 5 | 4
proposer_id=Node3 | 2 | 2 | 2
proposer_id=* | 10 | 10 | 9


This uses the same amount of space on Prometheus but gives us more insight.

- You get "mine" with `proposer_id=instance`.
- You get "total" with `proposer_id=*` ( the same  as not adding a filter for proposer_id)
- You can also read from the data that the failure which Node3 didn't participate/observe was owned by Node2.

Now, what I didn't consider is that `instance` uses a different name than `node_account_id`. So we either have to do some `instance->node_id` mapping on Grafana's side, or we add two labels (`proposer_id` and `node_account_id`) on the metric.

The second option would make the table 3 dimensional with lots of zero entries. So quite redundant again. But it's probably the easiest to keep queries on Grafana simple.

But if you just want two entirely separate metrics for mine and total, that's also good with me and we can merge this as you proposed.

volovyks: `node_account_id` is a stable identifier of the node provider. It never changes, unlike their setup (and instance), which may cause the loss of data integrity. But yes, we are wasting resources here. Mapping of instance -> node_account_id should be possible manually.

I will get back to this once we have request monitoring in place.

volovyks: @jakmeier  I've played with instances, since we are using K8s, the instance changes every time we reboot the node. Grafana recommends to use stable identifiers (node_account_id) In our case.

PR #560, opened 2025-10-13
Remove infra from MPC repo

volovyks:
@auto-mausx, please, approve once you think we are ready to merge it

PR #621, opened 2025-12-11
bidirectional: hydration indexer

ppca:
This works locally with https://github.com/galacticcouncil/hydration-node/pull/1189/files.
 
https://github.com/galacticcouncil/hydration-node/pull/1189/files has been closed, and hydration team is pushing forward with https://github.com/galacticcouncil/hydration-node/pull/1235/files#.

We can review this PR first, and I'll push new PR that aims to dynamically work with hydration runtime changes, i.e. https://github.com/galacticcouncil/hydration-node/pull/1235/files#.


chain-signatures/crypto/src/kdf.rs
```diff
@@ -28,6 +29,7 @@ impl Chain {
             Chain::Ethereum => "0x1",
             Chain::Solana => "0x800001f5",
             Chain::Bitcoin => "bip122:000000000019d6689c085ae165831e93",
+            Chain::Hydration => "polkadot:2034",
```
volovyks, line 32: `2034` is not a CAIP-2 identifier, it's a `ParaId` that identifies it inside of Polkadot network.
CAIP 2 is using the `genesis hash`. OS it should be `polkadot:afdc188f45c71dacbaa0b62e16a91f72`.

infra/scripts/generate_keys/src/main.rs
```diff
@@ -1,11 +1,22 @@
...
 use ethers::signers::Signer;
 use mpc_keys::hpke;
+use sp_core::crypto::{Ss58AddressFormatRegistry, Ss58Codec};
+use sp_core::{sr25519, Pair};
 use std::env;
 
 fn main() {
     let args: Vec<String> = env::args().collect();
 
+    let (hydration_pair, hydration_phrase, _seed) = sr25519::Pair::generate_with_phrase(None);
```
volovyks, line 11: The whole `/infra` folder is intended to be removed (https://github.com/sig-net/mpc/pull/560). Let's added changes there. @auto-mausx please, approve that PR if we are ready to move forward with it.

chain-signatures/primitives/src/lib.rs
```diff
@@ -133,18 +134,25 @@ impl Chain {
...
+            Chain::Hydration,
+        ]
     }
 
     pub fn checkpoint_interval(&self) -> Option<u64> {
         let (key, default) = match self {
             Chain::NEAR => return None,
             Chain::Ethereum => ("CHECKPOINT_INTERVAL_ETHEREUM", 20),
             Chain::Solana => ("CHECKPOINT_INTERVAL_SOLANA", 120),
+            Chain::Hydration => ("CHECKPOINT_INTERVAL_HYDRATION", 120),
```
volovyks: @ChaoticTempest remind me please, why Ethereum has smaller backup interval compared to Solana and now Hydration?

chain-signatures/node/src/rpc.rs
```diff
@@ -642,13 +674,131 @@ impl SolanaClient {
...
+        HydrationSignature {
+            big_r: HydrationAffinePoint { x, y },
+            s: s_bytes,
+            recovery_id: sig.recovery_id,
+        }
+    }
+
+    /// Call the Signet pallet's `respond()` extrinsic for a *single* request.
+    ///
+    /// You can extend this to true batching later; for now we just send 1-element
```
volovyks: You :)

chain-signatures/node/src/respond_bidirectional.rs
```diff
@@ -117,16 +118,13 @@ impl CompletedTx {
...
-            self.tx.key_version,
-            &self.tx.sender.to_string(),
-            &path,
-        );
+        let path = match chain {
+            Chain::Solana => SOLANA_RESPOND_BIDIRECTIONAL_PATH.to_string(),
+            Chain::Hydration => HYDRATION_RESPOND_BIDIRECTIONAL_PATH.to_string(),
+            _ => anyhow::bail!("Unsupported chain: {}", chain),
+        };
+        tracing::info!("requester to derive epsilon: {:?}", self.tx.sender);
```
volovyks: Why do we need this log?
And if we need it, do we want to specify the chain, since we have more then one now?

chain-signatures/node/src/indexer_sol.rs
```diff
@@ -229,11 +215,11 @@ impl SignatureEventTrait for SignatureRequestedEvent {
 
         // Call the existing derive_epsilon_sol function with the correct parameters
         // to match the TypeScript implementation
-        let epsilon = derive_epsilon_sol(self.key_version, &self.sender.to_string(), &self.path);
+        let epsilon = derive_epsilon_sol(self.key_version, &self.sender_string(), &self.path);
```
volovyks, line 216: I would say it should be sender(), not sender_string(). We should not specify type in the function name.

chain-signatures/node/src/indexer_sol.rs
```diff
@@ -229,11 +215,11 @@ impl SignatureEventTrait for SignatureRequestedEvent {
...
 
         // Use transaction signature as entropy
-        let mut entropy = [0u8; 32];
-        entropy.copy_from_slice(&tx_sig[..32]);
+        // let mut entropy = [0u8; 32];
```
volovyks: I guess this is not required anymore.

chain-signatures/node/src/indexer_hydration/mod.rs
```diff
@@ -0,0 +1,743 @@
...
+fn system_events_key() -> Vec<u8> {
+    let mut key = Vec::with_capacity(32);
+    key.extend_from_slice(&twox_128(b"System"));
+    key.extend_from_slice(&twox_128(b"Events"));
+    key
+}
+
+/// Fetch and *verify* the SCALE‑encoded `System::Events` bytes at a given block.
+///
+/// - Uses `state_get_read_proof` via `LegacyRpcMethods`.
```
volovyks, line 436: Legacy?

chain-signatures/node/src/indexer_hydration/mod.rs
```diff
@@ -0,0 +1,743 @@
...
+    let mut blocks = match hydration_api.blocks().subscribe_finalized().await {
+        Ok(blocks) => blocks,
+        Err(e) => {
+            tracing::error!("failed to subscribe to finalized blocks: {e}");
+            return;
+        }
+    };
+
+    while let Some(block_res) = blocks.next().await {
+        tracing::info!("received block from hydration rpc");
```
volovyks: logging block hash or height can be useful

volovyks (commented):
Great work! I have not found any major issues, let's move forward with this.

chain-signatures/node/Cargo.toml
```diff
@@ -39,7 +39,15 @@ opentelemetry-appender-tracing = "0.29.1"
 alloy-signer-local = "1.0.38"
 alloy-sol-types = "1.4.1"
 helios = { git = "https://github.com/a16z/helios", rev = "e1f9a50f73d7af6f6ae4acf2b663d04aab6adb19" }
-
+ed25519-zebra = { version = "4.1", default-features = false, features = ["alloc"] }
+futures = "0.3.31"
```
jakmeier:
```suggestion
```

I don't think you use this anywhere. I also don't know of any good reason to depend on this crate anymore ever since futures were integrated in std.

chain-signatures/node/Cargo.toml
```diff
@@ -39,7 +39,15 @@ opentelemetry-appender-tracing = "0.29.1"
...
+subxt = { version = "0.44", default-features = false, features = ["jsonrpsee","native","unstable-light-client"] }
+sp-core = { version = "38.1", default-features = false, features = ["std"] }
+sp-runtime = { version = "44.0.0", default-features = false, features = ["std"] }
+sp-trie = { version = "41.1.0", default-features = false, features = ["std"] }
+sp-state-machine = { version = "0.48", default-features = false, features = ["std"] }
+codec = { package = "parity-scale-codec", version = "3", features = ["derive"] }
```
jakmeier:
```suggestion
```

Is this used anywhere?

chain-signatures/node/src/indexer_common.rs
```diff
@@ -0,0 +1,455 @@
...
+    fn generate_sign_request(
+        &self,
+        entropy: [u8; 32],
+        total_timeout: Duration,
+    ) -> anyhow::Result<IndexedSignRequest>;
+    fn source_chain(&self) -> Chain;
+    fn sender_string(&self) -> String;
+}
+
+pub(crate) trait SignatureEvent: SignatureEventTrait + std::fmt::Debug {}
```
jakmeier: optional simplification: I don't think having two separate traits has a benefit here. I'd merge them into one.

```suggestion
pub(crate) trait SignatureEvent: std::fmt::Debug {
    fn generate_request_id(&self) -> [u8; 32];
    fn generate_sign_request(
        &self,
        entropy: [u8; 32],
        total_timeout: Duration,
    ) -> anyhow::Result<IndexedSignRequest>;
    fn source_chain(&self) -> Chain;
    fn sender_string(&self) -> String;
}
```

Then you could also remove all the blank `impl SignatureEvent for _ {}` lines.

chain-signatures/node/src/indexer_common.rs
```diff
@@ -0,0 +1,455 @@
...
+            backlog_tx,
+            sign_request_type,
+        )
+        .await;
+
+    if let Err(err) = sign_tx.send(Sign::Request(sign_request)).await {
+        // TODO: handle error to ensure 100% success rate
+        let chain = sign_event.source_chain();
+        tracing::error!(?err, chain = %chain, "Failed to send {} sign request into queue", chain.as_str());
+    } else {
```
jakmeier, line 275: Hm, so if this fails, we have a full channel buffer. A burst of incoming requests could easily trigger this. So anyone could try and abuser this to prevent a signature from getting through.

But the good news is, at this point we already have the request in the backlog. All we need is to ensure we start new attempts on backlog requests if no success is seen for a time. (cc @ChaoticTempest )

I know this is just moved from other files in this PR. But can we maybe link this to an issue or something to make it less likely this is just buried and forgotten?

chain-signatures/node/src/indexer_common.rs
```diff
@@ -0,0 +1,455 @@
...
+        }
+    };
+
+    tracing::info!(?sign_id, "bidirectional processing initial respond event");
+    let target_chain = Chain::from_str(&event.dest())
+        .map_err(|err| anyhow::anyhow!("unable to parse target chain from dest: {err:?}"));
+    let target_chain = match target_chain {
+        Ok(chain) => chain,
+        Err(_) => Chain::Ethereum,
+    };
```
jakmeier, line 344: We send to ETH if we don't know where it should go? Why?

chain-signatures/node/src/rpc.rs
```diff
@@ -1465,3 +1628,91 @@ async fn try_publish_sol(
 
     Ok(())
 }
+
+async fn try_publish_hydration(
+    hyd: &HydrationClient,
+    action: &PublishAction,
+    timestamp: &Instant,
+    signature: &Signature,
+    near_account_id: &AccountId, // you already pass this around for metrics labels
```
jakmeier:
```suggestion
    near_account_id: &AccountId,
```

chain-signatures/primitives/src/lib.rs
jakmeier: I think it's because the Ethereum block time is slower. It's better if checkpoints are never too old.

Maybe for Hydration it should be larger than for Solana?

jakmeier (dismissed):
Looking good!

chain-signatures/node/src/respond_bidirectional.rs
ppca: ah this is for my local debug. I'll remove it.

chain-signatures/node/src/indexer_hydration/mod.rs
ppca: this is for fetching proof, but I haven't been able to test it locally.

chain-signatures/node/src/indexer_common.rs
ppca: That's because now all the bidirectional tx's send actual transaction on Ethereum. This will need to change in the future.

chain-signatures/node/src/indexer_common.rs
ppca: sounds good!

chain-signatures/crypto/src/kdf.rs
ppca: this is matching the id that https://github.com/galacticcouncil/hydration-node/pull/1275 sets in the tests.

auto-mausx (approved):

chain-signatures/crypto/src/kdf.rs
volovyks: Then we need to change it there @esaminu and fix here

chain-signatures/primitives/src/lib.rs
volovyks: If Ethereum block time is slower, why do we need to back it up more often?

chain-signatures/node/src/indexer_hydration/mod.rs
volovyks: I'm saying that this API is marked as legacy (outdated), maybe there is new one that we should use.

chain-signatures/node/src/indexer_common.rs
volovyks: It would be safer to throw an error. I guess we are specifying target_chain=Ethereum anyway.

PR #622, opened 2025-12-11
test: signing with missing presignatures

jakmeier:
Add a component test where a node has all their presignatures removed.

The node itself detects that it does not have the required input for a signature, so they will reject the posits. The other nodes will succeed signing without the node that lacks presignatures.

### DIFF
--- a/integration-tests/src/mpc_fixture/fixture_tasks.rs
+++ b/integration-tests/src/mpc_fixture/fixture_tasks.rs
@@ -91,7 +91,7 @@ pub(super) fn test_mock_network(
                             )
                         },
                     };
-                    tracing::error!(target: "mock_network", ?action_str, "Received RPC action");
+                    tracing::info!(target: "mock_network", ?action_str, "Received RPC action");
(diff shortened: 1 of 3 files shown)


integration-tests/src/mpc_fixture/fixture_tasks.rs
```diff
@@ -91,7 +91,7 @@ pub(super) fn test_mock_network(
                             )
                         },
                     };
-                    tracing::error!(target: "mock_network", ?action_str, "Received RPC action");
+                    tracing::info!(target: "mock_network", ?action_str, "Received RPC action");
```
jakmeier, line 94: fly-by change: this shouldn't be logged as error and it annoyed me while analyzing what my test output

jakmeier:
This test seems to work. It shows that we have a decent robustness against nodes missing presignatures. Before posits, I think this test would have failed.

I suggest we add this test to ensure posits indeed handle this case.

ChaoticTempest (approved):

PR #623, opened 2025-12-11
test: interrupt signing

jakmeier:
This component test simulates when a node that accepted a posit fails and loses their presignatures.

The other nodes should eventually detect that the signature is not produced and start a new round of posits.

Note that this is not BFT, yet. The failing node behaves quite nice, it will reject future posit rounds. So, the other nodes should have no problem establishing a new set of participants that works.

PR #626, opened 2025-12-18
fix: align posit rounds

jakmeier:
Include the current round in posit messages and avoid mixing messages meant for different rounds.

When a timeout occurs, use the largest round that peer messages used, in order to catch up if other nodes are more than one round ahead.

Lastly, keep a buffer for messages that are sent to future rounds. Nodes with a deliberator might otherwise miss the Propose message if they enter the new round after the proposer.

### DIFF
--- a/chain-signatures/node/src/protocol/signature.rs
+++ b/chain-signatures/node/src/protocol/signature.rs
@@ -27,7 +27,7 @@ use mpc_primitives::{SignArgs, SignId};
 use rand::rngs::StdRng;
 use rand::seq::IteratorRandom;
 use rand::SeedableRng;
-use std::collections::{BTreeSet, HashMap};
+use std::collections::{BTreeSet, HashMap, VecDeque};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 use tokio::sync::{mpsc, watch, RwLock};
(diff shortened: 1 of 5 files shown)


jakmeier:
This PR combines the test from #623 with a fix for it.

You can look at CI logs for #623 to see how the test fails on the current develop branch.

Look at the last commit in this PR to see how I fixed it.

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -39,7 +39,7 @@ use near_account_id::AccountId;
 const ROUND_INTERVAL: usize = 512;
 
 /// The default timeout budget for organizing and posit phases.
-const ORGANIZE_POSIT_TIMEOUT: Duration = Duration::from_secs(60);
+const ORGANIZE_POSIT_TIMEOUT: Duration = Duration::from_secs(20);
```
jakmeier, line 42: 60 seconds seems way too long. If we can't organize posits in a few seconds, we should quickly change the proposer. I think 20s is still quite conservative.

With 60 seconds delay, the entire test runs 190s. The first round fails by design and the second happens to pick the bad node as the proposer. That explain at least 120s of it. There are also set up delays. Not sure about the remainder.

With 20s, it recovers much faster and the entire test only takes 71 seconds.

jakmeier:
See comments in https://github.com/sig-net/mpc/pull/623 for more details on the problems this fixes.

chain-signatures/node/src/protocol/signature.rs
ChaoticTempest: yeah, think we can go all the way down to probably a second

ChaoticTempest (approved):
very nice fix

PR #627, opened 2025-12-19
hydration: not rely on metadata.scale, dynamically index and respond

ppca:
this would ensure that, even if hydration runtime changes, i.e. protocol upgrade etc, as long as our signet pallet stays the same, we can always index requests and respond.


chain-signatures/hydration-pallet/src/lib.rs
```diff
@@ -0,0 +1,37 @@
...
+/// Signature type used by the Hydration pallet.
+///
+/// This structure matches the on-chain signature format, consisting of:
+/// * `big_r` - the public point component of the signature as an
+///   [`HydrationAffinePoint`],
+/// * `s` - the 32-byte scalar part of the signature,
+/// * `recovery_id` - a one-byte recovery identifier.
+/// It is encoded with `parity_scale_codec` to remain compatible with the
+/// pallet's on-chain signature type.
+#[derive(Clone, Debug, Encode, Decode)]
```
volovyks: Why do we need a new module just for a couple of structures?

chain-signatures/node/Cargo.toml
```diff
@@ -46,6 +46,7 @@ sp-runtime = { version = "44.0.0", default-features = false, features = ["std"]
 sp-trie = { version = "41.1.0", default-features = false, features = ["std"] }
 sp-state-machine = { version = "0.48", default-features = false, features = ["std"] }
 subxt-signer = "0.44"
+parity-scale-codec = { version = "3", features = ["derive"] }
```
volovyks: We already have this dependency in new module. If it is required here, let's follow workspace=true pattern.

volovyks (commented):
LGTM
Is this approach the "best practive" working with Hydration/Polkadot?

chain-signatures/hydration-pallet/src/lib.rs
ppca: These are the same as those defined in the pallet in hydration. So I figure it might be better to define them here in a separate module.

ppca: moving it into rpc

ppca:
> LGTM Is this approach the "best practive" working with Hydration/Polkadot?

not sure about best practice, but this should be resistant to runtime updates. I will check with Hydration team in the mean time

volovyks (approved):

PR #630, opened 2026-01-08
chore: remove google-datastore1 & ethabi

ChaoticTempest:
Removes google-datastore1 in favor of just using google-secretmanager1 since we no longer use datastore for anything. We can also just directly use alloy instead of ethabi

### DIFF
--- a/chain-signatures/node/src/indexer_hydration.rs
+++ b/chain-signatures/node/src/indexer_hydration.rs
@@ -6,7 +6,7 @@ use crate::node_client::NodeClient;
 use crate::protocol::{Chain, IndexedSignRequest, Sign, SignRequestType};
 use crate::rpc::ContractStateWatcher;
 use crate::sign_bidirectional::hash_rlp_data;
-use alloy::primitives::{keccak256, Bytes};
+use alloy::primitives::keccak256;
 use alloy_sol_types::SolValue;
 use anyhow::{anyhow, Result};
 use k256::elliptic_curve::sec1::FromEncodedPoint;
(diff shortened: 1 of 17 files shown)


jakmeier (dismissed):
Makes sense.

Although, judging by the Cargo.lock, we still pull in all the same dependencies indirectly. So the benefit of removing these direct dependencies is somewhat limited. But still useful IMO.

Oh and it seems you missed to update `indexer_hydration.rs`. Currently it doesn't compile.

ChaoticTempest:
ahh, yeah ethers is pulling in ethabi. I'll get rid of it in favor of purely alloy. The google-datastore1 isn't in our dependency tree anymore though

volovyks (commented):
Nice!

chain-signatures/node/src/indexer_hydration.rs
```diff
@@ -128,21 +127,32 @@ pub struct HydrationSignatureRequestedEvent {
...
-            Token::String(self.chain_id.clone()),
-            Token::String(self.algo.clone()),
-            Token::String(self.dest.clone()),
-            Token::String(self.params.clone()),
-        ]);
-        // Calculate keccak256 hash
-        let mut hasher = Keccak256::new();
-        hasher.update(&encoded);
-        hasher.finalize().into()
+        // Encode the event data in ABI format using alloy
```
ppca: would you mind checking if the old and new implementation generates the same request id? This is important as this need to be the same as the request id that our demo is checking for. Sometimes small changes in encoding changes the value of the request id. 

Thanks!!!!

A unit test for this function is a better approach in the longer run.

chain-signatures/node/src/indexer_sol.rs
```diff
@@ -165,17 +165,19 @@ pub struct SolSignRequest {
...
-            Token::String(self.sender.to_string()),
-            Token::Bytes(self.payload.to_vec()),
-            Token::String(self.path.clone()),
-            Token::Uint(self.key_version.into()),
-            Token::String(self.chain_id.clone()),
-            Token::String(self.algo.clone()),
-            Token::String(self.dest.clone()),
-            Token::String(self.params.clone()),
-        ]);
+        // Encode the event data in ABI format using alloy
```
ppca: same here.

chain-signatures/node/src/indexer_hydration.rs
ChaoticTempest: good call, the underlying alloy abi_encode actually appends an extra header so it definitely changed all the encodings. Let me see what I can do about that and I'll add a unit test

volovyks (approved):
Great reduction in dependencies! All the updates look reasonable to me.
Ethereum is not covered well with integration tests. Have you tested it locally?

ChaoticTempest:
this is a bit out of date now, so closing this in favor of https://github.com/sig-net/mpc/pull/812

PR #632, opened 2026-01-09
Fix digest calculation

volovyks:
The issue appeared to be the differance between debug and display representations of chiper_pk and AccountId.
Also, adding Blacksand digest.


volovyks:
@ChaoticTempest please, take a look

jakmeier (approved):
LGTM

PR #634, opened 2026-01-13
fix: only process completion events on finality

ChaoticTempest:
only process completion events on finality. Related to https://github.com/sig-net/mpc/pull/624 but goes a different direction where we will complete the request on finality instead of having the node track it.

### DIFF
--- a/chain-signatures/node/src/indexer_eth/mod.rs
+++ b/chain-signatures/node/src/indexer_eth/mod.rs
@@ -804,7 +804,6 @@ impl EthereumIndexer {
         let node_near_account_id_clone2 = node_near_account_id.clone();
         let requests_indexed_send_clone = requests_indexed_send.clone();
         let backlog_clone2 = backlog.clone();
-        let sign_tx_clone = sign_tx.clone();
         let client_clone = Arc::clone(&client);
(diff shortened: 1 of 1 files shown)


chain-signatures/node/src/indexer_eth/mod.rs
```diff
@@ -1278,6 +1275,10 @@ impl EthereumIndexer {
                     sign_tx.clone(),
                     node_near_account_id.clone(),
                 );
+
```
ppca, line 1278: not seeing where did we add to respond_logs?

ChaoticTempest: it's when we send it to the `requests_indexed` channel:
```rs
            requests_indexed
                .send(BlockAndRequests::new(
                    block_number,
                    block_hash,
                    sign_requests.clone(),
                    respond_logs,
                ))
```

ppca (approved):

ChaoticTempest:
gonna merge this one for now to get more data on devnet, but let me know if you guys find anything with this and will close the other PR

jakmeier:
@ChaoticTempest Does this change affect reported delays in metrics?

PR #635, opened 2026-01-14
Distinguish IN_TIME and DELAYED requests

volovyks:
- Refactored IN_TIME metric logic
- Added DELAYED metric and delayed watcher logic

### DIFF
--- a/chain-signatures/node/src/rpc.rs
+++ b/chain-signatures/node/src/rpc.rs
@@ -946,35 +946,20 @@ async fn execute_publish(
                     .map_err(|_| ())
             }
             ChainClient::Ethereum(eth) => {
-                try_publish_eth(
-                    eth,
-                    &action,
-                    &action.timestamp,
-                    &signature,
-                    &near_account_id,
-                )
-                .await
+                try_publish_eth(eth, &action, &action.timestamp, &signature).await
             }
-            ChainClient::Solana(sol) => try_publish_sol(
-                sol,
-                &action,
-                &action.timestamp,
-                &signature,
-                &near_account_id,
-            )
-            .await
-            .map_err(|_| ()),
-            ChainClient::Hydration(hyd) => try_publish_hydration(
-                hyd,
-                &action,
-                &action.timestamp,
-                &signature,
-                &near_account_id,
-            )
-            .await
-            .map_err(|_| ()),
+            ChainClient::Solana(sol) => {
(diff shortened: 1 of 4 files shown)


chain-signatures/node/src/rpc.rs
```diff
@@ -1001,6 +986,24 @@ async fn execute_publish(
...
+    let chain_str = chain.as_str();
+    let account_str = near_account_id.as_str();
+
+    if publish_result.is_ok() {
+        let elapsed = action.indexed.timestamp_sign_queue.elapsed();
+        if elapsed.as_secs() <= chain.expected_response_time_secs() {
+            crate::metrics::requests::NUM_SIGN_REQUESTS_MINE_IN_TIME
+                .with_label_values(&[chain_str, account_str])
+                .inc();
+        }
```
jakmeier, line 998: There is a race condition here for requests right around the timeout. We might count a request as delayed and as in-time.

Ideally the code always executes either one or the other. For example, when counting as delayed in the metrics, you could mark the request as delayed, then check here if it was already counted as delayed.

But maybe this is only theoretically a problem and we don't care about the rare miscounting. If it is cumbersome to add, I wouldn't bother.

jakmeier (approved):
LGTM, nice refactor and good addition for the metrics

volovyks:
@jakmeier  Good point, I will take a look and followup if it is not to much of a trouble. ow I just want to get some data to work with. Thanks!

chain-signatures/node/src/rpc.rs
volovyks: I've tried adding logic to prevent raceconditions here, but it looks like unnecessary complexity. Such situation can occur only when responses are sent exactly at finality + 60s and even then it will not be much of a trouble, DELAYED is still there.

jakmeier: > Such situation can occur only when responses are sent exactly at finality + 60s

Keep in mind that if a node is CPU starved for some reason, tokio tasks can be suspended for a long time. Then we might miscount even on larger margins. In such a situation, metrics can be important for debugging.

But yeah, it's really not too bad if we miscount anyway. Not worth the complexity. Thanks for looking into it!

volovyks: Good point, I will keep an eye on CPU usage in case we have drops in our success rate.

PR #636, opened 2026-01-14
Add backlog size metric

volovyks:


ChaoticTempest (approved):
LGTM, but I don't like that we keep sprinkling in our account id everywhere in our code just for metrics. So I made https://github.com/sig-net/mpc/pull/638

PR #637, opened 2026-01-15
fix: send tx on backlog recover

ChaoticTempest:
We weren't sending the recovered sign request back into the sign queue when we restart the node, so this will do that. Also added a unit test to make sure that we are indeed recovering.

Note, this is a breaking change on the checkpoint serialization/persistence since `SignArgs` was added so we could recover it into a `IndexedSignRequest`. Also bumped the sign channel to ~16000 so we don't drop anything beyond ~1000. Will probably make it unbounded in another PR so we can always have guaranteed signature delivery

Also looks like `total_timeout` seems to still be in the indexers so will remove that in a different PR.

### DIFF
--- a/chain-signatures/node/src/indexer_common.rs
+++ b/chain-signatures/node/src/indexer_common.rs
@@ -16,7 +16,6 @@ use crate::rpc::ContractStateWatcher;
 use crate::sign_bidirectional::BidirectionalTx;
 use crate::sign_bidirectional::BidirectionalTxId;
 use crate::sign_bidirectional::PendingRequestStatus;
-use crate::util::current_unix_timestamp;
 use anchor_lang::prelude::Pubkey;
 use k256::Scalar;
 use mpc_primitives::SignId;
@@ -246,6 +245,7 @@ pub(crate) async fn process_sign_event(
             source_chain: sign_event.source_chain(),
             status: PendingRequestStatus::AwaitingResponse,
             args: sign_request.args.clone(),
+            unix_timestamp_indexed: sign_request.unix_timestamp_indexed,
         }),
         SignRequestType::SignBidirectional(_event) => {
             // For bidirectional requests, start with a Sign transaction
@@ -255,6 +255,7 @@ pub(crate) async fn process_sign_event(
                 source_chain: sign_event.source_chain(),
                 status: PendingRequestStatus::AwaitingResponse,
                 args: sign_request.args.clone(),
+                unix_timestamp_indexed: sign_request.unix_timestamp_indexed,
             })
         }
         _ => anyhow::bail!("Unexpected sign request type"),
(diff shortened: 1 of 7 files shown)


chain-signatures/node/src/indexer_common.rs
```diff
@@ -298,9 +305,53 @@ pub(crate) async fn recover_backlog(
...
             .await;
+
+        // Re-enqueue any pending sign requests so the node processes them after recovery
+        let pending = pending.remove(&source_chain).unwrap_or_default();
+
+        for (sign_id, tx) in pending
+            .into_iter()
+            .filter(|(_, tx)| matches!(tx.status(), PendingRequestStatus::AwaitingResponse))
+        {
+            let BacklogTransaction::Sign(sign_tx_entry) = tx else {
```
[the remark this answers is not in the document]
ChaoticTempest: skipping Bidirectional is fine since Bidirectional tracked requests in the backlog merely serves to populate the execution watchers. The watchers are then checked on the indexer level later on where the actual `sign_tx.send` will be called

chain-signatures/node/src/indexer_common.rs
```diff
@@ -298,9 +303,53 @@ pub(crate) async fn recover_backlog(
...
+
+            if let Err(err) = sign_tx.send(Sign::Request(sign_request)).await {
+                tracing::error!(
+                    ?err,
+                    ?sign_id,
+                    ?source_chain,
+                    "failed to requeue sign request after recovery"
+                );
+            } else {
+                crate::metrics::requests::NUM_SIGN_REQUESTS
```
volovyks: This is a metric that represents indexed request, not recovered.
For backlog I've added BACKLOG_SIZE (per chain). That should be enough. It would be great to set it after the recovery is done, now it is only on insert/remove.

chain-signatures/node/src/indexer_common.rs
```diff
@@ -298,9 +303,53 @@ pub(crate) async fn recover_backlog(
...
+                    "sign type missing during backlog recovery"
+                );
+                continue;
+            };
+
+            let sign_request = IndexedSignRequest {
+                id: sign_id,
+                args: sign_tx_entry.args.clone(),
+                chain: sign_tx_entry.source_chain,
+                unix_timestamp_indexed: current_unix_timestamp(),
```
jakmeier: Hm, this feels wrong... The node indexed the request much earlier.

This is later used to report SIGN_TOTAL_LATENCY. We might end up with negative / zero latencies there after a recovery.

Would it make sense to persist `unix_timestamp_indexed` for recovery?

jakmeier (dismissed):
Good addition, nice test!

chain-signatures/node/src/indexer_common.rs
ChaoticTempest: persisting it now

volovyks (approved):

PR #638, opened 2026-01-15
feat: global node account id for metrics

ChaoticTempest:
I was getting annoyed by how we have the account id sprinkled around our code and constantly being injected and cloned across our codebase. This makes it so we no longer need to do that.

Instead, we have a global `node_account_id` that gets set at the very start of `Cli::Start`. This ID never changes between node runs anyways so we will never need to pass it around for the sake of metrics.


volovyks (commented):
good idea
I think we can make it even better by adding a [constant label](https://docs.rs/prometheus/latest/prometheus/struct.Opts.html#method.const_label)

This way we can avoid setting, then importanig account_id function and setting it in the metric.

jakmeier (dismissed):

ChaoticTempest:
I'll change it later to use const_label since it's a bit more involved than just calling into `const_label("node_account_id", node_account_id())`

volovyks (approved):

PR #639, opened 2026-01-15
fix: thundering herd issue for sign task

ChaoticTempest:
Fixes the case for sign tasks where the proposer expires and starts, but the deliberator also expires at the same time and is unable to process the start. This is due to us waiting for expiration on the deadline to start, so instead of starting on deadline, we will start on enough accepts (threshold amount). 

The caveat to this fix is that now all sign tasks will only have threshold participants ever, versus before where we would have anywhere from threshold nodes to all nodes as participants.

### DIFF
--- a/chain-signatures/node/src/protocol/signature.rs
+++ b/chain-signatures/node/src/protocol/signature.rs
@@ -532,14 +532,6 @@ impl SignPositor {
         let is_proposer = proposer == ctx.me;
         let is_deliberator = !is_proposer;
 
-        // Get the presignature participants - only these nodes participated in generating it
-        let presignature_participants = if let Some(ref taken) = presignature {
-            taken.artifact.participants.clone()
-        } else {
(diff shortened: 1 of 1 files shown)


jakmeier (approved):
Not waiting for expiration to start the protocol makes a lot of sense.

I don't see much of a problem with signing with fewer nodes, either. And if that is a problem, we can always wait for a bit longer for more responses. As long as we start well before the deliberator timeout.

Just one question: What happens with the deliberators who accepted but were not included as participants? Will they try and bump the round over and over, until the signature is completed and indexed?

ChaoticTempest:
The deliberators that accepted but do not get the start will simply just expire and go back into reorganization as expected because they should still be watchdogs just in case generation fails with the original participant set.

Later, if we want more participants, we can do something like: on enough participants, wait like 300ms, then start with the latest participant set

PR #643, opened 2026-01-19
Use unix_timestamp_indexed instead of timestamp_sign_queue for metrics

volovyks:
`timestamp_sign_queue` is reiintialized in request recovery, what we need is a real delay that is experienced by the user
The downside is that we are loosing the precision (1s now)

### DIFF
--- a/chain-signatures/node/src/rpc.rs
+++ b/chain-signatures/node/src/rpc.rs
@@ -979,7 +979,10 @@ async fn execute_publish(client: ChainClient, mut action: PublishAction, backlog
 
     let chain_str = chain.as_str();
     if publish_result.is_ok() {
-        let elapsed = action.indexed.timestamp_sign_queue.elapsed();
+        let elapsed = crate::util::duration_between_unix(
+            action.indexed.unix_timestamp_indexed,
+            crate::util::current_unix_timestamp(),
+        );
         if elapsed.as_secs() <= chain.expected_response_time_secs() {
             crate::metrics::requests::NUM_SIGN_REQUESTS_MINE_IN_TIME
                 .with_label_values(&[chain_str, node_account_id()])
@@ -1440,9 +1443,13 @@ async fn execute_batch_publish(
         };
         if publish.is_ok() {
             // Record metrics for successful batch publish
+            let current_timestamp = crate::util::current_unix_timestamp();
             for action in actions.iter() {
                 let chain = action.indexed.chain;
-                let elapsed = action.indexed.timestamp_sign_queue.elapsed();
+                let elapsed = crate::util::duration_between_unix(
+                    action.indexed.unix_timestamp_indexed,
+                    current_timestamp,
+                );
                 if elapsed.as_secs() <= chain.expected_response_time_secs() {
                     crate::metrics::requests::NUM_SIGN_REQUESTS_MINE_IN_TIME
                         .with_label_values(&[chain.as_str(), node_account_id()])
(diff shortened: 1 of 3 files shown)


volovyks:
@jakmeier I saw your comment about these fields somewhere, but I can not find it, tell me if there is anything you want to change here

ppca:
I think timestamp_sign_queue differentiates from timestamp indexed because we wait for finality?

chain-signatures/node/src/rpc.rs
```diff
@@ -1440,9 +1443,13 @@ async fn execute_batch_publish(
         };
         if publish.is_ok() {
             // Record metrics for successful batch publish
+            let current_timestamp = crate::util::current_unix_timestamp();
             for action in actions.iter() {
                 let chain = action.indexed.chain;
-                let elapsed = action.indexed.timestamp_sign_queue.elapsed();
+                let elapsed = crate::util::duration_between_unix(
```
jakmeier, line 1449: This is reported as `SIGN_TOTAL_LATENCY`. I thought that was to measure how long the signature generators took inside the MPC network, analogous to `TRIPLE_LATENCY_TOTAL`. But looking into it, it seems we have `SIGN_GENERATION_LATENCY` to measure that. And for presignatures, it is called `PRESIGNATURE_LATENCY`. So 3 different naming schemes.

But yes, assuming `SIGN_TOTAL_LATENCY` measures the delay experienced by the user, it makes sense to use `unix_timestamp_indexed` here.

jakmeier (approved):
The PR looks good to me, the changes make sense.

But I gotta admit, I find our metrics quite confusing at this point. The naming is not intuitive to me.

Also, I don't really see why we have  `unix_timestamp_indexed` and `timestamp_sign_queue`. Except for the new case of recovery, it seems we always set them at the same time.

> I think timestamp_sign_queue differentiates from timestamp indexed because we wait for finality?

That could make sense, yes. But I couldn't confirm this in the code.

chain-signatures/node/src/rpc.rs
volovyks: Here is a definition of SIGN_TOTAL_LATENCY:
```rust
pub(crate) static SIGN_TOTAL_LATENCY: LazyLock<HistogramVec> = LazyLock::new(|| {
    try_create_histogram_vec(
        "multichain_sign_latency_sec",
        "Latency of multichain signing, start from indexing sign request, end when publish() called.",
        &["chain", "node_account_id"],
        Some(exponential_buckets(0.001, 2.0, 20).unwrap()),
    )
    .unwrap()
});
```

Signature generation specifically is defined here:
```rust
// Signature metrics
pub(crate) static SIGN_GENERATION_LATENCY: LazyLock<Histogram> = LazyLock::new(|| {
    Histogram::new(
        "multichain_sign_gen_latency_sec",
        "Latency of multichain signing, from start signature generation to completion.",
        &["node_account_id"],
        Some(exponential_buckets(0.001, 2.0, 20).unwrap()),
    )
});
```

volovyks:
@jakmeier our metrics naming is misleading sometimes. Once I add a couple more things, I will apply a single standard to them. I was avoiding that not to brake our existing panels and history.

`timestamp_sign_queue` is initialized at the same place as ` unix_timestamp_indexed` and not used anywhere. I've kept it just in case we need it in the future (I will make a decision soon).

In the current setup it got nothing to do with finality, but I'm planning to track that using "staged" metrics. Working on that now.

PR #646, opened 2026-01-27
Request latancies

volovyks:
The goal of this PR is to unify, simplify, and extend the measurement of the request latencies. I've added a single metric (`multichain_sign_request_latency_sec`) with a `step` label for that purpose. Since Histogram creates `_count` metric automatically, I've removed many counters, as well as other histogram metrics that were measuring specific steps in request processing.

I recommend starting the review from the `SignRequestStep` struct. This PR does not implement measuring all the steps of the request processing. Current status is:
- Indexing: Only Ethereum has access to the block time, atm. Other chains record zeros. cc @ppca
- AwaitingGeneration: Done, but more of a placeholder. We can divide this step into Finalizing, Queueing, etc. Up for a discussion. cc @ChaoticTempest 
- Generating: Done
- Responding: Done
- Total: Done
- Other (delayed): Done

### DIFF
--- a/chain-signatures/node/src/indexer_common.rs
+++ b/chain-signatures/node/src/indexer_common.rs
@@ -273,10 +273,6 @@ pub(crate) async fn process_sign_event(
         // TODO: handle error to ensure 100% success rate
         let chain = sign_event.source_chain();
         tracing::error!(?err, chain = %chain, "Failed to send {} sign request into queue", chain.as_str());
-    } else {
-        crate::metrics::requests::NUM_SIGN_REQUESTS
-            .with_label_values(&[sign_event.source_chain().as_str()])
-            .inc();
     }
 
     Ok(())
--- a/chain-signatures/node/src/indexer_common.rs
+++ b/chain-signatures/node/src/indexer_common.rs
@@ -7,6 +7,7 @@ use crate::indexer_hydration::{
 };
 use crate::mesh::wait_threshold_active;
 use crate::mesh::MeshState;
+use crate::metrics::requests::{record_request_latency, SignRequestStep};
 
 use crate::node_client::NodeClient;
 use crate::protocol::Chain;
@@ -22,6 +23,7 @@ use k256::Scalar;
 use mpc_primitives::SignId;
 use mpc_primitives::Signature;
 use std::str::FromStr;
+use std::time::SystemTime;
 use std::time::{Duration, Instant};
 use tokio::sync::mpsc;
 use tokio::sync::watch;
@@ -232,6 +234,15 @@ pub(crate) async fn process_sign_event(
     backlog: Backlog,
 ) -> anyhow::Result<()> {
     let sign_request = sign_event.generate_sign_request(entropy, total_timeout)?;
+
+    record_request_latency(
+        sign_event.source_chain(),
+        SignRequestStep::Indexing,
+        "ok",
+        // Note: we need to pass actual indexing time here,
+        // may not be supported for all chains
+        SystemTime::now(),
+    );
 
     // Insert the transaction into the backlog when we first see the sign request
     let sign_id = sign_request.id;
--- a/chain-signatures/node/src/indexer_common.rs
+++ b/chain-signatures/node/src/indexer_common.rs
@@ -7,7 +7,7 @@ use crate::indexer_hydration::{
 };
 use crate::mesh::wait_threshold_active;
 use crate::mesh::MeshState;
-use crate::metrics::requests::{record_request_latency, SignRequestStep};
+use crate::metrics::requests::record_indexing_step_reached;
 
 use crate::node_client::NodeClient;
 use crate::protocol::Chain;
@@ -23,7 +23,6 @@ use k256::Scalar;
 use mpc_primitives::SignId;
 use mpc_primitives::Signature;
 use std::str::FromStr;
-use std::time::SystemTime;
 use std::time::{Duration, Instant};
 use tokio::sync::mpsc;
 use tokio::sync::watch;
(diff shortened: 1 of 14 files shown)


ppca:
> Indexing: Only added on Ethereum since that is where we have block time. If block time is not reported on some chains, we can send 0. cc @ppca, wdyt?

I guess this is index time stamp - block timestamp? Is it the same as indexer delay metric now? 
seems like Solana also has block time reported: https://solana.com/docs/rpc/http/getblocktime

volovyks:
@ppca, yes, SignRequestStep::Indexing is basically an INDEXER_DELAY.

I do not think this is a good idea to make an extra call just for a metric. But we can explore options.

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -48,10 +49,9 @@ pub struct IndexedSignRequest {
     /// Unix timestamp when the request was indexed by MPC node.
     /// Preserved across recoveries to maintain original request creation time.
     pub unix_timestamp_indexed: u64,
-    /// Monotonic system time when the request entered the signing queue.
-    /// Used for internal timeout enforcement and latency tracking.
-    /// Updated on recovery/requeue to current system time.
-    pub timestamp_sign_queue: Instant,
+    /// Monotonic system time when the request entered the system for processing.
+    /// Set during initial indexing or updated on recovery/requeue to current system time.
+    pub timestamp_indexed: Instant,
```
jakmeier: Having two fields named `unix_timestamp_indexed` and `timestamp_indexed` while the track slightly different things isn't great.

Unless you expect more changes to these fields soon, would it make sense to make the names more distinct? Maybe rename the second to `timestamp_created`, as in, that's when the `IndexedSignRequest` instance has been created?

chain-signatures/node/src/indexer_common.rs
```diff
@@ -233,6 +235,15 @@ pub(crate) async fn process_sign_event(
...
     let sign_request = sign_event.generate_sign_request(entropy, total_timeout)?;
 
+    record_request_latency(
+        sign_event.source_chain(),
+        SignRequestStep::Indexing,
+        "ok",
+        // Note: we need to pass actual indexing time here,
+        // may not be supported for all chains
+        SystemTime::now(),
+    );
```
jakmeier: Rather than collecting an invalid metric, why not just skip recording this step for chains that don't support it?

If this is to record the count, I would either mention it in the comment, or maybe even encapsulate it in a different method, such as `record_request_step_reached` instead of `record_request_latency`.

Otherwise, this might get removed by a well-meaning refactoring.

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -1129,29 +1132,27 @@ impl SignatureSpawner {
...
                 );
-                crate::metrics::requests::NUM_SIGN_REQUESTS_MINE_DELAYED
-                    .with_label_values(&[chain.as_str()])
-                    .inc();
+                record_request_latency(
+                    chain,
+                    crate::metrics::requests::SignRequestStep::Other,
+                    "delayed",
+                    unix_timestamp_indexed,
+                );
```
jakmeier: Is this metric still useful?

Wouldn't the reported latency always be equal to the expected response time?

And even just as a counting metric, I don't know what to make of an unspecific "other" step that was delayed. In what context would it be displayed?

chain-signatures/node/src/rpc.rs
```diff
@@ -1442,25 +1441,31 @@ async fn execute_batch_publish(
...
+                    );
                 }
-                crate::metrics::requests::SIGN_TOTAL_LATENCY
-                    .with_label_values(&[chain.as_str()])
-                    .observe(elapsed.as_secs_f64());
+                record_request_latency(
+                    chain,
+                    SignRequestStep::Responding,
+                    "ok",
+                    action.indexed.unix_timestamp_indexed,
```
jakmeier: This records the same latency for Responding as for Total.

Shouldn't it be `action.timestamp` or use the (now removed) parameter `start`?

jakmeier (dismissed):
Amazing clean up!

chain-signatures/node/src/protocol/signature.rs
volovyks: Good point, renamed

chain-signatures/node/src/indexer_common.rs
volovyks: Ok, I've added a new fn:
```rust
pub fn record_indexing_step_reached(chain: Chain)
```

chain-signatures/node/src/protocol/signature.rs
volovyks: "Delayed" is still useful for real-time SLI tracking.

Yes, it should be the same as the delay time, but what if it is not? I was thinking that it can address an [issue](https://github.com/sig-net/mpc/pull/635#discussion_r2716173752) you've mentioned earlier.

I do not like `Other` in SignRequestStep, since this is not an actual step in the request processing. I've added it here to have a single metric, but I do not mind separating it into its own counter or histogram.

Delayed is a key metric for SLI tracking.

chain-signatures/node/src/rpc.rs
volovyks: My bad,
Exactly that is why this field needs to be renamed
Fixed

chain-signatures/node/src/protocol/signature.rs
jakmeier: > "Delayed" is still useful for real-time SLI tracking.

I see. I though SignRequestStep::Total with status expired should cover this, but I guess this is not real-time.

> Yes, it should be the same as the delay time, but what if it is not? I was thinking that it can address an https://github.com/sig-net/mpc/pull/635#discussion_r2716173752 you've mentioned earlier.

Makes sense.

> I do not like Other in SignRequestStep, since this is not an actual step in the request processing. I've added it here to have a single metric, but I do not mind separating it into its own counter or histogram.

I also don't like `Other` but it doesn't matter too much. Your call on how you want to represent it.

jakmeier (dismissed):

chain-signatures/node/src/protocol/signature.rs
volovyks: I've reconsidered and replaced delayed with a separate counter
delay is printed in logs if we ever need it

jakmeier (dismissed):

jakmeier (approved):

PR #647, opened 2026-01-29
fix: solana indexer hang

### LINKED ISSUE #648: solana block height stuck
Image

ppca:
the reason why solana indexer block height stuck could be: 
1) get_transaction_with_config hangs, not getting response and hangs there forever
2) stream.next() hangs forever

in this PR, both will result in the function returning Err, then the solana client will resubscribe.

fix this issue:  https://github.com/sig-net/mpc/issues/648

### DIFF
--- a/chain-signatures/node/src/indexer_eth/mod.rs
+++ b/chain-signatures/node/src/indexer_eth/mod.rs
@@ -616,8 +616,7 @@ impl EthereumClient {
     ) -> Option<alloy::rpc::types::Block> {
         // Configure retry behaviour and delegate to shared retry_async helper.
         let retry_config = crate::util::retry::RetryConfig::default();
-        // the retry helper expects the operation closure to accept one argument (attempt/index)
-        let op = |_attempt: usize| async {
+        let get_block_op = |_attempt: usize| async {
             match self {
                 EthereumClient::Helios(client) => client.get_block(block_id).await,
                 EthereumClient::DirectRpc(client) => client.get_block(block_id).await,
@@ -626,7 +625,7 @@ impl EthereumClient {
 
         let res = crate::util::retry::retry_async(
             retry_config,
-            op,
+            get_block_op,
             |_attempt, _reason| true,
             |attempt, reason, sleep_duration| match reason {
                 crate::util::retry::RetryReason::Error(e) => {
--- a/chain-signatures/node/src/indexer_eth/mod.rs
+++ b/chain-signatures/node/src/indexer_eth/mod.rs
@@ -11,6 +11,7 @@ use crate::respond_bidirectional::CompletedTx;
 use crate::rpc::ContractStateWatcher;
 use crate::sign_bidirectional::PendingRequestStatus;
 use crate::storage::app_data_storage::AppDataStorage;
+use alloy::eips::BlockId;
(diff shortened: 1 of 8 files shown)


chain-signatures/node/src/indexer_sol.rs
```diff
@@ -327,6 +329,33 @@ impl SignatureEvent for SignBidirectionalEvent {
...
+}
+
+impl Default for TxFetchRetryConfig {
+    fn default() -> Self {
+        Self {
+            per_attempt_timeout: Duration::from_secs(3),
+            max_attempts: 5,
+            backoff_base: Duration::from_millis(300),
+            backoff_max: Duration::from_secs(5),
+            jitter_max_ms: 250,
```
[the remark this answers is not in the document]
volovyks: Ok, I see that `jitter_max_ms` is used to generate a random value, but still, the comment is a bit off.

volovyks (commented):
This is not the first place where we use exponential backoff. Can we create utils and use them across the project?

ppca:
> This is not the first place where we use exponential backoff. Can we create utils and use them across the project?

will refactor all the retries.

chain-signatures/node/src/indexer_eth/mod.rs
```diff
@@ -603,13 +603,74 @@ impl EthereumClient {
...
         &self,
         block_id: alloy::rpc::types::BlockId,
     ) -> Option<alloy::rpc::types::Block> {
-        match self {
-            EthereumClient::Helios(client) => client.get_block(block_id).await,
-            EthereumClient::DirectRpc(client) => client.get_block(block_id).await,
+        // Configure retry behaviour and delegate to shared retry_async helper.
+        let retry_config = crate::util::retry::RetryConfig::default();
+        // the retry helper expects the operation closure to accept one argument (attempt/index)
+        let op = |_attempt: usize| async {
```
volovyks: Let's not use clipping, especially when there is enough space for the full word. It makes reading code hard.

volovyks: In this case, it can even say what operation this is ("get_block" or something like that)

chain-signatures/node/src/rpc.rs
```diff
@@ -929,55 +929,65 @@ async fn execute_publish(client: ChainClient, mut action: PublishAction, backlog
...
-                    .map_err(|_| ())
+        },
+        |_attempt, _reason| true,
+        |attempt, reason, sleep| match reason {
+            RetryReason::Error(_) => {
+                tracing::warn!(
+                    ?sign_id,
+                    retry_count = attempt.saturating_sub(1),
+                    elapsed = ?action.timestamp.elapsed(),
+                    "failed to publish, retrying in {sleep:?}"
```
volovyks, line 974: Would we know what chain this is?

volovyks (dismissed):
Looks good! Nice error handling, it should give us more info in logs.

chain-signatures/node/src/rpc.rs
ppca: yes! I'll add it

volovyks:
@ppca compilation fails

volovyks (approved):

PR #650, opened 2026-01-29
Latest block for hydration

volovyks:
Also added statuses for LATEST_BLOCK, will implement later

### DIFF
--- a/chain-signatures/node/src/metrics/indexers.rs
+++ b/chain-signatures/node/src/metrics/indexers.rs
@@ -6,7 +6,7 @@ use super::try_create_int_gauge_vec_with_node_account_id;
 
 /// Possible status options:
 ///     - "indexed" - latest block number seen by the indexer
-///     - "finilized" - latest block number seen as finalized by the indexer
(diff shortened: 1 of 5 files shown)


chain-signatures/node/src/metrics/indexers.rs
```diff
@@ -4,11 +4,15 @@ use prometheus::IntGaugeVec;
 
 use super::try_create_int_gauge_vec_with_node_account_id;
 
+/// Possible status options:
+///     - "indexed" - latest block number seen by the indexer
+///     - "finilized" - latest block number seen as finalized by the indexer
```
jakmeier:
```suggestion
///     - "finalized" - latest block number seen as finalized by the indexer
```

jakmeier (dismissed):

chain-signatures/node/src/metrics/indexers.rs
volovyks: Thanks)

jakmeier (approved):

PR #656, opened 2026-02-05
Avoid bad timing on posits

jakmeier:
For presignature and triple generation, the proposer and deliberator roles time out at the same time.

The proposer tries to move forward on a timeout, if the threshold has been met. But a deliberator will simply abort. This leads to bad timing and non-determinism if even just a single node is not responding.

Increasing the deliberator timeout resolves the problem.

This also increases the frequency for checking the expiration, to speed up the process.

### DIFF
--- a/chain-signatures/node/src/protocol/posit.rs
+++ b/chain-signatures/node/src/protocol/posit.rs
@@ -345,8 +345,13 @@ impl<Id: Copy + Hash + Eq + fmt::Debug, S> Posits<Id, S> {
         timeout: Duration,
     ) -> Vec<(Id, PositInternalAction<S>)> {
         let mut expired = Vec::new();
-        for (id, (_, timestamp)) in &self.posits {
-            if timestamp.elapsed() > timeout {
+        for (id, (positor, timestamp)) in &self.posits {
+            // Deliberators need to wait longer than the proposer, otherwise
+            // they have a high chance of aborting just when the proposer
+            // decides to move forward.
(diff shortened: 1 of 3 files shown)


jakmeier:
@ChaoticTempest This is what I mentioned in today's meeting. I just saw a test (test_sign_contention_5_nodes) failed, I will need to look into that.

ChaoticTempest (dismissed):

jakmeier:
Unfortunately tests are non-deterministic.

`test_sign_contention_5_nodes` and `test_presignature_timeout` sometimes fails in CI

locally, I got `test_sign_contention_5_nodes` failing once but mostly it runs without issues.

I probably won't have time to dig deeper today. But we shouldn't merge this while the issue remains.

jakmeier:
I had to change the delay for deliberator expiration to make this work reliably.

In my initial fix, I used 2x the proposer timeout for deliberators. This makes it much more likely that round 1 of posits gets accepted. But I didn't consider that for round two, this creates another (perfectly) bad timing. 2x is pretty much the worst choice for the delay. Thus, resulting in the non-deterministic test failures I had.

Now I use a fixed 2s extra for deliberators to expire. That should be enough to receive the proposer START message in time. Combined with a proposer expiration of 10s, also leaves enough time for the next round to conclude.

volovyks (approved):

volovyks:
@ppca cases::mpc::test_sign_contention_5_nodes is still failing, often, so the issue is relevant
 cases::mpc::test_sign_requests_wait_for_presignatures is unstable as well

PR #657, opened 2026-02-05
show active posits in debug page

jakmeier:


jakmeier:
This adds another expandable section with posit tasks to the debug page.

Screenshot 2026-02-05 at 17-19-02 Debug Page

Most of the time, it shows no active posits, as they generally resolve quickly.

Screenshot 2026-02-05 at 17-18-28 Debug Page

But when there is a problem, it will show all posits that are stuck. On the page of the proposer, it also shows the number of accepts/rejects received so far.

Screenshot 2026-02-05 at 17-18-51 Debug Page

ChaoticTempest (approved):

PR #661, opened 2026-02-10
fix(audit): RUSTSEC-2026-0009

ChaoticTempest:


ChaoticTempest:
I updated it originally in cargo.lock but it required us to upgrade our near smart contract to edition 2024 with the newest time dependency. So I opted for ignoring it for now. It's a low attack vector but not completely out of the realms of possibility since we do not parse date into RFC 2822 format AFAIK. Our client could possibly but we can fix this by upgrading our near dependencies, which I will leave out for now. @volovyks didn't you start upgrading them recently in the last couple of months? Maybe it's time we get those back up

jakmeier (approved):

volovyks:
@ChaoticTempest  I've tried, but hit the very same issue. Let's get back to this once we fix our current stability issues.

PR #663, opened 2026-02-12
feat: ChainStream abstraction for indexers

ChaoticTempest:
## Current Implementation

This PR refactors our indexers under a common interface called `ChainStream`. This trait is defined as follows:
```rs
trait ChainStream {
    const CHAIN: Chain;

    fn next_event(&self) -> Option<ChainEvent>;
}
```

It purely has two entries `const CHAIN: Chain;` and `fn next_event() -> ChainEvent`. This makes it pretty simple, because each of the chains' concrete implementation purely just needs to process blocks, and emit the `ChainEvent` variants:
```rs
ChainEvent {
  Sign(event),
  Respond(event),
  RespondBidirectional(event),
  ExecutionConfirmed(...),
  Block(height),
}
```

The common event loop `run_stream` will have the common logic to then process these events and send the sign request if needed it to the signer for processing .

So at the end of the day, each new chain just processes the block and emits events. Fundamentally, not much has changed with each indexer, most of the code that was there to process the block and send a sign request is still there. But instead of sending a sign request, they emit ChainEvents instead. Some cleanup is still necessary but this is the interface has led to better testability. We now have tests against the common event stream loop, and a set of tests for each indexer. This interface makes it so we merely need to get the next event and make sure that it is indeed what we are expecting.

Most of the backlog related entries have been moved to the more common path. Not everything is on this happy common path yet (i.e. catchup and testing).

### Generalized Catchup

Catchup logic has not been generalized yet, but it's just orchestrating the following:
- get previous block height that was persisted.
- spin up live block subscription.
- use first block from live subscription.
- keep buffering the blocks from live subscription.
- catchup from previous block height to anchored block height.
- then start from buffered blocks and go all the way till we get to live subscription blocks.

So we can have something in the form of:

psuedocode:
```rs
run_stream(stream: ChainStream) {
    task = spawn({
        let buf_stream = stream.livestream().await;
        let anchor_height = buf_stream.initial().await;
        stream.catchup(anchor_height).await;

        buf_stream.run()
    })

    // ...
    // process next_event loop
    loop { stream.next_event() }
}

ChainStream {
    const CHAIN: Chain

    next_event() -> ChainEvent
    livestream() -> ChainBufferedStream
    catchup(BlockHeight) emits ChainEvent
}

ChainBufferedStream {
    initial() -> BlockHeight
    run() -> Cancellable emits ChainEvent
}
```

One thing to note is that we need to guarantee linear block processing, which is why we cannot have catchup and livestream be ran in parallel (this is unfortunately what ethereum indexing does right now).

### Generalize Indexing Testing

We should also generalize the set of tests for each of the chain streams so we don't need to rewrite the tests everytime. We can still have some chain specific tests, but the common path of checking whether our parsing logic works with each chain is best as a set of generic tests.

We can then have the following set of generalized tests:
1. parse Sign/SignBidirectional/Respond/RespondBidrectional events
2. emit block and do checkpoints regularly
3. linear catchup when starting behind
4. finality-based event processing (if possible, ethereum is hard to get finality in sandbox)
5. execution confirmation detection for bidirectional requests
6. handle concurrent submissions

With a common setup interface:
```rs
trait TestStreamSetup {
    const chain: Chain;

    fn spawn_sandbox() -> ChainSandbox;
    fn deploy_contract() -> ContractClient;
    fn stream() -> ChainStream;
}
```

## Notes

Only implemented ChainStream for Solana and Ethereum for now. So if we like what we see here, I can move onto hydration.

### DIFF
--- a/chain-signatures/node/src/stream/mod.rs
+++ b/chain-signatures/node/src/stream/mod.rs
@@ -40,7 +40,7 @@ pub enum ChainEvent {
         sign_id: mpc_primitives::SignId,
         source_chain: Chain,
         block_height: u64,
-        result: ExecutionResult,
+        result: ExecutionOutcome,
     },
 }
 
@@ -82,7 +82,7 @@ impl std::fmt::Debug for ChainEvent {
 }
 
 #[derive(Debug, Clone)]
-pub enum ExecutionResult {
+pub enum ExecutionOutcome {
     Success { output: Vec<u8> },
     Failed,
 }
--- a/chain-signatures/node/src/stream/mod.rs
+++ b/chain-signatures/node/src/stream/mod.rs
@@ -1,18 +1,20 @@
 use crate::backlog::Backlog;
-use crate::indexer_common::{
-    process_execution_confirmed, process_respond_bidirectional_event, process_respond_event,
-    process_sign_request, recover_backlog, RespondBidirectionalEvent, SignatureRespondedEvent,
-};
 use crate::mesh::MeshState;
 use crate::node_client::NodeClient;
 use crate::protocol::IndexedSignRequest;
 use crate::protocol::{Chain, Sign};
 use crate::rpc::ContractStateWatcher;
 use crate::sign_bidirectional::BidirectionalTxId;
+use crate::stream::ops::{
+    process_execution_confirmed, process_respond_bidirectional_event, process_respond_event,
+    process_sign_request, recover_backlog, RespondBidirectionalEvent, SignatureRespondedEvent,
+};
 
 use std::time::Duration;
 use tokio::sync::mpsc;
 use tokio::sync::watch;
+
+pub mod ops;
 
 pub const CHAIN_EVENT_STREAM_SIZE: usize = 16384;
 
@@ -21,6 +23,7 @@ pub fn channel() -> (mpsc::Sender<ChainEvent>, mpsc::Receiver<ChainEvent>) {
 }
 
 /// Unified event produced by a chain stream
+#[allow(clippy::large_enum_variant)]
 pub enum ChainEvent {
     SignRequest(IndexedSignRequest),
     Respond(SignatureRespondedEvent),
@@ -169,7 +172,6 @@ pub async fn run_stream<S: ChainStream>(
 mod tests {
     use super::*;
     use crate::backlog::Backlog;
-    use crate::indexer_common::SignatureRespondedEvent;
     use crate::mesh::MeshState;
     use crate::node_client::NodeClient;
     use crate::protocol::Chain;
@@ -177,6 +179,7 @@ mod tests {
     use crate::protocol::Sign;
     use crate::protocol::SignRequestType;
     use crate::rpc::ContractStateWatcher;
+    use crate::stream::ops::SignatureRespondedEvent;
     use crate::util::current_unix_timestamp;
     use k256::Scalar;
     use mpc_primitives::SignArgs;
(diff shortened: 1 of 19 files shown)


indexer_eth.rs.diff
```diff
@@ -0,0 +1,50 @@
...
++                        .await;
++                }
++            } else {
++                stuck_at_genesis_count = 0;
++            }
++
++            let new_finalized_block = match new_finalized_block {
+                 Some(block) => block,
+                 None => {
+                     tracing::warn!("Finalized block not found from Ethereum client");
```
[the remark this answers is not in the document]
ChaoticTempest: committed by mistake

setup.sh
```diff
@@ -19,16 +19,16 @@ if echo "$CARGO_CMD_ARGS" | grep -q "bench"; then
...
+# if [ -n "$CONTRACT_FEATURES" ]; then
+#     . $ROOT_DIR/build-contract.sh $CONTRACT_FEATURES
+# else
+#     . $ROOT_DIR/build-contract.sh
+# fi
 
-cargo build -p mpc-node --release $NODE_FEATURE_ARGS
+# cargo build -p mpc-node --release $NODE_FEATURE_ARGS
 
 exec $CARGO_CMD_ARGS
```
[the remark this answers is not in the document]
ChaoticTempest: also committed by mistake

chain-signatures/node/src/stream/ops.rs
```diff
@@ -182,33 +175,105 @@ impl RespondBidirectionalEvent {
...
+    /// from the provided `r` and `v` using field arithmetic (try x = r and
+    /// x = r + n) and take the solution whose y parity matches `v`.
     pub fn signature(&self) -> Signature {
         match self {
             SignatureRespondedEvent::Solana(event) => {
                 crate::indexer_sol::to_mpc_signature(event.signature.clone()).unwrap()
             }
             SignatureRespondedEvent::Hydration(event) => event.signature.clone(),
+            SignatureRespondedEvent::Ethereum(event) => {
+                // Try reconstructing a real signature; fall back to zeroed values
```
jakmeier: Not something to fix in this PR - but do you know why we have an invalid fallback signature? Seems sketchy 🤔

chain-signatures/node/src/stream/mod.rs
```diff
@@ -0,0 +1,529 @@
...
+        result: ExecutionResult,
+    },
+}
+
+#[derive(Debug, Clone)]
+pub enum ExecutionResult {
+    Success { output: Vec<u8> },
+    Failed,
+}
+// TODO: need to add sign_id to each event for better logging.
```
jakmeier: Good idea. While at it, maybe also add the chain id to log outputs where it is available? Before it was implicit from the file.

integration-tests/src/containers.rs
```diff
@@ -652,6 +654,8 @@ impl Solana {
             .arg("127.0.0.1")
             .arg("--reset")
             .arg("--quiet")
+            .stdout(async_process::Stdio::null())
+            .stderr(async_process::Stdio::null())
```
jakmeier: Why throw away the output? Wouldn't that make it harder to debug failing tests?

jakmeier (dismissed):
Wow, a great refactor and a ton more test coverage! Very nice!

chain-signatures/node/src/stream/ops.rs
ChaoticTempest: yeah, probably need to find a better way for this one. We're reconstructing bigR and it may fail here. The fallback was there just as a placeholder but really should not be possible. Probably should just bail at that point

integration-tests/src/containers.rs
ChaoticTempest: forget why I added this. Probably cause it was noisy. Will remove it

chain-signatures/node/src/stream/ops.rs
jakmeier: yeah bailing seems much better than returning invalid data

but let's keep the behavior the old way in this PR

ChaoticTempest: ahh it was using the old (v, r, s) fields but since we transitioned to the new (bigR, s, recovery_id) fields for the newest ethereum respond event, this is no longer needed at all. We just reconstruct it during parsing time, so it doesn't get to this point.

chain-signatures/node/src/stream/mod.rs
ChaoticTempest: it now shows chain in the debug as well

jakmeier (approved):

PR #664, opened 2026-02-12
log triple pair id, presig id as string

ppca:


jakmeier (approved):
Makes sense, thanks for the quick PR!

See https://github.com/sig-net/mpc-private/issues/12#issuecomment-3891294002 for context. But basically, if we log them as number types, the log display converts ids to `f64` which cannot represent the full range of `u64` precisely.

ChaoticTempest:
I think it's time we convert triple id and presignature id to newtypes tuples. It's a little less ergonomic for testing and more verbose but its pretty easy for us to forget about this and just set it as id instead of ?id in tracing. A newtype would solve that issue alongside having other pros with pattern matching

PR #665, opened 2026-02-18
Bidirectional State Sync

volovyks:
This is a step forward towards State Sync design + improved error handling and trasing

- Updated `/sync` API, now it returns a list of not found Ts and Ps for processing by the caller
- Implemented `not_found` on the storage layer 
- Improved error handling, now we are not threatening failed `/sync` calls as successful
- Now we are not returning specific error messages to the caller for security reason, that is a bad practice
- Added request ID middleware to track each request. In case we get an error, it should be easy to go and filter for logs with a specific ID on a specific node

The next step is to process `not_found` on the caller side.

### DIFF
--- a/chain-signatures/node/src/web/mod.rs
+++ b/chain-signatures/node/src/web/mod.rs
@@ -116,9 +116,7 @@ async fn request_id_middleware(mut req: Request<Body>, next: Next) -> Response {
     req.extensions_mut().insert(request_id.clone());
 
     let span = tracing::info_span!("request", %request_id);
-    let _guard = span.enter();
-
-    let mut response = next.run(req).await;
+    let mut response = next.run(req).instrument(span).await;
     if let Ok(value) = HeaderValue::from_str(&request_id) {
         response.headers_mut().insert(header_name, value);
     }
--- a/chain-signatures/node/src/web/mod.rs
+++ b/chain-signatures/node/src/web/mod.rs
@@ -34,6 +34,7 @@ use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::sync::Arc;
 use std::time::Instant;
+use tracing::Instrument;
 
 struct AxumState {
     node: NodeStateWatcher,
--- a/chain-signatures/node/src/web/mod.rs
+++ b/chain-signatures/node/src/web/mod.rs
@@ -269,13 +269,13 @@ async fn bench_metrics() -> Json<BenchMetrics> {
 async fn sync(
     Extension(state): Extension<Arc<AxumState>>,
     WithRejection(Cbor(update), _): WithRejection<Cbor<SyncUpdate>, Error>,
-) -> Result<Json<()>> {
+) -> Result<Cbor<SyncUpdate>> {
     let start = Instant::now();
-    state.sync_channel.request_update(update).await;
+    let response = state.sync_channel.request_update(update).await?;
     WEB_ENDPOINT_LATENCY
         .with_label_values(&["sync"])
         .observe(start.elapsed().as_millis() as f64);
-    Ok(Json(()))
+    Ok(Cbor(response))
 }
 
 #[derive(Debug, Deserialize)]
--- a/chain-signatures/node/src/web/mod.rs
+++ b/chain-signatures/node/src/web/mod.rs
@@ -17,8 +17,11 @@ use crate::web::cbor::Cbor;
 use crate::web::error::Result;
 
 use anyhow::Context;
+use axum::body::Body;
 use axum::extract::{DefaultBodyLimit, Query};
-use axum::http::StatusCode;
+use axum::http::{HeaderName, HeaderValue, Request, StatusCode};
+use axum::middleware::{self, Next};
+use axum::response::Response;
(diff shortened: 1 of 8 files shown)


chain-signatures/node/src/web/mod.rs
```diff
@@ -91,14 +94,37 @@ pub async fn run(
...
+        .and_then(|value| value.to_str().ok())
+        .map(|value| value.to_string())
+        .unwrap_or_else(|| hex::encode(rand::random::<u128>().to_be_bytes()));
+
+    req.extensions_mut().insert(request_id.clone());
+
+    let span = tracing::info_span!("request", %request_id);
+    let _guard = span.enter();
+
+    let mut response = next.run(req).await;
```
jakmeier: Holding an EnteredSpan like this across an await point generally leads to weird observations. See documentation here: https://docs.rs/tracing/latest/tracing/span/struct.EnteredSpan.html#in-asynchronous-code

Please use `instrument` here. I think this should work:


```suggestion
    let span = tracing::info_span!("request", %request_id);
    let mut response = next.run(req).instrument(span).await;
```

integration-tests/tests/cases/store.rs
```diff
@@ -153,9 +153,14 @@ async fn test_triple_persistence() -> anyhow::Result<()> {
     }
 
     // Let's say Node1 somehow used up triple 10, 11, 12 so we only have 13,14,15
-    let mut outdated = triple_storage.remove_outdated(node1, &[13, 14, 15]).await;
+    let result = triple_storage
+        .remove_outdated(node1, &[13, 14, 15, 99])
```
jakmeier, line 158: would be nice to update the comment to mention 99 as well

integration-tests/tests/cases/store.rs
```diff
@@ -292,11 +297,14 @@ async fn test_presignature_persistence() -> anyhow::Result<()> {
     }
 
     // Let's say Node1 somehow used up triple 10, 11, 12 so we only have 13,14,15
-    let mut outdated = presignature_storage
-        .remove_outdated(node1, &[13, 14, 15])
-        .await;
+    let result = presignature_storage
+        .remove_outdated(node1, &[13, 14, 15, 99])
```
jakmeier, line 303: same here about the comment

chain-signatures/node/src/web/mod.rs
```diff
@@ -91,14 +94,37 @@ pub async fn run(
...
+        .layer(middleware::from_fn(request_id_middleware))
+        .layer(Extension(Arc::new(axum_state)));
 
     let addr = format!("0.0.0.0:{port}");
     let listener = tokio::net::TcpListener::bind(&addr).await.unwrap();
     tracing::info!(?addr, "starting http server");
     axum::serve(listener, app).await.unwrap();
 }
 
+async fn request_id_middleware(mut req: Request<Body>, next: Next) -> Response {
```
jakmeier, line 108: Nit for next time: I think this middleware change could have been a separate PR, I don't see how it is linked to any of the state sync changes.

Reasoning: Aside from an easier review process, it's also better if something needs to be bisected and/or reverted later.

chain-signatures/node/src/web/mod.rs
```diff
@@ -91,14 +94,37 @@ pub async fn run(
...
+        .unwrap_or_else(|| hex::encode(rand::random::<u128>().to_be_bytes()));
+
+    req.extensions_mut().insert(request_id.clone());
+
+    let span = tracing::info_span!("request", %request_id);
+    let _guard = span.enter();
+
+    let mut response = next.run(req).await;
+    if let Ok(value) = HeaderValue::from_str(&request_id) {
+        response.headers_mut().insert(header_name, value);
```
jakmeier, line 122: Setting the header on the response makes sense. But is it actually used/seen anywhere so far? Would it make sense to include it in `RequestError`? (Can also be a new PR)

jakmeier (dismissed):
Nice step towards state sync going both ways!

I also like the addition of request ids for matching requests and responses in logs.

chain-signatures/node/src/web/mod.rs
volovyks: I've noticed that the storage layer can expose information we do not want to make public, since we blindly propagate the error to the top level. Request Id is a result of us losing error info. But I agree with you, it should be a separate PR.

jakmeier (approved):

chain-signatures/node/src/web/mod.rs
volovyks: Putting the request ID in the header is a standard practice, but I agree that having it in the body can be more transparent. I intended to use it for production debugging; the header should be sufficient for that. Let's get back to it in a separate PR if we have a good reason to do so.

jakmeier: I don't mean to put it in the body. I mean to extract it from the header and display it in the logs when a request sent through a reqwest HTTP client fails. (e.g. `RequestError::Unsuccessful` could be extended to `Unsuccessful(StatusCode, Body, RequestId)`

jakmeier (approved):

PR #669, opened 2026-02-23
Remove `stable` node state

volovyks:
- Removed `stable` as it is redundant now
- Made MeshState fields private to make sure we never mark a node as active during the sync process, etc.


jakmeier (approved):
Yeah I think `stable`has indeed become redundant. Seems cleaner to remove it. 👍 

I guess this is also a step towards triggering state sync upon (re-)establishing connections only?

volovyks:
@jakmeier, yes. From what I see, that should not be hard. The main issue is that we are still losing Ps; scheduled sync helps with that. We need to fix that first.

PR #670, opened 2026-02-24
Drop all connections when not a participant

volovyks:


jakmeier (approved):
Interesting, has this been a problem in practice? It seems it should be a rare case for participants to leave.

But yes, closing all open connections makes sense to me.

volovyks:
@jakmeier, no, we never faced this issue, more of a preventative measure.

PR #671, opened 2026-02-26
Bidirectional Sync (Part 2)

volovyks:
- Added bidirectional state sync. Now, when A calls B, B returns a list of Ids that were not found in node B's storage. Node A is removing node B from holders of these `not_found` artifacts.
- Now we distinguish `participants` and `holders`. Participants are those who participated in the generation, and holders are those who still have the shares. The list of participants is not used anywhere, but I decided to keep it. It can be useful for debugging, etc.
- Holders are not a part of the centralized artifact for efficiency.
- `fetch_owned` now returns Result, so we can get an empty list of owned artifacts and send it to other nodes
- While processing the sync response, Node A will prune artifacts if the number of holders is < T. If any of the artifacts were pruned, we may want to run the sync again to remove them from the remaining holders list, but I avoided that complexity for now.
- Now the sync process is not considered completed if any of the steps fail
- I've removed artifact reinsertion and added a check for the number of active participants when stockpiling presignatures to prevent wasting them (such a check was already added for triples)

We need to decide whether we want to include generating/reserved/used in the state sync. Details: https://github.com/sig-net/mpc/pull/671#issuecomment-4025655091 Can be addressed separately.


chain-signatures/node/src/protocol/sync/mod.rs
```diff
@@ -384,39 +574,3 @@ impl SyncChannel {
...
     }
 }
-
-#[cfg(test)]
-mod tests {
-    use super::*;
-    use crate::node_client::Options as NodeClientOptions;
-
-    #[tokio::test]
-    async fn test_broadcast_sync_on_empty_update() {
```
volovyks, line 394: This logic was unclear to me. When the owner sends an empty update, it means they no longer hold any Ts or Ps, and other nodes need to clean up their storage accordingly. This situation can occur if someone drops their storage or uses all of their owned Ts or Ps.

chain-signatures/node/src/storage/protocol_storage.rs
```diff
@@ -433,6 +433,34 @@ impl<A: ProtocolArtifact> ProtocolStorage<A> {
         self.used.read().await.contains(&id)
     }
 
+    /// Fetch an artifact from storage without removing it.
+    pub async fn fetch(&self, id: A::Id) -> Option<A> {
```
volovyks: Fetch/Update is a TMP dumb version of managing holders in storage. I'm working on a better implementation now.

integration-tests/tests/cases/sync.rs
```diff
@@ -176,134 +88,110 @@ async fn test_state_sync_e2e_large_outdated_stockpile() {
...
-) {
-    for id in range {
-        triples
-            .reserve(id)
-            .await
-            .unwrap()
-            .insert(dummy_pair(id), node)
-            .await;
+#[test_log::test(tokio::test)]
+async fn test_state_sync_e2e() {
```
volovyks, line 92: For now, I'm using a simple Integration test for State sync. I may work on the component layer implementation later.

chain-signatures/node/src/protocol/sync/mod.rs
volovyks: @ChaoticTempest is that an attempt to handle storage errors? `fetch_owned` returns a Vec no matter what:
```rust
pub async fn fetch_owned(&self, me: Participant) -> Vec<A::Id>
```

volovyks: Now fetch_owned returns a Result<>

ChaoticTempest: `Result<>` works, but I switched it to be `Option<Vec<A::Id>>` in the other PR. Just more convenient unless you want to define all the various redis errors that can occur. 

it's not super great having purely just `Vec<A::Id>` since returning `Vec::new()` could just mean connection failure, thus erasing all the non-owner shares if it were to be broadcasted. So empty update should just mean that we have no owned presignatures or triples. And that is right, either in case some drops storage or have no owned primitives

volovyks:
After I removed the union of `owned` and `reserved` in State Sync, the tests are passing successfully (except for cases::mpc::test_sign_contention_5_nodes).
That is expected, since `reserved` includes Ids that are not owned by this node.

Overall, the purpose of `used`, `reserved`, and `ArtifactSlot` is not fully clear to me. We need to at least fully document it.

Each T or P is going through "generating" -> "stored" -> "used" lifecycle, and `reserved` or any other additions adds complexity that may not be required.

chain-signatures/node/src/protocol/sync/mod.rs
```diff
@@ -27,6 +27,9 @@ pub const RECURRING_SYNC_INTERVAL: Duration = Duration::from_secs(3600 * 24);
 /// Timeout for waiting for a sync response from the sync task
 const SYNC_RESPONSE_TIMEOUT: Duration = Duration::from_secs(5);
 
+/// Timeout for the entire broadcast operation (waiting for all peers to respond)
+const BROADCAST_TIMEOUT: Duration = Duration::from_secs(10);
```
jakmeier, line 31: 10s feels a bit short, sync operations can require many DB reads for large state.
But we can try it and increase it if we run into it.

chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -56,14 +57,18 @@ impl FullPresignatureId {
 pub struct Presignature {
     pub id: PresignatureId,
     pub output: PresignOutput<Secp256k1>,
+    /// Original protocol participants
     pub participants: Vec<Participant>,
+    /// Nodes still holding their share of the artifact
+    pub holders: Option<Vec<Participant>>,
```
jakmeier, line 63: Why do we need to track these separately? Could we not just remove non-holders from the participants list?

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -623,8 +631,8 @@ impl SignPositor {
                         if counter.enough_rejects(ctx.threshold) {
                             tracing::warn!(?sign_id, ?round, ?from, "received enough REJECTs, reorganizing");
                             if let Some(taken) = presignature {
-                                tracing::warn!(?sign_id, "recycling presignature due to REJECTs");
-                                ctx.presignatures.recycle_mine(ctx.me, taken).await;
+                                tracing::warn!(?sign_id, "dropping presignature due to REJECTs");
```
jakmeier: Is this change required for the two way sync? Seems like an independent change.

jakmeier: Oh I see you just submitted https://github.com/sig-net/mpc/pull/697

So I guess it can be removed from this PR?

volovyks: Yes, just removed it

jakmeier:
> After I removed the union of `owned` and `reserved` in State Sync, the tests are passing successfully (except for cases::mpc::test_sign_contention_5_nodes). That is expected, since `reserved` includes Ids that are not owned by this node.
> 
> Overall, the purpose of `used`, `reserved`, and `ArtifactSlot` is not fully clear to me. We need to at least fully document it.
> 
> Each T or P is going through "generating" -> "stored" -> "used" lifecycle, and `reserved` or any other additions adds complexity that may not be required.

IIRC, reserved is to track ids that are in the generating state. I suggest renaming or removing it if we can track "generating" in other ways.

volovyks:
Here is the implementation of `reserve()`:
```rust
pub async fn reserve(&self, id: A::Id) -> Option<ArtifactSlot<A>> {
        let used = self.used.read().await;
        if used.contains(&id) {
            return None;
        }
        if !self.reserved.write().await.insert(id) {
            return None;
        }
        drop(used);

        let start = Instant::now();
        let Some(mut conn) = self.connect().await else {
            self.reserved.write().await.remove(&id);
            return None;
        };

        // Check directly whether the artifact is already stored in Redis.
        let artifact_exists: Result<bool, _> = conn.hexists(&self.artifact_key, id).await;
        let elapsed = start.elapsed();
        crate::metrics::storage::REDIS_LATENCY
            .with_label_values(&[A::METRIC_LABEL, "reserve"])
            .observe(elapsed.as_millis() as f64);

        match artifact_exists {
            Ok(true) => {
                // artifact already stored, reserve cannot be done, remove reservation
                self.reserved.write().await.remove(&id);
                None
            }
            // artifact does not exist, reservation successful
            Ok(false) => Some(ArtifactSlot {
                id,
                storage: self.clone(),
                stored: false,
            }),
            Err(err) => {
                self.reserved.write().await.remove(&id);
                tracing::warn!(id, ?err, ?elapsed, "failed to reserve artifact");
                None
            }
        }
    }
```

I'm afraid it is much more complicated than just "generating". I'm looking into it now, but I want to address it separately.

chain-signatures/node/src/protocol/presignature.rs
volovyks: We absolutely can. But I wanted to distinguish holders and participants. Mostly for debugging and storage analysis. Also, `holders` are not a part of the serialized object to avoid deserialization/serailzation of each artifact.

jakmeier: Okay, it just seemed like a lot of extra complexity with the separate tracking in Redis. But if you see enough value in having it, I have no problem with it.

chain-signatures/node/src/storage/protocol_storage.rs
```diff
@@ -186,21 +194,20 @@ impl<A: ProtocolArtifact> ProtocolStorage<A> {
...
-            .inspect_err(|err| {
+            .map_err(|err| {
                 tracing::warn!(?err, "failed to fetch my owned artifacts");
-            })
-            .unwrap_or_default();
+                StorageError::RedisFailed(err.to_string())
+            })?;
 
-        owned.union(&*self.reserved.read().await).copied().collect()
+        Ok(owned.into_iter().collect())
```
jakmeier, line 210: Hm, so if we move ahead with this change, we open the door for the race condition @ChaoticTempest described here https://github.com/sig-net/mpc/pull/649#issuecomment-3835978812

But I guess since all tests pass, it is no too common. I say we can  merge this as-is and address the race condition in a following PR.

jakmeier (approved):

jakmeier:
> I'm afraid it is much more complicated than just "generating". I'm looking into it now, but I want to address it separately.

Yes, it manages exclusive access to a redis entry. We should probably keep that as it is. Even if we can simplify it, I wouldn't do that together with these changes.

However, with respect to state sync, what we have marked as "reserved" can be treated the same as "Generating" or "Available". (See this table: https://github.com/sig-net/mpc/blob/develop/doc/mpc_node_specification.md#non-owner-action-on-state-sync)

So, the existing union kind of makes sense. But the problem is (as you [pointed out here](https://github.com/sig-net/mpc/pull/649#issuecomment-3999022891) that it also includes non-owned entries.

For Ts, until generation is done, we simply don't know who will be the owner. Maybe instead of a union, we should sync reserved ids in a separate list. The peer will then know not to delete local Ts but otherwise can ignore the list of reserved ids.

chain-signatures/node/src/storage/protocol_storage.rs
volovyks: Yes, reserved does not represent "owned by me now". For triples that are generating, that is not even possible. I'm looking into that now, not sure if that is a real concern.

PR #672, opened 2026-03-02
remove old non-cpi event subscription from solana

### LINKED ISSUE #519: Remove stale event!() parsing from Solana indexer
(after new contract and node release on all envs)

ppca:
https://github.com/orgs/sig-net/projects/2/views/1?pane=issue&itemId=133120528&issue=sig-net%7Cmpc%7C519

### DIFF
--- a/chain-signatures/node/src/indexer_sol.rs
+++ b/chain-signatures/node/src/indexer_sol.rs
@@ -6,15 +6,12 @@ use crate::util::retry::{retry_async, RetryConfig, RetryError, RetryReason};
 
 use std::collections::HashMap;
 use std::fmt;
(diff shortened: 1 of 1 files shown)


volovyks (commented):
Looks like our tests are failing, because they rely on an old logic
The comment from Copilot may be relevant

ppca:
tests not passing is not related to the change, those were signature tests and 1 test_solana_stream_concurrent_events. former is likely the unstableness that is known and it has nothing to do with indexer; latter is solana_sandbox() issue, which happens once in a while, i'm fixing it in another pr. @volovyks

volovyks (approved):

Issue #682, opened 2026-03-04
test_solana_stream_concurrent_events test failing


ppca:

volovyks:
@ppca is this relevant?

ppca:
already fixed

PR #683, opened 2026-03-04
fix test_solana_stream_concurrent_events

ppca:


volovyks:
Closing: #682 was closed in June and the test runs in the current stream harness; if it flakes again we should fix forward.

PR #685, opened 2026-03-04
Bump storage version v11

volovyks:
I wanted to do this in one of my PRs, but that will not be a clean experiment. I want to know for sure how resetting storage will affect our success rate.

### DIFF
--- a/chain-signatures/node/src/storage/mod.rs
+++ b/chain-signatures/node/src/storage/mod.rs
@@ -10,7 +10,7 @@ pub use protocol_storage::StorageError;
 pub use triple_storage::TripleStorage;
 
 // Can be used to "clear" redis storage in case of a breaking change
-pub const STORAGE_VERSION: &str = "v9";
+pub const STORAGE_VERSION: &str = "v11";
 
 /// Configures storage.
 #[derive(Debug, Clone, clap::Parser)]


chain-signatures/node/src/storage/mod.rs
```diff
@@ -10,7 +10,7 @@ pub use protocol_storage::StorageError;
 pub use triple_storage::TripleStorage;
 
 // Can be used to "clear" redis storage in case of a breaking change
-pub const STORAGE_VERSION: &str = "v9";
+pub const STORAGE_VERSION: &str = "v11";
```
jakmeier, line 13: Is there a reason to skip v10? I guess it doesn't matter, I'm just curious if there is a good reason.

jakmeier (approved):
Seems like a good experiment to do.

Just one question. Since it is somewhat costly to reset Ts and Ps in testnet and mainnet, would you revert this change if it doesn't show a positive effect? Reverting storage versions can be quite messy, that's why I'm asking.

chain-signatures/node/src/storage/mod.rs
volovyks: That is because we were adding a v10 patch with a version bump on Mainnet

volovyks:
@jakmeier I was not planning to revert it. There would be a ton of changes when we release the current develop on mainnet. I wanted to start from scratch. But not strong opiion here.

jakmeier:
> @jakmeier I was not planning to revert it. There would be a ton of changes when we release the current develop on mainnet. I wanted to start from scratch. But not strong opiion here.

Yeah makes sense, let's not revert it in any case.

PR #686, opened 2026-03-05
test: stability of test_sign_contention_5_nodes

jakmeier:
- increase stockpile_max to have more T+P generations in parallel
- increase time for T+P generation by 1 min
- increase time for signature generation by 1 min

None of these are perfect solutions.
The generation really should be faster.
But this might help with sporadic CI error that we currently face.


jakmeier:
It's not looking great for `test_sign_contention_5_nodes` stability`.

At least run 1 was successful. (`test_signature_rogue` did hit a documented race condition but that is unrelated)

But already run 2 hits a signature generation timeout again in `test_sign_contention_5_nodes`

Issue #681, opened 2026-03-04
cases::mpc::test_sign_contention_5_nodes and cases::mpc::test_sign_requests_wait_for_presignatures unstable


ppca:

jakmeier:
For `test_sign_contention_5_nodes`, if it's just too slow sometimes, I had better timing for it locally when I increased `STOCKPILE_MAX`. Then it runs more generators in parallel. Not sure how much it helps in CI, though.

jakmeier:
Continuing the discussion from https://github.com/sig-net/mpc/pull/656 here:

@volovyks @ppca Yeah, even with my supposed fix, timing can still be an issue. Specifically for presignatures. Which can be a reason for these tests to fail sporadically.

I believe the correct solution for that problem is to do what I did in https://github.com/sig-net/mpc/pull/626. That is, keep track of messages per round and allow for overlapping. I already wrote that this is the long-term solution [in comments](https://github.com/jakmeier/mpc/blob/365b9c47e0ff6fab4b040defaa1e79f6bf8a6938/chain-signatures/node/src/protocol/posit.rs#L343-L349).

But sometimes the test times out during the signature generation. That seems to be an unrelated problem, as my PR only changed posits for T and P generation.

The problem is that posits for T and P are currently handled differently from posits for signatures. For signatures, the entire flow of posits + the generator uses one task to handle all of the requests lifetime. (As outlined in https://github.com/sig-net/mpc/issues/559) For T and P, we still store the state for all requests in the shared spawner task.

On one hand, I'm tempted to say we should finish the refactor started in https://github.com/sig-net/mpc/pull/594. That would allow us resolve the P posit timing issues by apply the same solution that we had for signatures.

On the other hand, the said refactor for signature task is a suspected candidate for devnet instability we have right now.  And this _might_ be related to the cases when the tests timeout during signature generation. Therefore, I'd rather get a handle on the signature generation problem first before we resume the refactor.

But in the meantime, `test_sign_contention_5_nodes` is kind of a bad test, as it has at least two different non-deterministic failure modes. 😞 I'm testing a few tweaks in #686. If that passes CI more reliably, we could merge it as a temporary solution.

jakmeier:
I've analyzed both known error cases in `test_sign_contention_5_nodes` further.

# Presignature time out

When P generation times out, it has a very uneven distribution of Ps. I think it could be that we need a better way to assign T and P to owners.

```
(Waiting for 3 Ps each)
[...]

PR #687, opened 2026-03-05
run triple poke in blocking task

jakmeier:
Tokio futures should not run for long.
The poke on triples may take several seconds of CPU time. During this time, other tasks on the same executor will be blocked.

Spawning the work to a separate thread avoids this issue.

### DIFF
--- a/chain-signatures/node/src/protocol/triple.rs
+++ b/chain-signatures/node/src/protocol/triple.rs
@@ -42,7 +42,9 @@ struct TripleGenerator {
     me: Participant,
     proposer: Participant,
     participants: Vec<Participant>,
-    protocol: TripleProtocol,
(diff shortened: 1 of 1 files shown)


jakmeier:
Triple pokes blocking execution of other tasks would explain the observed START delay, as described in https://github.com/sig-net/mpc/issues/681#issuecomment-4006062800, see step 1 o signature timeouts.

The CI run so far confirms this suspicion. From 8s before, down to 0.75s. (But this could also be noisy data.)

And, of course, the test still fails for other reasons. :see_no_evil:

jakmeier:
On second try, it passed the test. :sweat_smile: 

Please review it, I would like to merge it as a partial fix to ongoing stability issues.

volovyks (approved):

PR #692, opened 2026-03-08
fix: stream respond error

ChaoticTempest:
This fixes the issue with the following error  `failed to process respond event`: `sign type not found for respond event (may have already been processed)`.

This is due to cases like sign type not being recovered during recovery of the checkpoint so it appears blank. The non-recovered sign type then proceeds to make the `process_respond_event` skip a respond event and not clearing the sign request from the backlog. Also added an idempotent test to make sure the same respond event does not trigger any errors.

The fix here is adding the sign type to be serialized and recovered now too. This means we have to serialize the bidrectional events unfortunately. This is a storage breaking change so CHECKPOINT_VERSION is now `v1`

### DIFF
--- a/chain-signatures/contract-sol/Cargo.toml
+++ b/chain-signatures/contract-sol/Cargo.toml
@@ -10,3 +10,4 @@ name = "signet_program"
 
 [dependencies]
 anchor-lang = { version = "0.31.1", features = ["event-cpi"] }
+serde.workspace = true
--- a/chain-signatures/node/src/storage/checkpoint_storage.rs
+++ b/chain-signatures/node/src/storage/checkpoint_storage.rs
@@ -10,7 +10,7 @@ use tokio::sync::RwLock;
 use std::collections::HashMap;
 use std::sync::Arc;
 
-const CHECKPOINT_VERSION: &str = "v0";
+const CHECKPOINT_VERSION: &str = "v1";
 
 #[derive(Clone, Debug)]
 pub enum CheckpointStorage {
--- a/chain-signatures/node/src/stream/ops.rs
+++ b/chain-signatures/node/src/stream/ops.rs
@@ -19,7 +19,7 @@ use std::str::FromStr;
 use std::time::{Duration, Instant};
 use tokio::sync::{mpsc, watch};
 
-#[derive(Debug, Clone, PartialEq, Eq, Hash)]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, serde::Serialize, serde::Deserialize)]
 pub enum SignBidirectionalEvent {
     Solana(signet_program::SignBidirectionalEvent),
     Hydration(HydrationSignBidirectionalRequestedEvent),
@@ -401,9 +401,12 @@ pub(crate) async fn process_respond_event(
     let sign_id = SignId::new(respond_event.request_id());
     let source_chain = respond_event.source_chain();
     let Some(entry) = backlog.get(source_chain, &sign_id).await else {
-        anyhow::bail!(
-            "sign request not found for respond event: {sign_id:?} on chain {source_chain:?}"
+        tracing::info!(
+            ?sign_id,
+            ?source_chain,
+            "respond event is already finalized or pruned; skipping"
         );
+        return Ok(());
     };
 
     let event = match entry.sign_type {
--- a/chain-signatures/node/src/stream/ops.rs
+++ b/chain-signatures/node/src/stream/ops.rs
@@ -421,10 +421,14 @@ pub(crate) async fn process_respond_event(
     let target_chain = Chain::from_str(&event.dest())
         .map_err(|err| anyhow::anyhow!("unable to parse target chain from dest: {err:?}"))?;
     if !matches!(entry.tx, BacklogTransaction::Sign(_)) {
-        let entry_type = entry.tx.typename();
-        anyhow::bail!(
-            "expected Sign transaction type for respond event, found different type={entry_type}: {sign_id:?}, {source_chain:?}, {target_chain:?}"
+        tracing::info!(
+            ?sign_id,
+            ?source_chain,
+            ?target_chain,
+            entry_type = %entry.tx.typename(),
+            "respond event backlog entry is already advanced; treating as processed"
         );
+        return Ok(());
     }
 
     let mpc_sig = respond_event.signature();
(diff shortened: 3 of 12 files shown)


chain-signatures/node/src/stream/ops.rs
```diff
@@ -418,15 +419,13 @@ pub(crate) async fn process_respond_event(
...
-    let target_chain = match target_chain {
-        Ok(chain) => chain,
-        Err(_) => Chain::Ethereum,
-    };
-
-    let Some(BacklogTransaction::Sign(_)) = backlog.get(source_chain, &sign_id).await else {
-        anyhow::bail!("bidirectional tx not found for advancement: {sign_id:?}");
-    };
+        .map_err(|err| anyhow::anyhow!("unable to parse target chain from dest: {err:?}"))?;
+    if !matches!(entry.tx, BacklogTransaction::Sign(_)) {
```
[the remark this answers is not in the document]
ChaoticTempest: to be considered for parsing `dest` and what format we should be considering for it

chain-signatures/contract-sol/Cargo.toml
```diff
@@ -9,4 +9,5 @@ crate-type = ["lib"]
 name = "signet_program"
 
 [dependencies]
-anchor-lang = { version = "0.31.1", features = ["event-cpi"] }
\ No newline at end of file
+anchor-lang = { version = "0.31.1", features = ["event-cpi"] }
```
volovyks, line 12: Do we need updates on the contract side?

chain-signatures/node/src/storage/checkpoint_storage.rs
```diff
@@ -10,7 +10,7 @@ use tokio::sync::RwLock;
 use std::collections::HashMap;
 use std::sync::Arc;
 
-const CHECKPOINT_VERSION: &str = "v0";
+const CHECKPOINT_VERSION: &str = "v1";
```
volovyks, line 13: What would be the starting point now? Can Indexer actually reach it?

integration-tests/tests/cases/mpc.rs
```diff
@@ -543,7 +543,7 @@ async fn test_sign_contention_5_nodes() {
     // Wait for presignatures to be generated - 5-node triple generation takes ~3-4 minutes
     // We wait for a modest per-owner count since distribution is not uniform
     tracing::info!("waiting for presignatures to be generated (triple gen takes ~3-4 min)...");
-    let timeout = Duration::from_secs(480); // 8 minutes for triple + presignature generation
+    let timeout = Duration::from_secs(600); // 10 minutes for triple + presignature generation
```
volovyks, line 546: What is the motivation behind this change?

chain-signatures/node/src/stream/ops.rs
```diff
@@ -418,15 +419,17 @@ pub(crate) async fn process_respond_event(
...
-        .map_err(|err| anyhow::anyhow!("unable to parse target chain from dest: {err:?}"));
-    let target_chain = match target_chain {
-        Ok(chain) => chain,
-        Err(_) => Chain::Ethereum,
-    };
-
-    let Some(BacklogTransaction::Sign(_)) = backlog.get(source_chain, &sign_id).await else {
-        anyhow::bail!("bidirectional tx not found for advancement: {sign_id:?}");
-    };
+        .map_err(|err| anyhow::anyhow!("unable to parse target chain from dest: {err:?}"))?;
```
volovyks, line 422: We agreed that @ppca will update this logic, but the only thing left is to use caip, not .dest(), on line 420.

volovyks: @ppca actually, there are also docs, etc. We can address that separately.

chain-signatures/contract-sol/Cargo.toml
ChaoticTempest: nope, purely typing for serialization on client side

integration-tests/tests/cases/mpc.rs
ChaoticTempest: this is because the contention test would not stockpile enough triples/presignatures in time

chain-signatures/node/src/stream/ops.rs
ChaoticTempest: I'm just leaving how it is for now, since bidirectional originally relied on dest. You can change this however you want after

integration-tests/tests/cases/mpc.rs
volovyks: This test is problematic; this change is only hiding the [issue](https://github.com/sig-net/mpc/issues/681). 
8 minutes must be plenty to stockpile triples and presignatures.

ChaoticTempest: yeah, but it's not a big issue IMO. We'd likely never face this issue on prod (even devnet). This is because we're resource constrained on the GH actions runners (especially in this case with 5 nodes running). I've never been able to have this test fail locally on my m1 macbook (around 30x runs).

volovyks: On my mac it's ~2m, seems too long
I understand that this is caused by low spec CI, but still, we definitely have an issue, so those issues are still relevant.
cc @jakmeier

volovyks (approved):

chain-signatures/node/src/storage/checkpoint_storage.rs
ChaoticTempest: only ethereum does catchup. It'll be as if there was no last processed block, therefore no catchup for ethereum once this new checkpoint storage rolls in

ChaoticTempest:
I'm going to wait on your review @jakmeier before I merge this one. I will also likely need to do a v2 of checkpoint storage after this one since we have a ton of complexity on the sign type and storing of it which is redundant with the enum variants in `BacklogTransaction { Sign, Bidirectional { ... } }`

chain-signatures/node/src/stream/ops.rs
```diff
@@ -831,6 +834,86 @@ mod tests {
         }
     }
 
+    #[tokio::test]
+    async fn process_respond_event_duplicate_ethereum_is_idempotent() {
```
jakmeier, line 838: Nice test!

integration-tests/tests/cases/mpc.rs
jakmeier: Yeah, even if this makes it work more often in CI, it definitely hides errors.  Even in CI, it should not take 10 min.

Looking at the logs shows that the delay is due to various problems. But it is annoying to debug, as it also almost never happens for me locally. (I'm using a bash loop to run it in repetition until it fails.)

But I'm fine with the increase in the timeout for this PR. We can investigate other problems separately.

jakmeier (approved):

jakmeier:
> This fixes the issue with the following error `failed to process respond event`: `sign type not found for respond event (may have already been processed)`.

Nice find, btw. I saw this a lot in devnet logs, it probably contributes to the observed instability on devnet. I look forward to see if this improves the situation.

PR #695, opened 2026-03-11
log hydration publish error more clearly

ppca:


volovyks (approved):

volovyks:
Closing: the hydration modules moved to mpc-chain-hydration (#1065), so this patch no longer applies; redo the logging there if still wanted.

PR #696, opened 2026-03-11
bidirectional: target chain from event.caip2_id instead of dest

ppca:


chain-signatures/node/src/stream/ops.rs
```diff
@@ -128,6 +127,15 @@ impl SignBidirectionalEvent {
             )),
         }
     }
+
+    pub fn target_chain(&self) -> anyhow::Result<Chain> {
```
volovyks: Returning Result<> here should not be required if we will parse string -> Cahin enum when constructing this Event.
It will reduce complexity (like `event.target_chain()?;`).

chain-signatures/primitives/src/lib.rs
```diff
@@ -176,12 +180,31 @@ impl Chain {
...
+
+    pub fn from_caip2_chain_id(caip2_chain_id: &str) -> Option<Self> {
+        mpc_crypto::kdf::Chain::from_caip2_chain_id(caip2_chain_id).map(Self::from_chain_crypto)
+    }
+
+    pub fn from_deprecated_chain_id(chain_id: &str) -> Option<Self> {
+        mpc_crypto::kdf::Chain::from_deprecated_chain_id(chain_id).map(Self::from_chain_crypto)
+    }
+
+    fn from_chain_crypto(chain_crypto: mpc_crypto::kdf::Chain) -> Self {
```
volovyks: We should have a single place to declare the supported Chains enum.
It should live in `chain_signatures/primitives` and be used everywhere in the project. But that may require refactoring, not a requirement for this PR.

chain-signatures/primitives/src/lib.rs
```diff
@@ -176,12 +180,31 @@ impl Chain {
             Chain::Ethereum => 15 * 60,
             Chain::Solana => 3,
             Chain::Hydration => 12,
+            Chain::Bitcoin => 60 * 60,
```
volovyks: It is worth mentioning that this is a probabilistic 6-block finality.
Also, we can add 10-20 minutes, since 60m is an average time for 6 bitcoint transactions.

chain-signatures/crypto/src/kdf.rs
```diff
@@ -22,25 +22,63 @@ pub enum Chain {
...
+            Chain::Ethereum => ChainMeta {
+                deprecated_chain_id: "0x1",
+                caip2_chain_id: "eip155:1",
+            },
+            Chain::Solana => ChainMeta {
+                deprecated_chain_id: "0x800001f5",
+                caip2_chain_id: "solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp",
+            },
+            Chain::Bitcoin => ChainMeta {
+                deprecated_chain_id: "bip122:000000000019d6689c085ae165831e93",
```
volovyks: What is the motivation behind this change?

volovyks (commented):
Let's not forget about docs. I was expecting that this PR would be mostly about documenting struct fields and flows.

chain-signatures/primitives/src/lib.rs
volovyks: https://github.com/sig-net/mpc/pull/707

chain-signatures/crypto/src/kdf.rs
ppca: removed

chain-signatures/contract-sol/src/lib.rs
```diff
@@ -218,19 +218,61 @@ pub struct SignatureRequestedEvent {
...
+    ///
+    /// Used by off-chain services (e.g., relayers, indexers) to filter and
+    /// verify events from the correct program.
+    ///
+    /// MUST be provided and MUST match the deployed program ID.
     pub program_id: Pubkey,
+
+    /// Schema used to deserialize the output of the signed transaction.
+    ///
+    /// MUST be provided.
```
volovyks: Why are we saying that this particular field is a requirement? There are no optional fields in this struct, all of them are required.

chain-signatures/contract-sol/src/lib.rs
```diff
@@ -218,19 +218,61 @@ pub struct SignatureRequestedEvent {
...
     pub dest: String,
+
+    /// Additional parameters encoded as a string (currently unused).
+    ///
+    /// Should be left empty (`""`).
     pub params: String,
+
+    /// The program ID of the Solana program that emitted this event.
+    ///
+    /// Used by off-chain services (e.g., relayers, indexers) to filter and
```
volovyks: I thought it is mostly used by us

chain-signatures/node/src/stream/ops.rs
```diff
@@ -128,6 +127,11 @@ impl SignBidirectionalEvent {
             )),
         }
     }
+
+    pub fn target_chain(&self) -> Chain {
+        // we can directly unwrap because we've checked that the chain id is valid during event deserialization in the indexer
+        Chain::from_caip2_chain_id(&self.caip2_id()).unwrap()
```
volovyks: Now I see that it is not possible to have a `Chain` type in a Bidirectional struct. Using unwrap() does not seem safe. There is no guarantee that the check in another part of the project will stay there, ot that we will not forget to add it in all indexer implementations. Handling Reulst seam as a lesser evil, but I'm happy to hear other thoughts.

volovyks (commented):
Maybe we could rely a bit less on Copilot and address the suggestions from it before requesting a review 🙂

chain-signatures/contract-sol/src/lib.rs
ppca: because some fields should be left blank, i.e. "".

chain-signatures/contract-sol/src/lib.rs
ppca: yeah, i can update the comments.

chain-signatures/node/src/stream/ops.rs
ppca: yeah it is more reasonable to have result as return type

integration-tests/tests/cases/solana_stream.rs
```diff
@@ -205,7 +205,7 @@ async fn test_solana_stream_parse_sign_bidirectional() -> Result<()> {
     solana
         .sign_bidirectional(
             &serialized_tx,
-            "solana:localnet",
+            "solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp",
```
volovyks: Let's not hardcode it. We have Chain caip2_chain_id.

integration-tests/tests/cases/chains.rs
```diff
@@ -99,7 +99,7 @@ async fn test_solana_eth_bidirectional_flow() -> anyhow::Result<()> {
         .solana()
         .bidirectional()
         .transaction_data(unsigned_rlp.clone())
-        .caip2_id("eip155:60")
+        .caip2_id("eip155:1")
```
volovyks: Let's not hardcode it. We have Chain caip2_chain_id.

integration-tests/src/containers.rs
```diff
@@ -883,7 +883,7 @@ impl Solana {
 
         // Call initialize function
         let signature_deposit = 1_000_000u64; // 0.001 SOL in lamports
-        let chain_id = "solana:localnet".to_string(); // CAIP-2 format for local testnet
+        let chain_id = "solana:5eykt4UsFv8P8NJdTREpY1vzqKqZKvdp".to_string(); // CAIP-2 format for local testnet
```
volovyks: Let's not hardcode it. We have Chain caip2_chain_id.

volovyks (commented):
Can you please merge the develop branch? I have conflicts locally. Not sure why GitHub is not showing them.

chain-signatures/primitives/src/lib.rs
```diff
@@ -234,6 +242,20 @@ impl Chain {
...
     }
+
+    pub fn from_caip2_chain_id(chain_id: &str) -> Result<Self, ChainFromError> {
+        Self::iter()
+            .into_iter()
+            .find(|chain| chain.caip2_chain_id() == chain_id)
+            .ok_or_else(|| ChainFromError::UnknownCaip2Id(chain_id.to_string()))
+    }
+
+    pub fn from_deprecated_chain_id(chain_id: &str) -> Result<Self, ChainFromError> {
```
volovyks: This function is never used

chain-signatures/node/src/indexer_sol.rs
```diff
@@ -586,7 +586,13 @@ fn parse_cpi_events(
             }
         } else if event_discriminator == SignBidirectionalEvent::DISCRIMINATOR {
             match <SignBidirectionalEvent as AnchorDeserialize>::deserialize(&mut &event_data[..]) {
-                Ok(ev) => acc.push(Box::new(ev) as SignatureEventBox),
+                Ok(ev) => {
+                    if let Err(e) = Chain::from_caip2_chain_id(&ev.caip2_id) {
```
volovyks, line 592: Why do we need this change? We are not using the caip2 value here.

chain-signatures/node/src/stream/ops.rs
```diff
@@ -134,6 +126,11 @@ impl SignBidirectionalEvent {
             )),
         }
     }
+
+    pub fn target_chain(&self) -> Result<Chain, mpc_primitives::ChainFromError> {
+        // we can directly unwrap because we've checked that the chain id is valid during event deserialization in the indexer
```
volovyks: There is no Unwrap here

volovyks:
This function also looks outdated:
```rust
fn create_test_event(dest: &str) -> SignBidirectionalEvent {
        let mut program_id = [0u8; 32];
        let prefix_len = dest.len().min(program_id.len());
        program_id[..prefix_len].copy_from_slice(&dest.as_bytes()[..prefix_len]);

        SignBidirectionalEvent::Solana(signet_program::SignBidirectionalEvent {
            sender: Default::default(),
            serialized_transaction: vec![],
            dest: dest.to_string(),
            caip2_id: format!("{dest}:test"),
            key_version: 0,
            deposit: 0,
            path: "".to_string(),
            algo: "".to_string(),
            params: "".to_string(),
            program_id: Pubkey::new_from_array(program_id),
            output_deserialization_schema: vec![],
            respond_serialization_schema: vec![],
        })
    }
```

chain-signatures/node/src/indexer_sol.rs
ppca: this is to not process the  SignBidirectionalEvent if its caip2_id is invalid

volovyks: Let's at least document that. This is a requirement for all indexers now;  it just happened that Solana is the only one we use to initiate the flow.

chain-signatures/node/src/sign_bidirectional.rs
```diff
@@ -50,6 +50,8 @@ pub struct BidirectionalTx {
     pub serialized_transaction: Vec<u8>,
     pub source_chain: Chain,
     pub target_chain: Chain,
+    // the mainnet caip2_id of the target chain where the signed transaction will be sent
+    // the value must be one of the values in: https://github.com/sig-net/mpc/blob/dbe8c5b3daa405d51e357ae16eb6df0711194e0e/chain-signatures/primitives/src/lib.rs#L191
```
volovyks: You are using a specific commit here. If we ever update the list of supported chains, people will still see it exactly as it is now. Overall, we should not post GitHub links in code. Mentioning that this must be a supported chain (added to Chain enum) should be enough.

volovyks (dismissed):

volovyks (approved):

chain-signatures/node/src/backlog/mod.rs
```diff
@@ -883,7 +883,7 @@ mod tests {
             serialized_transaction: vec![1, 2, 3],
             source_chain: Chain::Solana,
             target_chain: Chain::Ethereum,
-            caip2_id: "test_caip2_id".to_string(),
+            caip2_id: Chain::Ethereum.caip2_chain_id().to_string(),
```
ChaoticTempest, line 886: nit: can we just make the function name be `chain_id()`? It's more of a standard that we should be following by default, so we should not need to always have it be in the function name, and instead have it be a part of the doc strings for that function.

chain-signatures/node/src/stream/ops.rs
```diff
@@ -134,6 +126,10 @@ impl SignBidirectionalEvent {
             )),
         }
     }
+
+    pub fn target_chain(&self) -> Result<Chain, mpc_primitives::ChainFromError> {
+        Chain::from_caip2_chain_id(&self.caip2_id())
```
ChaoticTempest, line 131: nit: same for the from_* function

ChaoticTempest (approved):

Issue #472, opened 2025-08-04
Stop premature burning of presignatures


ChaoticTempest:
We currently burn a lot of presignatures when we try and make an attempt at a signature. This attempt includes posits, so when we have two nodes that think they are the proposer of this signature, they end up taking their own presignature as an attempt into the signature request. 

So instead when proposing signatures, the id for identifying who is proposing should be `(SignId, ProposerId)` instead of `(SignId, PresignatureId)`.

Note, this requires having the owner keep track of who has what presignature shares via sync, so https://github.com/sig-net/mpc/pull/371 needs to be done first and with it, consequently utilizing that list of participants to choose which presignature to use

volovyks:
The `PresignatureId` is unique and `(SignId, PresignatureId)` will always be a unique identifier. Unlike `(SignId, ProposerId)`, which can be reused when we iterate over all participants and start over.

Now we are reinserting Presignatures back into storage, so the  `(SignId, PresignatureId)` can be duplicated, but I've removed this in https://github.com/sig-net/mpc/pull/671.
The issue I see:
- Reinsertion is added because we are losing pre-signatures
- This issue is called `Stop premature burning of presignatures`, which is possibly caused by reinsertion

@ChaoticTempest can you please elaborate on this? From what I see, we should identify the real reason the pre-signatures are burned. We are recycling presignatures when:
- We are getting a posit deadline during signature generation (my bet is on this, should be related to our current instabilities)
- When we are getting enough REJECT (should rarely happen with State Sync)
- ```if participants.len() < ctx.threshold``` (should rarely happen, we are checking for active participants before starting stockpile logic)

cc @jakmeier

volovyks:
Fixed. Now we take the presignature from the storage only after the posit round was successful.

PR #697, opened 2026-03-12
Do not recycle presignatures

volovyks:
It was a part of the State Sync PR initially, but it does not belong there.
Recycling is not part of our design; rather, it's the opposite.
It can cause or hide issues, more motivation here: https://github.com/sig-net/mpc/issues/472#issuecomment-4044192555


chain-signatures/node/src/protocol/triple.rs
```diff
@@ -556,10 +556,6 @@ impl TripleSpawner {
     /// Stockpile triples if the amount of unspent triples is below the minimum
     /// and the maximum number of all ongoing generation protocols is below the maximum.
     async fn stockpile(&mut self, participants: &[Participant], cfg: &ProtocolConfig) {
-        if participants.len() < self.threshold {
```
volovyks, line 560: This check is already done before calling the stockpile() function

chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -720,21 +721,26 @@ impl PresignatureSpawner {
                     self.ongoing_owned.remove(&id);
                     let _ = ongoing_gen_tx.send(self.ongoing.len());
                 }
-                _ = stockpile_interval.tick(), if active.len() >= self.threshold => {
```
volovyks, line 735: Logic is the same here. I've only added logging.

jakmeier (commented):
LGTM but let's wait for @ChaoticTempest to see if the presiganture burning issue is still present.

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -287,7 +286,10 @@ impl SignOrganizer {
                         };
                         let participants = intersect_vec(&[holders, &active]);
                         if participants.len() < ctx.threshold {
-                            recycle.push(taken);
+                            tracing::warn!(
+                                ?sign_id,
+                                "discarding presignature due to inactive participants"
+                            );
```
ChaoticTempest, line 295: we should also show the active participants in this log

ChaoticTempest (commented):
I'm not entirely sure if this issue is still present. @volovyks can you add a test where we have 100 presignatures with 100 signature requests. If we can get through with ~95% of the signatures produced, then it'd be great. Ideally it should be as close to 100% chance as possible but we don't have to guarantee that with this PR

volovyks:
I do not think that having loose tests is a good idea. I've created a couple of very strict tests here: https://github.com/sig-net/mpc/pull/710
They pass and seem stable. Check details in the PR.

volovyks:
@ChaoticTempest @jakmeier do not lose triples, and presignatures tests are passing
I think that waste of presignatures is still possible if we have many round of signature generation, but I'm ok with merging this anyway

ChaoticTempest:
I would like to hold off merging this PR until https://github.com/sig-net/mpc/pull/723 is merged since that race conditions causes some sign requests to go through several 1000s of rounds

ChaoticTempest (approved):

PR #699, opened 2026-03-12
chore: make backlog use IndexedSignRequest & SignKind

ChaoticTempest:
Renames `SignRequestType` as `SignKind` which gets serialized alongside IndexedSignRequest.

This simplifies the encoding for the backlog entries that are serialized in checkpoints. The previous way had a lot of redundant information saved which is no longer the case with this PR. need to bump the checkpoint version again but the previous bump never made it to testnet/mainnet so this will be fine.


jakmeier (dismissed):
Great refactor!

volovyks (dismissed):

volovyks (approved):

PR #701, opened 2026-03-12
chore: metrics for queues and channels

ChaoticTempest:
This should fix the issue with protocols getting stalled due to backpressure from queues.

This bumps the default size for posits to 2^24. This is a lot so if we ever get to it, we would have other problems at that point.

`MessageInbox::{send, publish}` is no longer blocking so this alleviates the issue where sending one message could block another from being sent. This internally uses `try_send_lossy` which logs on full that we're dropping the message


chain-signatures/node/src/protocol/message/sub.rs
```diff
@@ -13,6 +13,7 @@ use crate::protocol::triple::TripleId;
 
 /// This should be enough to hold a few messages in the inbox.
 pub const MAX_MESSAGE_SUB_CHANNEL_SIZE: usize = 4 * 1024;
+pub const MAX_MESSAGE_POSIT_SUB_CHANNEL_SIZE: usize = 1 << 24;
```
jakmeier: So that would be ~16 M entries for every channel. I guess that is still better than unbound. But we will run into memory issues if these grow large and yes we still don't have a strong guarantee against the blocking.

To keep an eye on how much memory we need, can you add metrics for each channel capacity?

An estimation of buffered channel messages should be possible to read like this:

```rust

impl Subscriber {
    pub fn estimated_queue_len(&self) -> usize {
        match self {
            Self::Subscribed(tx) | Self::Unsubscribed(tx, _) => tx.max_capacity() - tx.capacity(),
            Self::Unknown => 0,
        }
    }
}
```

Then we can have a gague vector tracking how many messages are queue up per channel. 

That way, if we run into memory problems, we will see which channel is to blame. And even if it doesn't run into a full OOM crash, we can observe channel queue lengths and find out which tasks exactly are being stalled.

jakmeier (commented):
Hm, I think this could be seen as a temporary solution, mostly to check if this is the actual problem. Probably we have some cyclic dependencies between channels that completely lock up the system, which should be largely resolved with your change.

A better solution would be to identify cyclic dependencies and decouple them. Then, the worst thing that can happen with a full buffer is that things become really slow. But it shouldn't lock up the system.

In general, adding really large buffers to avoid backpressure problems just moves the problem to a different place. If more messages arrive at a channel than can be handled by the consumer, the number of messages in the channel will keep increasing until something stops it. In this case, OOM or hitting the limit and messages being dropped.

chain-signatures/node/src/protocol/message/sub.rs
ChaoticTempest: yep, I'll add in metrics

ChaoticTempest:
one of the thing this PR misses is why we start cascading into these full channels in the first place. I was talking with @volovyks the other day about this, but it seems like since sync now only utilizes `active` for synced participants, protocol generation (triples and presignatures) are no longer working if sync isn't working properly. So that could be one of the reason why we're filling up the sign channel since there's too many requests in the backlog. Once there is too many requests, the size of the channel will be full. So this wouldn't entirely fix that problem.

We could add back the  `stable` set so this wouldn't happen for purely protocol primitive generation but sync needs to be fixed either ways for signature generation

volovyks:
It seems like it is time to add a limit on the number of concurrent requests we can process. Others should be added to the backlog and processed later.

[This PR](https://github.com/sig-net/mpc/pull/702) should fix /sync (increased timeouts).

We shouldn't generate anything when the node is out of sync; that is the point.

chain-signatures/node/src/protocol/message/sub.rs
ChaoticTempest: added

ChaoticTempest:
added metrics now for channels. A channel name (i.e. triple_posit, triple) and a channel id (i.e. "singleton" or triple id) is used to specify the specific channel. The metrics for the channels will be removed for non singleton ones once they complete (that is not to say that they stop appearing in grafana but that the entries in our node will no longer report them)

volovyks (commented):
Can we move the metrics to a different PR? It would be great to see what the situation is now and how `lossy fix` or any other change affects it.

Also, I want to confirm we are not breaking the cardinality rule in metrics.

jakmeier (dismissed):
LGTM

I would also prefer merging just the metrics in a first PR, as @volovyks asked. But if untangling it is too much work, personally I am okay with merging it together.

One other note: You report queue lengths when sending but not when removing an element. Metrics might look a bit strange, staying at a high value even when the queue becomes empty, only updating once the next message is sent. I don't have a solution. But because of this quirk I think in dashboards we should label it as max queue size to avoid confusion.

ChaoticTempest:
this now just has the metrics for channel size

chain-signatures/node/src/protocol/message/mod.rs
```diff
@@ -280,15 +307,18 @@ impl MessageInbox {
...
                 SubscribeRequestAction::Subscribe(resp) => {
-                    let rx = self
+                    let sub = self
                         .signature
                         .entry((sign_id, presignature_id))
-                        .or_default()
-                        .subscribe();
+                        .or_insert_with(|| Subscriber::unsubscribed("sign_task"));
+                    let rx = sub.subscribe();
+                    sub.report_queue_len();
```
jakmeier: Here we report the length for a "sign_task" queue and I think they all such tasks share a single metric. I don't think a single gauge metric works in this case, it will randomly jump up and down and be very hard to read. Either the metrics has to to be labelled per task somehow but probably it should just be a different metric type, e.g. a histogram or a summary.

Same problem with other "triple_task" and "presign_task" I suppose.

jakmeier (commented):
Thanks for splitting this in two PRs. Please also update the PR title accordingly, it can be confusing right now.

chain-signatures/node/src/protocol/message/mod.rs
ChaoticTempest: that's true, but I'm not sure how we're going to solve the grafana cardinality issue otherwise. Maybe we just create a histogram instead of the current int guage

ChaoticTempest: now using histogram for tasks and int guage for posit channels

chain-signatures/node/src/metrics/messaging.rs
```diff
@@ -78,3 +79,48 @@ pub(crate) static WEB_ENDPOINT_LATENCY: LazyLock<HistogramVec> = LazyLock::new(|
     )
     .unwrap()
 });
+
+pub(crate) static CHANNEL_QUEUE_SIZE: LazyLock<IntGaugeVec> = LazyLock::new(|| {
+    super::try_create_int_gauge_vec_with_node_account_id(
+        "multichain_message_channel_queue_size",
+        "Estimated number of buffered messages queued per message channel",
```
volovyks: The description and the const name are misleading. What we are passing in practice is the remaining space in the channel, not its size.

chain-signatures/node/src/metrics/messaging.rs
```diff
@@ -78,3 +79,48 @@ pub(crate) static WEB_ENDPOINT_LATENCY: LazyLock<HistogramVec> = LazyLock::new(|
...
+    CHANNEL_QUEUE_SIZE
+        .with_label_values(&[channel])
+        .set(len as i64);
+}
+
+pub fn set_queue_len_tx<T>(name: &'static str, tx: &mpsc::Sender<T>) {
+    set_queue_len(name, channel_len(tx));
+}
+
+pub fn channel_len(tx: &mpsc::Sender<impl Sized>) -> usize {
```
volovyks: Let's not put `pub` by default everywhere. This function is not used outside of this file.

There are also a couple of ways we update this metric. Ideally, we should only use `set_queue_len_tx()` and other top-level functions and hide the direct access to the metric.

chain-signatures/node/src/metrics/messaging.rs
```diff
@@ -78,3 +79,48 @@ pub(crate) static WEB_ENDPOINT_LATENCY: LazyLock<HistogramVec> = LazyLock::new(|
...
+    .unwrap()
+});
+
+pub(crate) fn set_queue_len(channel: &str, len: usize) {
+    CHANNEL_QUEUE_SIZE
+        .with_label_values(&[channel])
+        .set(len as i64);
+}
+
+pub fn set_queue_len_tx<T>(name: &'static str, tx: &mpsc::Sender<T>) {
```
volovyks: Since the list of channels is predefined, I would create an enum and use it as an argument here. Just for clarity and to avoid mistakes.

chain-signatures/node/src/metrics/messaging.rs
```diff
@@ -78,3 +79,48 @@ pub(crate) static WEB_ENDPOINT_LATENCY: LazyLock<HistogramVec> = LazyLock::new(|
...
+
+pub fn channel_len(tx: &mpsc::Sender<impl Sized>) -> usize {
+    tx.max_capacity() - tx.capacity()
+}
+
+pub(crate) fn remove_queue_len(channel: &str) {
+    let _ = CHANNEL_QUEUE_SIZE.remove_label_values(&[channel]);
+}
+
+pub(crate) static TASK_QUEUE_LEN: LazyLock<HistogramVec> = LazyLock::new(|| {
```
volovyks: Representing it this way should not break the cardinality rule, but still, the number of messages per protocol is not big, and is predictable. Why not set the MAX to the expected number of messages in the protocol and do not have this metric?

(not blocking comment, I'm ok with monitoring this channel, but not sure if it is useful)

My API comments can be usefull here too

chain-signatures/node/src/protocol/message/ring.rs
```diff
@@ -0,0 +1,167 @@
...
+    (RingSender(shared.clone()), RingReceiver(shared))
+}
+
+impl<T: Send> RingSender<T> {
+    /// Send a message, dropping the oldest entry if the channel is at capacity.
+    ///
+    /// Returns `Ok(true)` if an entry was evicted to make room, `Ok(false)` if
+    /// the message was buffered without eviction, or `Err` if the receiver has
+    /// been dropped.
+    pub fn send_lossy(&self, msg: T) -> Result<bool, SendError<T>> {
```
volovyks: I would say we want to separate this from metrics PR.

volovyks (commented):
Metrics look good overall, just a couple of nits. It would be great to merge them separately.

chain-signatures/node/src/metrics/messaging.rs
ChaoticTempest: it is the size though: `max_capacity - capacity` is size not what is remaining. `max_capacity` here is the total capacity defined in `mpsc::channel(max_capacity)`, while `capacity` is what can be used

chain-signatures/node/src/metrics/messaging.rs
ChaoticTempest: it's a utility function, will move to a different module.

honestly, should we even care about `pub` at all. It's a binary at the end of the day, we're not building a library here so it doesn't really matter unless it's dead code

chain-signatures/node/src/metrics/messaging.rs
ChaoticTempest: not going to do this here, because at that point, we might as well just make a struct for the whole of metrics instead of just having to import everything every time we want to set the queue len in metrics

chain-signatures/node/src/protocol/message/ring.rs
ChaoticTempest: will remove, this got committed mistakenly

chain-signatures/node/src/metrics/messaging.rs
ChaoticTempest: where are you getting breaking the cardinality rule from? This is a histogram with buckets so why would that break the cardinality rule? I don't understand what you're saying about MAX expected number of messages here.

This is a metric on all tasks where we're reporting for all triple, presignatures, signatures their various buckets. We would get the metrics per category (triple, presignatures, signatures) and display them for each

ChaoticTempest:
this is taking a lot longer than I expect to get these metrics in. We might as well just get in https://github.com/sig-net/mpc/pull/788 so we can verify our backpressure related issues are fixed. I would rather not get in these metrics at this point if it takes any longer

ChaoticTempest:
or if we really want to wait for metrics, https://github.com/sig-net/mpc/pull/796 could fix our backpressure issues but in a different manner. My hypothesis is that when posit queue/channel gets overloaded, we get the backpressure and head of line blocking issues. So https://github.com/sig-net/mpc/pull/796 should fix the posit backpressure because we would not be sending as many posits anymore

chain-signatures/node/src/metrics/messaging.rs
volovyks: Correct, my bad
Is that what we want? That will require updating levels on Grafana each time we change any channel capacity

chain-signatures/node/src/metrics/messaging.rs
volovyks: We are building interfaces for each other and for future selves

chain-signatures/node/src/metrics/messaging.rs
volovyks: I'm saying that we are NOT breaking the [cardinality rule](https://grafana.com/docs/enterprise-logs/latest/get-started/labels/cardinality/) anymore.

I understand the purpose of the metric, but I have doubts that we need it. We can set the max expected number of messages in the protocol as its max channel size and throw an error if any of the protocols is receiving more than that. This MAX is not that big.

But again, this one is not blocking.

volovyks:
We're close enough to merge this PR.  I prefer to know what is wrong before applying fixes.
Reviewing 796

ChaoticTempest:
the thing is, this PR isn't even going to show us the problem because of how flaky the backpressure problem is. The moment we deploy and it restarts, the backpressure problem would go away, and then we wouldn't be able to see it again until it randomly pops up again.

IMO, in the future, we shouldn't be blocking a hotfix just for metrics

chain-signatures/node/src/metrics/messaging.rs
ChaoticTempest: > That will require updating levels on Grafana each time we change any channel capacity

how else would we do it then?

chain-signatures/node/src/metrics/messaging.rs
ChaoticTempest: honestly, it's not worth the energy to even consider. We control the whole stack end to end here. Our future selves won't care if something was `pub`, they'd care if it was non-pub because then they would have to make it `pub`

chain-signatures/node/src/metrics/messaging.rs
ChaoticTempest: ahh, didn't see NOT.

> We can set the max expected number of messages in the protocol as its max channel size and throw an error if any of the protocols is receiving more than that.

If you're saying that we don't need these metrics, and just error out in logs, then what was the point in wanting these metrics in the first place? We could've just shipped https://github.com/sig-net/mpc/pull/788 then because that's what it does when it's full.

chain-signatures/node/src/metrics/messaging.rs
volovyks: We can pass the remaining space in the channel. Then, in Grafana, we can monitor if any channel metrics are close to zero.

chain-signatures/node/src/metrics/messaging.rs
volovyks: I'm only talking about a metric that measures a single protocol channel.

chain-signatures/node/src/metrics/messaging.rs
ChaoticTempest: made it capacity

chain-signatures/node/src/metrics/messaging.rs
ChaoticTempest: yeah but we run into cardinality issues with that... so what do we really want here with this PR then?

chain-signatures/node/src/metrics/messaging.rs
volovyks: Should we make the other one capacity as well?

volovyks:
All I need from this one is CHANNEL_QUEUE_SIZE to become the capacity, for consistency and better representation on Grafana

jakmeier:
> the thing is, this PR isn't even going to show us the problem because of how flaky the backpressure problem is. The moment we deploy and it restarts, the backpressure problem would go away, and then we wouldn't be able to see it again until it randomly pops up again.

Exactly because we can't tell if a fix works or not it is doubly important to have metrics first. Otherwise we don't even know if the fix was valid.

> IMO, in the future, we shouldn't be blocking a hotfix just for metrics

I see your point. For obvious hotfixes, I agree. But this change isn't an obvious case IMO. Additionally, if the fix doesn't help, we probably don't want the change at all. Even if it helps, it can only be a temporary solution.

But I agree, in most cases, a fix should have priority over metrics.

> All I need from this one is CHANNEL_QUEUE_SIZE to become the capacity, for consistency and better representation on Grafana

Current queue length or current capacity left both are ok with me. But yes, consistency would be nice.

ChaoticTempest:
made it capacity for consistency now

chain-signatures/node/src/protocol/message/mod.rs
```diff
@@ -341,9 +371,11 @@ impl MessageInbox {
             tokio::select! {
                 _ = self.filter.update() => {}
                 Some(sub) = self.subscribe_rx.recv() => {
+                    set_channel_capacity_tx("subscribe", &self.subscribe_tx);
```
ppca, line 374: "subscribe" is used for a lot of things like presig and sig, should we differentiate them?

chain-signatures/node/src/protocol/message/mod.rs
```diff
@@ -121,42 +132,46 @@ impl MessageInbox {
...
             }
             Message::Signature(message) => {
-                let _ = self
+                let sub = self
                     .signature
                     .entry((message.id, message.presignature_id))
-                    .or_default()
-                    .send(message)
-                    .await;
+                    .or_insert_with(|| Subscriber::unsubscribed("sign_task"));
```
ppca, line 172: "sign_task" is also used for the posit inbox of signature spawner. I think it needs to be differentiated because they are two different channels for different functionalities. I opened a PR to change the name of signature spawner inbox here: https://github.com/sig-net/mpc/pull/827/changes 🙏

volovyks (approved):
LGTM, let's address what @ppca is talking about separately. We need to iterate on this.

chain-signatures/node/src/protocol/message/mod.rs
ChaoticTempest: not sure what you mean here, "subscribe" is only used on the channel for getting subscription requests

chain-signatures/node/src/protocol/message/mod.rs
ChaoticTempest: this one makes sense, let's merge your one separately

PR #702, opened 2026-03-12
Tune sync timeouts

volovyks:
Sync is failing on dev, because the default timeout on the client level is ~1s
I've also increased the waiting time in other places to make sure nodes have enough time to process all the artifacts

### DIFF
--- a/integration-tests/tests/cases/mpc.rs
+++ b/integration-tests/tests/cases/mpc.rs
@@ -516,6 +516,7 @@ async fn test_sign_requests_wait_for_presignatures() {
 /// This test generates triples and presignatures on-the-fly (slower but more realistic).
 /// Uses 5_nodes.json fixture for pre-shared keys only.
 #[test(tokio::test(flavor = "multi_thread"))]
+#[ignore]
 async fn test_sign_contention_5_nodes() {
     const NUM_NODES: u32 = 5;
     const THRESHOLD: usize = 4;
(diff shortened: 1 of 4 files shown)


jakmeier (dismissed):

volovyks:
cases::mpc::test_sign_contention_5_nodes issue is back...

volovyks:
I've checked our metrics, sync call takes ~15 seconds, and it is O(N), so this time can change. I will look into ways to make it more efficient.

volovyks:
Ok, I think that long /sync was caused by an outdated storage version. Now it is ~1s with a large stockpile of Ts and Ps. But I still think we can increase the limits, since /sync process is rare.

ChaoticTempest (dismissed):

integration-tests/tests/cases/mpc.rs
```diff
@@ -516,6 +516,7 @@ async fn test_sign_requests_wait_for_presignatures() {
 /// This test generates triples and presignatures on-the-fly (slower but more realistic).
 /// Uses 5_nodes.json fixture for pre-shared keys only.
 #[test(tokio::test(flavor = "multi_thread"))]
+#[ignore]
```
volovyks, line 519: Turning this test off temporarily to be able to merge PRs and fix dev

jakmeier: yeah probably a good idea, it really slows us down right now

jakmeier (approved):

Issue #688, opened 2026-03-05
Presignatures are not distributed evenely


jakmeier:
When generating new Ps, the proposer of the P posit generation becomes the owner of the new Presignature. Somehow, not all nodes get the same number of Ps in our tests.

These tests are known to fail because of slow or uneven P generation:

- test_sign_requests_wait_for_presignatures
- test_sign_contention_5_nodes
- test_sign_requests_wait_for_presignatures

Example assertion error, while waiting for 3 Ps per node:

```
2026-03-05T16:29:25.250615Z  INFO integration_tests::mpc_fixture::fixture_interface: Node Participant(0) has 12 Ps
2026-03-05T16:29:25.251657Z  INFO integration_tests::mpc_fixture::fixture_interface: Node Participant(1) has 8 Ps
2026-03-05T16:29:25.252524Z  INFO integration_tests::mpc_fixture::fixture_interface: Node Participant(2) has 18 Ps
2026-03-05T16:29:25.253070Z  INFO integration_tests::mpc_fixture::fixture_interface: Node Participant(3) has 3 Ps
2026-03-05T16:29:25.253654Z  INFO integration_tests::mpc_fixture::fixture_interface: Node Participant(4) has 2 Ps

thread 'cases::mpc::test_sign_contention_5_nodes' (159697) panicked at /home/runner/work/mpc/mpc/integration-tests/src/mpc_fixture/fixture_interface.rs:92:16:
should have enough presignatures: Elapsed(())
```
[full logs](https://github.com/sig-net/mpc/actions/runs/22725931619/job/65900923424?pr=686#step:16:40349)

Task: Find out why these are generated uneven and find a solution for a better distribution.

volovyks:
Here is the network setup in `test_sign_contention_5_nodes`.
```rust
    // Build network with pre-shared keys, generate triples/presignatures on the fly
    let network = MpcFixtureBuilder::new(NUM_NODES, THRESHOLD)
        .with_preshared_key()
        .with_preshared_triples()
        .with_min_triples_stockpile(STOCKPILE_MIN)
        .with_max_triples_stockpile(STOCKPILE_MAX)
        .with_min_presignatures_stockpile(STOCKPILE_MIN)
        .with_max_presignatures_stockpile(STOCKPILE_MAX)
        .build()
        .await;
```

[...]

PR #703, opened 2026-03-13
Set proposer as owner for Ts

volovyks:
Context: https://github.com/sig-net/mpc/issues/688#issuecomment-4051186897


ChaoticTempest:
see my comments in https://github.com/sig-net/mpc/issues/688#issuecomment-4072480723, we can't guarantee equal distribution by just making the introducer the owners. One node can possibly spam and we end up back at the our original problem but even worse cause it's very likely that a node would spam due to network congestion

jakmeier (dismissed):
I think this is a good idea. I outlined my reasons on the issue. (https://github.com/sig-net/mpc/issues/688#issuecomment-4088915370)

But let's not dismiss the reasoning of @ChaoticTempest, I value his opinion here. Thus, let's not merge this until we have reached an agreement on the discussion.

ChaoticTempest (dismissed):
It's not ideal with this but if you need to make progress, then we can see whether this works for the best or not. See my comment in https://github.com/sig-net/mpc/issues/688#issuecomment-4279038994 but overall it's just very difficult to get the generation of triples correct. Misconfiguring it could lead to huge resource starvation so tread very carefully

ChaoticTempest (approved):

PR #704, opened 2026-03-13
Ensure the signature we publish is valid

volovyks:


ppca:
stupid question: why would this make sure the signature we publish is valid?

volovyks:
@ppca, because to publish a signature, we need to create a Publish action, which is not possible if the signature is not valid.

ppca:
> @ppca, because to publish a signature, we need to create a Publish action, which is not possible if the signature is not valid.

are we not doing it now? i mean in what situation will an invalid signature be published in the current code? i want to know what the problem is so that i could more easily understand the fix

volovyks:
There is no problem, I'm adding tests and restructuring code in the context of https://github.com/sig-net/mpc/pull/705

ppca (approved):

ppca (approved):

PR #707, opened 2026-03-16
Merge Chain definitions

volovyks:


chain-signatures/crypto/src/types.rs
```diff
@@ -1,35 +1,9 @@
...
-        // The space of inputs is 2^256, the space of the field is ~2^256 - 2^129.
-        // This mean that you'd have to run 2^127 hashes to find a value that causes this to fail.
-        Scalar::from_bytes(hash).expect("Derived epsilon value falls outside of the field")
-    }
-}
+pub use mpc_primitives::{PublicKey, ScalarExt};
 
 #[test]
 fn scalar_fails_as_expected() {
+    use k256::Scalar;
```
ppca: should this test be moved to mpc-primitives then?

ppca (dismissed):

chain-signatures/crypto/src/types.rs
volovyks: Good point, moved the test

ChaoticTempest (approved):

ppca (approved):

PR #708, opened 2026-03-16
Change contention test from 5 to 3 nodes

volovyks:


ChaoticTempest:
why are we changing it to 5 nodes? The whole point is to have more contention with more nodes in this test. This would be just like any other test if we change it to 3 nodes

volovyks:
I'm trying to tackle this issue from different angles. I agree that reducing the number of nodes only hides the reason it is failing.
I've analyzed CI logs from a 5-node and a 3-node run. We have multiple issues, mostly related to timing. Both test and node setup. Magic numbers for delays do not work well.

image

PR #710, opened 2026-03-18
Do not loose artifacts tests

volovyks:
While writing tests, I've noticed that our pre-generated artifacts are not available on all nodes. Both 3 and 5 node setups. That is because when we generate them, not all participants are joining the protocol, especially on 5 nodes setup. I've regenerated Ts and Ps and increased their number.

Before:
3_nodes.json:
- 23 triples (18–21 per owner),
- 16 presignatures (15–16 per owner). Uneven node distribution.

5_nodes.json:
- 66 triples (30–38 per owner, except owner 4 has all 66)
- 15 presignatures (13–15 per owner). Uneven node distribution.

After:
3_nodes.json (3 nodes):
- Triples: 50 per owner, available on all nodes
- Presignatures: 25 per owner, available on all nodes

5_nodes.json (5 nodes):
Triples: 30 per owner, available on all nodes
Presignatures: 15 per owner,  available on all nodes

It allows us to write deterministic tests, like `test_sign_no_presignature_waste` and `test_presignature_no_triple_waste`.


jakmeier (approved):
Very cool!

1. Thank you for cleaning up the fixture data. This should make testing more deterministic.
2. Great addition of tests with `test_sign_no_presignature_waste` and `test_presignature_no_triple_waste`!

jakmeier (approved):

PR #713, opened 2026-03-19
Turn on 5 nodes consumption test

volovyks:
@ChaoticTempest @jakmeier, please, turn this test back on locally when working on stability fixes. I will do the same and merge develop on every stabilization step.


volovyks:
Not relevant, already turned on

PR #715, opened 2026-03-19
Ensure port uniqness in tests

volovyks:
Tests in https://github.com/sig-net/mpc/pull/697 failed because two of the three nodes got the same port.
```
ode is starting at http://127.0.0.1:42547  ← node 0-sxtmu
node is starting at http://127.0.0.1:42547  ← node 1-wdkyi
node is starting at http://127.0.0.1:45533  ← node 2-vhtvu
```

### DIFF
--- a/integration-tests/src/utils.rs
+++ b/integration-tests/src/utils.rs
@@ -6,8 +6,13 @@ use near_workspaces::{
     Account, AccountId, Worker,
 };
 use rand::Rng;
-use std::sync::Once;
+use std::collections::HashSet;
+use std::sync::{Mutex, Once};
 use tracing_subscriber::EnvFilter;
+
+/// Tracks ports already handed out by `pick_unused_port` within this process
+/// to prevent the OS from recycling the same ephemeral port for multiple nodes.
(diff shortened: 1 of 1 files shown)


integration-tests/src/utils.rs
```diff
@@ -143,15 +148,22 @@ pub async fn is_port_available(port: u16) -> bool {
...
     // Port 0 means the OS gives us an unused port
     // Important to use localhost as using 0.0.0.0 leads to users getting brief firewall popups to
     // allow inbound connections on macOS
     let addr = std::net::SocketAddrV4::new(std::net::Ipv4Addr::LOCALHOST, 0);
-    let listener = tokio::net::TcpListener::bind(addr).await?;
-    let port = listener.local_addr()?.port();
-    Ok(port)
+    loop {
+        let listener = tokio::net::TcpListener::bind(addr).await?;
+        let port = listener.local_addr()?.port();
```
jakmeier, line 159: Oh I see, by dropping `listener: TcpListener`, the socket and thus the port is freed on the OS level...

Ah well, I guess it would be nice if `solana-test-validator` could just pick an unused port itself and report it back.

Your solution seems like the best workaround.

jakmeier (approved):

PR #721, opened 2026-03-20
Set min T per node in tests to 1 pair

volovyks:
Our tests keep generating triples that are never used.


integration-tests/src/lib.rs
```diff
@@ -74,13 +74,13 @@ impl Default for NodeConfig {
                 max_concurrent_generation: 16,
                 max_concurrent_introduction: 2,
                 triple: TripleConfig {
-                    min_triples: 16,
-                    max_triples: 320,
+                    min_triples: 1,
```
volovyks, line 78: @ChaoticTempest This is actually a `min_triples_per_node`. With a "random" assignment of owner, we need to generate much more to meet this minimum.

PR #722, opened 2026-03-20
limit the number of protocols in contention test

volovyks:
More of an experiment, we did not had a setup for max running protocols in Mixture setup


integration-tests/src/mpc_fixture/builder.rs
```diff
@@ -69,6 +71,9 @@ struct FixtureConfig {
     node_min_presignatures: u32,
     network_max_presignatures: u32,
 
+    max_concurrent_introduction: u32,
```
ppca, line 74: these 2 values are not used?

volovyks:
@jakmeier I think we can merge this. This test is more of a metric on how much we can handle, and it looks like 8 is our limit atm.

integration-tests/src/mpc_fixture/builder.rs
volovyks: They are mapped to ProtocolConfig and passed to each node.

jakmeier (approved):

PR #723, opened 2026-03-23
feat/fix: checkpoint quorum with requeueing on catchup completed

ChaoticTempest:
This adds quorum checking on the selected checkpoint for recovery. Previously, we were just blindly accepting it at whatever threshold lowest checkpoint without checking with other nodes whether they have it or not. If a quorum isn't met, we go back to our local checkpoint only if the chain supports catchup (i.e. ethereum), where we would only requeue the entries of the recovered requests after catchup is completed. Otherwise, we throw away the checkpoint if no quorum is met and the chain does not support catchup (i.e. Solana/Hydration). This logic will be added to further chains once catchup is implemented generically for all.

This should fix some of the tasks that have long rounds due to prematurely sending the requests in for processing. There would potentially be a race condition where during restart, we would get the respond event, but during recovery we haven't requeued any of the tasks yet. Thus encountering a respond event would be a miss and we would still have the backlog entry & sign task in flight. This fixes that entirely by first recovering the backlog, then starting the indexer; verifying that our recovered backlog has been cleared before requeueing requests for processing.


volovyks (commented):
I understand separate pieces of this PR, but I'm struggling to understand if it gives as a guarantee of recovery, especially in case we do not have catch-up. Overall, it would be great to have a simple design doc for Backlog and specifically for backlog sharing between nodes before starting the implementation.

In my head, Backlog is always "block-based". We are always at a certain block height, and at this block height, we have a very specific set of requests. When nodes sync, they look for the lowest block height that meets T. Then, they can fetch the backlog from any node and check that its hash is correct. This design requires catch-up logic, which may be a blocker. Happy to hear other thoughts on this.

ChaoticTempest:
@volovyks one thing to note, chains that do NOT do catchup like solana currently do NOT do recovery. This is because for solana, we always start at the latest block. Later PRs, where we generalize catchup, this behavior should be better handled

volovyks:
My bad on saying we do not have a [design](https://docs.google.com/presentation/d/1A7JadOk4SLZ7xZmRRIvpMGlM5unD0hW9YtTQp_AGzLg/edit?pli=1&slide=id.g3736d5a6e75_0_43#slide=id.g3736d5a6e75_0_43).

jakmeier (commented):
This looks very promising, thank you for implementing that!

But given our stability problems, I feel uneasy on adding more potential points of failures at the moment. The quorum on checkpoint feature thus stands out to me.

Would it be possible to extract the part that addresses the race condition after a restart and merge only that for now? It would also make it easier to pin-point the impact of the change.

ChaoticTempest:
@jakmeier sounds good, will close this one in favor of https://github.com/sig-net/mpc/pull/733 which has only the changes for catchup

PR #724, opened 2026-03-23
Derive Delta discussion

volovyks:
@scafuro could you please clarify the requirements for the delta computation?

We’re integrating more chains, and each of them relies on a different entropy source, so it would be helpful to have a clear definition.

My understanding is that delta must be unique. To achieve this, we include `presignature_big_r`, which is unique per signature generation since pre-signatures are not reused. Because the requester does not know which presignature will be selected, the resulting delta should not be predictable.

However, node providers can still choose which presignature they own to use for generation. To mitigate this, we add an additional source of randomness that is specific to each chain - entropy. Let's discuss if what we have is sufficient.

### DIFF
--- a/chain-signatures/node/src/indexer.rs
+++ b/chain-signatures/node/src/indexer.rs
@@ -103,8 +103,8 @@ impl NearIndexer {
         let payload = pending_request.payload;
         let epsilon = pending_request.epsilon;
 
-        // no longer taking entropy from logs, but this is merely for integration tests, so
-        // it doesn't matter as much as long as the IT nodes agree on the entropy.
+        // No longer taking entropy from logs, but this is merely for integration tests, so
+        // it doesn't matter as much as long as the integration tests nodes agree on the entropy.
         let entropy = self.derive_entropy_from_sign_id(&sign_id);
         // NOTE: path is not used at all currently in signature.rs during signing, so hardcoding
         // it here won't matter.
--- a/chain-signatures/node/src/indexer_eth/mod.rs
+++ b/chain-signatures/node/src/indexer_eth/mod.rs
@@ -363,8 +363,11 @@ fn sign_request_from_filtered_log(log: Log) -> Option<IndexedSignRequest> {
         &event.path,
     );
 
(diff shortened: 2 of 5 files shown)


scafuro:
@volovyks 
The very short answer is: we do not use `entropy `to mitigate the ability to choose pre-signatures,  and uniqueness does not imply unpredictability.  
We use `entropy` to strengthen unpredictability, and `entropy` must be chosen properly from the block.  It will be all clear if you read the longer explanation below.

----------------------
Longer explanation:
The most important property we need from delta is that it is unpredictable up until the signature request is committed (i.e., posted on the blockchain).
This means: no-one should be able to know what delta it will be used to sign request sig_req, until sig_req is posted. We achieve this by ensuring that  delta depends on the `sig-req=  (h,PK,epsilon).`

However, we also use  other two inputs to computed delta. These inputs are  `entropy` and `presignature_big_r.`
 `delta=H(sig-request || big_r ||  entropy).`

Why?

`entropy `and `presignature_big_r` and are here for the following reasons:

-  `entropy` is there to strengthen unpredictability. Since the hash function is not a truly random function, we cannot rule out the possibility that an adversary might still be able to “control” delta and the payload in their favor. 
By adding  ` entropy`  in the computation of delta, we make delta dependent on  a variable that is  (hopefully) outside the adversary’s control (unless the adversary is the miner). This means that entropy should be a value that becomes  known to the adversary only after sig-request  is posted on the blockchain.
How do we set entropy so that it is satisfies that property? This will depend on the blockchain and what is contained in the block.
 

- `Presignature_big_r ` is used to guarantee  “uniqueness” : we want to make sure that there is a one-to-one mapping between a delta and presignature_big_r (in other words, we do not want that the same delta would work for many presignature_ big_r). This is why we add presignature_big_r


To conclude:
 - We use big_r to achieve uniqueness, but **uniqueness does not imply unpredictability.** 
- Providers can choose their big_r however delta still depends on entropy and payload, hence it does not give them an advantage in forgery 
- When we choose entropy we must make sure we select a field of the block that is unpredictable up until the block is validated.

PR #725, opened 2026-03-23
Unify IndexedSignRequest creation

volovyks:
Moved timestamp management into a function to reduce duplication and unify Indexed event creation. No changes in logic.

### DIFF
--- a/chain-signatures/node/src/protocol/signature.rs
+++ b/chain-signatures/node/src/protocol/signature.rs
@@ -56,6 +56,43 @@ pub struct IndexedSignRequest {
     pub sign_request_type: SignRequestType,
 }
 
+impl IndexedSignRequest {
+    pub fn new(
+        id: SignId,
+        args: SignArgs,
+        chain: Chain,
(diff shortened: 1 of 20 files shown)


chain-signatures/node/src/protocol/signature.rs
```diff
@@ -55,7 +55,9 @@ pub struct IndexedSignRequest {
 }
 
 impl IndexedSignRequest {
-    pub fn new(
+    /// Reconstruct a sign request from persisted data (e.g. backlog recovery or tests).
+    /// Preserves the original `unix_timestamp_indexed` rather than stamping the current time.
+    pub fn recover(
```
ChaoticTempest: the naming `recover` doesn't add anything to note here. We don't need to say that it's recovered, since this is just generic data we are inputting. `new` is much more better here, since it would fit more cases, otherwise calling `recover` would be incorrect for cases where you're just constructing the `IndexedSignRequest`. Also, `recover` is a verb so I would assume it does something under the hood, which this is not

PR #728, opened 2026-03-25
Improve eth config setup and logging

volovyks:


chain-signatures/node/src/indexer_eth/mod.rs
```diff
@@ -243,14 +243,52 @@ impl EthArgs {
...
+            if self.eth_consensus_rpc_http_url.is_none() {
+                missing.push("eth_consensus_rpc_http_url");
+            }
+            if self.eth_network.is_none() {
+                missing.push("eth_network");
+            }
+            if self.eth_helios_data_path.is_none() {
+                missing.push("eth_helios_data_path");
+            }
+        }
```
ChaoticTempest: hmm, all this custom logic for parsing args doesn't correct to do it here. We should really be doing this on the `clap` level within `EthArgs`

volovyks: Good point, added an update

ppca (approved):

PR #733, opened 2026-03-29
fix: race condition with requeuing on restarts

ChaoticTempest:
This is just the catchup portion of https://github.com/sig-net/mpc/pull/723

### DIFF
--- a/chain-signatures/node/src/backlog/mod.rs
+++ b/chain-signatures/node/src/backlog/mod.rs
@@ -869,6 +869,65 @@ mod tests {
         SignBidirectionalEvent::Solana(signet_program::SignBidirectionalEvent {
             sender: Default::default(),
             serialized_transaction: vec![],
+            dest: dest.to_string(),
+            caip2_id: format!("{dest}:test"),
+            key_version: 0,
+            deposit: 0,
+            path: "".to_string(),
+            algo: "".to_string(),
+            params: "".to_string(),
+            program_id: Pubkey::new_from_array(program_id),
+            output_deserialization_schema: vec![],
+            respond_serialization_schema: vec![],
+        })
+    }
+
+    fn create_test_args(id: u8) -> SignArgs {
+        SignArgs {
+            entropy: [id; 32],
+            epsilon: k256::Scalar::from(1u64),
+            payload: k256::Scalar::from(2u64),
+            path: "test".to_string(),
+            key_version: 1,
+        }
+    }
+
+    fn create_indexed_request(
+        sign_id: SignId,
+        chain: Chain,
+        args: SignArgs,
+        kind: SignKind,
+        unix_timestamp_indexed: u64,
+    ) -> IndexedSignRequest {
+        IndexedSignRequest::new(sign_id, args, chain, unix_timestamp_indexed, kind)
+    }
+
+    fn create_bidirectional_request(
+        sign_id: SignId,
+        chain: Chain,
+        dest: &str,
+        unix_timestamp_indexed: u64,
+    ) -> IndexedSignRequest {
+        IndexedSignRequest::sign_bidirectional(
+            sign_id,
+            create_test_args(sign_id.request_id[0]),
+            chain,
+            unix_timestamp_indexed,
+            create_test_event(dest),
+        )
+    }
+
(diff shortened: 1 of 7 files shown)


chain-signatures/node/src/stream/ops.rs
```diff
@@ -288,27 +288,48 @@ pub(crate) async fn recover_backlog(
...
+    for &sign_id in pending.keys() {
+        let Some(entry) = backlog.get(source_chain, &sign_id).await else {
+            continue;
+        };
+
+        if entry.status() != SignStatus::AwaitingResponse {
+            continue;
+        }
+
+        if entry.execution_tx().is_some() {
```
volovyks, line 331: It is worth adding a comment on what is checked here, since it is not obvious.

chain-signatures/node/src/backlog/mod.rs
```diff
@@ -753,27 +780,50 @@ impl BacklogEntry {
...
+                tracing::warn!(
+                    ?chain,
+                    block_height = local.block_height,
+                    "recovering from local checkpoint; requeue deferred until catchup"
+                );
+                RecoveryRequeueMode::AfterCatchup
+            } else {
+                RecoveryRequeueMode::Immediate
+            };
+            Some((local, requeue_mode))
```
volovyks: The only difference in `Some(_), Some(remote)` comparing to `(Some(local), Some(remote)) if local.block_height > remote.block_height ` is the log message. Is it our intention?

chain-signatures/node/src/backlog/mod.rs
```diff
@@ -753,27 +780,50 @@ impl BacklogEntry {
...
+        }
+        (None, Some(remote)) => Some((remote, RecoveryRequeueMode::Immediate)),
```
volovyks: What if the chain supports catch-up but does not have `local`? I'm expecting it to get the `remote`, but also get newer requests from chain.

chain-signatures/node/src/stream/mod.rs
```diff
@@ -115,6 +123,11 @@ pub async fn run_stream<S: ChainStream>(
     )
     .await;
 
+    // NOTE: we need to start after we recover entries from backlog and starting the run_stream task
```
volovyks, line 124: In the case of Ethereum, are we recovering what was saved, then catching up, then starting the stream of live events?

chain-signatures/node/src/stream/mod.rs
```diff
@@ -139,6 +152,23 @@ pub async fn run_stream<S: ChainStream>(
                     tracing::error!(?err, chain = %chain, "failed to process respond bidirectional event");
                 }
             }
+            ChainEvent::CatchupCompleted(catchup_chain) => {
+                if catchup_chain != chain {
+                    tracing::warn!(stream_chain = %chain, event_chain = %catchup_chain, "received mismatched catchup completion event");
```
volovyks: Looks like an error, not a warning

chain-signatures/node/src/stream/mod.rs
ChaoticTempest: not exactly. According to @ppca, we are doing catchup and livestream of events at the same time. This needs to be fixed so we can make it linear stream of events, where catchup leads to the livestream

chain-signatures/node/src/stream/mod.rs
ChaoticTempest: wait, i dont event remember why I had the chain parameter in `CatchupComplete(Chain) ` in the first place. It's kinda redundant

chain-signatures/node/src/backlog/mod.rs
ChaoticTempest: yeah, the logic didn't handle all cases. I made it so that we have requeue mode regardless of local or remote selected. So the choosing function is simpler now

chain-signatures/node/src/backlog/mod.rs
ChaoticTempest: yep, made changes so that we proceed with requeue_mode no matter if local or remote

chain-signatures/node/src/stream/mod.rs
volovyks: yeah,
That is why I want to have atomic block processing with block height check (new_block_height = processed_block_height + 1).

volovyks (approved):

PR #735, opened 2026-04-01
State Sync Component tests

volovyks:
- Added State Sync setup to Component test
- Added a couple of component tests to cover the main flows
- I intended to check sync on every step. We already have an integration test for end-to-end flows
- I had no intention to test active/need_sync states. Those are out of scope here
- Also renamed some functions and fields in test setup

### DIFF
--- a/integration-tests/src/mpc_fixture/builder.rs
+++ b/integration-tests/src/mpc_fixture/builder.rs
@@ -64,7 +64,7 @@ struct FixtureConfig {
     num_nodes: u32,
 
     use_preshared_triples: bool,
-    presignature_stockpile: bool,
+    use_preshared_presignatures: bool,
 
     node_min_triples: u32,
     network_max_triples: u32,
@@ -117,7 +117,7 @@ impl FixtureConfig {
             input: FixtureInput::load(num_nodes),
             num_nodes,
             use_preshared_triples: false,
-            presignature_stockpile: false,
+            use_preshared_presignatures: false,
             node_min_triples: 10,
             network_max_triples: 10 * num_nodes * 4,
             node_min_presignatures: 10,
@@ -295,8 +295,8 @@ impl MpcFixtureBuilder {
     }
 
     /// Use presignatures from fixture input
-    pub fn with_presignature_stockpile(mut self) -> Self {
-        self.fixture_config.presignature_stockpile = true;
+    pub fn with_preshared_presignatures(mut self) -> Self {
+        self.fixture_config.use_preshared_presignatures = true;
         self
     }
 
@@ -384,7 +384,8 @@ impl MpcFixtureBuilder {
     /// This setup will not attempt to stockpile triples or presignatures.
     pub fn only_generate_signatures(self) -> Self {
         self.with_preshared_key()
-            .with_presignature_stockpile()
+            .with_preshared_triples()
+            .with_preshared_presignatures()
             .with_node_min_triples(0)
             .with_node_min_presignatures(0)
     }
@@ -573,7 +574,7 @@ impl MpcFixtureNodeBuilder {
         let presignature_storage =
             Presignature::storage(&context.redis_pool, &self.participant_info.account_id);
 
-        if fixture_config.presignature_stockpile {
+        if fixture_config.use_preshared_presignatures {
             // removing here because we can't clone a presignature
             let my_shares = fixture_config.input.presignatures.remove(&self.me).unwrap();
             for (owner, presignature_shares) in my_shares {
--- a/integration-tests/src/mpc_fixture/builder.rs
+++ b/integration-tests/src/mpc_fixture/builder.rs
@@ -525,7 +525,7 @@ impl MpcFixtureNodeBuilder {
             context.contract_state,
             mpc_node::protocol::sync::SyncTask::synced_nodes_channel().0,
         );
-        let sync_task_handle = tokio::spawn(sync_task.run());
+        tokio::spawn(sync_task.run());
 
         let mut node = MpcFixtureNode {
             me: self.me,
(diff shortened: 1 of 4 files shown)


integration-tests/src/mpc_fixture/builder.rs
```diff
@@ -384,7 +384,8 @@ impl MpcFixtureBuilder {
     /// This setup will not attempt to stockpile triples or presignatures.
     pub fn only_generate_signatures(self) -> Self {
         self.with_preshared_key()
-            .with_presignature_stockpile()
+            .with_preshared_triples()
+            .with_preshared_presignatures()
```
jakmeier, line 390: Is there a reason to add `with_preshared_triples` when we already have `with_node_min_triples`?

I wonder if combining `with_preshared_triples` + `with_preshared_presignatures` can cause triple reuse problems. At least the stockpiled presignatures that I generated initially were from the triples that are also checked in.

integration-tests/src/mpc_fixture/builder.rs
```diff
@@ -512,6 +513,20 @@ impl MpcFixtureNodeBuilder {
             self.messaging.filter,
         );
 
+        // --- SyncChannel and SyncTask setup ---
+        use mpc_node::node_client::{NodeClient, Options as NodeClientOptions};
+        use mpc_node::protocol::sync::SyncTask;
```
jakmeier: Can we avoid mid-function `use` statements? I don't think the code base currently uses this style.

integration-tests/src/mpc_fixture/builder.rs
```diff
@@ -522,6 +537,8 @@ impl MpcFixtureNodeBuilder {
             triple_storage,
             presignature_storage,
             backlog: Backlog::new(),
+            sync_channel,
+            _sync_task_handle: sync_task_handle,
```
jakmeier: Please remove `_sync_task_handle`, there is no need to keep it around if you are not joining or aborting it anywhere.

integration-tests/src/mpc_fixture/fixture_interface.rs
```diff
@@ -152,6 +160,65 @@ impl MpcFixtureNode {
...
+    /// `triples` and `presignatures` are the lists of IDs the owner claims to hold.
+    /// Returns the SyncUpdate response (IDs missing on this node).
+    pub async fn sync(
+        &self,
+        from: cait_sith::protocol::Participant,
+        triples: Vec<u64>,
+        presignatures: Vec<u64>,
+    ) -> mpc_node::protocol::sync::SyncUpdate {
+        use mpc_node::protocol::sync::SyncUpdate;
+        let update = SyncUpdate {
```
jakmeier: style nit: I would either write it fully qualified both in return type and the constructor, or use a single `use` statement outside the function to use the short form `SyncUpdate` twice.

integration-tests/src/mpc_fixture/fixture_interface.rs
```diff
@@ -48,6 +50,12 @@ pub struct SharedOutput {
 }
 
 impl MpcFixture {
+    pub fn sorted_participants(&self) -> Vec<Participant> {
+        let mut p: Vec<_> = self.nodes.iter().map(|n| n.me).collect();
+        p.sort();
```
jakmeier, line 54: just a side-comment, no need to change anything:

It is a good habit to prefer `sort_unstable` over `sort` in cases like `sorted_participants`,  `owned_triples`, and `owned_presignatures`. We have a total order over these elements, so stable or unstable makes no logical difference. But stable sort in Rust adds an extra allocation. Doesn't really matter here in tests but just something to keep in mind.

jakmeier (dismissed):
Great test addition, thank you for implementing those!

I left a few comment but nothing major.

integration-tests/src/mpc_fixture/fixture_interface.rs
volovyks: Can we assume that with a new version of Rust `sort()` will get this improvement?

integration-tests/src/mpc_fixture/builder.rs
volovyks: `with_node_min_triples` ensures we do not generate new triples
`with_preshared_triples` adds pregenerated triples to the storage, which allows writing tests that require triples (like state sync)

When regenerating triples and presignatures, I've:
- generated triples
- then generated presignatures out of those triples
- then regenerated all triples
Should not be an issue.

integration-tests/src/mpc_fixture/builder.rs
volovyks: addressed, good point

integration-tests/src/mpc_fixture/builder.rs
volovyks: addressed

integration-tests/src/mpc_fixture/fixture_interface.rs
volovyks: addressed, thanks

integration-tests/src/mpc_fixture/fixture_interface.rs
jakmeier: No, this has nothing to do with language stability. It's whether the algorithm preserves the order of equal elements.

https://doc.rust-lang.org/std/primitive.slice.html#method.sort_unstable

integration-tests/src/mpc_fixture/builder.rs
jakmeier: Ok but what I meant is that we wouldn't really need triples if `only_generate_signatures` is called. Triples should not be used in any way.

jakmeier (approved):

PR #736, opened 2026-04-01
Remove participant id maping after removal

volovyks:
The resharing was successful, but I've noticed that lifted was still in the list if acc -> id map. I've applied a fix, not sure why we have not seen it before.

@ChaoticTempest having lifted in the list should not be a huge harm, but we may also remove it for sanity.

I've noticed that we are not updating ids. If we had 1, 2, 3 and removed node 2, we will have 1, 3. When we add a node we will have 1, 3, 4. That should not be a problem for Cait-sith, but we need to keep that in mind.

cc @auto-mausx

### DIFF
--- a/chain-signatures/contract/src/primitives.rs
+++ b/chain-signatures/contract/src/primitives.rs
@@ -143,6 +143,7 @@ impl Participants {
 
     pub fn remove(&mut self, account_id: &AccountId) {
         self.participants.remove(account_id);
+        self.account_to_participant_id.remove(account_id);
     }
 
     pub fn get(&self, account_id: &AccountId) -> Option<&ParticipantInfo> {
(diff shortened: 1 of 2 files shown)


volovyks:
@ChaoticTempest, I've added a fix; now we are reindexing all the nodes after each resharing.
It also means that the redundant lifted entry will be automatically removed after we add one more node.
@ChaoticTempest, please, deploy a new version of the contract on mainnet once this is merged and ping @auto-mausx

volovyks:
Looks like Cait-sith requires IDs to be stable. Going back to plan A.

ChaoticTempest (approved):

ChaoticTempest:
yep, which is why we have a counter that just increments forever for participants. We could potentially have a map of who can take what ID but that would add complexity and work right now

PR #738, opened 2026-04-02
Implement `generating` and `using` for State Sync

volovyks:
Relies on https://github.com/sig-net/mpc/pull/697, let's merge it first, cc @ChaoticTempest 
It would be great to merge and extend https://github.com/sig-net/mpc/pull/735 before merging this PR

I've replaced `reserved` and `used` with `generating` and `using. It is closer to the design and required by StteSync.


chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -732,21 +734,26 @@ impl PresignatureSpawner {
...
-                        .set(self.len_mine().await as i64);
-                    crate::metrics::storage::NUM_PRESIGNATURES_TOTAL
-
-                        .set(self.len_generated().await as i64);
-                    crate::metrics::protocols::NUM_PRESIGNATURE_GENERATORS_TOTAL
-
-                        .set(
-                            self.len_potential().await as i64 - self.len_generated().await as i64,
+                _ = stockpile_interval.tick() => {
+                    if active.len() >= self.threshold {
```
jakmeier, line 735: Interesting. This is a subtle change. Not sure if intentional. But previously the interval future was paused until we have enough active nodes. After that, I believe it would catch up all missed interval ticks in rapid succession. (See [MissedTickBehavior::Burst](https://docs.rs/tokio/latest/tokio/time/enum.MissedTickBehavior.html#variant.Burst))

What you do now will keep polling the interval even when the condition is not met, thus consuming wake up events. That avoids the burst of catch up wake up events. That should be good.

Even better would be to use `set_missed_tick_behavior(MissedTickBehavior::Skip)` to make it explicit that we don't want the burst.

chain-signatures/node/src/storage/protocol_storage.rs
```diff
@@ -210,51 +202,35 @@ impl<A: ProtocolArtifact> ProtocolStorage<A> {
...
             return None;
         }
-        drop(used);
-
-        let start = Instant::now();
-        let Some(mut conn) = self.connect().await else {
-            self.reserved.write().await.remove(&id);
+        if self.contains(id).await {
+            self.generating.write().await.remove(&id);
+            tracing::error!(id, "cannot create slot: artifact already exists in storage");
```
jakmeier: Don't you add a race condition by releasing the `using` guard before writing to `generating`? Multiple threads could be trying to insert the the new slot simultaneously.

I guess only one of them will succeed in inserting to `generating`. So maybe it's okay. But that would at least warrant a comment explaining the interaction.

chain-signatures/node/src/storage/protocol_storage.rs
```diff
@@ -336,20 +312,17 @@ impl<A: ProtocolArtifact> ProtocolStorage<A> {
...
-                    let mut used = self.used.write().await;
-                    for id in outdated.iter() {
-                        used.remove(id);
-                    }
-                }
+                // Filter out artifacts that are on this node but not in Redis:
+                // - `generating`: being generated, not yet persisted
+                // - `using`: taken from Redis, actively consumed by a protocol
+                let generating = self.generating.read().await;
+                let using = self.using.read().await;
```
jakmeier: Could you somewhere add a comment that defines the locking order between those two? (Always locking them in the same order avoids the most deadlock caused by multiple simultaneous locks.)

chain-signatures/node/src/storage/protocol_storage.rs
```diff
@@ -336,20 +312,17 @@ impl<A: ProtocolArtifact> ProtocolStorage<A> {
...
+                let not_found: Vec<_> = not_found
+                    .into_iter()
+                    .filter(|id| !generating.contains(id) && !using.contains(id))
+                    .collect();
+                drop(generating);
+                drop(using);
```
jakmeier: Do these explicit drops do anything anymore? Should be the same as the implicit drop behavior, no?

chain-signatures/node/src/storage/protocol_storage.rs
```diff
@@ -362,7 +335,8 @@ impl<A: ProtocolArtifact> ProtocolStorage<A> {
     /// Insert an artifact into storage under `owner`'s ownership set.
     /// Holders must be set on the artifact before calling this; they are
     /// persisted as a dedicated Redis set for later holder-tracking.
-    pub async fn insert(&self, artifact: A, owner: Participant) -> bool {
+    /// Private: callers must use `create_slot()` + `ArtifactSlot::insert()`.
+    async fn insert(&self, artifact: A, owner: Participant) -> bool {
```
jakmeier, line 436: 👍

jakmeier (commented):
I've reviewed the commits on top of #697, looking mostly good to me.

Please review the async + locking details where I added comments. But it's probably fine.

Not approving yet, as we need to wait for the other PR anyway.

chain-signatures/node/src/storage/protocol_storage.rs
volovyks: removed, thanks

chain-signatures/node/src/protocol/presignature.rs
volovyks: I was not aware of these strategies; it just looked cleaner. Yes, burst is definitely not what we want, added Skip.

chain-signatures/node/src/storage/protocol_storage.rs
volovyks: To make sure we never have issues with that, I've merged `generating` and `using` into a single struct inside of a RwLock. That should address both this and https://github.com/sig-net/mpc/pull/738#discussion_r3027295801. WDYT?

jakmeier: Very nice idea! Yes, that addresses both in an elegant way.

volovyks:
@ChaoticTempest @jakmeier I've added tests and a couple of fixes. For State Sync to work properly, we need to distinguish owned/not_owned generations and in-use artifacts.

chain-signatures/node/src/storage/protocol_storage.rs
```diff
@@ -210,51 +245,65 @@ impl<A: ProtocolArtifact> ProtocolStorage<A> {
...
+            }
+            let me = self.me().ok()?;
+            state.generating.insert(id, owner == me);
         }
-        if !self.reserved.write().await.insert(id) {
+        if self.contains(id).await {
+            self.reserved.write().await.generating.remove(&id);
+            tracing::error!(id, "cannot create slot: artifact already exists in storage");
             return None;
         }
```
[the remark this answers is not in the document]
volovyks: I would propose to allow each Storage function to fail (return Result) for consistency and robustness. But that deserves a separate PR.

chain-signatures/node/src/storage/protocol_storage.rs
```diff
@@ -68,37 +69,28 @@ pub trait ProtocolArtifact:
...
     fn drop(&mut self) {
-        self.unreserve();
+        let storage = self.storage.clone();
+        let id = self.id;
+        if let Ok(handle) = tokio::runtime::Handle::try_current() {
+            handle.spawn(async move {
+                storage.reserved.write().await.generating.remove(&id);
+            });
+        }
     }
```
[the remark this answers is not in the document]
volovyks: adressed

chain-signatures/node/src/storage/protocol_storage.rs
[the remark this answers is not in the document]
volovyks: Specifically:
- len_generated 
- contains 
- contains_by_owner 
- len_by_owner 
- is_empty

They return 0 or false by default, which is not great

integration-tests/tests/cases/state_sync.rs
```diff
@@ -0,0 +1,490 @@
...
+        let holders = node1.presignature_storage.fetch_holders(*id).await;
+        assert_eq!(
+            holders, all_participants,
+            "fixture presig {id} should still have all holders"
+        );
+    }
+}
+
+#[test(tokio::test(flavor = "multi_thread"))]
+async fn test_sync_matrix() {
```
jakmeier, line 283: Nice test setup!

jakmeier (dismissed):
Looking good! But I guess we still have to wait for other PRs first

ChaoticTempest (dismissed):

ChaoticTempest (approved):

PR #739, opened 2026-04-02
test: MPC + stream component test

jakmeier:
Create a MockStream that connects to the MPC fixture setup. This lets us test the indexer stream <-> MPC glue.


chain-signatures/node/src/cli.rs
```diff
@@ -214,7 +214,7 @@ pub async fn run(cmd: Cli) -> anyhow::Result<()> {
             );
             crate::metrics::nodes::CONFIGURATION_DIGEST.set(digest);
 
-            let (sign_tx, sign_rx) = mpsc::channel(16384);
+            let (sign_tx, sign_rx) = mpsc::channel(if cfg!(test) { 1 } else { 16384 });
```
volovyks: I guess this is more of an experiment, but we can make it configurable in tests.

volovyks: not related to this PR, but it definitely deserves a constant

integration-tests/src/mpc_fixture/mock_stream.rs
```diff
@@ -0,0 +1,140 @@
...
+        let mut guard = self.inner.lock().await;
+        guard.progress_block_height(steps)
+    }
+
+    pub async fn sign_requests(&self, requests: &[IndexedSignRequest]) {
+        let mut guard = self.inner.lock().await;
+        guard.sign_requests(requests)
+    }
+
+    pub async fn rpc_actions(&self, actions: &[RpcAction]) {
```
volovyks: It is not clear what we are doing with these actions. Are we queuing them? Processing? Adding to a specific block? I'm ok with a bit longer names. That makes the code a bit more readable IMO.

integration-tests/tests/cases/mpc_with_stream.rs
```diff
@@ -0,0 +1,93 @@
...
+    let request = [sign_request(0)];
+    network[0].mock_streams[0].sign_requests(&request).await;
+    network[1].mock_streams[0].sign_requests(&request).await;
+    network[2].mock_streams[0].sign_requests(&request).await;
+
+    network[0].mock_streams[0].progress_block_height(1).await;
+    network[1].mock_streams[0].progress_block_height(1).await;
+    network[2].mock_streams[0].progress_block_height(1).await;
+
+    let timeout = Duration::from_secs(10);
```
volovyks, line 41: nit: Such a pattern significantly slows down our tests; I would add a helper that actually waits for N events or specific events with a timeout.

volovyks (commented):
Nice! test_channel_contention should be very useful in our work on backlog and stability.

chain-signatures/node/src/cli.rs
jakmeier: yeah it's just an experiment that I will clean up before removing the draft label

integration-tests/tests/cases/mpc_with_stream.rs
jakmeier: Wait, doesn't this already do exactly what you ask?

```rust
let timeout = Duration::from_secs(10);
let actions = network.assert_actions(1, timeout).await;
```

```rust
    pub async fn assert_actions(
        &self,
        threshold_per_node: usize,
        timeout: Duration,
    ) -> HashSet<String> {
        let result = tokio::time::timeout(timeout, self.wait_for_actions(threshold_per_node)).await;
        if result.is_err() {
            self.print_actions().await;
        }
        result.expect("should produce enough signatures")
    }
```

jakmeier:
I'm still experimenting here. Getting some interesting results.

1. There is some luck involved if nodes can handle 50 signature requests at once coming from the Solana stream. However, channel buffers have no influence on that. So probably that is not problem.
2. Meanwhile, 50 requests seems to be a hard threshold for successful signatures. To get more (e.g. 51) I have to send way more requests. I think then they have better chances, with more requests running in parallel.
3. I often observed 25, 50 or 75 successful  signatures. Don't know why, yet. Could be an artifact from the test setup.

I will take a closer look again next week.

jakmeier:
Still investigating but this is interesting.

image

I see that 25 signatures get through in the first round. Then they wait for a timeout. (3s for faster iteration in the graph but it is the same with longer timeouts) One more is accepted in the third round. This is exactly the same for all participants.

After the third round, no more signatures are produced. They always fail in the posit phase with "proposer timeout waiting for presignature, reorganizing". Could be they are just out of presignatures. Still interesting that the need timeouts to make progress.

Retry round alignment seems to works fine, too. Only a short time window where different nodes have different rounds.

integration-tests/tests/cases/mpc_with_stream.rs
volovyks: Yes, but why do we have the timeout? What if it happens faster?

jakmeier: [`tokio::time::timeout`](https://docs.rs/tokio/latest/tokio/time/fn.timeout.html) returns early in that case

`wait_for_actions` keeps polling at a 100ms interval to check if there are enough actions

jakmeier:
The issue from my previous comment was that I didn't use good entropy in requests. Hence, at round 0, all requests went to participant 0 (first 25 requests) and after one timeout in round 2, all requests went to participant 2.

That's fixed now and it actually works rather stable, even when adding 1M requests all at once.

jakmeier:
The new problem I was able to identify is with delayed requests. With 1.5 times the posit timeout delay between when nodes see incoming blocks, they don't get a single request signed. It looks like they fail on achieving round alignment.

full_delay

Change that to only 0.5 times the posit timeout, and all signatures are generated as expected in round 0.

half_delay

This is reproducible with the pushed test `mpc_with_stream::test_channel_contention_multiple_blocks_at_once_delayed`.

integration-tests/tests/cases/mpc_with_stream.rs
```diff
@@ -0,0 +1,147 @@
...
+
+#[test(tokio::test(flavor = "multi_thread"))]
+async fn test_channel_contention_multiple_blocks_at_once() {
+    check_channel_contention(5, 10, 50, None).await;
+}
+
+#[test(tokio::test(flavor = "multi_thread"))]
+async fn test_channel_contention_multiple_blocks_at_once_delayed() {
+    // delay should be > ORGANIZE_POSIT_TIMEOUT
+    let delay = mpc_node::protocol::signature::organize_posit_timeout() * 3 / 2;
```
jakmeier: Changing this to less than a full posit delay make the test pass. Hence we know we only run into the issue when observations are apart more than a posit timeout (20s).


```suggestion
    let delay = mpc_node::protocol::signature::organize_posit_timeout() / 2;
```

jakmeier:
Somehow `test_sign_no_presignature_waste` now fails, as only 72 out of 75 signatures are produced.

I only changed the entropy of the signature requests. This suggests that "unlucky" proposer assignment can cause long delays, too. I tried increasing the timeout and using different seeds.

But nothing works except the entropy currently on develop. That one assigned all requests to the same participant in any given round. (round 0, proposer = Participant(0); round 1, proposer = Participant(1); round 2, proposer = Participant(2)) Apparently that's the only way we have no presignature waste :confused:

jakmeier:
`test_sign_no_presignature_waste` fails because we eagerly start the signature when the threshold is reached. Node may then becomes proposer for signatures that are already taken care of and we waste signatures.

Fix in https://github.com/sig-net/mpc/pull/765

jakmeier:
Update: The problem is that non-participants hold on to, any may dispose, presignatures while they attempt to become proposer for an already ongoing signature.

I think this shows we need 2 fixes. We should not waste presignatures in this case, and we should generally avoid posit round spamming when a participant is not selected as proposer.

I created https://github.com/sig-net/mpc/issues/793 for this.

jakmeier:
ah, the rebase broke this one :/

well, it was not going to pass tests anyway, I will add a MockIndexer for the new Stream abstraction later

jakmeier:
@ppca can you review this? It was blocked for a long time but now with https://github.com/sig-net/mpc/pull/819 merged, it should be ready for review.

This is an extension to the component testing framework to let us submit events from different chains in a controlled way. For example, we can simulate bad timing in how nodes observe events. This should be quite relevant to your current investigations into instability problems.

integration-tests/src/mpc_fixture/builder.rs
```diff
@@ -421,6 +424,22 @@ impl MpcFixtureBuilder {
...
             .with_node_min_presignatures(0)
     }
+
+    /// Add a mock stream to all nodes.
+    ///
+    /// Each node will have a independent deep-clone of the provided stream.
+    /// Events are thus delivered to all nodes.
+    pub async fn with_mock_stream(mut self, chain: Chain, stream: MockStream) -> Self {
+        for node in &mut self.prepared_nodes {
+            let cloned = stream.deep_clone().await;
```
ppca: does this mean we will not be able to test the situation where some nodes miss some events? 
I think we run into the issue of missing respond events on nodes sometimes with rpc sometimes rate limited.

ppca (dismissed):
lgtm. Let's merge this.

ppca:
Component tests were failing because of `test_channel_contention_1M_requests`.  Seems like what happened was the nodes are in organizing -> posit loop forever and zero signature was generated in 75 seconds. A lot of "deliberator timeout waiting for Propose, reorganizing" and "proposer timeout waiting for concurrency slot, reorganizing"

jakmeier:
Hm, this never failed locally with many runs. Probably the CI machines are too slow.

Increasing the timeout doesn't help. I guess this needs more investigation then.

> @ppca wrote:
> zero signature was generated in 75 seconds

Was it really zero? That would be quite bad.

The run I checked had 43 successful signatures. You can search the logs for a line like this.

`All published RPC actions: count=43`

integration-tests/src/mpc_fixture/fixture_interface.rs
```diff
@@ -178,11 +181,24 @@ impl MpcFixture {
...
+    /// Add a block that contains signature requesting events, visible to all
+    /// nodes, then progress the chain to execute it immediately.
+    pub async fn process_sign_requests(&self, chain: Chain, requests: &[IndexedSignRequest]) {
+        for node in &self.nodes {
+            let stream = node
+                .mock_streams
+                .get(&chain)
+                .expect("must have mock stream configured");
+            stream.prepare_block_of_sign_requests(requests).await;
+            stream.progress_block_height(1).await;
```
volovyks, line 199: Is it ok to hardcode the block height here?

integration-tests/tests/cases/mpc_with_stream.rs
```diff
@@ -0,0 +1,148 @@
...
+}
+
+#[test(tokio::test(flavor = "multi_thread"))]
+async fn test_channel_contention_multiple_blocks_at_once() {
+    check_channel_contention(5, 10, 50, None).await;
+}
+
+#[test(tokio::test(flavor = "multi_thread"))]
+async fn test_channel_contention_multiple_blocks_at_once_delayed() {
+    // TODO: delay should be > ORGANIZE_POSIT_TIMEOUT but right now the system can't handle it
```
volovyks, line 126: Good finding!

volovyks (dismissed):
Great setup!

integration-tests/src/mpc_fixture/fixture_interface.rs
jakmeier: This method progresses by exactly one block height, for example from 100 to 101. We can change the API later if tests prefer to let it progress to height X instead.

volovyks (approved):
@jakmeier, I will try to find the reason if it fails again

jakmeier:
> @jakmeier, I will try to find the reason if it fails again

Thanks! I think we should get the testing infrastructure merged in, so we can start using it for other cases. Even if some of my new test cases still fail.

jakmeier:
With problematic tests ignored, CI is green. I can take another look at those next week

PR #740, opened 2026-04-05
feat: add Canton chain integration

Pessina:
## Why

Canton has no MPC integration today. This PR adds Canton as a first-class chain in the MPC node so Canton ledger applications (e.g., `Erc20Vault`) can request threshold ECDSA signatures for cross-chain operations (Canton → EVM → Canton) using the same backlog, presignature, and signature protocols as Ethereum, Solana, and Hydration.

## Summary

- Adds Canton as a chain in the MPC node, following the existing plugin architecture.
- Canton supports the **bidirectional** sign flow only — `SignKind::Sign` is rejected (Daml's `Signer` contract does not expose a one-shot `Sign` choice yet).
- Canton's `SignBidirectionalEvent` does not include a `deposit` field; `deposit()` returns `0` until the Daml contract adds it.

## Defense-in-depth verification (`stream.rs::verify_sign_event`)

Three checks on top of Daml ledger guarantees before a `SignBidirectionalEvent` is admitted:

1. Every `operators` entry from the payload is a signatory on the `CreatedEvent`.
2. `requester` from the payload is a signatory on the `CreatedEvent`.
3. The same transaction contains an `ExercisedEvent` with `choice == "SignBidirectional"` on the pinned `signer_contract_id` — proves the event was created through the correct Daml code path and not a forged contract with a matching template ID.

## Key decisions

- `canton:global` as synthetic CAIP-2 ID (Canton has no registered ChainAgnostic namespace).
- Per-chain respond serialization: ABI for Canton, Borsh for Solana/Hydration.
- `sign_event_contract_id` is threaded through an opaque Borsh-encoded `CantonChainCtx` on `RespondBidirectionalTx.chain_ctx` so phase 2 (`RespondBidirectional`) can respond to the original Daml sign-event contract.
- DER signature encoding on the MPC side — required by Daml's `secp256k1WithEcdsaOnly`.

## Known limitations

- **DAR upgrades require node restart** — `signer_contract_id` and `signer_template_id` are pinned via CLI. After a `dpm build --all` + redeploy, both change and every inbound event is rejected until the node is restarted with the new IDs. Tracked in #774.

## Follow-up Tickets

- #770 — Golden-test Canton request ID hashing against TS/Daml.
- #771 — Cover EVM transaction param parsing edge cases.
- #772 — Cover template suffix matching near-misses.
- #773 — Golden-test Canton EVM transaction encoding against viem.
- #774 — Plan DAR upgrade handling without node downtime.
- #775 — Unit-test Canton sign event verification failures.


chain-signatures/node/src/indexer_canton/conn.rs
```diff
@@ -0,0 +1,133 @@
+use super::ledger_api::{
+    ActiveContractEntry, CumulativeFilter, EventFormat, GetActiveContractsRequest,
+    IdentifierFilter, JsCommands, LedgerEndResponse, PartyFilter,
+    SubmitAndWaitForTransactionRequest, SubmitAndWaitForTransactionResponse, TemplateFilterValue,
+};
+use jsonwebtoken::{encode, Algorithm, EncodingKey, Header};
+
+#[derive(serde::Serialize)]
+struct JwtClaims {
```
ppca: what does sub, iat, exp and nbf refer to? either more comments/doc or a name easier to understand would be needed.

chain-signatures/node/src/indexer_canton/conn.rs
```diff
@@ -0,0 +1,133 @@
...
+                        },
+                    }],
+                })?,
+                None => serde_json::json!({}),
+            };
+            filters.insert(party.to_string(), value);
+        }
+
+        let req = GetActiveContractsRequest {
+            active_at_offset: offset,
```
ppca: if the offset value of the ledger has changed between line 75 and here, would it be a problem?

chain-signatures/node/src/indexer_canton/contracts.rs
```diff
@@ -0,0 +1,188 @@
...
+
+/// EVM transaction parameters from the Signer contract.
+/// Address fields are 40-char hex (20 bytes); numeric fields are 64-char hex.
+#[derive(Clone, Debug, PartialEq, Eq, Hash, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct EvmTransactionParams {
+    pub to: String,
+    pub function_signature: String,
+    #[serde(default)]
+    pub encoded_args: String,
```
ppca: can we add some doc/comment on what encoded_args is?

chain-signatures/node/src/indexer_canton/ledger_api.rs
```diff
@@ -0,0 +1,399 @@
...
+    ArchivedEvent(ArchivedEvent),
+    ExercisedEvent(ExercisedEvent),
+}
+
+/// A contract creation event.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct CreatedEvent {
+    pub contract_id: String,
+    pub template_id: String,
```
ppca, line 96: may I ask again what's the difference between contract_id and template_id?

chain-signatures/node/src/indexer_canton/ledger_api.rs
```diff
@@ -0,0 +1,399 @@
...
+    pub created_event_blob: Option<String>,
+    #[serde(default)]
+    pub signatories: Vec<String>,
+    #[serde(default)]
+    pub witness_parties: Vec<String>,
+    /// Position of this event in the transaction tree (LEDGER_EFFECTS only).
+    #[serde(default)]
+    pub node_id: Option<u32>,
+    #[serde(default)]
+    pub package_name: Option<String>,
```
ppca, line 111: what does the package_name here refer to?

chain-signatures/node/src/indexer_canton/ledger_api.rs
```diff
@@ -0,0 +1,399 @@
...
+}
+
+/// Subscription message sent to `ws://.../v2/updates`.
+///
+/// Uses `updateFormat` (Canton 3.4+) instead of the deprecated
+/// `filter`/`verbose` top-level fields
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct GetUpdatesRequest {
+    pub begin_exclusive: u64,
```
ppca, line 151: what is this param for?

chain-signatures/node/src/indexer_canton/ledger_api.rs
```diff
@@ -0,0 +1,399 @@
...
+#[serde(rename_all = "camelCase")]
+pub struct OffsetCheckpoint {
+    pub offset: u64,
+}
+
+/// Request body for `POST /v2/parties`.
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct AllocatePartyRequest {
+    pub party_id_hint: String,
```
ppca, line 221: some comments on param is appreciated

chain-signatures/node/src/indexer_canton/ledger_api.rs
```diff
@@ -0,0 +1,399 @@
...
+pub struct AllocatePartyResponse {
+    pub party_details: PartyDetails,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct PartyDetails {
+    pub party: String,
+    #[serde(default)]
+    pub is_local: bool,
```
ppca, line 242: what is a local party?

chain-signatures/node/src/indexer_canton/ledger_api.rs
```diff
@@ -0,0 +1,399 @@
...
+    pub user: UserInfo,
+    #[serde(default)]
+    pub rights: Vec<UserRight>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(rename_all = "camelCase")]
+pub struct UserInfo {
+    pub id: String,
+    pub primary_party: String,
```
ppca, line 258: what is the primary_party?

ppca: is it the requester/observer?

chain-signatures/node/src/indexer_canton/request_id.rs
```diff
@@ -0,0 +1,70 @@
+use super::contracts::{EvmTransactionParams as CantonEvmTransactionParams, TxParams};
+use super::CantonSignBidirectionalRequestedEvent;
+use alloy::primitives::{keccak256, U256};
+use alloy_sol_types::SolValue;
+
+fn hex_u256(field: &'static str, hex: &str) -> anyhow::Result<U256> {
+    U256::from_str_radix(hex, 16).map_err(|e| anyhow::anyhow!("invalid hex in '{field}': {e}"))
```
ppca: field is not used in U256::from_str_radix(hex, 16), just to report error?

chain-signatures/node/src/indexer_canton/stream.rs
```diff
@@ -0,0 +1,527 @@
...
+
+    async fn next_event(&mut self) -> Option<ChainEvent> {
+        self.rx.recv().await
+    }
+}
+
+/// Main event loop with reconnection logic and exponential backoff.
+async fn run_canton_event_loop(
+    config: CantonConfig,
+    tx: mpsc::Sender<ChainEvent>,
```
ppca: might be better to rename to events_tx so it's clear it's a series of events not a single tx

chain-signatures/node/src/indexer_canton/stream.rs
```diff
@@ -0,0 +1,527 @@
...
+    // If already at or past the ledger end, emit CatchupCompleted immediately
+    if counter >= catchup_target {
+        tracing::info!(counter, catchup_target, "canton already caught up");
+        if tx.send(ChainEvent::CatchupCompleted).await.is_err() {
+            tracing::error!("canton event channel closed");
+            return;
+        }
+        catchup_completed = true;
+    } else {
+        tracing::info!(counter, catchup_target, "canton catching up");
```
ppca: can add in the target_counter too so we know how much the catchup is

chain-signatures/node/src/indexer_canton/stream.rs
```diff
@@ -0,0 +1,527 @@
...
+        format!("jwt.token.{jwt_token}, daml.ws.auth").parse()?,
+    );
+
+    let (ws_stream, _) = tokio::time::timeout(
+        std::time::Duration::from_secs(30),
+        tokio_tungstenite::connect_async(request),
+    )
+    .await
+    .map_err(|_| anyhow::anyhow!("canton WebSocket connect timed out"))??;
+    let (mut write, mut read) = ws_stream.split();
```
ppca: maybe change write/read to ws_write and ws_read?

chain-signatures/node/src/indexer_canton/stream.rs
```diff
@@ -0,0 +1,527 @@
...
+
+    loop {
+        let msg = tokio::select! {
+            maybe = read.next() => {
+                match maybe {
+                    Some(msg) => {
+                        last_ws_msg = tokio::time::Instant::now();
+                        msg?
+                    }
+                    None => break,
```
ppca, line 240: is there any websocket that needs to be dropped when we break?

chain-signatures/node/src/indexer_canton/stream.rs
```diff
@@ -0,0 +1,527 @@
...
+
+        if !*catchup_completed && *counter >= catchup_target {
+            tracing::info!(
+                counter = *counter,
+                catchup_target,
+                "canton catchup completed"
+            );
+            if tx.send(ChainEvent::CatchupCompleted).await.is_err() {
+                tracing::error!("canton event channel closed");
+                return Ok(());
```
ppca, line 309: why return Ok() here?

chain-signatures/node/src/indexer_canton/stream.rs
```diff
@@ -0,0 +1,527 @@
...
+                if msg.error.is_some() {
+                    tracing::warn!(error = ?msg.error, "canton ledger stream error");
+                }
+                continue;
+            }
+        }
+
+        if tx.send(ChainEvent::Block(*counter)).await.is_err() {
+            tracing::error!("canton event channel closed");
+            return Ok(());
```
ppca, line 298: this means we would never emit ChainEvent::Block(*counter) at this counter right? Because next time will start from counter + 1

chain-signatures/node/src/rpc.rs
```diff
@@ -893,6 +914,85 @@ impl HydrationClient {
...
+            read_as: vec![self.config.party_id.clone()],
+            commands: vec![Command::ExerciseCommand {
+                template_id: self.config.signer_template_id.clone(),
+                contract_id: self.config.signer_contract_id.clone(),
+                choice: choice.to_string(),
+                choice_argument,
+            }],
+            disclosed_contracts: vec![],
+        };
+        self.submit_and_wait(commands, &format!("canton {choice}"))
```
ppca, line 1074: we might need to add timeout and retries for this in the future if it can get stuck forever

chain-signatures/primitives/src/lib.rs
```diff
@@ -228,6 +250,7 @@ impl Chain {
             Chain::Solana => 3,
             Chain::Bitcoin => 60 * 60 + 20 * 60, // 6 confirmations at 10 minutes each, plus some buffer
             Chain::Hydration => 12,
+            Chain::Canton => 5,
```
ppca: this fast?

integration-tests/src/canton.rs
```diff
@@ -0,0 +1,602 @@
...
+) -> SignBidirectionalRequestedEvent {
+    let operators = vec![sandbox.operator_party.clone()];
+    let sender = compute_operators_hash(&operators);
+    SignBidirectionalRequestedEvent {
+        operators,
+        requester: sandbox.requester_party.clone(),
+        sig_network: sandbox.party_id.clone(),
+        sender,
+        tx_params: TxParams::EvmTxParams(test_evm_params(nonce.unwrap_or(0))),
+        caip2_id: "eip155:31337".to_string(),
```
ppca: this would need to be Chain::Ethereum.caip2_chain_id() when https://github.com/sig-net/mpc/pull/696 merges.

ppca (commented):
Mostly LGTM on the parts I'm familiar with. Nice!

chain-signatures/node/src/indexer_canton/conn.rs
Pessina: Sure, I will add docs. We can't change the name because that's standard for JWT.

.github/workflows/canton.yml
```diff
@@ -0,0 +1,180 @@
+name: Canton Integration Tests
+
+on:
```
volovyks, line 3: Why do we need a separate Job for Canton tests?

chain-signatures/node/src/indexer_canton/conn.rs
Pessina: No, it will fetch all contracts active on this offset and if we would like to fetch a newer one, we can call again. 

On the tests we wrap this function on a polling to wait for the MPC response.

.cargo/audit.toml
```diff
@@ -18,4 +18,11 @@ ignore = [
     "RUSTSEC-2026-0049",  # rustls-webpki 0.101.7/0.102.8 faulty CRL matching; no fix in those series, pinned by rustls 0.21/0.22 (reqwest 0.11, google-secretmanager).
     "RUSTSEC-2026-0098",  # rustls-webpki 0.101.7/0.102.8 URI name constraints incorrectly accepted; same dependency chain as RUSTSEC-2026-0049.
     "RUSTSEC-2026-0099",  # rustls-webpki 0.101.7/0.102.8 wildcard name constraints accepted; same dependency chain as RUSTSEC-2026-0049.
+    "RUSTSEC-2023-0071",  # rsa 0.9.10 Marvin Attack; no fix available, pulled in by jsonwebtoken (Canton JWT auth).
```
volovyks: Let's merge such things separately (not sure, but maybe most of them are in develop).

chain-signatures/node/src/indexer_canton/contracts.rs
Pessina: Good one, I will add a comment there.

chain-signatures/crypto/src/lib.rs
```diff
@@ -3,8 +3,8 @@ pub mod kdf;
 use k256::elliptic_curve::sec1::FromEncodedPoint;
 use k256::EncodedPoint;
 pub use kdf::{
-    check_ec_signature, derive_epsilon_eth, derive_epsilon_near, derive_epsilon_sol, derive_key,
-    x_coordinate,
+    check_ec_signature, derive_epsilon_canton, derive_epsilon_eth, derive_epsilon_hydration,
+    derive_epsilon_near, derive_epsilon_sol, derive_key, x_coordinate,
```
volovyks, line 7: Not sure why we need this change.

chain-signatures/node/src/indexer_canton/ledger_api.rs
Pessina: Sure, that's a confusing one. 

contract_id it's the equivalent of address in EVM, but with the caveat that all contracts in Canton are UTXOs, so if we need to mutate this contract, we will archive, create a new one and the contract_id will change

template_id it's more or less the hash of the Contract code, it's stable for it's lifetime, unless we change the code. 

1 template_id -> N contract_id (contract_id it's unique on chain)

chain-signatures/node/src/backlog/mod.rs
```diff
@@ -422,7 +422,6 @@ impl Backlog {
             tracing::warn!(
                 ?chain,
                 ?id,
-                ?status,
```
volovyks, line 453: nit: Why are we not logging status anymore?

chain-signatures/node/src/backlog/mod.rs
```diff
@@ -895,6 +894,7 @@ mod tests {
             request_id: [id; 32],
             from_address: Address::ZERO,
             nonce: 0,
+            chain_ctx: ChainContext::None,
```
volovyks: I've briefly looked into the code and have not found a good reason for having `chain_ctx` in `BidyrectionalTx` struct. Overall, we always want to keep chain-specific info on the integration layer.

chain-signatures/node/src/indexer_canton/ledger_api.rs
Pessina: The ledger_api.rs is hand-ported from the Canton JSON Ledger API v2 spec (no Rust SDK exists), that's why it doesn't have too many docs. 😅

I will add the Canton docs links on top of the file and we can check from there when we need more info.

Pessina: This comment is for all the comments on this file.

chain-signatures/primitives/src/lib.rs
```diff
@@ -35,6 +36,17 @@ impl ScalarExt for Scalar {
     }
 }
 
+/// The maximum valid scalar for the secp256k1 curve (group order minus one).
+pub static MAX_SECP256K1_SCALAR: LazyLock<Scalar> = LazyLock::new(|| {
```
volovyks, line 40: Nice! It belongs in primitives. But it is still defined a couple of times across the project. Let's use this one instead.

chain-signatures/node/src/indexer_canton/ledger_api.rs
Pessina: This is the offset (equivalent to block number) to start the websocket subscription

.github/workflows/canton.yml
volovyks: Hm, that may be a good idea if we only run it when the Canton package is updated.

chain-signatures/node/src/indexer_canton/ledger_api.rs
Pessina: This is the package name where our templates live, if you check the daml.yaml file it has the package name: https://github.com/sig-net/canton-poc/blob/docs/signer-vault-split-security/daml-packages/daml-signer/daml.yaml

I still have to rename this repo, but there you can find all the Templates that we will use in Canton

chain-signatures/node/src/stream/ops.rs
```diff
@@ -132,34 +168,46 @@ impl SignBidirectionalEvent {
                 &self.sender_string()?,
                 &self.path(),
             )),
+            SignBidirectionalEvent::Canton(event) => Ok(mpc_crypto::kdf::derive_epsilon_canton(
+                self.key_version(),
+                &event.sender,
```
volovyks: Why not use the same function for consistency?

chain-signatures/node/src/indexer_canton/ledger_api.rs
Pessina: This I'm not 100% sure, but a Party is Canton's equivalent of an EVM Wallet.

- `is_local: true` - "this" Canton Node hosts the party. This is our setup: the Canton Node hosts the `SigNetwork` party, and each MPC node authenticates as a user using a JWT.
- `is_local: false` - the party is known on the synchronizer but hosted by a "different" Canton Node

chain-signatures/node/src/stream/ops.rs
```diff
@@ -9,114 +13,146 @@ use crate::node_client::NodeClient;
...
+    Canton(CantonSignBidirectionalRequestedEvent),
 }
 
 impl SignBidirectionalEvent {
     pub fn sender(&self) -> [u8; 32] {
         match self {
             SignBidirectionalEvent::Solana(event) => event.sender.to_bytes(),
             SignBidirectionalEvent::Hydration(event) => event.sender,
+            // Canton party IDs are variable-length strings; hash fits them into [u8; 32].
+            SignBidirectionalEvent::Canton(event) => keccak256(event.sender.as_bytes()).into(),
```
volovyks: Is the fact that this is an irreversible process a reason we need a chain context in Bidirectional? Is it necessary to have the original sender?

chain-signatures/node/src/stream/ops.rs
```diff
@@ -9,114 +13,146 @@ use crate::node_client::NodeClient;
...
         }
     }
 
     pub(crate) fn sender_string(&self) -> anyhow::Result<String> {
```
volovyks, line 42: `sender_string(..)` is used in derivation; it is important to add chain id to the string

chain-signatures/node/src/stream/ops.rs
```diff
@@ -9,114 +13,146 @@ use crate::node_client::NodeClient;
...
             SignBidirectionalEvent::Hydration(event) => event.params.clone(),
+            SignBidirectionalEvent::Canton(event) => event.params.clone(),
         }
     }
 
     pub fn output_deserialization_schema(&self) -> Vec<u8> {
         match self {
             SignBidirectionalEvent::Solana(event) => event.output_deserialization_schema.clone(),
             SignBidirectionalEvent::Hydration(event) => event.output_deserialization_schema.clone(),
+            SignBidirectionalEvent::Canton(event) => {
```
volovyks: Why is this a special case?

chain-signatures/node/src/stream/ops.rs
```diff
@@ -9,114 +13,146 @@ use crate::node_client::NodeClient;
...
+            }
         }
     }
 
     pub fn respond_serialization_schema(&self) -> Vec<u8> {
         match self {
             SignBidirectionalEvent::Solana(event) => event.respond_serialization_schema.clone(),
             SignBidirectionalEvent::Hydration(event) => event.respond_serialization_schema.clone(),
+            SignBidirectionalEvent::Canton(event) => {
+                event.respond_serialization_schema.as_bytes().to_vec()
```
volovyks: Why is this a special case?

chain-signatures/node/src/stream/ops.rs
```diff
@@ -9,114 +13,146 @@ use crate::node_client::NodeClient;
...
             SignBidirectionalEvent::Hydration(event) => event.key_version,
+            SignBidirectionalEvent::Canton(event) => event.key_version,
         }
     }
 
     pub(crate) fn deposit(&self) -> u64 {
         match self {
             SignBidirectionalEvent::Solana(event) => event.deposit,
             SignBidirectionalEvent::Hydration(event) => event.deposit,
+            // Canton deposits can't be charged in Canton-native tokens the way
```
volovyks: No need to expose Canton specifics on the stream level. If `deposit` is not a thing, we will not use it.
Having such a comment is a good idea, but somewhere in the Canton integration layer.

chain-signatures/node/src/indexer_canton/ledger_api.rs
Pessina: Each User can control 1 or many Parties. The `UserInfo`/`CreateUserRequest` declares the User permissions(read/write) for each Party on the `rights` array

- The PrimaryParty is used when a tx is send and `act_as` is not specified. 
- If we wanna use another Party specified on the `rights` array, we need to pass `act_as`

This is used on test code to setup the Canton Node with the Users/Parties. In Production we will only have the sigNetwork Party and each MPC node will be registered a User on it with permission to read/write

chain-signatures/node/src/indexer_canton/request_id.rs
Pessina: Correct.

chain-signatures/node/src/stream/ops.rs
```diff
@@ -9,114 +13,146 @@ use crate::node_client::NodeClient;
...
         }
     }
 
     pub fn serialized_transaction(&self) -> Vec<u8> {
         match self {
             SignBidirectionalEvent::Solana(event) => event.serialized_transaction.clone(),
             SignBidirectionalEvent::Hydration(event) => event.serialized_transaction.clone(),
+            SignBidirectionalEvent::Canton(event) => match &event.tx_params {
+                crate::indexer_canton::contracts::TxParams::EvmTxParams(evm_params) => {
+                    crate::indexer_canton::rlp_encode_unsigned_eip1559(evm_params)
```
volovyks: Why is this serialization done here? Why not follow the existing pattern and use serialized_transaction?

integration-tests/src/cluster/spawner.rs
```diff
@@ -123,6 +123,8 @@ pub struct ClusterSpawner {
     pub redis: Option<containers::Redis>,
     pub worker: Option<Worker<Sandbox>>,
     pub solana: Option<containers::Solana>,
+    pub canton: Option<crate::canton::CantonSandbox>,
+    pub use_canton: bool,
```
volovyks: Why do we need this bool? How is Canton different from other chains? The option should be enough to say whether we want to use it or not.

integration-tests/fixtures/canton/README.md
```diff
@@ -0,0 +1,16 @@
+# Canton DAR Fixture
```
volovyks, line 1: This file seems off, or I do not understand it at all
Appreciate a simple explanation on what a .dar file is and why we need it.

chain-signatures/node/src/sign_bidirectional.rs
```diff
@@ -62,10 +75,16 @@ pub struct BidirectionalTx {
...
     pub from_address: Address,
     pub nonce: u64,
+    #[serde(default)]
+    pub chain_ctx: ChainContext,
 }
 
 impl BidirectionalTx {
     pub(crate) fn sender_string(&self) -> anyhow::Result<String> {
+        // Canton's [u8; 32] sender is an irreversible hash; read the original party ID from the Canton chain context.
+        if let ChainContext::Canton { sender, .. } = &self.chain_ctx {
```
volovyks: Same here, Canton should follow the same pattern and include source_chain

chain-signatures/node/src/respond_bidirectional.rs
```diff
@@ -24,6 +22,14 @@ pub enum SerDeserFormat {
     Abi,
 }
 
+fn respond_serialization_format(chain: Chain) -> SerDeserFormat {
```
volovyks: If this is something chain-specific, it can be added directly to the Chain struct.

integration-tests/tests/cases/canton_stream.rs
```diff
@@ -0,0 +1,545 @@
...
+            let Some(event) = stream.next_event().await else {
+                break None;
+            };
+            match event {
+                ChainEvent::SignRequest(req) => {
+                    saw_sign_request = true;
+                    backlog.insert(req).await;
+                }
+                ChainEvent::Block(height) => {
+                    if let Some(cp) = backlog
```
volovyks: nit: `cp` is not super readable. I've spent some time trying to understand what is going on here :)

integration-tests/tests/cases/canton_stream.rs
```diff
@@ -0,0 +1,545 @@
...
+        !bidir.serialized_transaction().is_empty(),
+        "RLP-encoded tx should not be empty"
+    );
+    Ok(())
+}
+
+#[ignore] // requires dpm
+#[serial]
+#[test(tokio::test)]
+async fn test_canton_rejects_unauthenticated_requests() -> Result<()> {
```
volovyks: Testing sandbox is not our responsibility, that is not a part of our project

integration-tests/tests/cases/canton_stream.rs
```diff
@@ -0,0 +1,545 @@
...
+        .status();
+    assert_eq!(status, 401, "invalid JWT should be rejected, got {status}");
+
+    Ok(())
+}
+
+#[ignore] // requires dpm + openssl
+#[serial]
+#[test(tokio::test)]
+async fn test_canton_rejects_jwt_signed_by_unconfigured_key() -> Result<()> {
```
volovyks: Same here. But maybe I'm losing something here.

volovyks (commented):
Great work! Nice test coverage!
(have not reviewed all the files, expect another part of the review from me)

.github/workflows/canton.yml
ChaoticTempest: if we only want canton be ran for this particular CI, then we should also ignore it in the integration.yml CI as well

chain-signatures/node/src/indexer_canton/conn.rs
```diff
@@ -0,0 +1,139 @@
...
+    if !resp.status().is_success() {
+        let status = resp.status();
+        let text = resp.text().await.unwrap_or_default();
+        anyhow::bail!("{context} failed: {status} {text}");
+    }
+    Ok(resp)
+}
+
+#[allow(async_fn_in_trait)]
+pub trait CantonConn {
```
ChaoticTempest: I see you did this trait to separate out the test client, but all the test client is doing is just point to different jwt related entries. So we can just use the `CantonClient` and set the same parameters in the `CantonConfig` that is passed in. So we don't need a separate `CantonTestClient` at all, and thus can also get rid of this `CantonConn` trait and implement these functions directly on `CantonClient`

chain-signatures/node/src/indexer_canton/mod.rs
```diff
@@ -0,0 +1,284 @@
...
+        // requires infallible request ID generation. We log and return zeros on
+        // failure — generate_sign_request will fail with a proper error anyway.
+        match compute_request_id(self) {
+            Ok(id) => id,
+            Err(e) => {
+                tracing::error!(%e, "failed to compute canton request_id");
+                [0u8; 32]
+            }
+        }
+    }
```
ChaoticTempest, line 109: This is an issue when we go to compute the request id, because the errors here means we get a pretty big mismatch in IDs. Which will be hard to debug.

Going further up the call chain, I see that this trying to parse the CantonSignBidirectionalRequestedEvent for the id, and it doesn't work, we error out here. Instead, on creation of the CantonSignBidirectionalRequestedEvent, we should already be parsing on those fields, so we can trust the inputs are valid when we actually go and utilize them. I see it's created in parse_sign_bidirectional_event so it's best to do all the parsing logic there. The CantonSignBidirectionalRequestedEvent doesn't need to change since this is the type we're looking at in indexer, but node side, we should probably create a new type where all these fields are already parsed into more refined types.

chain-signatures/node/src/backlog/mod.rs
ChaoticTempest: how are you creating the BidirectionalTx? Can't you just manually insert something into the sender field? That way we won't need chain_ctx

ChaoticTempest (commented):
did a cursory look, but mostly looking good! Just one parsing issue I noticed we can correct that I mentioned below

integration-tests/tests/cases/canton.rs
```diff
@@ -0,0 +1,188 @@
...
+        .create_contract(
+            &[&canton.operator_party, &canton.requester_party],
+            "#daml-signer:Signer:SignRequest",
+            serde_json::to_value(test_sign_request_payload(&expected_event))?,
+        )
+        .await?;
+    let (sign_request_cid, _) =
+        integration_tests::canton::find_created_contract(&sign_request, "SignRequest")?;
+
+    client
```
volovyks, line 128: Can you please add a comment on this and the previous step, explaining what is happening? Maybe with a link to an external doc.

integration-tests/tests/cases/canton.rs
```diff
@@ -0,0 +1,188 @@
...
+    let response_hash =
+        mpc_node::respond_bidirectional::calculate_respond_bidirectional_hash_message(
+            &hex::decode(&respond_payload.request_id)?,
+            &hex::decode(&respond_payload.serialized_output)?,
+        );
+
+    let respond_epsilon = mpc_crypto::derive_epsilon_canton(
+        LATEST_MPC_KEY_VERSION,
+        &expected_event.sender,
+        "canton response key",
```
volovyks: I think we have it defined in a const

integration-tests/src/canton.rs
```diff
@@ -0,0 +1,602 @@
...
+        key_version: LATEST_MPC_KEY_VERSION,
+        path: sandbox.requester_party.clone(),
+        algo: "ECDSA".to_string(),
+        dest: "ethereum".to_string(),
```
volovyks: Let's keep it empty to avoid confusion. Not used atm.

chain-signatures/node/src/indexer_canton/calldata.rs
```diff
@@ -0,0 +1,32 @@
+// TODO(test): golden-test build_calldata against viem's encodeFunctionData for:
```
volovyks: Are you planning to do this in this PR? Let's move it into an issue if not.

chain-signatures/node/src/indexer_canton/mod.rs
```diff
@@ -0,0 +1,284 @@
...
+/// Canton JSON Ledger API configuration.
+///
+/// # Contract migration
+///
+/// When the Signer DAR is upgraded and redeployed, both `signer_contract_id`
+/// and `signer_template_id` change (similar to deploying a new Ethereum
+/// contract — the old address/ID is gone). Currently the MPC node requires a
+/// restart with updated CLI args to pick up the new IDs.
+///
+/// TODO: decide how to handle DAR upgrades without MPC node downtime.
```
volovyks: I would move this and all other TODOs and challenges into an issue so we can prioritize and tackle them one by one.

integration-tests/fixtures/canton/README.md
volovyks: Now I understand, `canton-mpc-poc` is a repo. But it has a different name now.

chain-signatures/node/src/indexer_canton/stream.rs
Pessina: Agree, I will rename here.

chain-signatures/node/src/indexer_canton/stream.rs
Pessina: Agree, I will rename

chain-signatures/node/src/indexer_canton/stream.rs
Pessina: Adding.

chain-signatures/node/src/indexer_canton/stream.rs
Pessina: True, I’ll add an explicit ws close on our side too.

chain-signatures/node/src/rpc.rs
Pessina: When we create the client we add the timeout: https://github.com/sig-net/mpc/blob/44187b8615cd20bfa95eeba6d238177304b4c135/chain-signatures/node/src/rpc.rs#L934

And it's wrapped on the same retry logic as other chains: https://github.com/sig-net/mpc/blob/91e4f14aae8ff3cac58e11e999d612c6bbb71cdb/chain-signatures/node/src/rpc.rs#L1053

So should be good, right?

chain-signatures/node/src/indexer_canton/stream.rs
Pessina: I've added a more detailed explanation on the comment bellow

chain-signatures/node/src/indexer_canton/stream.rs
Pessina: I target to be like Solana. As far as I understood:

- Solana: if `events_tx.send(ChainEvent::Block(...))` fails, it logs the error and continues the loop: https://github.com/sig-net/mpc/blob/44187b8615cd20bfa95eeba6d238177304b4c135/chain-signatures/node/src/indexer_sol.rs#L720
- Canton: currently log and  retry the ws connection.

But I may be missing some MPC context here. From what I checked, this looks more like a downstream issue than a Canton-specific retry issue.

If `tx.send` fails, it means the receiver side of the channel was dropped. I didn't find a path that recreates this receiver while the node is still running, so I think the Solana/Canton loops would keep consuming resources, but no one would process those events.

The backlog should still point to the last block that was successfully received and checkpointed by the downstream stream consumer. So after a node restart, I think we recreate the channel and resume from that previous checkpoint.

Is it correct?

integration-tests/src/canton.rs
Pessina: Rebased and fixed :)

chain-signatures/primitives/src/lib.rs
Pessina: Yeah, safe to increase a bit, docs say finality it's 3-10s, but will put 15s for a buffer: https://docs.digitalasset.com/integrate/devnet/preparing-and-signing-transactions/index.html#submitting-the-transaction

.cargo/audit.toml
Pessina: Sounds good, I added coz the CI was failing.

chain-signatures/crypto/src/lib.rs
Pessina: I updated following the other chains pattern.

.github/workflows/canton.yml
Pessina: I did a separated file coz Canton have a lot of special dependencies like the Java, dpm sandbox, etc. So it's easier to organize. 

Regarding when to run, I thought here and I think we should run it on all PRs, to make sure that our changes on the underlying dependencies don't break Canton. what do you think?

chain-signatures/primitives/src/lib.rs
Pessina: Agree! I didn't updated before to keep the changes minimal and Canton focused. 

But will re-use it now

chain-signatures/node/src/stream/ops.rs
Pessina: Good point, I will change it.

.github/workflows/canton.yml
ChaoticTempest: Ahh I see it's already being ignored, and ran here.

Do we even need to schedule it to be ran daily? It's already going to be ran per PR anyways right?

chain-signatures/node/src/backlog/mod.rs
Pessina: My bad, I've added it back.

.github/workflows/canton.yml
Pessina: Ooh yeah, forgot to align on that also. Now it should runs on the same pattern as the integration.yml

chain-signatures/node/src/stream/ops.rs
Pessina: This function only returns the sender as a string, we don't need to add chain_id here, it will be added later at: https://github.com/sig-net/mpc/blob/26e40aa5021cc792566c00ba8c58eb045d4dd2e6/chain-signatures/node/src/sign_bidirectional.rs#L93

chain-signatures/node/src/stream/ops.rs
Pessina: Not only, but also coz Canton has the visibility system that we talked.

When we publish back the signatures, they need to be visible to the `operators` (DEX owners) and the `requester` (user depositing). This is not necessary on other chains coz everything is public, but on Canton we need to pass those parties again so the response event can add them as observers.

Also, in this flow `sender` only carries the `operators` hash, not the `requester`. That's why I'm struggling to remove the chain_ctx from the shared struct on the comment above, I'm thinking here how to to do it.

chain-signatures/node/src/indexer_canton/mod.rs
Pessina: Good point, I've moved the validation logic here: https://github.com/sig-net/mpc/blob/f619ddfdae8b0f17943ef3c0294564bb17ca360b/chain-signatures/node/src/indexer_canton/mod.rs#L47

So now request ID computation and tx serialization fail in one place before the event enters the shared MPC flow. How does that look?

chain-signatures/node/src/stream/ops.rs
Pessina: Update the code following @ChaoticTempest recommendation here: https://github.com/sig-net/mpc/pull/740#discussion_r3141722167

So now all the special cases are handled in one place

chain-signatures/node/src/stream/ops.rs
Pessina: Update the code following @ChaoticTempest recommendation here: https://github.com/sig-net/mpc/pull/740#discussion_r3141722167

So now all the special cases are handled in one place

chain-signatures/node/src/stream/ops.rs
Pessina: Update the code following @ChaoticTempest recommendation here: https://github.com/sig-net/mpc/pull/740#discussion_r3141722167

So now all the special cases are handled in one place

integration-tests/src/cluster/spawner.rs
Pessina: Updated to follow the Solana pattern

chain-signatures/node/src/stream/ops.rs
Pessina: Makes sense, moved the comment to the integration layer.

chain-signatures/node/src/sign_bidirectional.rs
Pessina: Update to follow the other chains pattern.

integration-tests/tests/cases/canton.rs
Pessina: True, gonna re-use it

chain-signatures/node/src/respond_bidirectional.rs
Pessina: Makes sense, moved it :)

integration-tests/tests/cases/canton_stream.rs
Pessina: Renamed

integration-tests/src/canton.rs
Pessina: Good point, default to empty now.

integration-tests/tests/cases/canton_stream.rs
Pessina: I think it's important to keep them as Canton has a different auth model via JWT on the Canton Node (sandbox). 

This tests make sure the sandbox auth it's properly configured. But I think it makes sense to move them to where we setup the sandbox. 

Same for comment bellow.

integration-tests/tests/cases/canton_stream.rs
Pessina: https://github.com/sig-net/mpc/pull/740#discussion_r3141813792

chain-signatures/node/src/indexer_canton/calldata.rs
Pessina: Created issues and added to the PR description

chain-signatures/node/src/indexer_canton/mod.rs
Pessina: Created issues and added to the PR description

integration-tests/tests/cases/canton.rs
Pessina: Cool, added a comment with an evm parallel explanation. Canton it's a bit confusing

chain-signatures/node/src/indexer_canton/conn.rs
Pessina: I partially agree, the `CantonConn` was the wrong abstraction, so I removed it and moved the shared stuff to `CantonClient`. But I think we should not leak test only code into our `CantonClient`.

So now the `CantonTestClient` it's a wrapper around the `CantonClient`. What do you think?

integration-tests/fixtures/canton/README.md
Pessina: Yeah my bad, updated it. Now it's correct pointing to the repo but the relevant PR there it's not merged, it's incoming.

chain-signatures/node/src/backlog/mod.rs
Pessina: That's a tricky one. The reason is that Canton is not a public blockchain where publish events back and let users find them by `requestId`. In Canton, we need to specify who can observe the response event. In our case, that means `operators + requester`, so we need to carry extra data that other chains don't need.

We can't put all `sender`, because `sender` is `hash(operators[])`. Even if we changed it to `sender = operators[]`, it still not enough coz we also need the `requester`.

I saw 3 options:

- Keep `chain_ctx`.
- Before posting the response, fetch the original `SignBidirectionalEvent` by `requestId` and look up the extra data from it. But that gets expensive when we have thousands of txs.
- Do what I did here: https://github.com/sig-net/mpc/pull/740/commits/3bd5076d69f0703bc2609ab379c331c1cd98fa2b. Instead of carrying all `operators + requester + any other data`, we only carry the `source_event_id` in the `RespondBidirectionalTx` and post it back to Canton. Canton can then resolve `operators` + `requester` on-chain.

`source_event_id` can be a tx hash, contract id... For Canton, it is the `SignBidirectionalEvent` contract id. This also gives us a generic lookup id for future chains that may need to fetch the source event that triggered the flow.

What do you think?

chain-signatures/node/src/indexer_canton/conn.rs
Pessina: https://github.com/sig-net/mpc/commit/f0351716b45ee502b8b3663c830f20e9d085f0e4

chain-signatures/node/src/indexer_canton/contracts.rs
Pessina: https://github.com/sig-net/mpc/commit/f0351716b45ee502b8b3663c830f20e9d085f0e4

chain-signatures/node/src/indexer_canton/ledger_api.rs
Pessina: https://github.com/sig-net/mpc/commit/f0351716b45ee502b8b3663c830f20e9d085f0e4

chain-signatures/node/src/indexer_canton/stream.rs
Pessina: https://github.com/sig-net/mpc/commit/a50e8ca1c5c63f08cbbdb40c5b7324f4f2978296

chain-signatures/node/src/indexer_canton/stream.rs
Pessina: https://github.com/sig-net/mpc/commit/a50e8ca1c5c63f08cbbdb40c5b7324f4f2978296

chain-signatures/node/src/indexer_canton/stream.rs
Pessina: https://github.com/sig-net/mpc/commit/a50e8ca1c5c63f08cbbdb40c5b7324f4f2978296

chain-signatures/node/src/indexer_canton/stream.rs
Pessina: https://github.com/sig-net/mpc/commit/c4fcceb94f21155cfef2b4172f2be3173b5382bc

chain-signatures/primitives/src/lib.rs
Pessina: https://github.com/sig-net/mpc/commit/26e40aa5021cc792566c00ba8c58eb045d4dd2e6

integration-tests/src/canton.rs
Pessina: https://github.com/sig-net/mpc/commit/69451ec20b252381ae1227aee47f063913bd94ee

.github/workflows/canton.yml
Pessina: https://github.com/sig-net/mpc/commit/26e40aa5021cc792566c00ba8c58eb045d4dd2e6

.cargo/audit.toml
Pessina: https://github.com/sig-net/mpc/commit/26e40aa5021cc792566c00ba8c58eb045d4dd2e6

chain-signatures/node/src/backlog/mod.rs
Pessina: https://github.com/sig-net/mpc/commit/26e40aa5021cc792566c00ba8c58eb045d4dd2e6

chain-signatures/primitives/src/lib.rs
Pessina: https://github.com/sig-net/mpc/commit/26e40aa5021cc792566c00ba8c58eb045d4dd2e6

chain-signatures/node/src/stream/ops.rs
Pessina: https://github.com/sig-net/mpc/commit/26e40aa5021cc792566c00ba8c58eb045d4dd2e6

chain-signatures/node/src/stream/ops.rs
Pessina: See: https://github.com/sig-net/mpc/pull/740#discussion_r3143413246

chain-signatures/node/src/stream/ops.rs
Pessina: https://github.com/sig-net/mpc/commit/ac0460741dc25b9c42982af22e7b15c744ea300b

chain-signatures/node/src/indexer_canton/mod.rs
Pessina: https://github.com/sig-net/mpc/commit/ac0460741dc25b9c42982af22e7b15c744ea300b

integration-tests/src/cluster/spawner.rs
Pessina: https://github.com/sig-net/mpc/commit/ac0460741dc25b9c42982af22e7b15c744ea300b

integration-tests/fixtures/canton/README.md
Pessina: https://github.com/sig-net/mpc/commit/90fb3639f5f2241622eb8312275270bb94c2f94e

chain-signatures/node/src/respond_bidirectional.rs
Pessina: https://github.com/sig-net/mpc/commit/28c3c9b089faeef542b4fb5663ba1377e52f4c9d

integration-tests/tests/cases/canton_stream.rs
Pessina: https://github.com/sig-net/mpc/commit/28c3c9b089faeef542b4fb5663ba1377e52f4c9d

integration-tests/tests/cases/canton_stream.rs
Pessina: https://github.com/sig-net/mpc/commit/93c8e00fb8f892a01aedf6f78c39011d7195a2ef

integration-tests/tests/cases/canton_stream.rs
Pessina: https://github.com/sig-net/mpc/commit/93c8e00fb8f892a01aedf6f78c39011d7195a2ef

chain-signatures/node/src/indexer_canton/conn.rs
Pessina: https://github.com/sig-net/mpc/commit/82df04b3202c7052aba899f464fc3ff7cf8cb5ec
https://github.com/sig-net/mpc/commit/7f1d255c8bbbcf3a71ba6eaf9cc6821e68cde3fc

integration-tests/tests/cases/canton.rs
Pessina: https://github.com/sig-net/mpc/commit/eb553f839e1e7410fbdb4d234d03a85caa3f54ce

integration-tests/tests/cases/canton.rs
Pessina: https://github.com/sig-net/mpc/commit/28c3c9b089faeef542b4fb5663ba1377e52f4c9d

integration-tests/src/canton.rs
Pessina: https://github.com/sig-net/mpc/commit/28c3c9b089faeef542b4fb5663ba1377e52f4c9d

chain-signatures/node/src/indexer_canton/calldata.rs
Pessina: https://github.com/sig-net/mpc/commit/593621e80fbd7e7f9d9809edaed67e61b4afe2a4

chain-signatures/node/src/indexer_canton/mod.rs
Pessina: https://github.com/sig-net/mpc/commit/593621e80fbd7e7f9d9809edaed67e61b4afe2a4

chain-signatures/node/src/rpc.rs
ppca: oh yes!

chain-signatures/node/src/indexer_canton/stream.rs
ppca: yeah sounds correct

chain-signatures/node/src/indexer_canton/conn.rs
ChaoticTempest: yep that works

chain-signatures/node/src/stream/ops.rs
volovyks: I see, let's continue the discussion in the mentioned thread

chain-signatures/node/src/backlog/mod.rs
volovyks: Ok, I see the motivation now.
With a new design, we are replacing strongly typed `chain_ctx` with a string (serialized struct?), that can contain any information this specific chain needs. I would not call it `source_event_id`, maybe something more generic? `chainCtx` works for me. Overall, I'm ok with this approach, it allows us to separate backlog and protocol layers from integrations.
cc @ChaoticTempest WDYT?

Pessina: No, the `source_event_id` it's more like the txHash that created the event.

On Canton, each contract (what we use as Event to trigger the sign flow) has an ID and this `source_event_id` stores the contractId and pass it back via args when posting both `Respond` and `RespondBidirectional`, so we can resolve the extra data `operators + requester` on the Contract. 

But what you said may also be a good idea, so it's more generic in case we need to pass generic data in the future.

volovyks: Yes, I mostly talk about abstractions. passing generic serialized struct (as a string) to Backlog, then decentralizing it on the way back. This struct can contain operators, ids, whatever this specific chain needs. No need to additionally fetch it.

Pessina: Sounds good to me, I will rename to `chainCtx` and each chain can use however it needs it.

Pessina: https://github.com/sig-net/mpc/pull/740/commits/2e975f795d1ca72cf7a25bcf66fc4148703c51e8

volovyks (approved):

Cargo.toml
```diff
@@ -40,6 +41,7 @@ serde_json = "1"
 sha3 = "0.10.8"
 thiserror = "1"
 tokio = { version = "1.45.1", features = ["full"] }
+tokio-tungstenite = { version = "0.29", features = ["native-tls"] }
```
ChaoticTempest, line 44: is there any reason to use `native-tls` over `rustls`? 

not that important for now, but in the future we should switch all of our stack to use rustls to have less dependence on openssl. Openssl keeps having security vulnerabilities and the dependency tree for it is quite bad for our build times.

chain-signatures/node/src/backlog/mod.rs
ChaoticTempest: This looks fine to me now. I guess the main issue we have is that we can have separate events per chain for SignBidirectional type but when it comes to RespondBidirectional, it's a more general type. I don't know if we should eventually turn it into a more refined type per chain but generalizing over a vector of bytes works for now

ChaoticTempest (approved):

Cargo.toml
Pessina: Cool, will move to issue and we can tackle later.

Pessina: Tracked in #782.

PR #743, opened 2026-04-08
feat: impl generic catchup and livestream

ChaoticTempest:
This implements catchup then livestream in a generic way for all chains but realistic only implemented for ethereum. This is outlined in https://github.com/sig-net/mpc/pull/663 in the generalized catchup section.

The issue with the previous ethereum catchup and live/final block polling was that all three of them were being done concurrently, so order is not guaranteed. We would end up with potentially out of order events here. So in this PR, ethereum processes block linearly.

Solana will be implemented if we like what we see here.

The basic gist of the generalized catchup & livestream is the following:
```
ethereum_indexer = ethereum_stream.start();

task::spawn({
   let buffered_stream = ethereum_indexer.livestream();
   let anchor_block = buffered_stream.initial();
   let anchor_height = ethereum_indexer.height(anchor_block);
   let catchup_range = ethereum_indexer.catchup_range(anchor_height);
   for i in catchup_range {
       ethereum_indexer.process_height(i);
   }
   while Some(block) = buffered_stream.next() {
      ethereum_indexer.process_block(block);
   }
})
```


chain-signatures/node/src/indexer_eth/mod.rs
```diff
@@ -722,213 +714,145 @@ impl EthereumClient {
...
-            .await
-            {
-                tracing::warn!(
-                    "Eth indexer failed to process block number {block_number}: {err:?}"
-                );
-                Self::add_failed_block(blocks_failed_send.clone(), block).await;
+
+            let mut block_number = next_block_number.unwrap_or(latest_block_number);
+            if block_number > latest_block_number {
+                tokio::time::sleep(Duration::from_millis(500)).await;
```
jakmeier:
```suggestion

            if next_block_number.is_some_and(|block_number| block_number > latest_block_number )  {
                tokio::time::sleep(Duration::from_millis(500)).await;
```

chain-signatures/node/src/indexer_eth/mod.rs
```diff
@@ -722,213 +714,145 @@ impl EthereumClient {
...
+                    break;
+                };
+
+                if let Err(err) = live_blocks.send(block).await {
+                    tracing::warn!(?err, block_number, "failed to buffer ethereum live block");
+                    return;
                 }
+
+                next_block_number = Some(block_number.saturating_add(1));
+                block_number = block_number.saturating_add(1);
```
jakmeier:
```suggestion
                block_number = block_number.saturating_add(1);
                next_block_number = Some(block_number);
```

chain-signatures/node/src/indexer_eth/mod.rs
```diff
@@ -722,213 +714,145 @@ impl EthereumClient {
...
             }
 
-            crate::metrics::indexers::LATEST_BLOCK_NUMBER
-                .with_label_values(&[Chain::Ethereum.as_str(), "indexed"])
-                .set(block_number as i64);
+            tokio::time::sleep(Duration::from_millis(500)).await;
```
jakmeier: why sleep here?

chain-signatures/node/src/stream/mod.rs
```diff
@@ -91,13 +91,145 @@ pub enum ExecutionOutcome {
...
+            }
+        }
+    }
+
+    let _ = catchup_completed_tx.send(());
+
+    let mut next_block = anchor_block;
+    loop {
+        if let Err(err) = indexer.process_buffered_block(next_block).await {
+            tracing::warn!(?err, %chain, "buffered block processing failed");
```
jakmeier: In this case we miss a block, don't we? Seems like this should be an error, not a warning. Or ideally, we shouldn't move on at all.

chain-signatures/node/src/stream/mod.rs
```diff
@@ -91,13 +91,145 @@ pub enum ExecutionOutcome {
...
+        None
+    }
+}
+
+#[async_trait]
+pub trait ChainBufferedStream: Send + 'static {
+    type Block: Send + 'static;
+
+    async fn initial(&mut self) -> Option<Self::Block>;
+    async fn next(&mut self) -> Option<Self::Block>;
```
jakmeier: From looking the rest of the PR, it seems to me these two are functionally identical, just different names. Wouldn't it be less confusing to have only one method?

chain-signatures/node/src/stream/mod.rs
```diff
@@ -91,13 +91,145 @@ pub enum ExecutionOutcome {
...
+        Ok(buffered) => buffered,
+        Err(err) => {
+            tracing::error!(?err, %chain, "failed to initialize livestream");
+            return;
+        }
+    };
+    let Some(mut buffered) = buffered else {
+        let _ = catchup_completed_tx.send(());
+        return;
+    };
```
jakmeier: I don't understand. If the stream doesn't implement buffering, we immediately claim catchup to be completed without doing anything?

chain-signatures/node/src/stream/mod.rs
```diff
@@ -91,13 +91,145 @@ pub enum ExecutionOutcome {
...
+    for height in catchup_range {
+        loop {
+            match indexer.process_catchup_on_height(height).await {
+                Ok(()) => break,
+                Err(err) => {
+                    tracing::warn!(?err, %chain, height, "catchup height processing failed; retrying");
+                    tokio::time::sleep(indexer.retry_delay()).await;
+                }
+            }
+        }
```
jakmeier:
```suggestion
        while let Err(err) = indexer.process_catchup_on_height(height).await {
                tracing::warn!(?err, %chain, height, "catchup height processing failed; retrying");
                tokio::time::sleep(indexer.retry_delay()).await;
            }
        }
```

chain-signatures/node/src/indexer_eth/mod.rs
jakmeier: Separate variables for `block_number`, `next_block_number`, `latest_block_number` makes this code rather hard to read.

Shouldn't it be enough to have one variable for where we are now (e.g. `buffer_tip`) and compare that to `client.get_latest_block_number()` directly?

jakmeier (dismissed):
Okay I just spent 2h+ looking at this and I didn't even get to the tests, yet. Your PR description is quite clear but the code somehow doesn't click for me.

Handling old blocks as catchup, while concurrently fetching blocks for new incoming events and buffering them, just seems a bit overcomplicated. And if I got that right, you start both before even defining the boundary between them (called anchor height in the code).
Wouldn't it be easier to just finish catchup first? I would only subscribe to new blocks after that. I.e. do catchup until processed_block_height == latest_block_height. Then subscribe to new blocks from latest_block_height+1.

Frankly, I think this might also need some refactoring. (Or maybe I just need a brain reset before taking another look.) Please also go through the Copilot comments.

Please re-request a review when I should take another look.

jakmeier:
> [jakmeier](https://github.com/jakmeier) approved these changes[ 1 minute ago](https://github.com/sig-net/mpc/pull/743#pullrequestreview-4080311950)

oops, sorry, I didn't mean to approve, just wanted to comment :shrug:

ChaoticTempest:
> Handling old blocks as catchup, while concurrently fetching blocks for new incoming events and buffering them, just seems a bit overcomplicated. And if I got that right, you start both before even defining the boundary between them (called anchor height in the code). Wouldn't it be easier to just finish catchup first? I would only subscribe to new blocks after that. I.e. do catchup until processed_block_height == latest_block_height. Then subscribe to new blocks from latest_block_height+1.

There are a couple issues this abstraction tries to solve:
- whether we can have a live feed or we have to fetch blocks one by one (current eth client gets blocks one by one, solana has live feed)
- if live feed, then we need to buffer because we can't stop once we start the solana websocket subscription.
- if get block one by one via height, then we don't need to buffer. I can simplify this part for ethereum where currently in this PR, we get the latest block and use it as anchor height but don't need to query the next block yet

chain-signatures/node/src/indexer_eth/mod.rs
ChaoticTempest: the diff blocks are a bit everywhere. There was a `interval.tick()` before, but I'll just go back to the interval since it's clearer

chain-signatures/node/src/stream/mod.rs
ChaoticTempest: yeah, this is a case I forgot to point out. Should we stall on not being able to process a block? In this case, our logic is incorrect in parsing or maybe something happened? Then would retrying be ideal here?

Actually thinking about it again, it's best to stall/retry so we can push a fix to unstuck an indexer if it ever becomes a problem. That way we know which block it gets stuck at

chain-signatures/node/src/stream/mod.rs
ChaoticTempest: yeah, you're right. Initially just had it to be clearer with naming but it could just be a variable name

chain-signatures/node/src/stream/mod.rs
ChaoticTempest: yep, I probably should put a comment here. It's because I haven't implemented this for solana, and solana doesn't do catchup nor the livestream logic here since it does it's own livestream of blocks in it's indexer right now. So this is a way of disabling the `catchup_and_livestream` feature for solana

chain-signatures/node/src/stream/mod.rs
```diff
@@ -161,36 +283,67 @@ pub async fn run_stream<S: ChainStream>(
                     .await;
                     recovered.pending.clear();
                 }
```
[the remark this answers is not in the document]
ChaoticTempest: if the sender is dropped, then we cannot progress the stream either ways because it is in shutdown mode. The sender is not dropped unless that is the case

chain-signatures/node/src/indexer_eth/mod.rs
```diff
@@ -722,213 +714,145 @@ impl EthereumClient {
...
-                        catchup_completed_emitted = true;
-                    }
+            while block_number <= latest_block_number {
+                let Some(block) = client
+                    .get_block(alloy::rpc::types::BlockId::Number(
+                        BlockNumberOrTag::Number(block_number),
+                    ))
+                    .await
+                else {
+                    tracing::warn!(block_number, "ethereum live block not yet available");
```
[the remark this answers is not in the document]
ChaoticTempest: reduced to just current_block_number and latest_block_number

ChaoticTempest:
Cleaned it up and addressed the comments.

We roughly have a simpler flow of just `ChainIndexer` & `ChainStream` traits now. So the high level of this now looks like:
```
indexer = stream.start();

task::spawn({
   anchor_height = indexer.livestream();
   for height in indexer.catchup_range(anchor_height) {
       indexer.process_height(height);
   }
   indexer.notify_catchup_completed();

   while Some(block) = indexer.next() {
      indexer.process_block(block);
   }
})
```

ChaoticTempest:
Additionally, the process_{event} functions for the general processing of events now takes a caught up parameter. This will make it so that we can have the same process functions for processing events for both catchup & live blocks. If we're not caught up yet, the process_{event} functions will simply not send `Sign::{Request, Completion}` related requests to the signer

chain-signatures/node/src/stream/ops.rs
```diff
@@ -274,8 +278,10 @@ pub(crate) async fn process_sign_request(
     backlog.insert(sign_request.clone()).await;
 
     let chain = sign_request.chain;
-    if let Err(err) = sign_tx.send(Sign::Request(sign_request)).await {
-        tracing::error!(?err, %chain, "failed to send sign request into queue");
+    if caught_up {
+        if let Err(err) = sign_tx.send(Sign::Request(sign_request)).await {
+            tracing::error!(?err, %chain, "failed to send sign request into queue");
+        }
     }
```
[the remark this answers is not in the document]
ChaoticTempest: not true, in CatchupCompleted, there is a flush

chain-signatures/node/src/indexer_eth/mod.rs
```diff
@@ -712,222 +674,146 @@ impl EthereumClient {
...
+    pub async fn new(
+        eth: EthConfig,
+        backlog: Backlog,
+        events_tx: mpsc::Sender<ChainEvent>,
+    ) -> anyhow::Result<Self> {
+        let client = Arc::new(EthereumClient::new(eth.clone()).await?);
+        let contract_address = format!("0x{}", eth.contract_address);
+        let contract_address = Address::from_str(&contract_address).map_err(|_| {
+            anyhow::anyhow!("failed to parse ethereum contract address: {contract_address}")
+        })?;
```
jakmeier:
```suggestion
        let contract_address = Address::from_str(&contract_address).with_context(|| {
            format!("failed to parse ethereum contract address: {contract_address}")
        })?;
```

nit: Prefer using `with_context` over `map_err` +`anyhow::anyhow!`. Using context will keep the source error. So when the anyhow error gets logged somewhere, it will include what parser error was encountered.

chain-signatures/node/src/indexer_eth/mod.rs
```diff
@@ -712,222 +674,146 @@ impl EthereumClient {
...
-            Arc::clone(&client),
-            blocks_failed_recv,
-            blocks_failed_send.clone(),
-            contract_address,
-            requests_indexed_send.clone(),
-            backlog.clone(),
-        ));
+            if current_block_number > latest_block_number {
+                continue;
+            }
```
jakmeier: Isn't this redundant with the loop that follows? (`while current_block_number <= latest_block_number { ... }`

chain-signatures/node/src/indexer_eth/mod.rs
```diff
@@ -712,222 +674,146 @@ impl EthereumClient {
...
-            client_clone,
-            requests_indexed_recv,
-            finalized_block_recv,
-            events_tx.clone(),
-            optimistic_requests,
-        ));
+        let mut current_block_number = start_block_number;
+        let mut interval = tokio::time::interval(Self::RETRY_DELAY);
+        loop {
+            interval.tick().await;
```
jakmeier, line 757: The default interval setting uses [MissedTickBehavior::Burst](https://docs.rs/tokio/latest/tokio/time/enum.MissedTickBehavior.html#variant.Burst).

So, if processing blocks takes longer than 2 times `RETRY_DELAY`, ticks are piling up and are released in a burst. We would end up with a burst of RPC calls, potentially hitting rate limits.

I think `MissedTickBehavior::Delay` would make more sense, always waiting `RETRY_DELAY` between each start of `loop` iterations.

chain-signatures/node/src/indexer_eth/mod.rs
```diff
@@ -712,222 +674,146 @@ impl EthereumClient {
...
-                    tracing::info!("Processed new block number {block_number}");
-                }
-            }
+    fn catchup_start_block_number(
+        last_processed_block: Option<u64>,
+        anchor_height: BlockNumber,
+    ) -> BlockNumber {
+        let requested_start = last_processed_block
+            .map(|height| height.saturating_add(1))
+            .unwrap_or(anchor_height);
```
jakmeier: Hm, what is the desired behavior when the node doesn't have a last processed block number?

The current logic just skips catchup and starts with live processing wherever that happens to be, right? That doesn't seem right to me. Nodes might start with different views depending on when they start tracking a chain for the first time.

For correctness, it seems a node without known last processed block should start at the genesis block height of the given chain. More practically, I suggest we define our own "signet genesis" height per chain, which is where the MPC network starts tracking requests.

Code wise, it might be best to push this responsibility up. That is, `last_processed_block: u64` instead of `last_processed_block: Option<u64>`. The caller has to figure out where we need to start. In the absence of a persisted `last_processed_block`, that would have to be the last checkpoint we can bootstrap from, or worst-case, genesis.

chain-signatures/node/src/indexer_eth/mod.rs
```diff
@@ -712,222 +674,146 @@ impl EthereumClient {
...
-                .set(block_number as i64);
+        if requested_start < oldest_supported {
+            tracing::warn!(
+                requested_start,
+                anchor_height,
+                oldest_supported,
+                "ethereum catchup start is older than supported range; clamping"
+            );
+            oldest_supported
+        } else {
```
jakmeier: Ah that seems like an easy way to miss requests on some nodes and get out-of-sync with backlogs / checkpoints.

I know this is existing behavior and hard to fix in a real way. But could we make it clear in comments and/or the logged warning that this is actually quite a big problem? Ideally also with a linked github issue that tracks what we are going to do about this.

chain-signatures/node/src/indexer_eth/mod.rs
```diff
@@ -936,13 +822,13 @@ impl EthereumIndexer {
...
-        let block_receipts = client
+        let block_receipts = self
+            .client
             .get_block_receipts(block_number.into())
             .await
             .map_err(|err| {
                 anyhow::anyhow!(
-                    "Failed to get block receipts for block number {block_number}: {:?}",
-                    err
+                    "Failed to get block receipts for block number {block_number}: {err:?}",
```
jakmeier: nit: this could also be nicer using `with_context` instead of manually formatting the error

chain-signatures/node/src/indexer_eth/mod.rs
```diff
@@ -712,222 +674,146 @@ impl EthereumClient {
...
     }
 
-    async fn process_block(
-        client: Arc<EthereumClient>,
-        block: alloy::rpc::types::Block,
-        contract_address: Address,
-        requests_indexed: mpsc::Sender<BlockAndRequests>,
-        backlog: Backlog,
-    ) -> anyhow::Result<()> {
+    async fn parse_block(&self, block: &Block) -> anyhow::Result<BlockAndRequests> {
```
jakmeier, line 815: Initial thought: If this only parses, why is it `async`?

Maybe change the name to something reflecting that it also collects request details from external sources.

chain-signatures/node/src/indexer_eth/mod.rs
```diff
@@ -1148,221 +1025,182 @@ impl EthereumIndexer {
...
-                        if let Err(err) = events_tx.send(exec_event).await {
-                            tracing::error!(?err, "failed to emit ExecutionConfirmed event");
-                        }
-                    }
-                }
+        for req in indexed_requests {
+            self.events_tx
+                .send(ChainEvent::SignRequest(req))
+                .await
+                .map_err(|err| anyhow::anyhow!("failed to emit SignRequest event: {err:?}"))?;
```
jakmeier: this one could also make use of anyhow context


```suggestion
                .context("failed to emit SignRequest event")?;
```

chain-signatures/node/src/indexer_eth/mod.rs
```diff
@@ -1148,221 +1025,182 @@ impl EthereumIndexer {
...
             .await
-            {
-                tracing::warn!("Retry failed for block {block_number}: {err:?}");
-                Self::add_failed_block(blocks_failed_tx.clone(), block).await;
-            } else {
-                tracing::info!("Successfully retried block: {block_number}");
-            }
-        }
-    }
+            .map_err(|err| anyhow::anyhow!("failed to emit block event: {err:?}"))?;
```
jakmeier: dito

```suggestion
            .context("failed to emit block event")?;
```

chain-signatures/node/src/stream/mod.rs
```diff
@@ -91,13 +94,135 @@ pub enum ExecutionOutcome {
...
+    }
+
+    async fn notify_catchup_completed(&mut self) -> anyhow::Result<()> {
+        Ok(())
+    }
+
+    async fn catchup_range(&mut self, anchor_height: u64) -> Range<u64> {
+        let _ = anchor_height;
+        // TODO: this disables the catchup range, will be removed in the future once
+        // all chains like solana support catchup & livestream.
```
jakmeier, line 114: nit: a linked github issue would be lovely

chain-signatures/node/src/stream/mod.rs
```diff
@@ -212,24 +352,243 @@ mod tests {
...
+        assert!(matches!(observed[3], ChainEvent::Block(5)));
+    }
+
+    #[tokio::test]
+    async fn test_run_linearized_source_retries_without_reordering() {
+        let mut stream = TestLinearStream::new(
+            TestLinearControl::new(Some(1), vec![4, 5])
+                .fail_catchup_once(3)
+                .fail_live_once(4),
+        );
```
jakmeier, line 566: Great test setup!

chain-signatures/node/src/stream/mod.rs
```diff
@@ -91,13 +94,135 @@ pub enum ExecutionOutcome {
...
+        if let Err(err) = indexer.notify_catchup_completed().await {
+            tracing::warn!(?err, %chain, "failed to signal catchup completion");
+        }
+        return;
+    };
+
+    let catchup_range = indexer.catchup_range(anchor_height).await;
+    for height in catchup_range {
+        while let Err(err) = indexer.process_catchup_on_height(height).await {
+            tracing::warn!(?err, %chain, height, "catchup height processing failed; retrying");
```
jakmeier, line 214: Note: When implementing the catchup / livestream for Solana, we have deal with missed slots causing block heights to be missing. This iteration over all heights looks like a potential issue but I didn't look too deep into it.

jakmeier (commented):
Thank you for putting in the extra work, this looks much cleaner now!

I left a couple of nits but also a few deeper questions for discussion. I don't expect all of them to be resolved before we merge this but I would like to understand what we are going to do about it and ideally have GitHub issues to track them.

chain-signatures/node/src/indexer_eth/mod.rs
ChaoticTempest: yeah it is, will remove

chain-signatures/node/src/indexer_eth/mod.rs
ChaoticTempest: > what is the desired behavior when the node doesn't have a last processed block number?

Yep, if there's no last process block, it'll just be the anchor_height or basically the latest block

> The current logic just skips catchup and starts with live processing wherever that happens to be, right? That doesn't seem right to me. Nodes might start with different views depending on when they start tracking a chain for the first time.
>
> For correctness, it seems a node without known last processed block should start at the genesis block height of the given chain. More practically, I suggest we define our own "signet genesis" height per chain, which is where the MPC network starts tracking requests.

yeah that's a good idea with the signet genesis, but additionally if the last processed block or checkpoint was empty, we should be able to query other nodes and select a checkpoint. Only issue is how should we define such a config value and how do we differentiate between networks: https://github.com/sig-net/mpc/issues/777

> Code wise, it might be best to push this responsibility up. That is, last_processed_block: u64 instead of last_processed_block: Option<u64>. The caller has to figure out where we need to start. In the absence of a persisted last_processed_block, that would have to be the last checkpoint we can bootstrap from, or worst-case, genesis.

done

chain-signatures/node/src/indexer_eth/mod.rs
ChaoticTempest: I converted this into `client.clamp_oldest_supported` so it's clearer

chain-signatures/node/src/stream/mod.rs
ChaoticTempest: done: https://github.com/sig-net/mpc/issues/778

chain-signatures/node/src/stream/mod.rs
ChaoticTempest: I think we can have it return something else in the future, like an iterator instead of a range, so shouldn't be too much of a problem. Also need to do ethereum batched transaction lookup for catchup too which will have a different shape for this catchup range

jakmeier (dismissed):

jakmeier (approved):

PR #744, opened 2026-04-09
fix: catchup requeueing events that are completed

ChaoticTempest:
The previous fix https://github.com/sig-net/mpc/pull/733 didn't entirely fix the catchup problem. This should now truly fix it where Respond events during catchup will remove events from the recovered backlog entries. The previous fix got a slice/view of the recovered entries but didn't remove entries when catchup completed at all

We should be able to merge https://github.com/sig-net/mpc/pull/697 after this one completes with some verification. But probably still need to bump the storage version one more time since the backlog entries on devnet are stale

### DIFF
--- a/chain-signatures/node/src/backlog/mod.rs
+++ b/chain-signatures/node/src/backlog/mod.rs
@@ -9,7 +9,7 @@ use crate::storage::checkpoint_storage::CheckpointStorage;
 
 use anyhow::Context;
 use mpc_primitives::{PendingTx, SignId};
-use std::collections::{hash_map, HashMap};
+use std::collections::{hash_map, HashMap, HashSet};
 use std::hash::{Hash, Hasher};
 use std::sync::Arc;
 use std::time::{Duration, Instant};
@@ -185,6 +185,7 @@ struct HistoricalCheckpoint {
 pub struct Backlog {
     storage: CheckpointStorage,
     requests: Arc<RwLock<HashMap<Chain, PendingRequests>>>,
+    recovered_requests: Arc<RwLock<HashMap<Chain, HashSet<SignId>>>>,
     execution_watchers: Arc<RwLock<HashMap<Chain, ExecutionWatchers>>>,
     /// Historical checkpoints kept for 30 minutes, indexed by chain
     historical_checkpoints: Arc<RwLock<HashMap<Chain, Vec<HistoricalCheckpoint>>>>,
@@ -195,12 +196,6 @@ pub enum RecoveryRequeueMode {
     #[default]
     Immediate,
     AfterCatchup,
-}
-
-#[derive(Debug, Default)]
-pub struct RecoveredChainRequests {
-    pub pending: HashMap<SignId, BacklogEntry>,
-    pub requeue_mode: RecoveryRequeueMode,
 }
 
 impl Default for Backlog {
@@ -218,6 +213,7 @@ impl Backlog {
         Self {
             storage,
             requests: Arc::new(RwLock::new(HashMap::new())),
+            recovered_requests: Arc::new(RwLock::new(HashMap::new())),
             execution_watchers: Arc::new(RwLock::new(HashMap::new())),
             historical_checkpoints: Arc::new(RwLock::new(HashMap::new())),
         }
@@ -235,6 +231,7 @@ impl Backlog {
         };
 
         self.observe_backlog_size(chain, len);
+        self.unmark_recovered_request(chain, &id).await;
         prev
     }
 
@@ -247,6 +244,7 @@ impl Backlog {
         };
 
         self.observe_backlog_size(chain, len);
+        self.unmark_recovered_request(chain, id).await;
         removed
     }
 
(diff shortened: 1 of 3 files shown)


chain-signatures/node/src/backlog/mod.rs
```diff
@@ -279,6 +277,50 @@ impl Backlog {
...
+        }
+    }
+
+    async fn mark_recovered_requests(&self, chain: Chain, sign_ids: HashSet<SignId>) {
+        let mut recovered_requests = self.recovered_requests.write().await;
+        recovered_requests.insert(chain, sign_ids);
+    }
+
+    /// Takes the set of requests that were once marked as recovered and has not been
+    /// unmarked/removed from the backlog since. This also includes only requests that
```
jakmeier:
```suggestion
    /// unmarked/removed from the backlog since. This also includes requests that
```

Not 100% clear to me what you want to say but I think the word `only` looks to be there by mistake

chain-signatures/node/src/backlog/mod.rs
```diff
@@ -279,6 +277,50 @@ impl Backlog {
...
+    /// haven't been marked for execution watching & are still awaiting signature response,
+    /// which means they're safe to be requeued for processing.
```
jakmeier: what does "safe" mean here?

chain-signatures/node/src/backlog/mod.rs
```diff
@@ -279,6 +277,50 @@ impl Backlog {
...
+        recovered.remove(id);
+        if recovered.is_empty() {
+            recovered_requests.remove(&chain);
+        }
+    }
+
+    async fn mark_recovered_requests(&self, chain: Chain, sign_ids: HashSet<SignId>) {
+        let mut recovered_requests = self.recovered_requests.write().await;
+        recovered_requests.insert(chain, sign_ids);
+    }
```
jakmeier: This replaces the HashMap entry and silently drops previous values.

I think it should either extend an existing entry, or return an error if there was already something stored for the given chain.

jakmeier (commented):
> Respond events during catchup will remove events from the recovered backlog entries

I see, this makes sense. We definitely need that functionality.

But the `recovered_requests` with mark/unmark logic feels a bit like an unnatural layer tagged on top of the existing flow.

 I wonder, wouldn't it be simpler to have an order like:

1. Read checkpoint from persistent storage and use it to recover a Backlog. (Without queuing, yet)
2. Catch up and remove requests for Respond events from the Backlog as usual
3. Once catch up has finished, requeue all requests still in the Backlog
4. Start normal work:
    - Start processing new blocks
    - Start participating in signing protocols

At least that sounds more straight forward and maintainable to me. But here is a good chance I'm missing something. As always, I'm open to change my mind.

ChaoticTempest:
> But the `recovered_requests` with mark/unmark logic feels a bit like an unnatural layer tagged on top of the existing flow.
> 
> I wonder, wouldn't it be simpler to have an order like:
> 
> 1. Read checkpoint from persistent storage and use it to recover a Backlog. (Without queuing, yet)
> 2. Catch up and remove requests for Respond events from the Backlog as usual
> 3. Once catch up has finished, requeue all requests still in the Backlog
> 4. Start normal work:
>    
>    * Start processing new blocks
>    * Start participating in signing protocols
> 
> At least that sounds more straight forward and maintainable to me. But here is a good chance I'm missing something. As always, I'm open to change my mind.

@jakmeier maybe I'm not understanding you correctly, but 1, 2, and 4 are what we do here. All this PR is doing is marking whichever requests were recovered in 1. The catchup is in parallel to the main chain event loop in `run_stream` where we're enqueueing requests so it eventually removes items that are recovered. 

For 3, `Once catch up has finished, requeue all requests still in the Backlog`. We can't exactly do this, because catchup adds SignRequests to the backlog too. That request already gets enqueued in the main chain event loop. So we need mark which requests were recovered to re-enqueue. I think for consistency eventually, we should have re-enqueueing emit to the main chain event loop to deduplicate logic but not an issue for now.

chain-signatures/node/src/backlog/mod.rs
ChaoticTempest: Maybe I should just rename this to be `set_recovered_requests`. It's only purpose is to be used once during the startup of a node for recovery of the backlog per chain, so it should never overwrite any existing entries. I will add a safeguard logging otherwise though

ChaoticTempest:
or we could make 3 work, but then that would require waiting for catchup to complete, enqueue all events to our channel (would need to buffer), then actually start the regular chain stream event loop

chain-signatures/node/src/backlog/mod.rs
ChaoticTempest: Yeah this one is a bit wordy, umm. Let me just rephrase it here before I push anything. But I'm basically trying to say that requests that were initially marked as recovered events during startup and the same ones that haven't been advanced to watching execution and still have the AwaitingResponse status, then we should get these requests and requeue them.

```rs

    /// Takes the set of requests that were marked as recovered during startup and has 
    /// not been unmarked or removed from the backlog. This includes requests that haven't 
    /// been marked for execution watching & are still awaiting signature response.

```

chain-signatures/node/src/backlog/mod.rs
ChaoticTempest: safe here means to be able to requeue it to the sign tasks, but it would be confusing to say it in rust context so gonna get rid of it

jakmeier:
> The catchup is in parallel to the main chain event loop in run_stream where we're enqueueing requests so it eventually removes items that are recovered.

Hm yes. My suggestion is to not have these two things in parallel. Finish catch up first, only then start normal operation.

> For 3, `Once catch up has finished, requeue all requests still in the Backlog.` We can't exactly do this, because catchup adds SignRequests to the backlog too. That request already gets enqueued in the main chain event loop. So we need mark which requests were recovered to re-enqueue. I think for consistency eventually, we should have re-enqueueing emit to the main chain event loop to deduplicate logic but not an issue for now.

I think it would be better not to enqueue anything during catchup. Chances are we find responses later that cancels out requests.

> or we could make 3 work, but then that would require waiting for catchup to complete, enqueue all events to our channel (would need to buffer), then actually start the regular chain stream event loop

Yeah that's exactly what I had in mind.

chain-signatures/node/src/backlog/mod.rs
jakmeier: That already reads better!

But I honestly still find it easier reading the code than the comment :laughing: 
The terms "marked", "unmarked" and "recovered" all require context to make sense. So I end up reading the code to decipher the comment. :sweat_smile:

Maybe it's just too much details for the comment? Something shorter and focused on intention rather than details would be more useful as a comment to me personally. Maybe in the direction of  "removes recovered requests for a chain and returns a list of them filtered to only those that should be enqueued for processing"

chain-signatures/node/src/backlog/mod.rs
jakmeier: Yeah renaming to `set` is good, that makes the code behave as I would expect. And all callers want to use `set` functionality anyway.

I don't think logging is necessary but wouldn't hurt I guess.

chain-signatures/node/src/backlog/mod.rs
ChaoticTempest: ahh, I'll shorten it to your version

ChaoticTempest:
looking at it again, we can't do 3 until the linearization of ethereum (with https://github.com/sig-net/mpc/pull/743). Not easily at least since it would introduce more duplicate code than I would like due that chain event loop doing a lot of heavy lifting with backlog manipulations. So it would be best to keep it all in one place.

Why duplicate the event loop? Because due to ethereum not linearly sending events with this PR, we can have live events interleave into the chain loop where we also process catchup related chain events too. To not have it interleave would mean I need a separate chain event loop just for catchup.

It would be a lot simpler when https://github.com/sig-net/mpc/pull/743 lands in, but I feel like that PR might take some time and I'd like to get this one in first to see if it fixes some things. @jakmeier what do you think? It's pretty easy to remove all this marked related stuff afterwards since it's just code cruft

jakmeier:
I see, thanks for explaining!

>  I'd like to get this one in first to see if it fixes some things. @jakmeier what do you think? It's pretty easy to remove all this marked related stuff afterwards since it's just code cruft

True, it would be easy to remove again. I don't love the idea that we need to merge things to develop to test if it fixes a given problem but I suppose this is a more general problem.

But yeah as far as I'm concerned we can move ahead with this before https://github.com/sig-net/mpc/pull/743 and consider cleaning up the architecture afterwards. I don't want to delay things or force you to duplicate the event loop.

ChaoticTempest:
alright should be good to go now, just need an approval

jakmeier (approved):

PR #747, opened 2026-04-15
Make sure all nodes participate in T and P protocols

volovyks:
I was suspicious of the fact that when I was generating triples and presignatures locally, the list of participants was often not full. I think that under normal circumstances, all nodes must be able to ACCEPT and be included in all protocols.


volovyks:
It was passing locally, but on CI `cases::mpc::test_everybody_participate_triple_protocol` failed.

PR #749, opened 2026-04-15
Fixture improvements

volovyks:
- moved all the logic to build() for consistency
- require threshold when loading pregenerated keys/artifacts


jakmeier (approved):
Thank you for the clean up!

PR #753, opened 2026-04-16
eth indexer: fix block height metric dips

### LINKED ISSUE #752: eth indexer: block height dips from time to time
for some reason eth block height seem to go backwards, shown by the dips in value (probably without restarts too because we didn't deploy anything)

Image

ppca:


volovyks (approved):
Make sense. Going through the channel can add a bit of a delay, but that is a nice single place in the project to consider requests as finalized. cc @ChaoticTempest

ChaoticTempest (approved):

PR #757, opened 2026-04-16
test: MPC + stream component test (part 1)

jakmeier:
Create a MockStream that connects to the MPC fixture
setup. This lets us test the indexer stream <-> MPC glue.

This is part 1, with just the tests adjusted to pass with the current configuration.

See https://github.com/sig-net/mpc/pull/739 for tests that show current issues that need fixing.

### DIFF
--- a/chain-signatures/contract-sol/src/lib.rs
+++ b/chain-signatures/contract-sol/src/lib.rs
@@ -197,6 +197,7 @@ pub struct SignatureRespondedEvent {
 }
 
 #[event]
+#[derive(Clone)]
 pub struct RespondBidirectionalEvent {
     pub request_id: [u8; 32],
     pub responder: Pubkey,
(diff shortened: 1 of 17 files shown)


jakmeier:
This is no longer blocked with #765 merged.

@volovyks I think you already reviewed most of this in the draft (#739).
This is the same but with test settings changed that they pass.

jakmeier:
@volovyks sorry, `test_sign_no_presignature_waste` apparently is still flaky in CI, I will ping you again when it's resolved

jakmeier:
just a quick update: it succeeds locally but fails in CI, probably it is the same problem as with `test_channel_contention_multiple_blocks_at_once_delayed` in #739. Round alignment problems. It worked in CI with `ACCEPT_POSIT_TIMEOUT = 500ms` but fails again with  `ACCEPT_POSIT_TIMEOUT = 100ms`.

Basically, the current solution is still quite fragile. I need to look in to this more to decide if we can just increase the timeout for the tests, or if we can solve the underlying problem. (100ms for network delay should be plenty, even on slow CI machines)

jakmeier:
in theory, this should be passing tests now... but I did a bad rebase (clean in git but causes rust conflicts)

The `trait ChainStream` was changed on develop. I will have to fix that.

PR #759, opened 2026-04-20
fix: sign queue size staleness for metric

ChaoticTempest:
The sign task queue metric was not being updated properly on task exit, so it cloud lead to some stale reportings of sign queue size


volovyks (approved):
It may be worth setting  `SIGN_QUEUE_SIZE`  in the `SignatureSpawnerTask` `abort()` function to 0, since all the tasks are dropped there. But that only makes sense on a state transition.

ChaoticTempest:
state transition as in going from Node state of Running to Resharing? In that case, we should still have tasks paused so it wouldn't be zero IMO

PR #765, opened 2026-04-23
feat: try to include all deliberators in signing

jakmeier:
Instead of starting eagerly, give all deliberators some time to response with accept.

This avoids that with every signature, N-threshold nodes fall into the bad state where they keep trying to form a new posit round, even when they could have been
included in the first round.


ChaoticTempest (approved):
nice, thanks for adding this in

PR #767, opened 2026-04-24
solana indexer: combine respond and sign task

ppca:


chain-signatures/node/src/indexer_sol.rs
```diff
@@ -712,15 +584,34 @@ where
...
-                                tracing::warn!("Failed to parse cpi events for {}: {}", signature, e);
-                                continue;
+                        } else {
+                            let (respond_bidirectional_events, respond_events) = match parse_cpi_respond_events(tx_res, &program_id) {
+                                Ok(v) => v,
+                                Err(err) => {
+                                    tracing::warn!(?err, sig = %signature, "failed to parse respond events (will skip this signature)");
+                                    continue;
+                                }
+                            };
```
ChaoticTempest: can we also get `looks_like_cpi_respond_event` for this? Otherwise, I feel like we're going to get a lot of warnings here

ChaoticTempest: can we also create a looks_like_cpi_respond_event function for this? Otherwise, I feel like we're going to get a lot of warnings here

ppca: yep sure!

ChaoticTempest (approved):

PR #783, opened 2026-04-30
ci: remove redundant steps

jakmeier:
Installing rust stable and also 1.81.0 doesn't
seem to have a specific purpose, it's just extra
work.

This also now uses 1.81.0 everywhere, which
means we will no longer have unplanned
Rust version updates in CI checks, which leads
to new clippy warnings popping up without
us changing anything.

Lastly, I removed the `cargo check` step.
`cargo clippy` does the same and more,
and even `cargo test` would catch the same
errors.

### DIFF
--- a/.github/workflows/canton.yml
+++ b/.github/workflows/canton.yml
@@ -43,12 +43,6 @@ jobs:
           export PATH="$HOME/.dpm/bin:$PATH"
           dpm --version
 
-      - name: Install stable toolchain
-        uses: actions-rs/toolchain@v1
-        with:
-          profile: minimal
-          toolchain: stable
-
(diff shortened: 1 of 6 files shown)


ChaoticTempest (approved):
I wonder if we should bump all our near deps so we can go to version 1.86

jakmeier:
> I wonder if we should bump all our near deps so we can go to version 1.86

Not sure how high of priority that is. Anything you need from newer Rust?

Also, that's still 10 versions behind the latest rustc version. That should be resolved soon, so if anything, I would wait until that's resolved upstream.

ChaoticTempest:
ahh if it's going to be resolved soon upstream let's just wait then

PR #784, opened 2026-04-30
ci: run component tests in parallel

jakmeier:
1. Create two different jobs for (slow and sequential) cluster-based integration tests vs (parallel and hopefully faster) fixture-based tests.
2. Add `cargo nextest` as the test runner to make this easier to set up in CI as well as for running tests locally

If there are reasons not to use `cargo nextest`, I can revert the second commit and just split it into two jobs.

But I recommend `cargo nextest` for anyone running tests locally.

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,0 +1,20 @@
+# Always limit cluster tests to a single thread
+[test-groups]
+cluster = { max-threads = 1 }
+fixture = { max-threads = "num-cpus" }
+
+[[profile.default.overrides]]
+filter = 'test(cases::mpc) | test(cases::state_sync)'
+test-group = 'fixture'
+
+[[profile.default.overrides]]
+filter = 'not (test(cases::mpc) | test(cases::state_sync))'
+test-group = 'cluster'
+
+# Usage: cargo nextest run -p integration-tests --profile fixture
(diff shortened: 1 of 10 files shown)


jakmeier:
note: this builds on top of #783 to avoid merge conflicts

jakmeier:
Interesting, the improvement here is rather modest.

| | Latest run on develop | This PR |
| -- | -- | -- |
| All 56 Integration Tests | 41m | - |
| 22 Component Tests | - | 16m |
| 34 Integration Tests without Component Tests | - | 37m |

Only gaining 4min (~10%) in total CI time is disappointing, I was hoping for more. But I guess it makes sense, all the heavy tests still run in sequence.

The main benefit I see is for local development. Rerunning only component tests takes only 50s. And their coverage is starting to be quite decent. At least for me, lately they caught more bugs in my new code than the integration tests, as they test more specific and more tricky cases than the integration tests.

At some point, we could also run integrations tests on a nightly basis, if waiting for CI is too annoying.

ChaoticTempest:
I'm trying to actually get all our integration-tests to be ran in parallel (but `--test-threads 4`) so let's see if that improves anything

ChaoticTempest (approved):

PR #786, opened 2026-04-30
fix(audit): RUSTSEC-2023-0071

ChaoticTempest:
fixes RUSTSEC-2023-0071 by making jsonwebtoken only use specific features

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3530,7 +3530,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "8d162beedaa69905488a8da94f5ac3edb4dd4788b732fadb7bd120b2625c1976"
 dependencies = [
  "data-encoding",
- "syn 2.0.101",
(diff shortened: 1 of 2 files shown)


jakmeier (dismissed):
Nice way to resolve it!

ChaoticTempest:
guess canton integration hates this change though... Surprised it's not on the compilation level where this would've been caught

jakmeier (approved):

ChaoticTempest:
Just in case someone tries to merge this PR, I don't like that the `aws_lc_rs` backend since it pulls in non-rust dependencies and possibly adds to our build time. I'm going to try and get us back to rust_crypto and fix the audit in a different manner. The `rsa` crate is currently being updated to a higher version, so maybe we can wait until that is released, but if we need to make a release, we can ship with `aws_lc_rs` instead

ChaoticTempest:
no longer necessary since jsonwebtoken was removed as of latest develop

PR #787, opened 2026-05-05
feat: parallel integration tests (test-threads 4)

ChaoticTempest:
adds in the ability to run integration-tests in parallel (at least 4 tests at a time).


ChaoticTempest:
integration went from 37m to 21m in this PR. We can tune for higher than thread count 4 but might run into other issues with resource constraints

integration-tests/src/cluster/spawner.rs
```diff
@@ -5,17 +5,56 @@ use near_account_id::AccountId;
...
+}
+
+fn thread_network_name(docker: &DockerClient) -> String {
+    THREAD_NETWORK_NAME.with(|name_cell| {
+        let mut name = name_cell.borrow_mut();
+        if let Some(name) = name.as_ref() {
+            return name.clone();
+        }
+
+        let slot = NEXT_NETWORK_SLOT.fetch_add(1, Ordering::Relaxed);
```
jakmeier, line 45: Is the intention here to reuse existing docker networks across tests? Because `cargo nextest` runs tests in separate processes, unlike the default `cargo test` harness which runs them in separate threads of the same process.

I think this creates one network (and then destroys it) for every test. That's a bit wasteful but should be correct.

jakmeier (approved):
Nice, I like how it works with any number of threads in cli arguments.

integration-tests/src/cluster/spawner.rs
ChaoticTempest: yep, it's because each test are their own separate environment. So stuff like spinning up a redis container would conflict with another test if we used the same docker network

PR #788, opened 2026-05-07
fix: backpressure drop messages on full

ChaoticTempest:
second part of https://github.com/sig-net/mpc/pull/701

Should hotpatch this into testnet as well


jakmeier (commented):
I'm still not certain if this is the issue we have in testnet. At least I wasn't able to force this problem to happen in my testing a few weeks ago, even with very small buffer sizes and many simultaneous requests.

Would it make sense to merge https://github.com/sig-net/mpc/pull/701 first and confirm that channel queue lengths actually hit their capacity?

Unless you are already certain that this must be the issue, then we can merge this right after the metrics.

ChaoticTempest:
yep, best to merge 701 first. I'm not fully sure, but at least the unit test here at least confirms that we block messages from going when posit messages are full

ChaoticTempest:
also just noticed that our devnet was able to successfully fulfill a huge backlog of requests after restart with parallel integration tests, so I think it is this backpressure issue. Since a lot of sign tasks positing can fill up the channels

chain-signatures/node/src/protocol/message/sub.rs
```diff
@@ -190,20 +193,41 @@ impl<T> Subscriber<T> {
...
         }
     }
-}
 
+    pub fn try_send_lossy(&self, msg: T) -> Result<(), mpsc::error::SendError<T>> {
+        let result = match &self.kind {
+            SubscriberKind::Subscribed(tx) | SubscriberKind::Unsubscribed(tx, _) => {
+                match tx.try_send(msg) {
+                    Ok(()) => Ok(()),
+                    Err(mpsc::error::TrySendError::Full(_)) => {
```
ppca, line 206: I think this is good except in the case where: let's say there are 3 nodes. node 0 indexed request A first, nodes 1 and 2 took a while to catch up  (assuming they were down for a while), and while nodes 1 and 2 are catching up, they will keep receiving posit messages from node 0, and in this change, node 1 and node 2 will drop these messages. Will it be a problem for node 0 and 1 to later complete request A? 
I think what you proposed https://github.com/sig-net/mpc/issues/854#issuecomment-4653038086 is a better idea: dropping mesasges with round value < current highest. That way we are keeping the latest and useful messages, dropping old unuseful ones.

ChaoticTempest: so what I suggested is a bit more complicated to implement, because we'd then have to have the ability to flush the receiver somehow. The simplest thing I can think of is to do a ring buffer so newer messages overwrites older messages. There's no out of the box solution for this unfortunately. So we can add this for now while we think of how to replace this with a ring buffer or the ability to flush

ppca: can you add a TODO comment in code here and create an issue? Let's make sure not to forget about this because it can cause real issues for us.

ppca (dismissed):

chain-signatures/node/src/protocol/message/sub.rs
ChaoticTempest: yep

ppca (approved):

PR #792, opened 2026-05-07
bidirectional: replace eth_call with debug_traceTransaction

### LINKED ISSUE #795: Using `eth_call` for transaction output data is not reliable
### Description

Location: `chain-signatures/node/src/respond_bidirectional.rs`, `extract_success_tx_output` method:

```rust
let transaction_output = match output_deserialization_format {
    SerDeserFormat::Abi if is_contract_call => {
        let to_address = tx_info.inner.to().ok_or_else(|| {
            anyhow::anyhow!("Transaction {:?} missing destination", tx.id)
        })?;
        let call_block = block_number.saturating_sub(1);
        let call_result = client
            .call(tx.from_address, to_address, data.clone(), call_block)
            .await?;
        TransactionOutput::from_call_result(output_deserialization_schema, &call_result)?
    }
    _ => TransactionOutput::non_function_call_output(),
};
```

Using `eth_call` is not a reliable since it simulates transaction in vacuum and does not account for MEV, transaction ordering, etc. Simulation might return success while in reality transaction was reverted, unbacked tokens minted as a result. Node should check actual transaction status and check logs to guarantee consistency

ppca:
context in https://github.com/sig-net/mpc-private/issues/23
this is ready to go for dev. Mainnet we need to switch infura to quicknode

### DIFF
--- a/chain-signatures/node/src/indexer_eth/indexer_eth_helios.rs
+++ b/chain-signatures/node/src/indexer_eth/indexer_eth_helios.rs
@@ -70,7 +70,7 @@ impl HeliosEthereumClient {
         tx_hash: alloy::primitives::B256,
     ) -> anyhow::Result<Bytes> {
         tracing::warn!(
-            tx_hash = ?tx_hash,
+            ?tx_hash,
             "debug_traceTransaction is not supported by Helios; refusing to fall back to eth_call"
         );
 
--- a/chain-signatures/node/src/indexer_eth/indexer_eth_helios.rs
+++ b/chain-signatures/node/src/indexer_eth/indexer_eth_helios.rs
@@ -65,6 +65,20 @@ impl HeliosEthereumClient {
         })
     }
 
+    pub async fn trace_transaction_output(
+        &self,
+        tx_hash: alloy::primitives::B256,
+    ) -> anyhow::Result<Bytes> {
+        tracing::warn!(
+            tx_hash = ?tx_hash,
+            "debug_traceTransaction is not supported by Helios; refusing to fall back to eth_call"
+        );
+
+        anyhow::bail!(
+            "debug_traceTransaction is not supported by Helios; use direct RPC, e.g. Alchemy, for bidirectional transaction output extraction"
+        )
+    }
+
     pub async fn call(
         &self,
         from: Address,
--- a/chain-signatures/node/src/indexer_eth/mod.rs
+++ b/chain-signatures/node/src/indexer_eth/mod.rs
@@ -650,6 +650,16 @@ impl EthereumClient {
         }
     }
 
+    pub async fn trace_transaction_output(
+        &self,
+        tx_hash: alloy::primitives::B256,
+    ) -> anyhow::Result<alloy::primitives::Bytes> {
+        match self {
+            EthereumClient::Helios(client) => client.trace_transaction_output(tx_hash).await,
+            EthereumClient::DirectRpc(client) => client.trace_transaction_output(tx_hash).await,
+        }
+    }
+
     pub async fn call(
         &self,
         from: Address,
@@ -915,7 +925,7 @@ impl EthereumIndexer {
         );
 
         let result = if receipt_succeeded {
-            let completed_tx = CompletedTx::new(pending_tx.clone(), block_number);
+            let completed_tx = CompletedTx::new(pending_tx.clone());
             match completed_tx.extract_success_tx_output(&self.client).await {
                 Ok(serialized_output) => {
                     tracing::info!(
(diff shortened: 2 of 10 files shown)


chain-signatures/node/src/indexer_eth/indexer_eth_helios.rs
```diff
@@ -65,6 +65,20 @@ impl HeliosEthereumClient {
         })
     }
 
+    pub async fn trace_transaction_output(
+        &self,
+        tx_hash: alloy::primitives::B256,
+    ) -> anyhow::Result<Bytes> {
+        tracing::warn!(
+            tx_hash = ?tx_hash,
+            "debug_traceTransaction is not supported by Helios; refusing to fall back to eth_call"
```
ChaoticTempest, line 105: nit: you don't need to extra specify the variable naming if it's the same in tracing

```suggestion
        tracing::warn!(
            ?tx_hash,
            "debug_traceTransaction is not supported by Helios; refusing to fall back to eth_call"
```

chain-signatures/node/src/indexer_eth/mod.rs
```diff
@@ -928,26 +948,33 @@ impl EthereumIndexer {
...
                         ?sign_id,
                         ?err,
-                        "Failed to extract transaction output for bidirectional tx, using empty output"
+                        "Failed to extract transaction output for bidirectional \
+                         tx; leaving watcher pending for retry. Common causes: \
+                         trace RPC unavailable, malformed trace response, or \
+                         invalid output/response serialization schema."
                     );
-                    ExecutionOutcome::Success { output: vec![] }
+                    return None;
```
ChaoticTempest, line 1115: can we get some tests on CompletedTx so we can directly test this behavior?

ChaoticTempest (commented):
The question remains is that should we support a fallback option or should we only limit ourselves to RPC providers that support `debug_traceTransaction`. If our approach is multiple rpc providers over light client, then it would be a bit of an issue to find enough RPC providers.

Does having nodes using this `debug_traceTransaction` vs other nodes using `eth_call` leave us in a bad state? If not, then we probably should have a fallback in the case we are limited by RPC providers

ppca:
> The question remains is that should we support a fallback option or should we only limit ourselves to RPC providers that support `debug_traceTransaction`. If our approach is multiple rpc providers over light client, then it would be a bit of an issue to find enough RPC providers.
> 
> Does having nodes using this `debug_traceTransaction` vs other nodes using `eth_call` leave us in a bad state? If not, then we probably should have a fallback in the case we are limited by RPC providers

@ChaoticTempest Good question on the fallback! I'd say we can't fall back to `eth_call`, but rpc provider constraint is real. And I have a proposal to improve that. 

Why not fall back to `eth_call`?

Basically `eth_call` would give us incorrect result if there are other tx in the same block before this tx that touches the state that this tx needs/alters. So if the nodes using `eth_call` come together to generate a signature, then it's a pretty bad state for us. 
So for correctness sake, for this PR, I'd keep the strict requirement: use debug_traceTransaction, and fail on helios, rather than silently falling back to something unreliable.


Here is a planned follow-up: local revm replay.

Instead of asking an RPC to trace the transaction, each node re-executes it in-process with [revm](https://github.com/bluealloy/revm):
  - Fetch the block + preceding txs, replay them, then the target tx — the same execution context
  debug_traceTransaction reconstructs, just done locally.
  - Only needs standard RPC methods (eth_getStorageAt / eth_getCode / eth_getBalance / eth_getBlockBy*) — no
  debug namespace. These works on direct RPC and Helios. We could switch between them any time we want.

So the trajectory is:
1. this PR ships debug_traceTransaction (strict, no unreliable fallback)
2. mainnet switch infura to quicknodes
3. follow-up that implements local revm replay: https://github.com/sig-net/mpc/issues/818

volovyks:
How many providers support `debug_traceTransaction`? Do we have enough?

Agree that we can not fall back on something that can produce incorrect results.

If `revm` is reliable, fast, and supported by all providers (I have not investigated that), why not use it as our main and only option?

ppca:
> How many providers support `debug_traceTransaction`? Do we have enough?
> 
> Agree that we can not fall back on something that can produce incorrect results.
> 
> If `revm` is reliable, fast, and supported by all providers (I have not investigated that), why not use it as our main and only option?

Yes we have alchemy, and can add quicknode. aside from quicknode, chainstack and tenderly also supports it.
In theory revm is ideal solution, I'm proposing to do it as followup because it might take some time to get it right, and low chance but not impossible that while implementing I find out something is not as easy as it seems. @volovyks

volovyks:
@ppca make sense, please, finalize the PR and tests so we can merge it

ChaoticTempest (approved):

PR #794, opened 2026-05-07
feat: only fetch artifacts after posit succeeds

jakmeier:
WIP: THIS IS NOT COMPLETE. CURRENTLY IT FETCHES THE FULL ARTIFACT, NOT JUST AN ID.

Removing an artifact from redis before we know if the posit succeeds can lead to wasted triples and presignatures.

This introduces a "reserved" state, meaning that an artifact is reserved for being used. Only owned artifacts can be reserved.

This is different from the previous "reserved" state previously renamed to "generating" in https://github.com/sig-net/mpc/pull/738.

The extra state allows to avoid wasted P and T without recycling them.

For state sync, an artifact with a reserved id is still treated as AVAILABLE.


jakmeier:
Closing in favor of #819

PR #798, opened 2026-05-09
fix(evm): canonicalize transaction signature scalar encoding

Pessina:
## Summary
- Canonicalize EVM signed transaction RLP encoding for signature scalars in the MPC sign-and-hash path.
- Apply the same fix to EIP-1559 and legacy transaction reconstruction.
- Add Alloy-oracle regression tests for leading-zero `r`/`s` scalars, including the EIP-1559 access-list shape that exposed the issue.

## Why
Ethereum transaction signature fields `r` and `s` are RLP integers. The MPC node was appending fixed-width 32-byte scalar slices directly when manually reconstructing signed transaction bytes. If either scalar began with one or more zero bytes, the watcher hashed non-canonical bytes while Anvil hashed the canonical raw transaction, causing deterministic `BidirectionalTxId` mismatches at broadcast time.

Using the existing `U256::to_be_bytes_trimmed_vec()` conversion makes the manually encoded scalar payload match Alloy/Anvil canonical integer encoding without adding a local trimming helper.

### DIFF
--- a/chain-signatures/node/src/sign_bidirectional.rs
+++ b/chain-signatures/node/src/sign_bidirectional.rs
@@ -292,8 +292,8 @@ pub fn sign_and_hash_eip1559_from_unsigned(
     }
     let y: u8 = if y_parity { 1 } else { 0 };
     srlp.append(&y);
-    srlp.append(&r);
-    srlp.append(&s);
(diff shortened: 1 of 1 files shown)


ppca:
Do you know why this have worked for the bidirectional flow from solana and hydration but not canton? @Pessina

Issue #751, opened 2026-04-16
investigate low eth in-time request rate and slow solana requests


ppca:
Problems:
1. Eth requests have < 50% in-time success rate ([dashboard](https://signetwork.grafana.net/d/chain_signatures_main_dashboard/chain-signatures-new?orgId=1&from=now-24h&to=now-20h&timezone=utc&var-node_account_id=$__all&var-environment=dev&refresh=5s&viewPanel=panel-163))

<img width="1206" height="552" alt="Image" src="https://github.com/user-attachments/assets/e2ea0ac9-4871-442c-841e-a4c66d44afab" />

2. Solana requests can take more than 1 minute ([dashboard](https://signetwork.grafana.net/d/chain_signatures_main_dashboard/chain-signatures-new?orgId=1&from=now-24h&to=now-20h&timezone=utc&var-node_account_id=$__all&var-environment=dev&refresh=5s&viewPanel=panel-49))

<img width="1203" height="586" alt="Image" src="https://github.com/user-attachments/assets/926b346a-3b6c-4412-83cf-152c93302b57" />

ppca:
I digged into the Eth requests and find out:

**1) why in-time request success rate is low:** 

the setup of in time request metric `<grafana_link>` is wrong. 

Currently in-time request rate = sum count of `multichain_sign_request_latency_sec(stage=total, status="in_time")` / (sum count of `multichain_sign_request_latency_sec(stage=total, status="in_time")` + sum of `multichain_sign_request_delayed` )

`multichain_sign_request_latency_sec(stage=total, status="in_time")`: only signature proposer can emit because it's only emitted at signature publish;
`multichain_sign_request_delayed`:  emitted by every node that indexed the request and have somehow failed to complete the signature generation in time (a lot of them have deliberator timeout). 

So the issue is, for one sign request, there could be at most 1 count of `multichain_sign_request_latency_sec(stage=total, status="in_time")`, but can have many ``multichain_sign_request_delayed`. 

The best way to fix this is: only emit multichain_sign_request_delayed if the node is proposer；

**2) why eth has so many delayed requests:** 
with eth, the respond event will only be processed when the respond event gets finalized, which will be another 20 mins after signature published. while waiting for that respond event, the nodes that failed to complete signature generation will keep retrying, and they will emit the multichain_sign_request_delayed metric even tho the request is already completed by other nodes.

[...]

PR #800, opened 2026-05-11
request latency observability: latency metrics for organizing, posit and generating

ppca:
as pointed out in https://github.com/sig-net/mpc/issues/751, we need more visibility into the time spent looping between organizing and posit.

By looking at code, it turns out, besides going from posit -> organizing when posit fails, it is also possible to go from generating -> organizing when generating fails.

This change captures the possibility that organizing, posit and generating can each be happening multiple times before a sign request is completed, and sum up all the time spent in each attempt in the corresponding phase per sign request.

Example:
let's say what happened is: organizing(1s) -> posit (which failed, 2s) -> organizing(1s) -> posit(2s) -> generating (failed, 10s) -> organizing(1s) -> posit(2s) -> generating(10s) -> completed, then the request latency emitted for stage `organizing` will be the 3 organizing attempts added together = 1 + 1 +1 = 3s, the latency emitted for stage `posit` = 2 + 2 +2 = 6s, the latency metric emitted for stage `Generating` = 10 +10 = 20s.

This would better observe the requests where we had multiple attempts, and the request stages' latency would add up towards total request latency much more closely.

Caveat:
we could still lose time in idle waiting, i.e. waiting while contract state changes, we omit them for now.

### DIFF
--- a/chain-signatures/node/src/metrics/requests.rs
+++ b/chain-signatures/node/src/metrics/requests.rs
@@ -47,7 +47,7 @@ impl SignRequestStep {
     }
 }
 
-pub(crate) static SIGN_REQUEST_LATENCY: LazyLock<HistogramVec> = LazyLock::new(|| {
+static SIGN_REQUEST_LATENCY: LazyLock<HistogramVec> = LazyLock::new(|| {
     try_create_histogram_vec_with_node_and_version(
         "multichain_sign_request_latency_sec",
         "Latency of multichain sign request processing with step and status specification.",
--- a/chain-signatures/node/src/metrics/requests.rs
+++ b/chain-signatures/node/src/metrics/requests.rs
@@ -117,7 +117,7 @@ pub(crate) static SIGN_REQUEST_DELAYED: LazyLock<CounterVec> = LazyLock::new(|| 
 /// the loop-back: organizing (self-loop on no presignature / no active peers),
 /// posit (consensus failed / timeout), or generating (generator construction
 /// or MPC run failed).
-pub static SIGN_REQUEST_LOOPS: LazyLock<CounterVec> = LazyLock::new(|| {
+pub(crate) static SIGN_REQUEST_LOOPS: LazyLock<CounterVec> = LazyLock::new(|| {
     try_create_counter_vec_with_node_and_version(
         "multichain_sign_request_loops_total",
         "Number of back-edges to organizing in the sign request state machine, by chain and source phase.",
--- a/chain-signatures/node/src/metrics/requests.rs
+++ b/chain-signatures/node/src/metrics/requests.rs
@@ -1,4 +1,5 @@
 use std::sync::LazyLock;
+use std::time::Duration;
 
 use prometheus::{exponential_buckets, CounterVec, HistogramVec, IntGauge};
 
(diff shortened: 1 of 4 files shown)


chain-signatures/node/src/metrics/requests.rs
```diff
@@ -16,7 +17,20 @@ pub enum SignRequestStep {
     Indexing,
     /// Time from indexing to signature generation start (Status: ok)
     AwaitingGeneration,
-    /// Time to generate the signature (Status: ok, error)
+    /// Cumulative time spent in the organizing phase across all attempts for
+    /// this request (Status: ok). Summed across loop iterations. In the
+    /// absence of governance pauses, Indexing + AwaitingGeneration +
+    /// Organizing + Posit + Generating + Responding == Total; resharing or
+    /// other transitions out of `Running` mid-request show up only in Total,
+    /// so the equality holds as `<=` in that case.
```
volovyks: This explanation is not relevant to Organizing. Also, it is obvious that the sum of all steps should be equal to the total.

chain-signatures/node/src/metrics/requests.rs
```diff
@@ -32,14 +46,16 @@ impl SignRequestStep {
...
+            Self::Posit => "posit",
             Self::Generating => "generating",
             Self::Responding => "responding",
             Self::Total => "total",
         }
     }
 }
 
-static SIGN_REQUEST_LATENCY: LazyLock<HistogramVec> = LazyLock::new(|| {
+pub(crate) static SIGN_REQUEST_LATENCY: LazyLock<HistogramVec> = LazyLock::new(|| {
```
volovyks: I do not think we need this change

ChaoticTempest (dismissed):
lgtm

chain-signatures/node/src/metrics/requests.rs
```diff
@@ -80,6 +104,18 @@ pub(crate) static SIGN_REQUEST_DELAYED: LazyLock<CounterVec> = LazyLock::new(||
     .unwrap()
 });
 
+/// Counts back-edges to organizing in the sign request state machine.
+/// `from_phase` identifies the source: organizing (self-loop), posit
+/// (consensus failed/timeout), or generating (construction or MPC failed).
+pub(crate) static SIGN_REQUEST_LOOPS: LazyLock<CounterVec> = LazyLock::new(|| {
```
volovyks, line 110: I was sceptical about adding this metric, because this is more about debugging than measuring. But since the request processing is so fundamental, let's have it.

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -1234,12 +1236,7 @@ impl SignGenerator {
                         "completed signature generation"
                     );
 
-                    record_request_latency(
-                        self.indexed.chain,
-                        SignRequestStep::Generating,
-                        "ok",
-                        self.created,
-                    );
+                    // SignRequestStep::Generating is emitted cumulatively from SignTask::run; the per-attempt protocol metrics below are unchanged.
```
volovyks: Do we really need this in our code base?

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -1810,6 +1883,72 @@ mod tests {
         assert_eq!(sign_task.governance.me, Participant::from(0));
     }
 
+    #[test]
+    fn phase_durations_default_is_zero() {
```
volovyks: PhaseDuration does not even have a custom default implementation. Do we really need this test?

volovyks (dismissed):
loogs good,
I like the additive design

volovyks (approved):

PR #801, opened 2026-05-13
fix: more nondeterminism issues with solana

ChaoticTempest:
should fix more issues with solana being non-deterministic due to the port reservations not being correct


volovyks:
Closing: targets the pre-rewrite Solana indexer; the nondeterminism concerns were addressed in the poll-based redesign (#1158, #1169).

PR #802, opened 2026-05-13
chore: feature gate and disable helios by default

ChaoticTempest:
this feature gates and disables helios by default. Our binaries still ship helios when we go to deploy them, it's just that for faster local iterability, helios has been feature gated and disabled. Why build something when we barely ever touch it

building the `mpc-node` binary went from 1505 dependencies and 3m 26s build time to 1324 dependences and 2m 39s build time.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -20,7 +20,7 @@ rustflags = [
 runner = "./setup.sh"
 
 [alias]
-build-node = "build -p mpc-node --release"
+build-node = "build -p mpc-node --release --features helios"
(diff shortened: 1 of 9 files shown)


volovyks (dismissed):
I like the idea, but want @auto-mausx to check it and decide what the best setup is. deployment, etc).
From what I see, now we are turning off Helios locally but bundling it when creating the release image.

ppca (approved):

ChaoticTempest:
merging this now, since the Dockerfile here automatically builds with helios flags so our deployments are good to go

PR #804, opened 2026-05-13
fix(ci): target ARC runner scale set

auto-mausx:
## Summary
- update the scheduled k6 load test to target the current ARC runner scale set label (`arc-runner-set`)
- replace the stale `arc-runners-arc-runner-set` label left over from earlier ARC wiring

## Verification
- `git diff --check origin/develop..HEAD`
- confirmed live ARC listener config in `gke_`<gcp_project_dev>`_europe-west1_dev` reports `runner_scale_set_name: arc-runner-set` and `runner_scale_set_id: 3`


auto-mausx:
This should fix our broken scheduled load tests

volovyks (approved):

PR #806, opened 2026-05-14
fix: sync staleness

ChaoticTempest:
was debugging through testnet redis storage, and these are some bugs I've noticed if we're missing metadata:
- if owner mapping, artifact mapping, or holders metadata is missing in remove_outdated, then the entry should be marked as not_found
- in remove_holder_and_prune, if holders metadata does not exist we should remove the artifact itself.

These are all for artifacts that do not have these metadata. This is likely due to us adding these metadata as we went on without deleting prior old state or not bumping the versions


volovyks (approved):
Nice catch

PR #807, opened 2026-05-14
chore: disable integration-test parallelism

ChaoticTempest:
I don't have time to get integration test parallelism to be running properly, so let's just disable it for now and not have it be a pain point for other PRs. 

If someone wants to pick this up, feel free too. Otherwise, some stuff in https://github.com/sig-net/mpc/pull/801 would be needed to confirm that we're no longer failing non-deterministically.


volovyks (approved):
We can decrease the number of threads to 2 or 3, but I'm ok with reverting for now.

PR #816, opened 2026-05-21
feat: canton catchup and livestream

ChaoticTempest:
Did the minimal amount of changes to implement catchup + livestream for canton.

Added `CantonConnection` which does all the logic for connecting to the websocket stream of Canton. Since Canton streaming can start from any point, we start at where we left off in the last checkpoint. Catchup still happens here because we don't want to do work for that's already been done by other nodes on their views of what is latest on canton.

How this `CantonIndexer` differs from the solana one is that for catching up, on `process_catchup(offset)` we have to receive from the websocket stream the `Update` till height or `offset`. So not too different, but the code can look weird here.

This `ChainIndexer` and `ChainStream` interface now is reaching its limits a bit. I will need to refactor the type after this PR so we can make this more concise and clear. Probably with something like a cursor that can explicitly mutate the websocket stream objects:
```rs
pub trait ChainIndexer {
    const CHAIN: Chain;
    type CatchupCursor: Send;
    type LiveEntry: Send;

    async fn start(&mut self) -> anyhow::Result<CatchupPlan<Self::CatchupCursor>>;
    async fn process_catchup(
        &mut self,
        cursor: &mut Self::CatchupCursor,
    ) -> anyhow::Result<CatchupStep>;
    async fn next(&mut self) -> Option<Self::LiveItem>;
    async fn process(&mut self, item: Self::LiveEntry) -> anyhow::Result<()>;
    async fn notify_catchup_completed(&mut self) -> anyhow::Result<()>;
}
```

### DIFF
--- a/chain-signatures/node/src/indexer_canton/stream.rs
+++ b/chain-signatures/node/src/indexer_canton/stream.rs
@@ -287,6 +287,11 @@ impl CantonIndexer {
     }
 
     async fn process_catchup_offset(&mut self, target_offset: u64) -> anyhow::Result<()> {
+        // If we're already at or past the target offset, we're done
+        if self.last_seen_offset >= target_offset {
+            return Ok(());
+        }
+
         loop {
             let Some(update) = self.next_update().await else {
                 anyhow::bail!("canton WebSocket closed during catchup; reconnecting");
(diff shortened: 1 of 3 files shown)


chain-signatures/node/src/indexer_canton/stream.rs
```diff
@@ -69,287 +60,298 @@ impl ChainStream for CantonStream {
...
+    type Iter = Range<u64>;
+
+    async fn livestream(&mut self) -> anyhow::Result<Option<u64>> {
+        let checkpoint = self
+            .backlog
+            .processed_block(Chain::Canton)
+            .await
+            .unwrap_or(0);
+        self.last_seen_offset = checkpoint;
+        let anchor_height = self.client.fetch_ledger_end().await?;
```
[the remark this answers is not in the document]
ChaoticTempest: I don't think this matters as much. ledger_end should be the latest update/block from canton. So this should be handled the same as every other chain

chain-signatures/node/src/indexer_canton/stream.rs
```diff
@@ -69,287 +60,298 @@ impl ChainStream for CantonStream {
...
+                    transaction_shape: "TRANSACTION_SHAPE_LEDGER_EFFECTS".to_string(),
+                    event_format: ledger_api::EventFormat {
+                        filters_by_party,
+                        verbose: true,
+                    },
+                },
+            },
+        };
+        tokio::time::timeout(
+            Duration::from_secs(30),
```
jakmeier: optional, nit: this is the second place where a magic 30 seconds is used for the timeout, could be a named const

jakmeier (dismissed):
looks good, nice to see things coming together

> This ChainIndexer and ChainStream interface now is reaching its limits a bit. I will need to refactor the type after this PR so we can make this more concise and clear.

:+1: for this

chain-signatures/node/src/indexer_canton/stream.rs
ChaoticTempest: replaced

jakmeier (approved):

Pessina (approved):

Issue #793, opened 2026-05-07
Prevent non-participant from blocking resources


jakmeier:
When a successful posit does not include all potential participant, those who are not chosen will keep sending posits.

`test_sign_no_presignature_waste` becomes rather unstable due to this problem. Whenever non-participants are trying to be a proposer, they will block a presignature and therefore prevent the test from using up all existing presignatures.

Right now, this even deletes the old presignature. We can resolve that by reserving ids and only removing artifacts when the posit has succeeded, as described in #794. But that would still block a presignature (or triple) temporarily, still causing the test to be flaky.

In practice, blocking temporarily isn't a big deal. as long as we have plenty of presignatures. But we should aim for stability even when we run low in presignatures.

We should try to detect as non-participant that a signature is on-going and ignore that signature for some time. For example, the winning proposer could send a message to non-participants.

jakmeier:
With https://github.com/sig-net/mpc/pull/819 and https://github.com/sig-net/mpc/pull/832, this should no longer be an issue. As far as I know, the test `test_sign_no_presignature_waste` has been stable.

PR #819, opened 2026-05-21
feat: avoid presignature waste on posit rounds

jakmeier:
As described in #793, for every extra posit round for signatures is needed, we waste a presignature.

When not all nodes are included as participants, this is particularly bad, as the remaining nodes will keep trying to run more posit rounds until they see the finalized resulting signature on chain.

This PR avoid the presignature waste. For the posit round, we only mark it as "pending" and don't remove it from Redis. Only when we actually start using the presignature is it removed from Redis.

This preserves the invariants that:

- no signing algorithm starts before we remove the presignature from Redis
- once removed from redis, we never put a presignature back


jakmeier:
@volovyks @ChaoticTempest I added you both as reviewers, as we discussed the issue of recycled presignatures before.

volovyks (approved):
LGTM
I was concerned it might break the state sync, but it appears it only shifts the moment the artifact is considered "taken".

Tst coverage would be great though.

ChaoticTempest (approved):

PR #827, opened 2026-05-27
change name for signature spawner inbox size

ppca:
differentiate signature spawner posit inbox capacity from "sign_task"


chain-signatures/node/src/protocol/signature.rs
```diff
@@ -65,6 +59,9 @@ const ACCEPT_POSIT_TIMEOUT: Duration = Duration::from_millis(if cfg!(feature = "
     500
 });
 
+/// Metric channel label shared by every entry in `SignatureSpawner.inboxes`.
+const SIGNATURE_SPAWNER_POSIT_INBOX_LABEL: &str = "signature_spawner_posit_inbox";
```
ChaoticTempest: nit: looks a bit long of naming just to have a label. Probably this would be better
```suggestion
const SIGN_POSIT_INBOX_LABEL: &str = "sign_posit_inbox";
```

ChaoticTempest (dismissed):

ChaoticTempest (approved):

PR #830, opened 2026-05-28
feat: more granular SignStatus & requeue publishing to RPC

ChaoticTempest:
This makes our SignStatus more granular:
```
enum SignStatus {
    PendingGeneration,
    PendingPublish { publish: PublishState },
    PendingExecution { tx: BidirectionalTx },
    PendingGenerationBidirectional,
    PendingPublishBidirectional { publish: PublishState },
}
```

Where we will now explicitly handle the publishing part where if a node were to restart in the middle of a publishing, they would restart from there instead of going back to signature generation phase. This however does not solve the issue of watchdogging over rpc publishing where if the proposer fails to republish, the other participants could publish themselves

### DIFF
--- a/chain-signatures/node/src/backlog/mod.rs
+++ b/chain-signatures/node/src/backlog/mod.rs
@@ -76,7 +76,9 @@ impl PendingRequests {
     }
 
     fn pending_execution(&self, id: &SignId) -> Option<&BacklogEntry> {
-        self.requests.get(id).filter(|entry| entry.status().is_pending_execution())
+        self.requests
+            .get(id)
+            .filter(|entry| entry.status().is_pending_execution())
     }
 
     fn pending_executions(&self) -> Vec<(SignId, BacklogEntry)> {
@@ -378,6 +380,14 @@ impl Backlog {
             .unwrap_or_default()
     }
 
+    pub async fn pending_execution(&self, chain: Chain, id: &SignId) -> Option<BacklogEntry> {
+        self.requests
+            .read()
+            .await
+            .get(&chain)
+            .and_then(|requests| requests.pending_execution(id).cloned())
+    }
+
     pub async fn len_by_chain(&self, chain: Chain) -> usize {
         self.requests
             .read()
@@ -462,7 +472,7 @@ impl Backlog {
 
     /// Get the set of bidirectional transactions currently awaiting execution on the
     /// specified destination chain.
-    pub async fn pending_execution(
+    pub async fn execution_watchers(
         &self,
         chain: Chain,
     ) -> HashMap<BidirectionalTxId, (SignId, BidirectionalTx)> {
@@ -1169,6 +1179,7 @@ mod tests {
         let backlog = Backlog::new();
 
         // Add transactions with different statuses to Ethereum
+        let tx0 = create_test_tx(0);
         let tx1 = create_test_tx(1);
         let tx2 = create_test_tx(2);
         let tx3 = create_test_tx(3);
@@ -1211,12 +1222,9 @@ mod tests {
 
         // Filter Ethereum by Pending
         let eth_pending = backlog
-            .get_by_status(
-                Chain::Ethereum,
-                pending_execution_status(&create_test_tx(3)),
-            )
+            .pending_execution(Chain::Ethereum, &SignId::new(tx3.request_id))
             .await;
-        assert_eq!(eth_pending.len(), 1);
+        assert!(eth_pending.is_some());
 
         let eth_awaiting = backlog
             .get_by_status(Chain::Ethereum, SignStatus::PendingGeneration)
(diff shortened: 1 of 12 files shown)


chain-signatures/node/src/backlog/mod.rs
```diff
@@ -766,10 +853,11 @@ impl BacklogEntry {
         &mut self,
         bidirectional_tx: BidirectionalTx,
     ) -> Result<(), BacklogError> {
-        match (&self.request.kind, self.status) {
-            (SignKind::SignBidirectional(_), SignStatus::AwaitingResponse) => {
-                self.status = SignStatus::PendingExecution;
-                self.execution = Some(bidirectional_tx);
+        match (&self.request.kind, self.status.clone()) {
+            (SignKind::SignBidirectional(_), SignStatus::PendingPublish { .. }) => {
```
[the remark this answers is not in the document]
jakmeier: @ChaoticTempest isn't this a legit problem? This would hit if we recover `PendingGeneration` from the checkpoint and then during catch-up we find the signature is already published. (Quiet common case if the node is offline for a while.)

Don't we need to set the status to `PendingExecution` here in such a case?

chain-signatures/node/src/sign_bidirectional.rs
```diff
@@ -31,19 +32,69 @@ struct AbiField {
...
+    pub signature: Signature,
+    pub participants: Vec<Participant>,
+    pub is_proposer: bool,
+}
+
+impl PublishState {
+    fn digest_bytes(&self, tag: u8) -> Vec<u8> {
+        let mut bytes = vec![tag];
+        bytes.extend(borsh_to_vec(&self.signature).expect("signature serialization is infallible"));
+        ciborium::ser::into_writer(&self.participants, &mut bytes)
```
jakmeier: nit: mixing borsh and ciborium is a bit ugly but I guess it works just fine

chain-signatures/node/src/backlog/mod.rs
```diff
@@ -331,18 +412,32 @@ impl Backlog {
...
+        let entry = pending
+            .requests
+            .get(id)
+            .ok_or(BacklogError::NotFound { chain, id: *id })?;
+
+        match (&entry.request.kind, &entry.status) {
+            (SignKind::SignBidirectional(_), SignStatus::PendingPublish { .. })
+            | (SignKind::RespondBidirectional(_), SignStatus::PendingPublishBidirectional { .. }) =>
+            {
+                tracing::info!(?chain, ?id, success, status = ?entry.status, "publish attempt recorded");
```
jakmeier: What exactly do we record? It looks like we only validate the transition but I don't see any state modifications.

jakmeier (commented):
I think the extra state make sense to add. But it is getting a bit complicated I mus admit 😅 I left some comments/questions, PTAL.

ChaoticTempest:
yeah, this is getting a bit rough. Hopefully after this PR I can turn it into the typestate which lets us do the state transitioning better

chain-signatures/node/src/backlog/mod.rs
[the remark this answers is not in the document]
ChaoticTempest: ahh yeah it is. Skimmed over this one, but added the ability to transition from PendingGeneration as well

chain-signatures/node/src/sign_bidirectional.rs
ChaoticTempest: fixed, no longer using either ciborium or borsh, but our own custom one. This way we don't have any sort of infallibility

chain-signatures/node/src/backlog/mod.rs
ChaoticTempest: yeah, this one feels a bit weird to have now that I think about it. So removed `mark_published`, but instead have `mark_publishing` for the transition from PendingGeneration to PendingPublish

chain-signatures/node/src/backlog/mod.rs
```diff
@@ -762,14 +841,32 @@ impl BacklogEntry {
         self.request = request;
     }
 
+    pub fn mark_publishing(&mut self, publish: PublishState) -> Result<(), BacklogError> {
+        match (&self.request.kind, self.status.clone()) {
+            (SignKind::Sign | SignKind::SignBidirectional(_), SignStatus::PendingGeneration) => {
```
volovyks, line 846: Why is it possible to transition from PendingGeneration to PendingPublish without going through PendingGeneration? Should we be more strict?

Ah, I see your and Jakob's comments now. Let's at least document this.

chain-signatures/node/src/storage/checkpoint_storage.rs
```diff
@@ -10,7 +10,7 @@ use tokio::sync::RwLock;
 use std::collections::HashMap;
 use std::sync::Arc;
 
-const CHECKPOINT_VERSION: &str = "v6";
+const CHECKPOINT_VERSION: &str = "v7";
```
volovyks, line 13: Some day we will need to stop doing this

volovyks (approved):
Definitely a move in the right direction!

chain-signatures/node/src/backlog/mod.rs
ChaoticTempest: yep, would be fixed in the future when we have better typestate

chain-signatures/node/src/storage/checkpoint_storage.rs
ChaoticTempest: yeah, the trade off is if we want to invest in doing migration

volovyks:
image

This PR improved consistency in our Backlog and reduced the post-stage time significantly. Great work @ChaoticTempest!
cc @ppca @jakmeier

PR #832, opened 2026-05-28
feat: pause proposing while other nodes are generating

jakmeier:
This is the second step to address #793.
Potentially the last step.

### DIFF
--- a/chain-signatures/node/src/protocol/message/types.rs
+++ b/chain-signatures/node/src/protocol/message/types.rs
@@ -4,7 +4,7 @@ use std::hash::{DefaultHasher, Hash as _};
 use cait_sith::protocol::{MessageData, Participant};
 use serde::{Deserialize, Serialize};
 
-use crate::protocol::posit::PositAction;
+use crate::protocol::posit::{PositAction, PositRejectReason};
 use crate::protocol::presignature::{FullPresignatureId, PresignatureId};
 use crate::protocol::triple::TripleId;
 use crate::types::Epoch;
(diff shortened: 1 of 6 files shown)


jakmeier:
only the last commit is relevant, the rest is from #739

that needs to go in first, as otherwise the entropy in tests changes again

jakmeier:
@ChaoticTempest This is ready for a review now

chain-signatures/node/src/protocol/message/types.rs
```diff
@@ -46,7 +46,9 @@ impl PositMessage {
                 participants.len() * std::mem::size_of::<Participant>()
             }
             PositAction::Accept => 0,
+            #[allow(deprecated)]
             PositAction::Reject => 0,
+            PositAction::RejectWithReason(_reason) => std::mem::size_of::<PositRejectReason>(),
```
ChaoticTempest, line 51: optional: so we're probably going to fail deserializing `PositAction::RejectWithReason` anyways, so might as well just replace `PositAction::Reject` entirely unless there are places that still need it without a reason

ChaoticTempest (approved):
LGTM, but will likely break the serialization for the messages. We can bump PROTOCOL_VERSION for this but at that rate, we should fix what I mention in the comments as well

ppca (approved):

chain-signatures/node/src/protocol/message/types.rs
volovyks: followup and PROTOCOL_VERSION bump discussion: https://github.com/sig-net/mpc/pull/834

PR #833, opened 2026-05-30
fix: solana catchup missing entry

### LINKED ISSUE #828: lingering already generated signIds: missed respond events during catchup
This is just one scenario of lingering already generated signIds.

When a node gets restarted, its indexers will need to catch up. However, it happens a lot that we fail to fetch some blocks, especially Solana. And it becomes a problem when we have successfully found a sign request in a catchup block, but failed to find its respond event because the block where that respond event was emitted failed to be fetched. This signId will keep retrying because it will never see its respond event, and eventually it will clog other nodes' signature spawner inbox.

The lingering signid will be on the node forever as it is preserved in the backlog.

Below is a case of failed catchup blocks in solana:

Image

ChaoticTempest:
fixes https://github.com/sig-net/mpc/issues/828

solana catchup will now no longer skip blocks in case where they are empty, and retry until it can get those blocks. Originally caused by: 
> Transaction version (0) is not supported by the requesting client. Please try the request again with the following configuration parameter: \"maxSupportedTransactionVersion\": 0

Which is fixed by having proper configuration


volovyks (dismissed):
LGTM
PS, what do you think about adding git hooks for Clippy?

ChaoticTempest:
could work but only if the git hooks are when you're pushing to a branch, but there's a lot of times where I'm just storing stuff commits on github to revisit later. Not ideal in those situations

ppca (approved):

PR #834, opened 2026-06-01
Remove reject

volovyks:
I [agree](https://github.com/sig-net/mpc/pull/832#discussion_r3327123682) with Phuong that keeping Reject does not make much sense since serialization issues exist anyway.

I considered bumping the PROTOCOL_VERSION, but realized that it isn't the best option.
Why:
- Let's say we have 7 nodes, T=5
- 1, then 2, upgrade, all fine
- node 3 upgrades - system stops functioning
- node 4 upgrades - still down
- node 5 upgrades - system is functioning again

The time between 3 and 5 updates can be long. We can not afford that.


ChaoticTempest (approved):

ppca (approved):
lgtm

PR #835, opened 2026-06-01
Check respond signature

volovyks:
Anybody can create a response event without providing a valid signature. Signature check must be a part of our flow.

### DIFF
--- a/chain-signatures/node/src/stream/mod.rs
+++ b/chain-signatures/node/src/stream/mod.rs
@@ -172,15 +172,9 @@ pub async fn catchup_then_livestream<I: ChainIndexer>(mut indexer: I) {
     let chain = I::CHAIN;
     tracing::info!(%chain, "starting ChainStream catchup then livestream");
 
-    // TODO: on failure, we currently send catchup_completed due to some streams not enabling
-    // this particular catchup_then_livestream function (i.e. Solana & Hydration). Once
-    // those are implemented, we can remove the catchup_completed sending on error here.
     let anchor_height = match indexer.livestream().await {
         Ok(anchor_height) => anchor_height,
         Err(err) => {
-            if let Err(err) = indexer.notify_catchup_completed().await {
-                tracing::warn!(?err, %chain, "failed to signal catchup completion");
-            }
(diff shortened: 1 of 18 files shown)


chain-signatures/node/src/stream/ops.rs
```diff
@@ -529,16 +553,30 @@ pub(crate) async fn process_respond_event(
...
+
+    if !matches!(entry.request.kind, SignKind::RespondBidirectional(_)) {
+        anyhow::bail!(
+            "unexpected sign type for RespondBidirectionalEvent: {:?}",
+            entry.request.kind
+        );
+    }
+
+    let root_public_key = contract_watcher.wait_public_key().await;
+    verify_entry_signature(root_public_key, &entry, &event.signature(), sign_id)?;
```
ChaoticTempest: we shouldn't be waiting for the public key inside processing the respond event. It should be initially fetched before processing events in the event loop and then on change, we update it

chain-signatures/node/src/stream/mod.rs
```diff
@@ -924,7 +923,9 @@ mod tests {
             .set_status(
                 Chain::Solana,
                 &sign_id,
-                SignStatus::PendingExecution { tx: execution },
+                SignStatus::PendingExecution {
+                    tx: execution.clone(),
```
ChaoticTempest: nit: we don't need to clone this just for the id that is used later

ChaoticTempest (commented):
nice catch, we shouldn't be processing respond events if the signatures aren't correct. The only issue now is that what do we do if in the case our public key changes in the middle of this? Not something we need to solve right now, but something to keep in mind

chain-signatures/node/src/stream/ops.rs
volovyks: Good point, now we are getting the PK once the stream starts.
As for managing key rotation, this is not as simple as keeping the PK up to date with the contract. PK version depends on the request parameters (key version), not only on the state of the contract. Key rotation is not supported by our system yet, so I do not want to pretend it does.

ChaoticTempest (approved):

PR #841, opened 2026-06-03
component test: missed respond event lead to nodes stop signature generation

ppca:
- Adds a MockChain test infrastructure that simulates a shared chain: sign requests and respond events are distributed to all nodes' streams, with per-node filters for simulating event misses
- Adds two component tests that demonstrate a production issue where a single missed respond event can clog the signature spawner inbox, both are expected to pass:
-- test_missed_respond_event_clogs_inbox: node 2 is excluded from generation for a single request(bad_request) and misses the respond event. Its stale task keeps sending posit messages that fill inboxes of this sign_id on the nodes 0 and 1, eventually blocking their SignatureSpawner and preventing all future signature generation.
-- test_respond_event_prevents_clog: Same setup but the respond event IS delivered. It cleans up the stale task and all 20 requests complete — proving the respond event is the key factor.
- Reduces POSIT_INBOX_CHANNEL_SIZE to 4 under test-feature so the lingering signId's inbox fills quickly in tests (production keeps 4096, which will typically fill in ~8 hrs if stale task persists)

### DIFF
--- a/chain-signatures/node/src/protocol/message/sub.rs
+++ b/chain-signatures/node/src/protocol/message/sub.rs
@@ -18,6 +18,15 @@ use crate::util::channel_len;
 /// This should be enough to hold a few messages in the inbox.
 pub const MAX_MESSAGE_SUB_CHANNEL_SIZE: usize = 4 * 1024;
 
+/// Channel size for posit inboxes in SignatureSpawner::handle_posit.
+/// Kept small under test-feature so dead-letter inboxes (created for completed
+/// sign_ids) fill quickly and expose the clogging cascade in component tests.
+pub const POSIT_INBOX_CHANNEL_SIZE: usize = if cfg!(feature = "test-feature") {
+    4
+} else {
+    MAX_MESSAGE_SUB_CHANNEL_SIZE
+};
+
 pub enum SubscribeId {
     Generating,
     Resharing,
--- a/chain-signatures/node/src/protocol/message/sub.rs
+++ b/chain-signatures/node/src/protocol/message/sub.rs
@@ -18,9 +18,7 @@ use crate::util::channel_len;
 /// This should be enough to hold a few messages in the inbox.
 pub const MAX_MESSAGE_SUB_CHANNEL_SIZE: usize = 4 * 1024;
 
-/// Channel size for posit inboxes in SignatureSpawner::handle_posit.
-/// Kept small under test-feature so dead-letter inboxes (created for completed
-/// sign_ids) fill quickly and expose the clogging cascade in component tests.
+/// Small under test-feature so dead-letter inboxes fill quickly in clog tests.
 pub const POSIT_INBOX_CHANNEL_SIZE: usize = if cfg!(feature = "test-feature") {
     4
 } else {
--- a/chain-signatures/node/src/protocol/signature.rs
+++ b/chain-signatures/node/src/protocol/signature.rs
@@ -1708,10 +1708,12 @@ impl SignatureSpawner {
         if from == me {
             return;
         }
-        let inbox = self
-            .inboxes
-            .entry(sign_id)
-            .or_insert_with(|| Subscriber::unsubscribed(SIGN_POSIT_INBOX_LABEL));
+        let inbox = self.inboxes.entry(sign_id).or_insert_with(|| {
+            Subscriber::unsubscribed_with_capacity(
+                SIGN_POSIT_INBOX_LABEL,
+                crate::protocol::message::POSIT_INBOX_CHANNEL_SIZE,
+            )
+        });
         let _ = inbox
             .send(SignTaskMessage::PositMessage {
                 presignature_id,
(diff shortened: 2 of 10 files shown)


chain-signatures/node/src/protocol/signature.rs
```diff
@@ -1708,10 +1708,12 @@ impl SignatureSpawner {
...
-        let inbox = self
-            .inboxes
-            .entry(sign_id)
-            .or_insert_with(|| Subscriber::unsubscribed(SIGN_POSIT_INBOX_LABEL));
+        let inbox = self.inboxes.entry(sign_id).or_insert_with(|| {
+            Subscriber::unsubscribed_with_capacity(
+                SIGN_POSIT_INBOX_LABEL,
+                crate::protocol::message::POSIT_INBOX_CHANNEL_SIZE,
+            )
+        });
```
jakmeier, line 1716: I think this is no inconsistent with SignatureSpawner::spawn_task, which also can create the same inbox.

chain-signatures/node/src/protocol/message/sub.rs
```diff
@@ -18,6 +18,13 @@ use crate::util::channel_len;
 /// This should be enough to hold a few messages in the inbox.
 pub const MAX_MESSAGE_SUB_CHANNEL_SIZE: usize = 4 * 1024;
 
+/// Small under test-feature so dead-letter inboxes fill quickly in clog tests.
+pub const POSIT_INBOX_CHANNEL_SIZE: usize = if cfg!(feature = "test-feature") {
+    4
+} else {
+    MAX_MESSAGE_SUB_CHANNEL_SIZE
+};
+
```
jakmeier, line 27: I wonder if we should apply this more broadly. Maybe set `MAX_MESSAGE_SUB_CHANNEL_SIZE` to 4, instead of making a special case for posit inboxes.

integration-tests/src/mpc_fixture/mock_chain.rs
```diff
@@ -0,0 +1,110 @@
...
+            return;
+        }
+        self.inner.lock().await.distribute_events(&events).await;
+    }
+
+    fn rpc_action_to_events(action: &RpcAction) -> Vec<ChainEvent> {
+        use elliptic_curve::sec1::ToEncodedPoint;
+        use mpc_node::protocol::SignKind;
+        use mpc_primitives::Chain;
+        use solana_sdk::pubkey::Pubkey;
```
jakmeier, line 65: nit(style): usually we group most imports at the top of the file

integration-tests/src/mpc_fixture/mock_chain.rs
```diff
@@ -0,0 +1,110 @@
...
+
+        let RpcAction::Publish(publish_action) = action;
+
+        if publish_action.indexed.chain != Chain::Solana {
+            return vec![];
+        }
```
jakmeier, line 71: Why only handle this for Solana?

Should we log a warning that other chains aren't supported?

integration-tests/src/mpc_fixture/mock_chain.rs
```diff
@@ -0,0 +1,110 @@
...
+}
+
+impl MockChainInner {
+    async fn distribute_events(&mut self, events: &[ChainEvent]) {
+        for (i, stream) in self.node_streams.iter().enumerate() {
+            let filtered: Vec<ChainEvent> = events
+                .iter()
+                .filter(|event| {
+                    if let Some(filter) = self.filters[i].as_mut() {
+                        matches!(filter(event), EventDelivery::Deliver)
```
jakmeier, line 98: nit: Personally I would prefer an exhaustive match here. In case another delivery variant is added later (e.g. `EventDelivery::Random`), the compiler will force us to think about this case.

integration-tests/tests/cases/mpc_with_stream.rs
```diff
@@ -152,3 +152,282 @@ async fn test_channel_contention_1M_requests() {
...
+async fn run_stale_task_test(drop_respond_event: bool) {
+    use cait_sith::protocol::Participant;
+    use integration_tests::mpc_fixture::message_collector::CollectMessages;
+    use integration_tests::mpc_fixture::mock_chain::EventDelivery;
+    use mpc_node::protocol::message::{PositProtocolId, SendMessage};
+    use mpc_node::protocol::Message;
+    use mpc_node::stream::ChainEvent;
+    use std::collections::HashMap;
+    use std::sync::Arc;
+    use tokio::sync::Mutex;
```
jakmeier, line 178: nit(style): usually we group most imports at the top of the file

jakmeier (approved):
Very nice! Great finding and well implemented in a test.

As we add more such tests, at some point we probably should refactor them and figure out how to make them less verbose. But for now I think this works great.

> /// A missed respond event leaves a stale task that clogs other nodes' inboxes.

I'm not quite sure yet what the correct behavior would look in the case of missed respond events. If a node doesn't see a response, it has to keep trying to get the response to be generated. But of course, it shouldn't clog other node's inboxes 😅 Fixing that should be our first priority.

ppca:
for the sake of not waiting on another review, I'll merge this and address the comments in another PR. issue open: https://github.com/sig-net/mpc/pull/841

PR #846, opened 2026-06-03
more buckets for TASK_QUEUE_CAPACITY and add inbox count metric

ppca:
- Currently we do not know if the inbox capacities are decreasing until they are under 2000. Added more buckets between 2000 and 4096
- add inbox count per channel so we know if number of inboxes are increasing, which may lead to ram decreasing

### DIFF
--- a/chain-signatures/node/src/metrics/messaging.rs
+++ b/chain-signatures/node/src/metrics/messaging.rs
@@ -109,7 +109,7 @@ pub(crate) static TASK_QUEUE_CAPACITY: LazyLock<HistogramVec> = LazyLock::new(||
         "Distribution of queue capacities across message channels",
         &["channel"],
         Some(vec![
-            0.0, 1.0, 10.0, 50.0, 100.0, 200.0, 500.0, 1000.0, 2000.0, 2500.0, 3000.0, 3500.0,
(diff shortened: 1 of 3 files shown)


chain-signatures/node/src/metrics/messaging.rs
```diff
@@ -109,7 +109,8 @@ pub(crate) static TASK_QUEUE_CAPACITY: LazyLock<HistogramVec> = LazyLock::new(||
         "Distribution of queue capacities across message channels",
         &["channel"],
         Some(vec![
-            0.0, 1.0, 10.0, 50.0, 100.0, 200.0, 500.0, 1000.0, 2000.0, 4096.0,
+            0.0, 1.0, 10.0, 50.0, 100.0, 500.0, 1000.0, 1500.0, 2000.0, 2500.0, 3000.0, 3500.0,
+            4000.0, 4096.0,
```
jakmeier, line 113: In my local testing I did some time ago, I think I only saw capacity going down 1 or 2, so even the 4000 vs 4096 might not show up as a different bucket.

But that's probably fine. Keep in mind that for a [histogram](https://prometheus.io/docs/concepts/metric_types/#histogram) you can always look at the total, which is a sum of all observed values without rounding them to buckets first. Divide it by the count and you get an accurate average.

If we wanted to have an accurate median (or the 90th percentile), we should use a [summary metric](https://prometheus.io/docs/concepts/metric_types/#summary) instead.

jakmeier (approved):

Issue #785, opened 2026-04-30
Generate mpc signature for consensus on checkpoint


ChaoticTempest:
Talked offline with @jakmeier and @volovyks so making this issue to track what discussed.

We should reach consensus on a specific checkpoint by being able to have the MPC network generate a signature on it.

How this is achieved should be weighed carefully. New mechanism for doing this or rely on previous mechanisms:
- sign request on the existing chain to have this flow?
- internal node endpoints that would trigger this sign?
- stream/backlog based where when we create a new checkpoint, we send a new sign task and have the posit flow work with this?

### Issues

What happens when we have a checkpoint consensus on a checkpoint that doesn't match our node's local checkpoint for that specific block height? Should we regress to that checkpoint? And if so then should we catchup again?

volovyks:
stale: implemented as voting instead ([#1046](https://github.com/sig-net/mpc/issues/1046)): vote_checkpoint at [lib.rs:912-955](vscode-webview://0djktbu94qo63qp3922bpiojkt7vdclturpfabdecfv329a3tr3l/chain-signatures/contract/src/lib.rs#L912-L955).

PR #847, opened 2026-06-04
feat: checkpoint consensus

ChaoticTempest:
implements https://github.com/sig-net/mpc/issues/785 with Checkpoint consensus.

This is a lot of changes but they really cannot be done separately:
* New `ChainPipeline` that replaces `catchup_then_livestream` where there are different modes to transition to.
  * allows us to cycle back to recovery in cases of regression
* `ChainStream { Recovery, Catchup, Live }` that will handle each case properly.
* MPC sign tasks now generates signatures via `Sign::Checkpoint` which is created whenever a checkpoint is made through block intevals.
* near contract changes:
  * Latest checkpoint digests per chain are stored on the near contract (requires migration for testnet and mainet, devnet is handle automatically via CI).
  * new `read` function which batches `state`, `config` and `checkpoints`.
  * `respond_checkpoint` which is the equivalent to `respond` but just purely for checkpoints.
* removes checkpoint selection in favor of grabbing the latest checkpoint digest from the governance chain.
* new checkpoint regression, where on checkpoint divergence from the checkpoint the node tracks, we regress to that consensus checkpoint.


chain-signatures/contract/src/lib.rs
```diff
@@ -733,6 +790,76 @@ impl VersionedMpcContract {
...
+            views.push(view);
+        }
+
+        views
+    }
+
+    #[handle_result]
+    pub fn respond_checkpoint(
+        &mut self,
+        checkpoint: ConsensusCheckpoint,
```
jakmeier: ~~What happens if `ConsensusCheckpoint` is too large? Transactions with a total size above a threshold (I think 1.5MB or 2MB) are rejected on Near.~~

Ah this is only the hash and chain id. Maybe rename it to `ConsensusCheckpointDigest` or something but not critical.

chain-signatures/contract/src/lib.rs
```diff
@@ -733,6 +790,76 @@ impl VersionedMpcContract {
...
+        }
+
+        self.mutable_checkpoints().insert(
+            checkpoint.chain,
+            SignedCheckpoint {
+                checkpoint,
+                signature,
+            },
+        );
+        Ok(())
```
jakmeier, line 862: Will `latest_checkpoints` just grow forever? Shouldn't we have a sliding window or something to keep the state size bounded?

chain-signatures/node/src/backlog/consensus.rs
```diff
@@ -0,0 +1,226 @@
...
+        Ok(None) => {
+            tracing::info!(?source_chain, "no local checkpoint found");
+        }
+        Err(err) => {
+            tracing::warn!(?source_chain, %err, "failed to load local checkpoint");
+        }
+    }
+
+    // Align with consensus
+    let _ = align_backlog_with_consensus(
```
jakmeier: Why assign to `_`? I usually only see this for a `Result` that should be explicitly ignored. But this function only returns an `Option`.

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -127,13 +127,35 @@ impl IndexedSignRequest {
...
+    pub fn checkpoint(checkpoint: mpc_primitives::ConsensusCheckpoint) -> Self {
+        let payload = checkpoint.sign_payload_scalar();
+        let epsilon = k256::Scalar::ZERO;
+        let entropy = [0u8; 32];
+        let id = SignId::checkpoint(&checkpoint.sign_payload_hash());
+        let args = SignArgs {
+            entropy,
+            epsilon,
+            payload,
+            path: "checkpoint".into(),
```
jakmeier: hm, would it make sense to discuss the design decision of the path with the team briefly? Maybe we want a common prefix for all "system" paths.

chain-signatures/node/src/cli.rs
```diff
@@ -348,6 +350,19 @@ pub async fn run(cmd: Cli) -> anyhow::Result<()> {
...
+            };
+            let mut checkpoints_tx: HashMap<Chain, watch::Sender<CheckpointDigest>> =
+                HashMap::new();
+            let mut checkpoints_rx: HashMap<Chain, watch::Receiver<CheckpointDigest>> =
+                HashMap::new();
+            for chain in Chain::iter() {
+                let (tx, rx) = watch::channel(default_digest.clone());
+                checkpoints_tx.insert(chain, tx);
+                checkpoints_rx.insert(chain, rx);
+            }
```
jakmeier: side note: I kinda like [enum_map](https://crates.io/crates/enum-map) for such use cases. It can statically guarantee that every enum variant has an entry and its macros make initialization a bit more straight forward. (Of course, avoiding another dependency and keeping the code simpler by keeping it to std collections is also nice.)

chain-signatures/primitives/src/lib.rs
```diff
@@ -382,18 +391,86 @@ impl fmt::Debug for PendingTx {
 #[borsh(crate = "near_sdk::borsh")]
 pub struct Checkpoint {
     pub chain: Chain,
-    pub block_height: u64,
+    pub height: u64,
```
jakmeier: This rename breaks the json serialized format persisted on redis. Please add `#[serde(alias = "block_height")]`, undo the renaming, or bump the checkpoint version.

chain-signatures/node/src/stream/pipeline.rs
```diff
@@ -0,0 +1,252 @@
...
+                }
+                new_state = wait_state_change_or_regression(
+                    &mut self.state_rx,
+                    &mut self.checkpoints_rx,
+                    &self.state_tx,
+                    &self.backlog,
+                    chain,
+                    current_state,
+                ) => {
+                    return Some(new_state);
```
jakmeier, line 195: With this return here, aren't we missing a `notify_catchup_completed` call to signal CatchupCompleted?

chain-signatures/node/src/backlog/consensus.rs
```diff
@@ -0,0 +1,226 @@
...
+    // Ignore the default zero-digest (no consensus checkpoint observed yet).
+    if checkpoint_digest.digest == [0u8; 32] {
+        return None;
+    }
+
+    // No mismatch/divergence, we are aligned with the consensus.
+    let current_checkpoint = backlog.checkpoint(chain).await;
+    if current_checkpoint.digest() == checkpoint_digest.digest {
+        return None;
+    }
```
jakmeier, line 34: I believe this is buggy.

The digest only commits to the pending requests set, not the height or the chain.

Consequently, the idle state represented by an empty set of pending requests will be the same on all chains and on different heights.

Say a restarting node has a checkpoint at Solana height=1000 with an empty pending set.
The consensus checkpoint is at height 2000, also with an empty set.
This code would conclude that `current_checkpoint.digest() == checkpoint_digest.digest` and return early, missing that we were 1000 blocks behind.

Perhaps it would be better to include the chain and the height in the digest?

chain-signatures/node/src/stream/ops.rs
```diff
@@ -659,6 +641,38 @@ pub async fn process_execution_confirmed(
...
+    let consensus_checkpoint = mpc_primitives::ConsensusCheckpoint {
+        chain,
+        height: checkpoint.height,
+        digest,
+    };
+    let indexed = IndexedSignRequest::checkpoint(consensus_checkpoint);
+    let sign = Sign::Checkpoint(indexed);
+    if let Err(err) = sign_tx.send(sign).await {
+        tracing::error!(?err, %chain, "failed to enqueue checkpoint sign request");
+    }
```
jakmeier: Shouldn't we check if `caught_up = false`? I think we might start signing old checkpoints again.

chain-signatures/node/src/backlog/consensus.rs
```diff
@@ -0,0 +1,226 @@
...
+        node_client,
+        chain,
+        checkpoint_digest.digest,
+        checkpoints_rx,
+        my_account_id,
+    )
+    .await?;
+
+    let height = fetched_checkpoint.height;
+    if let Err(err) = backlog.recover_by_checkpoint(fetched_checkpoint).await {
```
jakmeier, line 69: I don't see a check if `fetched_checkpoint.height` is higher than the locally recovered checkpoint? Is it possible that this undoes local progress and goes back to an older state?

jakmeier (commented):
Nice to see this progress!

I gave my best to review it thoroughly but it is possible some of my comments are just due to a lack of understanding on my part. Please let me know where my understanding is wrong if this is the case.

Also, I wonder what the benefits are from storing aligned checkpoint hashes on chain. What would be lost if we instead directly query the latest checkpoints from peers?

Given we still need to query the actual checkpoint from peers, having the hashes stored on chain doesn't help against a case where all nodes lose their storage, right? I think we also don't get better alignment properties, since a majority has to sign the same checkpoint anyway, even before it is stored on chain.

chain-signatures/contract/src/lib.rs
ChaoticTempest: `latest_checkpoints` is an `IterableMap<Chain, SignedCheckpoint>`, where the key here is the chain so it'll just overwrite

chain-signatures/node/src/backlog/consensus.rs
ChaoticTempest: will remove

chain-signatures/node/src/protocol/signature.rs
ChaoticTempest: yep, whatever "system" path we want works. @volovyks @ppca what do you guys think would be a good system path

chain-signatures/node/src/stream/pipeline.rs
ChaoticTempest: nope, normally when we have a new transition like this, it would mean that we're going to transition to prior state like Recovery. I probably should just get rid of checking against state_rx in that function though, since state_tx is only used within the pipeline anyways

chain-signatures/node/src/backlog/consensus.rs
ChaoticTempest: nice catch, will include it in the digest

chain-signatures/node/src/cli.rs
ChaoticTempest: I'll take a look at enum_map and see if it doesn't add too much dependencies

chain-signatures/node/src/stream/ops.rs
ChaoticTempest: true, I'll add it

chain-signatures/node/src/backlog/consensus.rs
ChaoticTempest: yeah I think I need to add persistence of all historical checkpoints so we can validate the consensus checkpoint is in our history if our latest local checkpoint is higher than the consensus checkpoint.

In the case that it is in our history and digests matches, we don't overwrite. But if it doesn't match then we should overwite and start from consensus checkpoint

chain-signatures/node/src/cli.rs
ChaoticTempest: added enum_map

chain-signatures/crypto/src/kdf.rs
```diff
@@ -48,6 +48,16 @@ fn keccak(derivation_path: impl AsRef<[u8]>) -> Scalar {
     Scalar::from_non_biased(hash)
 }
 
+pub fn derive_epsilon_checkpoint(
```
volovyks: There are a couple of important cryptography-related decisions inside and outside of this function. Let's move them inside of this function and make them more explicit. Specifically:
- height used as derivation path, once extra let where we say that `path = height` would be great
- sender should be hardcoded in crypto crate (const SENDER: &str = "checkpoint|sender";). No need to pass it around. Also helps to make it clearer.

chain-signatures/crypto/src/kdf.rs
```diff
@@ -48,6 +48,16 @@ fn keccak(derivation_path: impl AsRef<[u8]>) -> Scalar {
...
+    key_version: u32,
+    chain: Chain,
+    sender: &str,
+    height: u64,
+) -> Scalar {
+    let derivation_path = derivation_path(key_version, chain, sender, &height.to_string());
```
volovyks: Hm.. I do not think that using `height` as `path` is what we need here. It means that nodes will use a different key for each checkpoint. That is not a problem by itself, but intuitively, I expect it all to be signed by a single "checkpoint" key. It means we can hardcode both the user and the path in this function.

Also, I do not understand how the signature check in the contract works, since it is checking against the root key. Maybe I'm losing something here, but if not, let's add a test that covers this.

integration-tests/tests/cases/store.rs
```diff
@@ -299,3 +299,75 @@ async fn test_presignature_persistence() -> anyhow::Result<()> {
...
+        pending_requests: vec![],
+    };
+    storage.persist(&cp1).await?;
+
+    // 3. Verify latest and history
+    let latest = storage.load_latest(Chain::Solana).await?.unwrap();
+    assert_eq!(latest.height, 10);
+    let history = storage.load_history(Chain::Solana).await?;
+    assert_eq!(history.len(), 1);
+    assert_eq!(history[0].height, 10);
```
volovyks: It would be great to have requests in the checkpoint and make sure they are stored correctly

chain-signatures/contract/src/lib.rs
jakmeier: Oh I see, it only stores one checkpoint per height. Somehow I completely misread the code

chain-signatures/node/src/backlog/consensus.rs
jakmeier: Oh I see, this is more complicated than just a height comparison.

Since this is more of an optimization, maybe this is better suited as a follow up.

integration-tests/tests/cases/store.rs
```diff
@@ -299,3 +299,75 @@ async fn test_presignature_persistence() -> anyhow::Result<()> {
...
+    assert_eq!(heights, vec![10, 20]);
+
+    // 6. Test expiration/pruning: modify score of cp1 to be expired (e.g. now - 2000s)
+    let mut conn = pool.get().await?;
+    let history_key = format!(
+        "party0.near:checkpoint:history:{}:Solana",
+        mpc_node::storage::checkpoint_storage::CHECKPOINT_VERSION
+    );
+    let cp1_str = serde_json::to_string(&cp1)?;
+    let expired_score = mpc_node::util::current_unix_timestamp() - 2000;
```
volovyks: Is a 30m window enough? What if the system goes offline for an hour? Then all the history will be cleaned up.
I was not expecting history at all on the first iteration, since we have signatures and on-chain storage of hash-signature information. It should be sufficient for a single node to hold the system checkpoint for recovery on all nodes. But if we are not sure that happens, then the history windows should be bigger.

Also, some chains' finality is > 30m. There would be no history at all.

chain-signatures/node/src/backlog/consensus.rs
jakmeier: never mind, I see you've already implemented it :D

integration-tests/tests/cases/mpc_with_stream.rs
```diff
@@ -39,10 +39,29 @@ async fn test_sign() {
         .await;
 
     let timeout = Duration::from_secs(10);
-    let actions = network.assert_actions(1, timeout).await;
-
-    assert_eq!(actions.len(), 1);
-    let action_str = actions.iter().next().unwrap();
+    let start = std::time::Instant::now();
+    let actions = loop {
```
volovyks, line 43: Why do we need this change?

chain-signatures/node/src/backlog/consensus.rs
```diff
@@ -0,0 +1,258 @@
...
+    }
+
+    // If our current height is greater than consensus height,
+    // check if we have a historical checkpoint that matches the consensus digest.
+    if current_checkpoint.height > checkpoint_digest.height {
+        if let Some(historical) = backlog
+            .find_checkpoint_by_digest(chain, checkpoint_digest.digest)
+            .await
+        {
+            if historical.height == checkpoint_digest.height {
```
jakmeier: nit: this if condition is no longer needed now that the digest includes the hieght

jakmeier (approved):
Looks like all important comments have been addressed, I'll approve this now.

There are still 2 open comments / discussions, which I think could be resolved in a follow up, they don't necessarily need to be discussed in the context of the full PR.

chain-signatures/primitives/src/lib.rs
```diff
@@ -93,6 +94,17 @@ impl SignId {
         let request_id: [u8; 32] = hasher.finalize().into();
         Self { request_id }
     }
+
+    pub fn checkpoint(chain: Chain, height: u64, payload: &[u8; 32]) -> Self {
```
volovyks: Having a "checkpoint" function under SignId looks a bit strange. Oh, I got it, this is another constructor. For consistency, should it be "from_checkpoint"?

chain-signatures/node/src/indexer_sol.rs
```diff
@@ -757,6 +757,29 @@ async fn subscribe_to_program_events(
 
     let (mut stream, _unsubscriber) = pubsub_client.logs_subscribe(filter, config).await?;
 
+    // The WS subscription is now live. Fetch the current confirmed slot via RPC as the
+    // anchor: this is the correct boundary because the subscription is already buffering
+    // all new events, so catchup can safely cover [persisted_block, anchor) via RPC history
+    // with no gaps. We do not wait for the first WS event because that could deadlock if
+    // no program-mentioning transactions arrive (e.g. in tests after a single sign call).
+    if let Some(anchor_tx) = anchor_tx.take() {
```
volovyks, line 637: Is this somehow relevant to this PR?

chain-signatures/node/src/indexer_canton/stream.rs
```diff
@@ -290,12 +290,34 @@ impl CantonIndexer {
         }
 
         loop {
-            let Some(update) = self.next_update().await else {
-                anyhow::bail!("canton WebSocket closed during catchup; reconnecting");
-            };
-            let offset = self.process_update(&update).await?;
-            if offset >= target_offset {
-                return Ok(());
+            // Try to get the next update with a conservative timeout during catchup.
```
volovyks, line 307: Are these changes related to checkpoints?

chain-signatures/node/src/storage/checkpoint_storage.rs
```diff
@@ -75,7 +115,98 @@ impl CheckpointStorage {
...
+
+        // 1. Clean storage returns None / empty history
+        assert!(storage.load_latest(Chain::Solana).await?.is_none());
+        assert!(storage.load_history(Chain::Solana).await?.is_empty());
+
+        // 2. Persist first checkpoint
+        let cp1 = Checkpoint {
+            chain: Chain::Solana,
+            height: 10,
+            pending_requests: vec![],
```
volovyks, line 182: Adding a couple of requests can be a good idea

chain-signatures/node/src/backlog/consensus.rs
```diff
@@ -0,0 +1,258 @@
...
+    contract_watcher: &mut ContractStateWatcher,
+    mesh_state: &mut watch::Receiver<MeshState>,
+    node_client: &NodeClient,
+    source_chain: Chain,
+    checkpoints_rx: &mut watch::Receiver<CheckpointDigest>,
+) {
+    // Recover backlog before doing anything.
+    // Wait for threshold to be available
+    let threshold = contract_watcher.wait_threshold().await;
+    if threshold == 0 {
```
volovyks: Threshold cannot be 0

volovyks (commented):
Overall, this looks good. I like the contract-related part and the consensus layer. The history component makes things a bit more complicated, though. Not sure if it's required.

A metric for the latest checkpoint height would be very useful for detecting issues.

chain-signatures/node/src/protocol/signature.rs
volovyks: Epsilon is using something else for "path"; I've left a comment. A single path for all checkpoint signatures seems correct, but I would create a constant and use it across the project.

integration-tests/tests/cases/mpc_with_stream.rs
ChaoticTempest: because there is a new Sign::Checkpoint that we need to ignore now

chain-signatures/node/src/indexer_sol.rs
ChaoticTempest: yes, our tests caught this particular case where the contract does not have any transactions thus stalling our websocket stream and never giving us the initial anchor height

chain-signatures/node/src/indexer_canton/stream.rs
ChaoticTempest: yep, related to one of the other comments I mentioned about stalling websocket if there's no transactions. This is because we're only listening for the specific contract

chain-signatures/crypto/src/kdf.rs
ChaoticTempest: it doesn't hurt to have `height` as path since we don't store any sort of derived key anyways. It does enhance the security here where we now we have a new key per checkpoint, so an attacker would have to compromise the next derived key at a particular height to submit a checkpoint for consensus

chain-signatures/crypto/src/kdf.rs
ChaoticTempest: consolidated them with your recent changes

integration-tests/tests/cases/store.rs
ChaoticTempest: that's a good point for the 30m window. I'll change it to be more of the most recent 32 checkpoints. That sounds like a good number.

> I was not expecting history at all on the first iteration, since we have signatures and on-chain storage of hash-signature information. It should be sufficient for a single node to hold the system checkpoint for recovery on all nodes.

It's required for sanity checks in the case that a node is ahead in their indexer checkpoint. The latest stored checkpoint is sufficient for recovery, not sufficient for regression.

ChaoticTempest:
@volovyks let me know if you have any other thoughts on this one, otherwise I think we should be good to merge

volovyks (approved):

ChaoticTempest:
we aren't aborting sign tasks that get orphaned with this PR though. This issue has existed before where SignTasks exist on one node, while another does not have it. So rather than snowball this PR, I will just leave it up to a separate PR to more thoroughly handle this case in general

PR #851, opened 2026-06-05
chore: cleanup transient types that are the same across chains

ChaoticTempest:
removes a couple types over multiple chain that are a bit verbose. These don't break serialization so it's good to go

### DIFF
--- a/chain-signatures/node/src/stream/ops.rs
+++ b/chain-signatures/node/src/stream/ops.rs
@@ -1,11 +1,7 @@
 use crate::backlog::Backlog;
-use crate::indexer_canton::{
-    CantonRespondBidirectionalEvent, CantonSignBidirectionalRequestedEvent,
-    CantonSignatureRespondedEvent,
-};
+use crate::indexer_canton::{CantonRespondBidirectionalEvent, CantonSignatureRespondedEvent};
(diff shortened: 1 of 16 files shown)


chain-signatures/node/src/stream/ops.rs
```diff
@@ -16,262 +9,140 @@ use crate::rpc::{ContractStateWatcher, RpcChannel};
...
-            SignatureRespondedEvent::Solana(event) => event.request_id,
-            SignatureRespondedEvent::Hydration(event) => event.request_id,
-            SignatureRespondedEvent::Ethereum(event) => event.request_id,
-            SignatureRespondedEvent::Canton(event) => event.request_id,
-        }
+        self.request_id
     }
 
-    /// Convert the contained event into an `mpc_primitives::Signature`.
     pub fn signature(&self) -> Signature {
```
volovyks: There are many places where access functions no longer seem necessary.

volovyks (dismissed):
Good simplification!

chain-signatures/node/src/stream/ops.rs
ChaoticTempest: yep, got rid of them

volovyks (approved):

Issue #347, opened 2025-04-30
Protect secrets with `secrecy` crate


volovyks:
### Description
Crate: https://crates.io/crates/secrecy
We can do our best protecting secrets using SecretManager or other solutions and then incidentally log or expose it other way. This crate is intended to prevent such mistakes.

Specifically I'm worried about tracing instrumentation. It logs all the function parameters by default.

volovyks:
Superseded by #1269, which tracks the full in-memory secrets policy (doc/SECRET_MANAGEMENT.md) including the `secrecy` wrapping proposed here.

PR #862, opened 2026-06-09
fix: prevent `eth_account_sk` leakage to stdout

isSerge:
I noticed there is a risk of accidental leak of `eth_account_sk` to stdout (If Clap fails to parse an argument or someone logs the `Cli` or `Config` structs).
This PR adds `secrecy` crate and introduces `SecretString` for `eth_account_sk`. Plus few other minor secret-related enhancements

### DIFF
--- a/chain-signatures/primitives/src/lib.rs
+++ b/chain-signatures/primitives/src/lib.rs
@@ -110,7 +110,7 @@ pub struct SignArgs {
 impl std::fmt::Debug for SignArgs {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("SignArgs")
-            .field("entropy", &hex::encode(self.entropy))
+            .field("entropy", &hex::encode(&self.entropy[..4])) // not a secret atm, but better truncate for readability and logging safety in the future (if it becomes one)
             .field("epsilon", &self.epsilon)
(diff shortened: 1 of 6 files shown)


chain-signatures/primitives/src/lib.rs
```diff
@@ -110,7 +110,7 @@ pub struct SignArgs {
 impl std::fmt::Debug for SignArgs {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("SignArgs")
-            .field("entropy", &hex::encode(self.entropy))
+            .field("entropy", &hex::encode(&self.entropy[..4])) // not a secret atm, but better truncate for readability and logging safety in the future (if it becomes one)
```
volovyks, line 113: Entropy always comes from the chain, so it's public by default. But no harm in truncating.

volovyks (approved):
Good move. I was postponing [this](https://github.com/sig-net/mpc/issues/347) for a long time.

This PR does fix the mentioned issue, but we will need to revisit it to better protect other secrets, especially root key shares. We can be more strict in places like [this](https://github.com/sig-net/mpc/pull/862/changes#diff-7920121bf12ccd32f2c7c205c7a6ef92c36b49c39e9d5779f0966f33ede9f856R301). There are several ways, beyond Debug, that could introduce problems.

PR #865, opened 2026-06-10
Revert "fix: backpressure drop messages on full"

volovyks:
This fix was discussed many times, and it is not ideal. Let's get rid of it once we have the proper fix.
@ChaoticTempest @ppca, please, let's define a proper way of fixing it.

### DIFF
--- a/chain-signatures/node/src/protocol/message/mod.rs
+++ b/chain-signatures/node/src/protocol/message/mod.rs
@@ -2,6 +2,8 @@ mod filter;
 mod sub;
 mod types;
 
+pub use sub::{Subscriber, POSIT_INBOX_CHANNEL_SIZE};
+
 use super::contract::primitives::{ParticipantMap, Participants};
 use super::presignature::PresignatureId;
 use super::triple::TripleId;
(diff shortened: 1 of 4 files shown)


ChaoticTempest:
So, I'd say the fix for this still adds value. The rest of the system does not stall and we can progress instead of blocking things. Our metrics will still show that our capacity is full in this case, but it doesn't stall everything else. I'd say we should keep it in the mean time. It doesn't hurt us in any way as far as I see

ChaoticTempest:
actually, if anything we can just decrease the posit channel capacity back down to sub 10k

jakmeier:
> So, I'd say the fix for this still adds value. The rest of the system does not stall and we can progress instead of blocking things. Our metrics will still show that our capacity is full in this case, but it doesn't stall everything else. I'd say we should keep it in the mean time.

For now, it still seems useful, yes. But if we implement another way that prevents a full channel from blocking other parts of the system, then I don't really see the value of this fix anymore.

> It doesn't hurt us in any way as far as I see

Nothing immidiately critical but I can think if 2 long term  problems.
- Large channels may cause OOM issues and they potentially make any RAM related investigations trickier.
- Lossy message delivery makes it very hard to argue about liveness guarantees of our system.

 > actually, if anything we can just decrease the posit channel capacity back down to sub 10k

Agreed, that's probably the best to do for now. That gets rid of the first problem I mentioned.

volovyks:
I will get back to this once we have a complete design.

PR #866, opened 2026-06-10
chore: CI optimizations

isSerge:
This PR addresses some of the issues in our CI pipeline

Main changes:

- Stripped Debug Symbols in `cargo.toml` for relevant profiles (reduces cache size)
- Replace generic `WarpBuilds` cache with `Swatinem/rust-cache@v2` (it does a lot of Rust-specific optimizations)
- Prevent PR cache thrashing: PRs will no longer overwrite the baseline cache, which caused cold start
- Force system OpenSSL across all workflows (previously it was building from source, which took extra ~4min)
- Node.js optimizations (`npm ci` + cache)
- Replaced few manual `curl` scripts with GitHub actions
- Update and unify GitHub actions versions across workflows 

Following improvements achieved:

- Cache size: slashed from 14GB to 7GB (cache was previously discarded due to 10 GB GitHub limit)
- Somewhat reduced build time for `Integration` job (the slowest):
  - Current baseline: Component Test ~22 min, Integration Test ~30-32 min
  - New cold start: Component Test ~19min, Integration Test ~28min
  - New warm start: Component Test ~12min, Integration Test ~25min
  
  To get further build time improvements we need to address actual integration tests (will address separately)

### DIFF
--- a/.github/workflows/canton.yml
+++ b/.github/workflows/canton.yml
@@ -19,7 +19,7 @@ jobs:
     runs-on: warp-ubuntu-latest-x64-4x
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v6
 
       - name: Login to GitHub Container Registry
         uses: docker/login-action@v1
@@ -88,7 +88,7 @@ jobs:
     runs-on: warp-ubuntu-latest-x64-4x
 
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v6
 
       - name: Login to GitHub Container Registry
         uses: docker/login-action@v1
--- a/.github/workflows/canton.yml
+++ b/.github/workflows/canton.yml
@@ -22,7 +22,7 @@ jobs:
       - uses: actions/checkout@v6
 
       - name: Login to GitHub Container Registry
-        uses: docker/login-action@v1
+        uses: docker/login-action@v4
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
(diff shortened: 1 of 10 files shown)


.github/workflows/canton.yml
```diff
@@ -55,21 +55,26 @@ jobs:
...
+          cache-dependency-path: chain-signatures/contract-eth/package-lock.json
 
-      - uses: WarpBuilds/cache@v1
+      # Install system OpenSSL explicitly here so the build uses system OpenSSL instead of building OpenSSL from source
+      - name: Install system OpenSSL
+        run: sudo apt-get update && sudo apt-get install -y libssl-dev pkg-config
+      - name: Set OpenSSL to system
+        run: echo "OPENSSL_NO_VENDOR=1" >> $GITHUB_ENV
+
+      - uses: Swatinem/rust-cache@v2
```
ChaoticTempest, line 67: So our github actions runner is on WarpBuilds. Any reason to change to `Swatinem/rust-cache@v2`? AFARIK the warpbuilds cache should be better as long as we're running on WarpBuild

ChaoticTempest (commented):
Mostly looking good to me besides the comment I made

.github/workflows/canton.yml
isSerge: With WarpBuilds cache cargo was still recompiling everything because restore fingerprints did not match. Plus it caches entire `/target` including historical artifacts, which was often more than 10 GB (beyond GitHub limit), `Swatinem` action is Rust-aware, it does clean up before caching.

ChaoticTempest: I see, looks good to me then

ChaoticTempest (approved):

PR #867, opened 2026-06-10
feat: secret storage enhancements

isSerge:
Some changes to `storage/secret_storage.rs` to improve correctness and durability. No public API changes.

1. Replace `SecretNodeStorageBox` and dynamic dispatch with `SecretNodeStorageVariant` enum - not a huge runtime impact, but seems better for type visibility.
2. Update `store` method for `DiskNodeStorage` to use atomic rename (avoid risk of secret key lost or corrupted)
3. Update `load` method for `DiskNodeStorage` to differentiate between `NotFound` and other errors (PermissionDenied, Corrupted JSON, etc.)
4. Implement unit tests + add doc comments

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6577,9 +6577,9 @@ dependencies = [
 
 [[package]]
 name = "libc"
-version = "0.2.172"
+version = "0.2.186"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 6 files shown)


ChaoticTempest (approved):
Even though at the type level, this was still needed. Thanks for picking it up!

PR #850, opened 2026-06-04
fix(sol-indexer): update sdk to parse new formats

jakmeier:
Fixes the bug reported in #844.

This is a minimal Solana RPC client dependencies to get https://github.com/anza-xyz/solana-sdk/pull/410 included.

- solana-client 2.2.7 -> 2.3.13
- solana-sdk 2.2.2 ->2.3.1
- solana-transaction-status 2.2.6 -> 2.3.13

We should consider updating to 4.0 dependencies in the near future.

PR #870, opened 2026-06-11
fix: solana fetch issues on multiple nodes due to format change

ChaoticTempest:
A couple of nodes' solana indexers are stalling on catchup due to the issues that was fixed in https://github.com/sig-net/mpc/pull/850. This can be seen in our logs with:
> "failed to fetch Solana block range for catchup; retrying", with startSlot: 468042241 and endSlot: 468542952 (or higher).

Before the fix landed in, the indexer was paused for a while and did not process events. So it created a gap larger than 500k slots where solana RPC only supports up to 500k slots. This fixes that by batching/chunking the rpc request.

### DIFF
--- a/chain-signatures/node/src/indexer_sol.rs
+++ b/chain-signatures/node/src/indexer_sol.rs
@@ -446,15 +446,10 @@ impl SolanaIndexer {
         Ok(())
     }
 
-    /// Fetches blocks from start_slot to end_slot. The range is inclusive `[start_slot, endslot]`
-    async fn fetch_sparse_blocks(
-        &self,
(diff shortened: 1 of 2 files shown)


ChaoticTempest:
also since we have nodes that were able to do catchup, we have them growing in size in the backlog which is expected. threshold nodes are not able to participate in the signing process. This also makes it so that our signature_init channel gets full due to the active nodes that are live not able to progress and keeps reorganizing with constant propose messages until our channel is full

ppca (approved):
lgtm

jakmeier (approved):

Issue #873, opened 2026-06-11
[Proposed Guideline] Don't use channels as implicit work queues


jakmeier:
We make frequent use of Rust's mpsc channels as communication between tokio tasks. Naturally, this leads to the problem of handling backpressure. 

Here I propose a strategy to avoid blocking backpressure on a mpsc::channel. It boils down to a simple principle that we can gradually apply to our usage of channels.

# Principle: Don't use channels as implicit work queues

If instead we only use channels as pure communication bridges, and not as queues for incoming work, then we avoid backpressure problems entirely. Mpsc channels are ill-suited as work queues because they lack the tools to observe or edit the queued worked items.

In practice, this means mpsc channel consumers must always pull the channel within a short time frame, even if they can't currently process the work. If the consumer doesn't want to handle the work right now, it should put it in an actual work queue and treat messages as it makes sense in the specific context. Sometimes, that means dropping messages. Sometimes, we rather want an unbounded queue of work items that we may put on disk to avoid memory issues. Usually what to do is quite obvious when thinking about a specific case.

Example: The signature posit consumer should poll incoming messages at all times. If the node is not currently in a state to process such events, it can keep the latest posit message per sender in a queue and drop older messages.

```rust
struct SignPositWorkQueue {
    messages: HashMap<Participant, SignTaskMessage>,
}

let posit_queue = SignPositWorkQueue::new();
tokio::spawn(async move {
    while let Some(msg) = rx.recv().await {
        posit_queue.push(msg);
    }
});

fn push(&self, msg: SignTaskMessage) {
    if msg.round >= self.messages[msg.from].round {
        // replace buffered message
        self.messages[msg.from] = msg;
    }
}
```

The actual consumers will now await on new messages from `SignPositWorkQueue` instead of directly pulling a `Receiver<SignTaskMessage>`

This is better than the general two options we have with mpsc channels.

- `channel(BUFFER_SIZE)` blocks the producer and may lead to a locked up system as we had in #854
- `unbounded_channel()` has unbounded memory requirement

ChaoticTempest:
[...]

Issue #864, opened 2026-06-10
On channel full should drop oldest instead of newest


ChaoticTempest:
In https://github.com/sig-net/mpc/pull/788, we mitigate backpressure by dropping messages on full which drops the newest messages from entering the channel. This is not great since older messages are more likely to be invalid than newer ones.

We need to investigate the best way of handling this. It's not necessary to drop the newest message but rather have some sort of filter so we can only maintain messages that are necessary.

Alternatives to this are mentioned in https://github.com/sig-net/mpc/issues/854#issuecomment-4656890536

volovyks:
I'm against dropping messages overall and hope that #788 is [reverted](https://github.com/sig-net/mpc/pull/865).
We need to look for causes, not apply patches and treat symptoms. Dropping messages can help, but it can also hide bad architectural decisions.

For example, the `sign posit` channel has a capacity of ~4000 messages. The posit message is only relevant for ~1 second. Even with 4x redundancy, that is 1000 signature protocols per second, and we are far from processing that volume.

I will look for reasons as well.

jakmeier:
I mostly agree with @volovyks here. I propose a general solution to this kind of problem in #873.

If you like that approach, I also have a PR that implements it for the signature posit inbox specifically ready in #874. I think that would cleanly resolve the issue at hand.

Issue #854, opened 2026-06-08
testnet incident Jun 07: signature generation stop


ppca:
https://signetwork.grafana.net/a/grafana-irm-app/incidents/2

Signature generation stops completely around 07:30 Jun 07.

<img width="1134" height="714" alt="Image" src="https://github.com/user-attachments/assets/726e29a1-0cb0-4d84-b65f-f80e5ecd293a" />

ppca:
signature spawner inbox  `<grafana_link>`was full for node 0 at 3:22, full for node 6 at 3:36, full for node 2 at 5:38:

<img width="1163" height="732" alt="Image" src="https://github.com/user-attachments/assets/f20f4036-dc48-4964-8824-8bf69ece3dca" />

<img width="1138" height="635" alt="Image" src="https://github.com/user-attachments/assets/bed5280e-61f9-4f56-88d7-ad8a74b60636" />

<img width="1147" height="653" alt="Image" src="https://github.com/user-attachments/assets/0bf48cbc-0a05-48bc-b489-60c4922522b7" />

checking the awaiting generation solana indexed requests, these nodes plateau `<grafana_link>` at exactly same timestamp. Also these nodes' signature generator rate `<grafana_link>` becomes 0 at matching timestamps.

After a few hours around 8 - 8:30, node 0, 6 and 2's signature posit channel becomes full `<grafana_link>`, i.e., the indexed requests has not been drained by signature spawner because spawner's inboxes were clogged, so the signature posit channel becomes full over time. We can also see that received messages `<grafana_link>` becomes 0 for these nodes accordingly.

ppca:
The above is aligned with investigation in https://github.com/sig-net/mpc/issues/814. where our theory is: Lingering stale signIds on some nodes keep sending posit messages -> signature posit inbox full -> sigantureSpawner loop blocks (signature generation stop)-> eventually signature_init box of MessageInbox full -> no messages get received anymore (triple/presig generation stop)

One example of such a signId is this `<gcp_link>`:
```jsonPayload.signId="SignId(\"ca2f719227ab9a85d2ea2f9ef810820d7ec1dcdc81cf8f79c2b61a9c95e0ec02\")"```

<img width="1436" height="659" alt="Image" src="https://github.com/user-attachments/assets/d10a250e-8153-44e8-a8ca-ecb70757b0b0" />

ppca:
[...]

PR #874, opened 2026-06-11
feat: always drain signature posit messages

jakmeier:
Avoid cases where a channel for posit messages is not drained because the SignatureTask is in a different state.

This implements the princinple described in #873.

This is a great example because we already had buffered messages forincoming `SignTaskMessage` but we didn't properly decouple the work item queue from the coommunication channel. Following the principl, this PR always drains the communicaion channel and moves the responsibility of a classical work item queue to struct SignPositWorkQueue.

This is a proposed solution for the problem discovered in #854. It is a third alternative to already proposed solutions in #864 and [this comment](https://github.com/sig-net/mpc/issues/854#issuecomment-4656890536)-

### DIFF
--- a/chain-signatures/node/src/protocol/signature.rs
+++ b/chain-signatures/node/src/protocol/signature.rs
@@ -34,11 +34,12 @@ use mpc_primitives::{SignArgs, SignId};
 use rand::rngs::StdRng;
 use rand::seq::IteratorRandom;
 use rand::SeedableRng;
-use std::collections::{BTreeSet, HashMap, VecDeque};
+use std::collections::hash_map::Entry;
+use std::collections::{BTreeSet, HashMap};
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, RwLock};
 use std::time::{Duration, Instant};
-use tokio::sync::{mpsc, watch, OwnedSemaphorePermit, Semaphore};
+use tokio::sync::{mpsc, watch, Notify, OwnedSemaphorePermit, Semaphore};
 use tokio::task::JoinHandle;
 
 /// The round interval to search for a proposer in the organizing phase.
@@ -272,8 +273,8 @@ struct SignState {
     /// would not have sent a posit message before the proposer proposes.
     ///
     /// INVARIANT: All messages stored here are for `highest_seen_round`. Must
-    /// be cleared when `highest_seen_round` changes.
-    buffered_messages: VecDeque<SignTaskMessage>,
+    /// be cleared when `highest_seen_round` changes. One slot per sender.
+    buffered_messages: HashMap<Participant, SignTaskMessage>,
     /// When Some, another group is already generating this signature.
     /// The timestamp is when proposing can be resumed.
     pause_proposing: Option<std::time::Instant>,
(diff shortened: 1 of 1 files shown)


jakmeier:
@ChaoticTempest what do you think of this kind of approach?

ChaoticTempest:
overall I think this helps but it doesn't solve the issue of messages at rest (i.e. no associated SignTask to process the channel of messages in inboxes). That's why I think https://github.com/sig-net/mpc/issues/854#issuecomment-4656890536 is the best way moving forward for signatures

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -272,8 +273,8 @@ struct SignState {
     /// would not have sent a posit message before the proposer proposes.
     ///
     /// INVARIANT: All messages stored here are for `highest_seen_round`. Must
-    /// be cleared when `highest_seen_round` changes.
-    buffered_messages: VecDeque<SignTaskMessage>,
+    /// be cleared when `highest_seen_round` changes. One slot per sender.
+    buffered_messages: HashMap<Participant, SignTaskMessage>,
```
volovyks, line 277: A single participant can try to initiate multiple signature protocols. We may want to have a `HashMap<(Participant, SignId), SignTaskMessage>` to allow parallel protocol creation.

volovyks: I misunderstood. I thought that this was the SignatureSpawner state, not SignTask state. That is why I proposed to have `HashMap<(Participant, SignId), SignTaskMessage>` message buffer. I still think we should have one, but on the SignatureSpawner level.

From what I understand, the main problem we have is that some of the required SignTasks are not spawned, since the request was not indexed. As a result, the messages are stuck forever on the SignatureSpawner mpsc channel.
cc @ChaoticTempest

ChaoticTempest: yep, we need to handle this more deliberately at the SignatureSpawner level. Our SignatureSpawner is already inherently a work queue that preprocesses the messages and forwards them

ppca (approved):
LGTM! this means we will never clog inbox for a certain sign_id, and we will not stop processing sign requests because of this. Can help with memory too since we are draining the inbox continuously, and work queue by design can only be as big as count of participants per sign id, versus previously inbox can keep growing for sign_ids, which becomes much more bloated as rounds++.
No sign task dealing with a sign id is a separate problem. More related to our backlog/indexer consistency, we should address separately.

jakmeier:
> overall I think this helps but it doesn't solve the issue of messages at rest (i.e. no associated SignTask to process the channel of messages in inboxes). That's why I think [#854 (comment)](https://github.com/sig-net/mpc/issues/854#issuecomment-4656890536) is the best way moving forward for signatures

You are talking about Subscriber::Unsubscribed, right? When a channel has been created but not SignTask has subscribed to it yet. Yeah my PR doesn't really solve that.

Maybe merging the posit and generating inboxes would indeed be the right change to address that specifically.

Generally speaking, I honestly think how the `Subscriber` abstraction deals with unsubscribed channels right now hurts us. It hides the exact details from the sender that the sender should be dealing with.

I would rather have `Unsubscribed` replaced by an small limited buffer _outside_ the channel. Senders are then forced to deal with overflowing work but they can also inspect already queued messages.

volovyks:
@ChaoticTempest, how is your investigation? Can we merge this one?

ChaoticTempest:
yep, we can merge this one, and go from here

ChaoticTempest (approved):

volovyks:
@jakmeier please, merge `develop` to fix the audit issues

jakmeier:
closing in favor of #949

PR #875, opened 2026-06-11
refactor: remove runner script, make integration tests setup process explicit, add `justfile` for integration tests

isSerge:
Remove the implicit Cargo runner used for integration tests and make artifact generation explicit. Full local integration suite now runs faster (~1360s → ~1044s, ~23% improvement). CI behavior remains the same.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -14,11 +14,6 @@ rustflags = [
   "-C", "link-arg=dynamic_lookup",
 ]
 
-# This is basically a prebuild script used to build the node/contract before running
-# integration-tests.
-[target.'cfg(not(target = "wasm32-unknown-unknown"))']
-runner = "./setup.sh"
-
 [alias]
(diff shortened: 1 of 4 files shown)


ChaoticTempest:
nice to see the speed up for the CI, but does `cargo nextest` invoke the setup.sh script or parts of it? If not, that would be an issue for local development, because a lot of times as you are iterating, you forget to rebuild your artifacts or don't remember if you've changed anything. This setup.sh script was used so that we have parity between what we're testing and what we changed in the node/contract

isSerge:
@ChaoticTempest we could add a `just` or `make` command:
```
integration-test:
    ./setup.sh
    cargo nextest run -p integration-tests
```

ChaoticTempest:
yeah that works too, and hopefully we can pass arguments like `test_signature_basic` or modules so cargo nextest can run specific parts of the integration test

ChaoticTempest (approved):
nicely done!

PR #881, opened 2026-06-14
refactor: make cluster integration tests parallel

isSerge:
Few tweaks to make cluster integration tests run in parallel:

- Make Docker network names unique across test processes
- Use unique network name to isolate temp directories for each test
- Update `nextest.toml` to use `max-threads = "num-cpus"` for cluster tests
- Increase timeout to 90 sec in `test_checkpoint_recovery_after_offline` (original 30 sec is not enough to fully process signatures and clear backlog, probably due to CPU starvation)

Cluster integration tests time reduced from ~28 min to ~15 min

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,5 +1,5 @@
 [test-groups]
-cluster = { max-threads = 1 }
+cluster = { max-threads = "num-cpus" }
 fixture = { max-threads = "num-cpus" }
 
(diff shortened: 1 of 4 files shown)


volovyks (approved):
Imprecive! I have not seen our Integration tests succeed in 15m for a while.
(I will rerun all the tests to make sure there are no race conditions)

PR #884, opened 2026-06-15
refactor: Update Solana indexer to fetch blocks concurrently

isSerge:
- Replace `AsyncCatchupIter` with `futures_util::stream::Stream` in `ChainIndexer` trait. Update indexers accordingly
- Update Solana indexer to fetch catchup blocks concurrently using `futures::stream` API (will update other indexers in separate PRs)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6016,6 +6016,7 @@ dependencies = [
  "ecdsa",
  "elliptic-curve",
  "futures",
+ "futures-util",
  "generic-array",
  "hex",
  "hyper 0.14.32",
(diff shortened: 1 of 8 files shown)


ChaoticTempest (approved):
Looks good changing the AsyncCatchupIter to futures-util. The changes you made with solana will likely be removed in https://github.com/sig-net/mpc/pull/889 though since that deals with the rate limiting

isSerge:
@ChaoticTempest I think we can have both. Do they limit concurrency or request per sec?

isSerge:
If we use Alchemy we could utilize [`governor`](https://docs.rs/governor/latest/governor/)

ChaoticTempest:
>  think we can have both. Do they limit concurrency or request per sec?

Possibly, but the code changes I have changes everything about how we do RPC requests. They rate limit based on a compute budget which is hard to estimate exactly, but batch RPC requests

ChaoticTempest:
Let's merge this one first, and then I can just fix whatever changes

PR #885, opened 2026-06-15
Key derivation standard

volovyks:
We already have 3 types of keys in the system, and it is a good time to standardize them (see @jakmeier comment [here](https://github.com/sig-net/mpc/pull/847#discussion_r3356163373)).

### DIFF
--- a/chain-signatures/crypto/src/kdf.rs
+++ b/chain-signatures/crypto/src/kdf.rs
@@ -1,3 +1,4 @@
+use crate::types::{Address, KeyVersion, Path, Purpose};
 use crate::{PublicKey, ScalarExt};
 use anyhow::Context;
 use k256::{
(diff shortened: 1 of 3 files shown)


chain-signatures/crypto/src/kdf.rs
```diff
@@ -48,29 +72,71 @@ fn keccak(derivation_path: impl AsRef<[u8]>) -> Scalar {
...
+pub fn derive_epsilon_bitcoin(key_version: KeyVersion, address: &str, path: &str) -> Scalar {
+    derive_epsilon(&DerivationParams::UserAccount(
+        key_version,
+        Chain::Bitcoin,
+        address.to_string(),
+        path.to_string(),
+    ))
+}
+
+pub fn derive_epsilon_checkpoint() -> Scalar {
```
volovyks, line 138: @ChaoticTempest I've added a function you can use in your PR

ChaoticTempest (approved):

PR #887, opened 2026-06-15
1.16.4 and CHECKPOINT_VERSION v10

volovyks:
@ChaoticTempest, we can postpone this merge if we want to make sure we fix backlog issues without resetting it.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,7 @@ exclude = ["infra/scripts/generate_keys"]
 resolver = "2"
 
 [workspace.package]
-version = "1.16.3"
+version = "1.16.4"
 
 [workspace.dependencies]
 alloy = { version = "=1.0.38", features = ["contract", "json"] }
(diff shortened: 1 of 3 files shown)


ChaoticTempest:
yes let's postpone this. Ideally the fix should fix the backlog so this wouldn't be needed

volovyks:
> yes let's postpone this. Ideally the fix should fix the backlog so this wouldn't be needed

This was already released on testnet, so we need to merge it. But yes, let's fix dev issues first.

Cargo.toml
```diff
@@ -12,7 +12,7 @@ exclude = ["infra/scripts/generate_keys"]
 resolver = "2"
 
 [workspace.package]
-version = "1.16.3"
+version = "1.16.5"
```
ChaoticTempest: isn't this supposed to be 1.16.4?

ChaoticTempest: I guess it should be 1.16.4 since your branch and naming of the PR says so, so I'll just make this change
```suggestion
version = "1.16.4"
```

ChaoticTempest (approved):

PR #889, opened 2026-06-16
fix: solana RPC rate limits on large catchup window

ChaoticTempest:
This should clear up issues on devnet where we have a large catchup window. There's a new ignored test `test_solana_pipeline_devnet` that proves that we should be able to handle from latest to being a week behind specifically for catchup.

* new SolanaClient for both rpc and indexer for the most common operations on solana. Needed for batching requests that the regular `solana_client::RpcClient` does not support out of the box.
* HTTP JSON-RPC 2.0 Batching for block retrievals (getBlock) which fetches up to 50 blocks in a single HTTP request (reducing HTTP call overhead by 50x to stay under Alchemy/RPC rate limits)


isSerge:
Not sure if you saw, I have another PR here - https://github.com/sig-net/mpc/pull/884
I think they complement each other - I am replacing `AsyncCatchupIter` with `futures::Stream` to fix sequential bottleneck.

ppca (approved):
LGTM!

ppca (approved):

PR #890, opened 2026-06-16
test: Bypass CPU exhaustion in integration tests

isSerge:
Few changes to address CPU spikes due to integration test running in parallel:

- Enable prestockpile in `test_signature_ethereum` and `test_proper_indexer_checkpoint` (removed `.disable_prestockpile()`). I think it's ok for these two to use pre-computed Triples, let me know if I am wrong.
- Replace `.signable()` with `.nodes_running()` in `test_checkpoint_recovery_after_offline`. We only care if Indexer can catch up.
- Increase timeouts in `test_checkpoint_recovery_after_offline` (the slowest test, often times out)
- Implement `spawn_sandbox_with_retry` helper to absorb random CPU spikes

### DIFF
--- a/integration-tests/tests/cases/ethereum.rs
+++ b/integration-tests/tests/cases/ethereum.rs
@@ -19,7 +19,7 @@ use tokio::time::Duration;
 
 #[test(tokio::test)]
 async fn test_signature_ethereum() -> Result<()> {
-    let cluster = cluster::spawn().disable_prestockpile().ethereum().await?;
+    let cluster = cluster::spawn().ethereum().await?;
     cluster.wait().signable().await?;
 
     let ctx = cluster.nodes.ctx();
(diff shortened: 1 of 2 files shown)


integration-tests/tests/cases/ethereum.rs
```diff
@@ -19,7 +19,7 @@ use tokio::time::Duration;
 
 #[test(tokio::test)]
 async fn test_signature_ethereum() -> Result<()> {
-    let cluster = cluster::spawn().disable_prestockpile().ethereum().await?;
+    let cluster = cluster::spawn().ethereum().await?;
```
volovyks, line 22: I'm not sure how this helps with CPU.

volovyks (approved):

integration-tests/tests/cases/ethereum.rs
isSerge: less polynomial math to generate Triples?

volovyks: Yep, correct. Removing "disable_x" actually enables x, my bad.

PR #893, opened 2026-06-17
refactor: Decouple `backlog` from indexers

isSerge:
This PR decouples Canton, Solana and Ethereum indexers from `Backlog` in order to extract indexers into isolated crates later. 

- Introduce `StateManager` trait in `mpc-primitives`. Add implementation for `Backlog`. Currently has only two methods: `get_processed_block` and `get_execution_watchers` (what indexers currently need). Indexer structs now use generics instead of `Backlog` directly.
- Remove `ChainStream` trait and specific implementations. Channel is created within `Indexer::new` method, which now returns `(Self, Receiver<ChainEvent>)`. `run_stream` takes `(I: ChainIndexer, Receiver<ChainEvent>)`

Hydration indexer is not covered here since it does not implement `ChainIndexer` yet (will do in a separate PR)


isSerge:
Closing due to conflicts, will submit smaller targeted PRs separately

PR #894, opened 2026-06-17
fix: add lazy catchup process blocks to avoid OOM issues

### LINKED ISSUE #869: Catchup should iterate instead of fetching everything at once
Some chain stream like solana will currently fetch everything at once for catchup and create a Vec or Map for it. Could be expensive to hold all in memory. So we should transition it to an iterator as much as possible

ChaoticTempest:
This makes it so that we only fetch `MAX_CHUNK_SIZE` amount of blocks at a time when we go to process catchup. This avoid OOM issues when we have a large catchup window where before we would fetch all blocks. We still however fetch all slots in that window first, since we can only walk backwards for fetching slots relating to a program id.

Also resolves https://github.com/sig-net/mpc/issues/869 for solana

### DIFF
--- a/chain-signatures/node/src/indexer_sol.rs
+++ b/chain-signatures/node/src/indexer_sol.rs
@@ -1,7 +1,7 @@
 use crate::backlog::Backlog;
 use crate::protocol::Chain;
 use crate::sign_bidirectional::hash_rlp_data;
-use crate::solana_client::{SolanaCatchupBlock, SolanaClient, MAX_CHUNK_SIZE};
+use crate::solana_client::{SolanaCatchupBlock, SolanaClient, MAX_CONCURRENT_CHUNK_SIZE};
 use crate::stream::{ChainIndexer, ChainStream};
 use crate::util::ethabi_request_id;
 use crate::util::retry::{retry_async, RetryConfig, RetryError, RetryReason};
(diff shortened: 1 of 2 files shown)


chain-signatures/node/src/indexer_sol.rs
```diff
@@ -263,8 +263,32 @@ impl ChainIndexer for SolanaIndexer {
...
+                loop {
+                    if let Some(block) = current_chunk.pop_front() {
+                        return Some((block, (remaining_slots, client, current_chunk)));
+                    }
+                    if remaining_slots.is_empty() {
+                        return None;
+                    }
+
+                    let chunk_slots: BTreeSet<u64> = remaining_slots
+                        .drain(..std::cmp::min(MAX_CHUNK_SIZE, remaining_slots.len()))
```
jakmeier: I think now `fetch_blocks_for_slots` only gets a chunk of size `MAX_CHUNK_SIZE` at the time, thus `buffer_unordered(MAX_CONCURRENT_FETCH)` is ineffective because it only every has a single chunk.

Would it make sense to use `MAX_CHUNK_SIZE * MAX_CONCURRENT_FETCH` here? Or maybe there is a smart way to rewrite it to ensure concurrency across the two layers.

jakmeier (approved):
Yeah this should solve the OOM.

But I believe we lose concurrency here, would be nice to resolve this before merging.

chain-signatures/node/src/indexer_sol.rs
ChaoticTempest: good catch, added `MAX_CONCURRENT_CHUNK_SIZE` to be `MAX_CHUNK_SIZE * MAX_CONCURRENT_FETCH`

PR #903, opened 2026-06-22
refactor: decouple indexers from metrics

isSerge:
Decouple metric updates from indexers:

- Introduce `ChainTelemetry` trait, add implementations: `PrometheusChainTelemetry` (default), `NoopTelemetry` (for tests)
- Update Ethereum, Solana, Canton, Hydration indexers and `run_stream` to accept generic `telemetry` parameter (`T: ChainTelemetry`)
- Remove `Indexed` update from Near indexer since it's only used for integration tests

### DIFF
--- a/chain-signatures/node/src/indexer.rs
+++ b/chain-signatures/node/src/indexer.rs
@@ -183,11 +183,6 @@ async fn poll_pending_requests(ctx: &mut Context) -> anyhow::Result<()> {
     // Update timestamp to indicate we're still running
     ctx.indexer.update_timestamp();
 
-    // Update metrics
-    crate::metrics::indexers::LATEST_BLOCK_NUMBER
-        .with_label_values(&[Chain::NEAR.as_str(), "indexed"])
-        .set(latest_height as i64);
-
     // Send all new requests
     for request in new_requests {
         tracing::info!(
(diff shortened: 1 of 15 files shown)


ppca (approved):
lgtm

chain-signatures/node/src/indexer.rs
```diff
@@ -183,11 +183,6 @@ async fn poll_pending_requests(ctx: &mut Context) -> anyhow::Result<()> {
     // Update timestamp to indicate we're still running
     ctx.indexer.update_timestamp();
 
-    // Update metrics
```
volovyks, line 186: Are we doing this in some other place now?

volovyks (approved):
Looks good!
Which metrics for Canton and Solana have now been added?

chain-signatures/node/src/indexer.rs
isSerge: this is Near indexer, do we need to update metrics here since it's only for integration tests?

isSerge:
@volovyks no metrics for Canton and Solana added since they are not currently updating any, but maybe it's worth adding here

chain-signatures/node/src/indexer.rs
volovyks: Got it. It is not an active integration. We can skip it.

volovyks:
@isSerge from what I see they are `<grafana_link>` added. At least Solana. That is why I'm asking which metric you are specifically talking about.

volovyks:
I see now. That is done on the stream level. We need to make it consistent. At least for chains that use stream.

volovyks (approved):

PR #904, opened 2026-06-22
refactor: decouple CLI args from indexers

isSerge:
This RP focuses on separation of CLI configuration from indexers. 
Main motivation: indexer should not know anything about CLI, `clap`, etc.

Just moving structs to different folders and updating imports, no new logic:
- Specific CLI argument structs are moved to `/cli/args/`
- `indexer_sol` create a dedicated folder similar to the rest, extract client into a separate modules
- Extract indexer config structs into separate files


chain-signatures/node/src/indexer_hydration/config.rs
```diff
@@ -0,0 +1,18 @@
...
+    /// Hydration RPC ws URL
+    pub rpc_ws_url: String,
+    /// Hydration signer URI
+    pub signer_uri: String,
+}
+
+impl fmt::Debug for HydrationConfig {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("HydrationConfig")
+            .field("rpc_ws_url", &self.rpc_ws_url)
```
volovyks: It can contain API keys

volovyks (approved):

PR #905, opened 2026-06-22
metrics: Indexer checkpoint

volovyks:


chain-signatures/node/src/stream/ops.rs
```diff
@@ -389,11 +389,13 @@ pub(crate) async fn process_block_event<T: ChainTelemetry>(
     caught_up: bool,
     telemetry: &T,
 ) {
+    telemetry.block_finalized(block);
```
volovyks, line 392: Moving `block_finilized` out of the checkpoint condition. Otherwise, it is only emitted when we create a checkpoint.

isSerge: we have it twice, second one is end of the method

volovyks: removed

isSerge (approved):

Issue #871, opened 2026-06-11
Generalize RPC (signature publishing) for each chain


ChaoticTempest:
Currently, a singular RPC task is responsible for handling the publishing of all signatures to their respective chains.

There are some fundamental issues with this task:
- It does not infinitely retry and stops after 6 attempts (we do however persist the signature on publishing, so a restart would attempt the publishing again)
- It does not handle the proposer node going offline well (https://github.com/sig-net/mpc/issues/837)
- It's not properly separated and there's a lot of repeated logic (retries, clients, ...)
- Handling of publishing errors is not consistent between chains
- some have special handling where we need to wait additionally or batch responding

So we need to generalize this to make it more manageable and we would be able to handle some of these cases better and actually guarantee the signature gets published

isSerge:
I think all of these points were resolved in #926, except node offline handling, which has a separate issue

volovyks:
True, great work @isSerge !

PR #908, opened 2026-06-23
refactor: centralize RPC retry logic

isSerge:
This PR standardizes the retry logic for RPC and publishing operations across the Ethereum, Solana, and cross-chain publisher components. It replaces custom, ad-hoc retry loops with a centralized `retry_rpc!` macro built on top of the `backon` crate (we were already using it in some places).

**Core retry infra:**

- Replaced custom retry logic in `util/retry.rs` with `retry_rpc!` macro, which provides both a standard form (auto-logging) and a custom closure form (for detailed context). Distinction between timeout and all attempts exhausted preserved: the macro unifies both errors into standard `anyhow` errors seamlessly (no need to do explicit match)
- Add HTTP 4xx Suppression using `is_retryable` utility. This ensures we immediately fail on terminal client errors (HTTP 400-405) while correctly retrying rate limits (429) or timeouts (408) (previously we were retrying all)
- Introduce `RetryConfig` struct which maps to `backon::ExponentialBuilder`, so we can use different retry strategies for different use-cases 

**Update Ethereum client:**

- Update methods to use `retry_rpc!`
- Add `retry_strategy` field, which holds `RetryConfig` (we have default and faster version for unit tests)
- Add some unit tests, introduce `test_utils` and implemented `create_test_ethereum_client()` helper (with fast retries), used for both client and indexer tests

**Update Solana client:**

- Update methods to use `retry_rpc!`
- Add `retry_strategy` field, which holds `RetryConfig`
- `subscribe_and_buffer_live_events` now accepts fully configured `SolanaClient` (with retries)

**Update RPC and publishing:**

- Replace magic numbers with constants for different operations (Publishing, Nonce Fetching, Tx Sending, Mempool Polling, Receipt Polling, etc.)
- Update methods to utilize `retry_rpc!`

I only covered places where we already had some retry logic (`rpc.rs`, Solana and Ethereum clients), it does not include Canton and Hydration. These should be addressed separately.

### DIFF
--- a/chain-signatures/node/src/indexer_sol/client.rs
+++ b/chain-signatures/node/src/indexer_sol/client.rs
@@ -422,4 +422,297 @@ impl SolanaClient {
     }
 }
 
-// TODO: add unit tests
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use solana_sdk::pubkey::Pubkey;
+    use solana_sdk::signature::Signature;
+
+    /// Helper to create a SolanaClient for testing with mockito
+    fn test_client(url: &str) -> SolanaClient {
+        SolanaClient::for_indexer(
+            url.to_string(),
+            url.replace("http", "ws"),
+            Pubkey::new_unique(),
+        )
+        .with_fast_retry()
+    }
+
+    /// Helper to create a mock JSON-RPC response for getSlot
+    fn slot_response(slot: u64) -> String {
+        format!(r#"{{"jsonrpc":"2.0","id":1,"result":{slot}}}"#)
+    }
+
+    /// Helper to create a mock JSON-RPC response for getBlock
+    fn make_block_response(id: usize, slot: u64) -> serde_json::Value {
+        serde_json::json!({
+            "id": id,
+            "result": {
+                "blockHeight": slot,
+                "blockTime": null,
+                "blockhash": "11111111111111111111111111111111",
+                "parentSlot": slot.saturating_sub(1),
+                "previousBlockhash": "11111111111111111111111111111111",
+                "transactions": [],
+                "rewards": []
+            }
+        })
+    }
+
+    /// Helper to create a mock JSON-RPC response for getSignaturesForAddress
+    fn signature_entry(slot: u64, sig: &str) -> serde_json::Value {
+        serde_json::json!({
+            "signature": sig,
+            "slot": slot,
+            "err": null,
+            "memo": null,
+            "blockTime": null,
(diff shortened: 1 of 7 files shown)


chain-signatures/node/src/indexer_sol/client.rs
```diff
@@ -117,149 +151,182 @@ impl SolanaClient {
...
-                    jsonrpc: "2.0",
-                    id: i,
-                    method: "getBlock",
-                    params: json!([slot, config_val]),
-                });
-            }
+        let max_attempts = self.retry_strategy.max_times;
+        let res = retry_rpc!(
+            SOL_BATCH_TIMEOUT,
+            self.retry_strategy,
```
ppca: I think we want to keep retrying until it succeeds, but this limits retry counts?

chain-signatures/node/src/indexer_sol/client.rs
```diff
@@ -117,149 +151,182 @@ impl SolanaClient {
...
-                &ExponentialBuilder::default()
-                    .with_min_delay(Duration::from_millis(500))
-                    .with_max_delay(Duration::from_secs(10))
-                    .with_max_times(usize::MAX),
-            )
-            .notify(|err, delay| {
+    ) -> anyhow::Result<Vec<RpcConfirmedTransactionStatusWithSignature>> {
+        retry_rpc!(
+            SOL_RPC_TIMEOUT,
+            self.retry_strategy,
```
ppca: same here, we should retry forever until success

chain-signatures/node/src/indexer_sol/client.rs
```diff
@@ -117,149 +151,182 @@ impl SolanaClient {
...
+                    .map_err(|e| anyhow::anyhow!(e))
+            }
+        )
+    }
+
+    pub async fn get_block(&self, slot: u64) -> anyhow::Result<UiConfirmedBlock> {
+        let max_attempts = self.retry_strategy.max_times;
+        retry_rpc!(
+            SOL_RPC_TIMEOUT,
+            self.retry_strategy,
```
ppca: Here too, infinite retry

chain-signatures/node/src/indexer_sol/client.rs
isSerge: thanks for catching up

volovyks (approved):
LGTM
Using uint::MAX is not really an infinite retry, but enough for now
I would rather iterate on this in the context of #871 in separate PRs

PR #911, opened 2026-06-23
refactor: centralize `IndexedSignRequest` metrics update and decouple Eth indexer from telemetry

isSerge:
This PR focuses on final decoupling of indexers from Node's Prometheus metrics and centralization of telemetry updates for `IndexedSignRequest`:

- `ChainEvent::SignRequest` - add optional `block_timestamp` field. Currently only Ethereum provides `Some(block_timestamp)`, the rest are just `None`
- Remove `metrics::requests::record_request_latency_since` from Ethereum indexer. Request metrics are now being updated in one place - `run_stream` using either `request_indexed_at` or `request_indexed` methods from `ChainTelemetry`

### DIFF
--- a/chain-signatures/primitives/src/traits.rs
+++ b/chain-signatures/primitives/src/traits.rs
@@ -30,9 +30,11 @@ pub trait ChainTelemetry: Send + Sync + Clone + 'static {
     /// Records that a checkpoint was created
     fn checkpoint_created(&self, block_number: u64);
 
-    /// Report that a request was indexed.
-    /// Pass `Some(timestamp)` if the chain supports block timestamps (Ethereum), otherwise `None`.
-    fn request_indexed(&self, block_timestamp: Option<u64>);
+    /// Report that a request was indexed at the given block timestamp (e.g. for Ethereum)
+    fn request_indexed_at(&self, block_timestamp: u64);
+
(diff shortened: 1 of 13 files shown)


chain-signatures/primitives/src/traits.rs
```diff
@@ -29,6 +29,12 @@ pub trait ChainTelemetry: Send + Sync + Clone + 'static {
 
     /// Records that a checkpoint was created
     fn checkpoint_created(&self, block_number: u64);
+
+    /// Report that a request was indexed at the given block timestamp (currently used for Ethereum due to ~15 min finality delay)
+    fn request_indexed_at(&self, block_timestamp: u64);
+
+    /// Report that a request was indexed without a block timestamp (faster chains, e.g. for Solana, Canton, or Hydration)
```
volovyks, line 36: I would not specify specific chains, since the list may change and nobody will update this comment.

volovyks (approved):

chain-signatures/primitives/src/traits.rs
isSerge: Otherwise might be confusing why we need two methods: one with timestamp and another without. Most new chains will probably go to a faster category anyway

volovyks: I do not think that this is about chain speed. Time may or may not be in the RPC API.
Not a blocker, feel free to merge.

Issue #859, opened 2026-06-09
testnet incident Jun 04 UTC 11:00 - 12:40


ppca:
https://signetwork.grafana.net/a/grafana-irm-app/alert-groups/IW1LJY3G3PNJY

we sent load test traffic to testnet for 20 mins from 11:35 - 11:55, and we observe that signature generation success rate dips and not recovering until 12:40.

We redeployed around 12:40

ppca:
**Context**: 
We send load test Solana traffic for 20 mins starting 11:36. The signature in time success rate `<grafana_link>` drops significantly, and becomes 0 and did not recover until we redeploy testnet at 12:40.

**Observations**:
1. signature success: The mpc system stopped having any signature success `<grafana_link>` starting 11:58, and started having some success starting 12:21.
2. signature generators vanishing(grafana `<grafana_link>`): node 4 stopped having new signature generators starting 11:46, other nodes stopped having new signature generators starting 11:56. 
3. signature generators appearing again(grafana `<grafana_link>`): All nodes except node 4 started having new signature generator starting 12:21: node 0 at 12:24, node 1 at 12:25, node 2 at 12:28, node 3 at 12:21, node 5 at 12:27, node 6 at 12:27, node 7 at 12:21. Node 4 started having new signature generator at 12:38.
4. presignature generators(grafana `<grafana_link>`): node 4 stopped having new presignature generators at 11:37, all other nodes stopped having new presignature generators at 11:56. New generators started showing up at 12:38 on all nodes.
5. presignature posit channel(grafana `<grafana_link>`): node 4's presignature posit channel capacity quickly shrinks starting 11:37, around 11:44, it approaches 0. Then it becomes intermittently 0 `<grafana_link>`, sometimes non-zero small values like 1 or 4, sometimes 0, until 11:55, then it starts climbing back up. For other nodes, we see drop in presignature posit channel capacity between 11:36 and 11:56, but much much smaller magnitude and they climb back up quickly.
6. mine triples(grafana `<grafana_link>`): node 4's count of mine triples keeps plunging starting 11:36, approaches 0 at around 11:56 and stayed around 0. Other nodes' count of mine triples decreased between 11:36 and 11:56 too but climbed back up.
7. signature logs: between 11:56 and 12:38, all nodes are still sending logs in the signature target. They are processing new sign ids, but always looping between organizing and posit, never gets to generation.
[...]

PR #913, opened 2026-06-24
component test: test_presig_stall_from_slow_redis

ppca:
**This PR adds a component test that reproduces and verifies a presignature generation stall** (issue #859).

issue we want to reproduce: when Redis is under contention (e.g. heavy take_mine calls from the stockpile loop), latency spills over into contains() checks during presig posit processing. Posits expire before accepts arrive, wasting triples and stalling presig generation on the affected node — while triple generation (which has no Redis IO in posit processing) continues fine.

The PR has three parts:
1. Configurable posit timeouts (presignature.rs) — Extracts the hardcoded 10s posit timeout and 2s extra delay into constants (PRESIGNATURE_POSIT_TIMEOUT, PRESIGNATURE_POSIT_EXTRA_DELAY) that shrink to 3s/500ms under the test-feature flag, so the stall manifests faster in tests.
2. Injectable Redis operation delays (protocol_storage.rs) — Adds a set_operation_delay(operation, duration) method behind #[cfg(feature = "test-feature")] that lets tests inject artificial latency on specific Redis operations (take_mine, insert, contains). Shared across clones via Arc<Mutex<...>>.
3. The test itself (mpc.rs: test_presig_stall_from_slow_redis) — Spins up a 3-node MPC fixture, injects 100ms delays on take_mine/insert and 50ms on contains for node 0 only, then runs in two phases:
    - Phase 1 (detect stall): Polls every 10s, asserts that node 0's stored presig count stops growing while healthy nodes keep producing presigs and all nodes still generate triples. Must hold for 3 consecutive polls.
    - Phase 2 (verify recovery): Once healthy nodes hit their presig target and stop proposing, node 0's posit backlog clears and it resumes generating. Asserts node 0's presig count grows for 3 consecutive polls.

### DIFF
--- a/chain-signatures/node/src/protocol/presignature.rs
+++ b/chain-signatures/node/src/protocol/presignature.rs
@@ -29,6 +29,19 @@ use tokio::sync::{mpsc, watch};
 use tokio::task::JoinHandle;
 use tokio::time;
 
+pub const PRESIGNATURE_POSIT_TIMEOUT: Duration =
+    Duration::from_millis(if cfg!(feature = "test-feature") {
+        3_000
+    } else {
+        10_000
+    });
+pub const PRESIGNATURE_POSIT_EXTRA_DELAY: Duration =
+    Duration::from_millis(if cfg!(feature = "test-feature") {
+        500
+    } else {
+        2_000
+    });
+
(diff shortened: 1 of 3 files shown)


volovyks:
@ppca is this still relevant?

PR #914, opened 2026-06-24
feat: add `mpc-indexer-core` crate

isSerge:
- Create new crate `mpc-indexer-core`. It collects all traits that an indexer has to implement or relies on.
- Move `ChainIndexer`, `ChainStream`, `ChainTelemetry`, `StateManager` here

Benefits:
- `mpc-primitives` is now mostly pure structs, no behavioral traits
- `mpc-indexer-core` acts as a unified "indexer framework"


chain-signatures/indexer-core/src/state.rs
```diff
@@ -0,0 +1,18 @@
+use std::collections::HashMap;
+
+use mpc_primitives::{BidirectionalTx, BidirectionalTxId, Chain, SignId};
+
+/// Interface for the Indexer to query and update state.
+/// Currently implemented by the Backlog
+#[async_trait::async_trait]
+pub trait StateManager: Send + Sync + Clone + 'static {
```
ChaoticTempest, line 8: Since we're using StateManager for backlog, it's a bit too ambiguous what state it's trying to manage. Something like RequestManager might be better but we can triage this a bit more.

I guess the whole point of having this trait in the first place is to be able to generalize our indexer logic, while not having to move the `Backlog` out of the mpc-node code. It's probably best to eventually move it closer to indexers since indexers are the only code that utilizes backlogs a lot. The signing code might touch it in a single spot

ChaoticTempest (approved):

chain-signatures/indexer-core/src/state.rs
isSerge: My understanding is that in the future indexers might query not just requests, but also state of the system in general in order to decide whether to push blocks further to the queue

ChaoticTempest: I don't think we would need to do that. Indexers should just be independent from the state of the system. So it should just act as a faucet for request input. How we limit that faucet is a detail the indexers shouldn't care about for now until we actually need it

But even in the case that the indexers can limit itself, putting it all in one place/interface can start to look weird. I would prefer not having this interface at all since there's only one implementor (the Backlog), so it's really not needed eventually when we move the backlog to the indexer crates

isSerge: You mean one backlog per chain integration instead of a centralized one?

ChaoticTempest: could work as either one backlog per chain integration or a centralized one, whichever is simplest

PR #915, opened 2026-06-24
feat: abort sign tasks on regression

ChaoticTempest:
on chain consensus regression detected, the MPC node immediately halts all active signature generation and consensus checkpoint tasks for the regressed chain. This prevents the node from wasting compute resources and publishing stale/abandoned signatures or checkpoints on-chain.

## Changes
* `Sign::AbortChain(chain)` notifies SignatureSpawner of a consensus regression.
* On Sign::AbortChain, the spawner immediately aborts running tasks, removes their posit inboxes, and cancels delayed watcher handles.
* dead_ids LruCache in SignatureSpawner to track recently completed/aborted task IDs.
* Late-arriving peer posit messages for stale requests are silently dropped if the request ID is in the cache, preventing orphaned inbox memory leaks.
  * Automatically clears a task's ID from the cache upon spawn_task so that re-queued requests (after catchup and alignment) can successfully process new peer messages.

### DIFF
--- a/chain-signatures/node/src/protocol/signature.rs
+++ b/chain-signatures/node/src/protocol/signature.rs
@@ -23,6 +23,7 @@ use crate::types::SignatureProtocol;
 use crate::util::{AffinePointExt, JoinMap, TimeoutBudget};
 
 use cait_sith::protocol::{Action, InitializationError, Participant};
+use lru::LruCache;
 use cait_sith::PresignOutput;
 use chrono::Utc;
 use k256::Secp256k1;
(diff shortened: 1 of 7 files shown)


volovyks (approved):
We are adding more and more things to signature.rs. We need to start decoupling everything it contains.

ChaoticTempest:
yeah we need to refactor signature.rs eventually, just that we have a lot of moving things in the work right now with signature.rs so I think small chunks at a time would be best

PR #916, opened 2026-06-24
Do not check the proposer

volovyks:


chain-signatures/node/src/protocol/signature.rs
```diff
@@ -792,8 +779,11 @@ impl SignPositor {
 
                     if is_deliberator {
                         if let PositAction::Start(participants) = action {
-                            if from != proposer {
-                                tracing::warn!(?sign_id, ?round, ?from, ?proposer, "received Start from non-proposer, ignoring");
+                            // We accepted a Propose from `accepted_proposer`; the
+                            // Start must come from that same node, not from an
+                            // unrelated peer.
+                            if from != accepted_proposer {
```
jakmeier, line 785: Hm, rejecting here means the `ACCEPT` sent earlier pretty much lost its meaning.

The cases that currently fail due to bad round alignment on the development branch, this PR will let them slip through the ACCEPT phase but then two or more proposers will send START and try to generate a signature with fewer nodes responding than necessary. That's worse than what we have today.

I think switching to accepting any proposer means we have to allow multiple parallel signature generations for the same request. So, we need to support multiple accepted proposers and multiple ongoing signature generations for a single request id.

jakmeier (commented):
Thanks for implementing what we discussed! But I think we need more changes to make it work. (see my inline comment)

PR #919, opened 2026-06-25
fix: persist only consensus checkpoint

ChaoticTempest:
This PR makes checkpoint consensus more robust where it addresses edge cases in the checkpoint consensus alignment logic.

We now only have `pending_checkpoints` which is an in-memory list of checkpoints that have yet to reach consensus. And a `latest_checkpoint` which is the latest consensus checkpoint that we have confirmed to be in the contract.

#### Backpressure on Checkpoint Creation
- Added an in-memory `pending_checkpoints` map to track checkpoints awaiting consensus confirmation.
- This is limited by `MAX_PENDING_CHECKPOINTS = 32`. When this cap is reached, `ChainPipeline` block processing (`handle_live` and `handle_catchup` loops) stalls until we are able to get a consensus checkpoint signature. This makes it so that if our signature pipeline is stalled somewhere, we won't continually keep chugging our checkpoint creation.

#### Checkpoint Eviction
- On consensus confirmation, we clear all checkpoints with height `<= confirmed_height` from `pending_checkpoints`.

#### Storage
- Removed whole pending history persistence from `CheckpointStorage`. We now persist only the latest confirmed consensus checkpoint.

##### New Unit Tests
  - `test_checkpoint_stalls_at_cap` and `test_checkpoint_unblocks_after_confirmation`.
  - `test_on_consensus_confirmed_evicts_older_pending_checkpoints`: evicts all checkpoints older than a certain point.
  - `test_find_checkpoint_by_digest_falls_back_to_storage`.
  - matrix test `test_consensus_alignment_matrix` covering all alignment cases:
     - Case 1 (No Local, Has Remote): Triggers recovery and fetches checkpoint from peer.
     - Case 2 (Has Local, No Remote): Zero digest is ignored; local checkpoint is preserved.
     - Case 3 (No Local, No Remote): Zero digest is ignored; returns None.
     - Case 4 (Both Present, Matches): Match confirmed; persisted to storage.
     - Case 5 (Ahead but Aligned): Confirms and persists the earlier matching checkpoint.
     - Case 6 (Both Present, Divergent): Overwrites storage with remote checkpoint fetched from peer.

##### New Component Integration Tests
  - `test_web_server_serves_checkpoint`, `test_consensus_alignment_peer_fetch`, and `test_consensus_alignment_consensus_changes_while_fetching`


chain-signatures/node/src/storage/checkpoint_storage.rs
```diff
@@ -126,38 +84,17 @@ impl CheckpointStorage {
...
-            CheckpointStorage::InMemory { latest, .. } => {
-                Ok(latest.read().await.get(&chain).cloned())
-            }
+            CheckpointStorage::InMemory { latest } => Ok(latest.read().await.get(&chain).cloned()),
         }
     }
 
+    /// Load the latest consensus checkpoint for the given chain.
+    /// Returns a single-element vec if one exists, empty vec otherwise.
     pub async fn load_history(&self, chain: Chain) -> anyhow::Result<Vec<Checkpoint>> {
```
volovyks: Since we do not have history (multiple checkpoints) on the storage level now, it can be renamed to `load` or something similar.

volovyks (approved):
LGTM, in memory history seems more reasonable.
We will need to have docs once we stabilize the design, it is hard to understand all the edge cases in the checkpoint, stream, and backlog state machine.

ChaoticTempest:
@volovyks yep, planning on having a spec for all this backlog and checkpoint related stuff

chain-signatures/node/src/storage/checkpoint_storage.rs
ChaoticTempest: actually, this function is not really needed since it just calls load_latest. Will get rid of it

PR #920, opened 2026-06-25
feat: do not limit checkpoint sign tasks

ChaoticTempest:
The checkpoint sign task should not be limited by the SignLimiter. That way, they can complete as fast as needed in cases where we have millions of sign requests/tasks. 

Checkpoint tasks are inherently limited to 32 per chain if https://github.com/sig-net/mpc/pull/919 merges in


volovyks (approved):
A comment would be nice

jakmeier (approved):

ChaoticTempest:
closing this in favor of https://github.com/sig-net/mpc/pull/1046 where we no longer rely on the mpc signing flow and instead use voting, so we no longer need to unlimit the checkpoint signing flow

PR #926, opened 2026-06-27
feat: introduce `ChainPublisher` trait

isSerge:
This PR addresses some of the concerns mentioned in #871:
- Introduce `ChainPublisher` trait with `publish_signature` method. `RpcExecutor` is now chain-agnostic. 
- Batching is now encapsulated inside `EthClient` (no other clients need it). Executor does not care about the internal publishing mechanism of each client
- All clients publish methods now return `anyhow::Result<()>`, logging is done using `inspect_err`
- Update retry mechanism: retry infinitely (`usize::MAX`) with a capped exponential backoff (1 min)
- Metrics update is extracted into `record_publish_metrics`, which is now used within each client instead of `execute_publish` free function: Ethereum client updates metrics after batch is published successfully, others after single action was published.
- Add `test_utils` and unit tests for `RpcExecutor`

### DIFF
--- a/chain-signatures/node/src/cli/mod.rs
+++ b/chain-signatures/node/src/cli/mod.rs
@@ -320,13 +320,13 @@ pub async fn run(cmd: Cli) -> anyhow::Result<()> {
             let near_client =
                 NearClient::new(&near_rpc, &my_address, &network, &mpc_contract_id, signer);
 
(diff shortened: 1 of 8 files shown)


ppca (approved):
lgtm

volovyks:
> Batching is now encapsulated inside EthClient (no other clients need it).

@isSerge we do need batch on all chains to publish faster and cheaper. Not a top priority, though.

PR #928, opened 2026-06-29
refactor: decouple `CompletedTx` from from the Ethereum JSON-RPC client

isSerge:
This PR extracts the `extract_success_tx_output` functionality and its supporting `is_contract_call` method out of `src/respond_bidirectional.rs` into `src/indexer_eth/mod.rs`. This ensures that `CompletedTx` only cares about generating follow-up signature requests from raw bytes, while the Ethereum indexer is responsible for determining the `ExecutionOutcome::Success { output: Vec<u8> }`

### DIFF
--- a/chain-signatures/node/src/respond_bidirectional.rs
+++ b/chain-signatures/node/src/respond_bidirectional.rs
@@ -195,15 +195,6 @@ mod tests {
         Bytes::from(buf.to_vec())
     }
 
-    // #[test]
-    // fn is_contract_call_detects_calldata() {
-    //     assert!(!is_contract_call(&Bytes::new()));
-    //     assert!(!is_contract_call(&Bytes::from(vec![0u8; 2])));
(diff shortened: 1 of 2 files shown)


chain-signatures/node/src/respond_bidirectional.rs
```diff
@@ -247,14 +195,14 @@ mod tests {
...
 
-    #[test]
-    fn is_contract_call_detects_calldata() {
-        assert!(!is_contract_call(&Bytes::new()));
-        assert!(!is_contract_call(&Bytes::from(vec![0u8; 2])));
-        assert!(is_contract_call(&Bytes::from(vec![
-            0xa9, 0x05, 0x9c, 0xbb, 0x00
-        ])));
-    }
+    // #[test]
```
ppca: i guess this can be removed?

ppca (approved):

Issue #888, opened 2026-06-16
lingering signIds: some nodes miss request due to no catchup when reconnecting websocket


ppca:
Nodes miss solana sign requests because they paused 1 sec before reconnecting to solana websocket and we haven't yet implemented catchup for this case([code](https://github.com/sig-net/mpc/blob/8e2ea385db74afe1001652f5685e176919bb5001/chain-signatures/node/src/indexer_sol.rs#L655))

```jsonPayload.signId="SignId(\"6b7e3ddb094949d9a644df0f9035980e2867147323f00b50b919c105c65971d0\")"```
[log](https://console.cloud.google.com/logs/query;query=--labels.%22k8s-pod%2Fapp%22%3D%22multichain-6%22%0A%0AjsonPayload.signId%3D%22SignId%2528%5C%226b7e3ddb094949d9a644df0f9035980e2867147323f00b50b919c105c65971d0%5C%22%2529%22%0AjsonPayload.target%3D%22mpc_node::indexer_sol%22;lfeCustomFields=jsonPayload%252Ftarget;cursorTimestamp=2026-06-10T23:05:36.040869204Z;startTime=2026-06-10T23:03:35.948Z;endTime=2026-06-10T23:07:35.948Z?project=near-cs-testnet)

this signId was only indexed by nodes 4, 0,3,6. 
looking at the logs around then, the other nodes reported ```Live solana subscription failed: solana logs subscription stalled: no ws message for 60s``` around the time of indexing.

another example: ```jsonPayload.signId="SignId(\"aaade1de64198b8e50e476d905a12e961c1f36daad623f59de96e5aeeb2a534e\")"```,
```jsonPayload.signId="SignId(\"ad69567caf428c3e968bafb826ced14ff73284b369100c4eb220faaf04cc1b88\")```,
```jsonPayload.signId="SignId(\"00d15cbf06d82393a066e047386cef8751582cbc8f05c853282fd0b9556e086c\")"```,
```jsonPayload.signId="SignId(\"e0672682c3589f2adf2b861c7c5e731c80890b4482075f9e74dd97efab09ab50\")"```

ppca:
fixing https://github.com/sig-net/mpc/issues/811 will resolve this

volovyks:
[...]

PR #929, opened 2026-06-29
solana catchup on ws reconnect

### LINKED ISSUE #811: solana needs more-than-one catchup logic on websocket stream failing
on the websocket stream failing, solana will need addtional catchup logic between restarts of the websocket stream. This is due to the limitations of websocket streaming only starting from the latest slot. So the time between restarting the websocket, the blocks would have advanced without us knowing

ppca:
## Summary

  Fix Solana websocket reconnect handling so missed slots are backfilled before live forwarding resumes.

  Previously, when the Solana WS subscription dropped, the indexer resubscribed but did not run catchup from the last processed slot to the new subscription anchor. Any sign requests emitted during the disconnect window could be skipped permanently.

  This PR moves reconnect handling into the shared chain pipeline. On reconnect, the pipeline establishes a new live subscription, anchors it, replays missed blocks up to that anchor, then marks catchup complete and resumes live processing.

  ## Key Changes

  - Add `LiveStreamStatus::{Continue, Reconnect, Shutdown}` so indexers can distinguish normal progress, reconnectable stream closure, and terminal shutdown.
  - Add a `Reconnect` pipeline state that re-establishes the livestream and transitions through catchup before returning to `Live`.
  - Make Solana WS subscriptions one-shot: when the subscription ends, the live channel closes and the pipeline owns  reconnect/catchup retry.
  - Add `CatchupInProgress` as an in-band `ChainEvent`, ordered on the same event channel as sign/respond/block events.
  - Replace `run_stream`’s state-channel `select!` with a single ordered event loop.
  - Keep `state_rx` for pipeline state observation/readiness in tests, but no longer use it for forwarding correctness.
  - Add `pipeline_caught_up`, an out-of-band atomic veto used only for regression recovery, where already-buffered live  events may come from a forked chain.
  - Keep reconnect semantics ordered: buffered live events before `CatchupInProgress` are forwarded, replayed catchup events  are buffered, and pending work is requeued after `CatchupCompleted`.
  - Add regression detection while reconnecting, so checkpoint changes are not blocked by reconnect retry loops.
  - Add temporary Hydration compatibility because Hydration still calls shared stream ops outside `ChainPipeline`.
  - Restore integration-test readiness waits for stream helpers that submit transactions immediately after startup.

  ## Why Both `CatchupInProgress` And `pipeline_caught_up`?

  Reconnect and recovery have different safety requirements.

  For reconnect, the chain has not regressed; the websocket dropped. Events already buffered before the catchup marker were produced while live and should be processed as live. The in-band `CatchupInProgress` marker gives the correct ordering.

  For recovery, a consensus checkpoint regression may mean buffered events came from a forked chain. The in-band marker may  be queued behind those events, so the pipeline flips `pipeline_caught_up` to `false` immediately at recovery start. Forwarding then uses `ctx.caught_up && pipeline_caught_up`, preventing stale events from reaching signing while recovery realigns the backlog.

  Closes #811, #888


chain-signatures/primitives/src/events.rs
```diff
@@ -19,6 +19,10 @@ pub enum ChainEvent {
     /// Catchup has completed and live events may be forwarded to the signer.
     CatchupCompleted,
 
+    /// The stream is entering catchup/replay; events after this marker are not live
+    /// until a subsequent `CatchupCompleted`.
+    NotCaughtUp,
```
ChaoticTempest: I would prefer we name this `CatchupInProgress`, because `NotCaughtUp` here would mean that something happened for us to have not been caught up instead

chain-signatures/indexer-core/src/indexer.rs
```diff
@@ -21,6 +28,9 @@ pub trait ChainIndexer: Send + 'static {
         Ok(())
     }
 
+    /// Emit an ordered marker on the same event stream used by catchup replay.
+    async fn notify_not_caught_up(&mut self) -> anyhow::Result<()>;
```
ChaoticTempest: wait, why do we need a not caught up or a catchup in progress event at all? This is the part I'm not understanding

chain-signatures/node/src/stream/mod.rs
```diff
@@ -89,6 +90,9 @@ pub async fn run_stream<S: ChainStream, T: ChainTelemetry>(
                         requeue_pending_sign_requests(&backlog, chain, sign_tx.clone()).await;
                         resume_pending_publish_requests(&backlog, chain, &contract_watcher, &rpc).await;
                     }
+                    ChainEvent::NotCaughtUp => {
+                        caught_up = false;
+                    }
```
ChaoticTempest: So `NotCaughtUp` is for setting caught_up to be false. Hmm, I guess we can simplify this to just have a `watch::Receiver<CatchupStatus { CaughtUp, CatchingUp }>` to handle this better instead of adding all this code to emit the not caught up event

chain-signatures/indexer-core/src/indexer.rs
ppca: The event exists to solve an ordering race between two channels in `run_stream`. When reconnect triggers catchup, the sign requests and respond events that are seen during catchup are sent to`events_rx`, and state transition (like `Reconnect` → `Catchup`) are sent to `state_rx`. Since `run_stream` uses `select!` over both channels, there's no guarantee which one it reads first when both are ready. 
  
Without the marker event, this can happen:
1. Pipeline enters `Reconnect`, sends state update on `state_rx`
2. catchup running, sending sign and respond events to `events_rx`
3. `run_stream`'s `select!` picks the `events_rx` first — processes the replayed `SignRequest` with `caught_up = true`, forwarding it to signing prematurely
4. `run_stream` then reads `state_rx` and sets `caught_up = false` — too late
  
By sending `NotCaughtUp` on the same `events_rx` channel *before* the sign and respond events during catchup, we guarantee `caught_up` is cleared before any sign or respond arrive, regardless of `select!` ordering.

chain-signatures/primitives/src/events.rs
ppca: sounds good. will change

chain-signatures/node/src/stream/mod.rs
ppca: A separate `watch` channel would have the same ordering race described above — `watch` and `events_rx` are independent channels, so `select!` doesn't guarantee which one `run_stream` reads first. 

But agree this can be simplified a great deal. 

tokio::select! supports biased. I think to simplify the implementation, we can do this: add biased to the existing select! in run_stream, put state_rx.changed() before events_rx.recv(), and then we can remove `NotCaughtUp` event, the `notify_not_caught_up` trait method, and all the associated code.

chain-signatures/indexer-core/src/indexer.rs
ppca: yeah but using tokio::select! with biased can simplify this greatly.

chain-signatures/primitives/src/events.rs
ppca: won't need this if we use select! with biased. Let me change code.

chain-signatures/node/src/stream/mod.rs
ppca: After thinking about this more, I've moved away from the `biased; select!` approach. The fundamental issue is that
  `state_rx` and `events_rx` are two separate channels with no shared ordering guarantee. `biased` only controls polling priority within a single `select!` invocation — it can't enforce a consistent ordering between messages produced on independent channels by a different task.

This leads to two subtle problems:
### 1. Watch channel coalescing
In a fast reconnect scenario:
1. Pipeline sends `Reconnect` state on `state_rx`
2. Catchup runs and completes quickly
3. Pipeline sends `Live` state on `state_rx` (coalesces over `Reconnect` — watch only keeps the latest value)
4. Meanwhile, catchup replay events are queued on `events_rx`

When `run_stream` next polls: `state_rx.changed()` resolves with `Live` (the intermediate `Reconnect`/`Catchup` states were coalesced away), so `caught_up` is never set to `false`. But `events_rx` has replay sign requests that should have been buffered. Result: replayed sign requests are forwarded to signing with `caught_up = true`.

### 2. Stale events from the previous connection

Even without coalescing, `events_rx` may still have unconsumed sign/respond events from *before* the reconnect. When the `biased` select reads `state_rx` first and sets `caught_up = false`, that only affects the *next* `events_rx.recv()`. But those stale events were already in the channel buffer with the expectation that `caught_up = true` (they were produced while live). Now they get processed with `caught_up = false` — they go to backlog instead of being forwarded to signing, which is incorrect in the opposite direction. Because the two channels are independent, there's no way to know which events in `events_rx` belong to "before the state change" vs "after."

---

Both problems stem from the same root cause: two separate channels cannot provide a total ordering of their messages. The new design avoids this entirely by putting `CatchupInProgress` on the **same ordered mpsc channel** as the sign/respond events. Since it's the same channel, it can't be coalesced or reordered — `CatchupInProgress` appears at exactly the right position in the event stream: after any stale live events and before any replay events. This also lets us remove `state_rx` from `run_stream` entirely (and from `ChainPipeline`, since it had no other production consumer), simplifying the select to a plain event loop.

ppca:
I'll resolve merge conflicts and fix whatever is broken. don't review yet

chain-signatures/chain-solana/src/indexer.rs
```diff
@@ -71,7 +71,27 @@ pub struct SolanaIndexer<S: StateManager, T: ChainTelemetry> {
...
+    }
+}
+
+impl Drop for AbortOnDrop {
+    fn drop(&mut self) {
+        if let Some(handle) = self.0.take() {
+            handle.abort();
+        }
+    }
 }
```
ChaoticTempest: nit: should move this to utils or mpc-primitives since it's pretty general primitive

chain-signatures/chain-solana/src/indexer.rs
```diff
@@ -71,7 +71,27 @@ pub struct SolanaIndexer<S: StateManager, T: ChainTelemetry> {
...
-    live_task: Option<tokio::task::JoinHandle<()>>,
+    pub live_task: Option<tokio::task::JoinHandle<()>>,
+}
+
+struct AbortOnDrop(Option<tokio::task::JoinHandle<()>>);
+
+impl AbortOnDrop {
+    fn new(handle: tokio::task::JoinHandle<()>) -> Self {
+        Self(Some(handle))
+    }
```
ChaoticTempest: we should make this `AbortOnDrop::spawn` or add it so we don't need to call tokio::spawn in addition

chain-signatures/chain-solana/src/indexer.rs
```diff
@@ -71,7 +71,27 @@ pub struct SolanaIndexer<S: StateManager, T: ChainTelemetry> {
...
+
+    fn into_handle(mut self) -> tokio::task::JoinHandle<()> {
```
ChaoticTempest: Should just rename to take and return an option since you wrap it around an optional anyways

chain-signatures/chain-integration-core/src/indexer.rs
```diff
@@ -3,6 +3,13 @@ use std::time::Duration;
 use futures_util::Stream;
 use mpc_primitives::{Chain, ChainEvent};
 
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum LiveStreamStatus {
```
ChaoticTempest: Maybe ChainStreamStatus might be better since it could be used for the whole of the stream instead of just livestreaming

chain-signatures/node/src/stream/pipeline.rs
```diff
@@ -308,14 +329,92 @@ impl<I: ChainIndexer> ChainPipeline<I> {
...
-                // Watchdog: if no block is processed within the chain-specific timeout,
-                // the background producer is assumed stalled and the pipeline restarts.
                 _ = tokio::time::sleep(timeout) => {
                     return self.handle_watchdog_timeout("live block").await;
                 }
             }
         }
     }
+
+    async fn handle_reconnect(&mut self) -> Option<ChainStreaming> {
```
ChaoticTempest: Hmm, I would expect us to not have a specific handle reconnect branch at all. In my head, reconnect should trigger recovery which then retriggers catch-up then live branches. So we don't need to repeat any logic. So is this actually required then to have this? Are we doing something special with this? If so, then can we add it to the recovery branch?

ChaoticTempest (commented):
Did a single pass so might have missed some things

chain-signatures/node/src/stream/pipeline.rs
ppca: Agree it is more elegant to use handle_recovery directly and it is also correct.
And I was doing a different handle_reconnect mostly since reconnect is for the case where the indexer stalled for a short while like 60s, and the node has memory where it was left off so I intentionally did not want handle_reconnect to load and restore from a checkpoint and catch up from the checkpoint. Especially when we only store checkpoint every 20-30 mins on Solana previously, if handle_reconnect start from a checkpoint, it would be 20-30 mins back in history.
Also catching up from a checkpoint means we repeatedly process some blocks for sure, and it is alright when it is necessary for example in case of recovery, but I was hoping to avoid it when it is not necessary.

ppca:
looks like the previous timing issue with select! between state change and events are gone in the new design, so we don't need the CatchupInProgress and pipeline_caught_up now.
I simply changed the loop of `subscribe_to_program_events` to ending if it errs so live_tx gets dropped and run() will bail and run_supervised will restart the indexer.run().
@ChaoticTempest

ChaoticTempest (approved):

PR #930, opened 2026-06-29
feat: create `utils` within `mpc-indexer-core`

isSerge:
This PR consolidate existing utils re-used across different indexers and publishers under `mpc-indexer-core` crate

- Move retry utils (`retry_rpc!` and related structs) to `mpc_indexer_core::utils::retry`
- Create `mpc_indexer_core::utils::hashing`:
  - Replace `ethabi_request_id` with `compute_request_id`. This implementation is utilizing `alloy-primitives` and `alloy_sol_types`, both crates are small, seems cleaner)
  - Replace `hash_rlp_data` with `hash_payload` (same implementation)
  - Add unit test for both utils
- Update `Chain` imports to use `mpc_primitives` in relevant modules

This finally enables decoupling of indexers and publishers from the `mpc-node`


chain-signatures/indexer-core/src/utils/hashing.rs
```diff
@@ -0,0 +1,199 @@
+use alloy_primitives::{keccak256, U256};
+use alloy_sol_types::SolValue;
+
+/// Computes a request ID based on the provided parameters. The request ID is generated by encoding the parameters using ABI encoding and then hashing the result using Keccak256.
+#[allow(clippy::too_many_arguments)]
+pub fn compute_request_id(
```
ChaoticTempest, line 6: so this could be pretty dangerous changing the underlying code to compute request id. Since this was moved to using different alloy functions, we should have it match some golden values in testing such that we can confirm nothing has changed from underneath us. This is pretty important as this would conflict with our sign id mapping, and that changes how we assign sign tasks and do checkpoints. I don't see any tests for this, so it would be great if we can have some before merging this PR

isSerge: We have `request_id_matches_ethabi` tests for Hydration and Solana, which I used to verify new implementation is functionally the same. Can add similar test for Ethereum and Canton too

ChaoticTempest: ahh that's right, we had these before. Just didn't pop up in the changes. We don't need it for ethereum then, so this should be good to go

ChaoticTempest (approved):

PR #941, opened 2026-07-01
hydration indexer: reconnect if no events received for a while

### LINKED ISSUE #940: hydration: reconnect when remote websocket gets dropped
when the remote WS server closes the connection (many Substrate/Polkadot RPC nodes enforce max connection lifetimes around 12-24 hours), the hydration indexer will hang.

my plan here is to do a simple fix first, and once we've merged the change for reconnect catchup https://github.com/sig-net/mpc/pull/929, we can integrate hydration with that pattern (if hydration needs it): https://github.com/sig-net/mpc/issues/942

ppca:
resolves https://github.com/sig-net/mpc/issues/940
deliberately not integrating with the stream and chain pipeline design yet to keep hydration related stuff light
*Summary*

- The hydration indexer's WS connection silently drops after ~20 hours. When `subscribe_finalized()` returns `None`, the `run()` function exits and the spawned task dies permanently — no reconnect, no logging, no recovery.
- This PR wraps the connection + subscription + block processing in a reconnect loop with a 60s stall watchdog (matching the Solana indexer pattern), so all three drop modes are handled: clean close (`None`), transport error (`Some(Err)`), and silent stall (no data for 60s).
- Also fixes: runtime updater task accumulation across reconnects (now aborted before each retry), and stale `root_pk` (now refreshed from `contract_watcher` on every block).

**Known limitation**

Blocks finalized during the disconnect window are skipped — there is no backfill from the last seen height. Tracked as a TODO for follow-up.

### DIFF
--- a/chain-signatures/node/src/indexer_hydration/mod.rs
+++ b/chain-signatures/node/src/indexer_hydration/mod.rs
@@ -360,10 +360,8 @@ pub async fn run<T: ChainTelemetry>(
 
     let ws_url: &str = hydration.rpc_ws_url.as_str();
     const RECONNECT_DELAY: Duration = Duration::from_secs(5);
+    const STALL_TIMEOUT: Duration = Duration::from_secs(60);
     let mut runtime_updater_handle: Option<tokio::task::JoinHandle<()>> = None;
-
-    // stall watchdog
-    let stall_timeout = Duration::from_secs(60);
 
     loop {
         if let Some(handle) = runtime_updater_handle.take() {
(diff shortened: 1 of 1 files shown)


chain-signatures/node/src/indexer_hydration/mod.rs
```diff
@@ -379,221 +356,287 @@ pub async fn run<T: ChainTelemetry>(
...
-        Ok(blocks) => blocks,
-        Err(e) => {
-            tracing::error!("failed to subscribe to finalized blocks: {e}");
-            return;
+    let ws_url: &str = hydration.rpc_ws_url.as_str();
+    const RECONNECT_DELAY: Duration = Duration::from_secs(5);
+    let mut runtime_updater_handle: Option<tokio::task::JoinHandle<()>> = None;
+
+    // stall watchdog
+    let stall_timeout = Duration::from_secs(60);
```
isSerge: very minor: this could be a constant similar to `RECONNECT_DELAY`

isSerge (approved):

chain-signatures/node/src/indexer_hydration/mod.rs
ppca: done!n added STALL_TIMEOUT

PR #944, opened 2026-07-01
Make treshold dynamic

volovyks:


volovyks:
cc @DavidM-D @auto-mausx, this is an important change for the business side and deployment

volovyks:
I've realized that node-side changes are also required. I'll take a look at that.

ChaoticTempest (approved):

PR #947, opened 2026-07-02
Delta fix

volovyks:
Thanks for finding this, @isSerge !

- moved `derive_delta(...)` to `/crypto`
- fixed https://github.com/sig-net/mpc-private/issues/29
- added some tests on top


chain-signatures/crypto/src/kdf.rs
```diff
@@ -292,6 +292,44 @@ pub fn generate_signature(
...
+) -> Scalar {
+    use hkdf::Hkdf;
+    use near_primitives::hash::CryptoHash;
+
+    // Constant prefix that ensures delta derivation values are used specifically for
+    // near-mpc-recovery with key derivation protocol vX.Y.Z.
+    const DELTA_DERIVATION_PREFIX: &str = "near-mpc-recovery v0.1.0 delta derivation:";
+
+    let hk = Hkdf::<Sha3_256>::new(None, &entropy);
+    let info = format!("{DELTA_DERIVATION_PREFIX}:{}", CryptoHash(request_id));
```
isSerge, line 317: Not related to the fix, but maybe we should consider using more light-weight alternatives to `near-primitives` for trivial things in the future. It seems like a pretty heavy crate, which takes almost 20 sec to compile

isSerge (approved):

chain-signatures/crypto/src/kdf.rs
volovyks: Yeah, can not agree more
Once we move all the chain-specific logic to the integrations layer, we should try to get rid of near and all the other such crates completely

PR #949, opened 2026-07-02
feat: always drain signature posit messages

jakmeier:
Avoid cases where a channel for posit messages is not drained because the SignatureTask is in a different state.

This implements the principle described in #873.

This is a great example because we already had buffered messages for incoming `SignTaskMessage` but we didn't properly decouple the work item queue from the communication channel. Following the principle, this PR always drains the communication channel and moves the responsibility of a classical work item queue to struct SignPositWorkQueue.

This is a proposed solution for the problem discovered in #854. It is a third alternative to already proposed solutions in #864 and [this comment](https://github.com/sig-net/mpc/issues/854#issuecomment-4656890536).


jakmeier:
replaces #874 because the recent refactor made a git merge largely impractical

volovyks (approved):

PR #964, opened 2026-07-06
Simplify bumping rounds in signature generation

volovyks:


chain-signatures/node/src/protocol/request/task.rs
```diff
@@ -188,8 +185,9 @@ impl SignTask {
                     }
                     is_running = self.refresh_governance(&contract_state);
                     if is_running {
-                        // Back in running; reset to Organizing.
-                        phase = SignPhase::Organizing(OrganizingPhase);
+                        // Back in running; the interrupted attempt is abandoned and we
```
volovyks, line 188: This is the only place where we were not bumping the round, and I don't think it's worth the exception, since the contract state change is very rare.

jakmeier (approved):

Issue #952, opened 2026-07-03
indexer: solana block height stall after running for a while


ppca:
<img width="363" height="226" alt="Image" src="https://github.com/user-attachments/assets/d231eb70-454b-4cb8-a1a5-1d3dd5f8b49e" />

ppca:
seems like what had happened was that solana indexer became stuck, no block processed for 303s, then restarting pipeline. Then restarting catchup then nothing. Likely something went wrong during this pipeline restart

<img width="1422" height="536" alt="Image" src="https://github.com/user-attachments/assets/01340ab8-f66b-4f1b-8fa2-3c055c794f3b" />

ppca:
ok I looked into the logs and grafana, and i think this is what most likely has happened.
TLDR: locally accumulated 32 Solana checkpoints because none of the checkpoints get written to near contract, so we are not able to process next block or wait for a checkpoint change.

Observations: 
- Our last checkpoint creation for solana was at 22:16:28 (log `<gcp_link>`), and a bit more than 5 mins later, restarting pipeline because the live block timeout (Log `<gcp_link>`).

- And when we look at grafana, the finalized solana block height gets stuck at ~22:17 (grafana `<grafana_link>`).

- Also when checking logs, I'm not seeing "failed to persist consensus checkpoint" or "consensus checkpoint confirmed and persisted" in logs, this means Solana consensus checkpoints are never being published to the checkpoints_rx watcher. So in `wait_detected_regression`([code](https://github.com/sig-net/mpc/blob/c536a54d608a40610a0189df0750d2906b6b1fb5/chain-signatures/node/src/stream/pipeline.rs#L287)), it will wait ```if checkpoints_rx.changed().await.``` forever.

Theory:
[...]

PR #968, opened 2026-07-07
metric: observe pending checkpoints count

ppca:
part of debugging https://github.com/sig-net/mpc/issues/952.
Suspect that we hit maximum number of pending checkpoints


isSerge (approved):

volovyks:
Metrics aren't for debugging. What is the motivation for having it instead of a log?

ppca:
> Metrics aren't for debugging. What is the motivation for having it instead of a log?

I think it's worthwhile to have this metric so we can monitor if the checkpoints are accumulating for whatever reason

PR #969, opened 2026-07-07
fix: implement watchdog timeout for `ChainPipeline`

isSerge:
- Add `watchdog_timeout` field in `ChainPipeline`, used for both `handle_live` and `handle_catchup`, add relevant tests
- Add `pub async fn pending_checkpoint_count(chain)` for `Backlog` (to provide additional context for warn logs below)
- Add tracing warning for both Ethereum and Solana if no advance for a specified time


chain-signatures/chain-ethereum/src/indexer.rs
```diff
@@ -609,6 +609,11 @@ impl<S: StateManager, T: ChainTelemetry> EthereumIndexer<S, T> {
         let mut last_final_block_number: Option<BlockNumber> = None;
         let mut consecutive_failures = 0u32;
 
+        // Warn if the finalized block number has not advanced for this long
+        const STALL_WARN_SECS: u64 = 600;
```
volovyks: Is it enough for Mainnet Ethereum finality?

volovyks: I feel like we need to generalize this and use `Chain::expected_finality_time_secs()`

isSerge: One Ethereum epoch is ~6.4 min, so we warn only if gap is clearly more than one epoch. Can increase to 900 sec (cover 2 epochs) if warnings get too noisy. I think early warning is safer

ppca:
We already have watchdog timeout and restart for the hydration and for solana websockets, those should already fix a stalled client. We shouldn't need to timeout and restart the chain pipeline.
Also this won't fix the current stall because after watchdog timeout and pipeline restart, we will still be stuck because we still have pending_checkpoints filled and no checkpoint change on near contract.

isSerge:
@ppca I agree that timeout alone won't fix the stall due to pending checkpoints, but it is still valuable for other potential cases as a standard defense. We might enhance it with `has_checkpoint_slot` context at fire-time so it is clear which layer is stuck.

ppca:
> @ppca I agree that timeout alone won't fix the stall due to pending checkpoints, but it is still valuable for other potential cases as a standard defense. We might enhance it with `has_checkpoint_slot` context at fire-time so it is clear which layer is stuck.

makes sense. I'll approve but please update the timeout using the finality time as Serhii mentioned because on sepolia it can take 20+ minutes per epoch sometimes that's why we had a long finality time set for ethereum.

chain-signatures/chain-ethereum/src/indexer.rs
ppca: sepolia can go up to 20+ mins.

ppca (approved):

ppca (approved):

Issue #977, opened 2026-07-08
checkpoint: solana checkpoint too sparse


ppca:
solana checkpoints can be unsteady and sparse, as opposed to ~5min for eth.

<img width="1118" height="335" alt="Image" src="https://github.com/user-attachments/assets/509150a5-cedc-4f74-aea5-fa11a4ed91eb" />

ppca:
The Solana indexer only emits ChainEvent::Block(slot) for slots where the MPC program is mentioned. If the slot that's an exact multiple of 120 doesn't have an MPC transaction, it's never seen by the indexer and that checkpoint boundary is skipped entirely. The next checkpoint can only be created at the next multiple of 120 that happens to have an MPC transaction.

For example, during a non-load-test period, we send 1 solana sign request per min. 

<img width="555" height="311" alt="Image" src="https://github.com/user-attachments/assets/92a21b4e-edb9-4f3c-9d35-efd83c52b2f9" />

This plus some response/other program transactions, the indexer sees maybe ~4 distinct slots per minute out of ~150 total. With interval=120, the chance of any visible slot being an exact multiple is 1/120, so:
- ~4 visible slots/minute × 1/120 probability = 1 checkpoint every ~30 minutes

<img width="556" height="318" alt="Image" src="https://github.com/user-attachments/assets/7a049e4a-37a8-46aa-90f0-8e05ace4bc3f" />

Eth doesn't have this problem because it processes every block, so it always hits the exact multiples of 20.

That's why Solana checkpoints accumulate slowly and in irregular bursts caused by traffic change — the indexer has to get lucky that a transaction lands on a slot that's % 120 == 0.

ppca:
This is a bigger problem when we are on mainnet where requests are very sparse

PR #980, opened 2026-07-09
 fix: checkpoint on interval boundary crossing instead of exact multiple

ppca:
## Summary

Solana indexer only observes slots containing program-relevant transactions may never land on an exact multiple of the checkpoint interval. With sparse traffic, a pending request could sit uncheckpointed indefinitely because no visible slot happened to be divisible by 120. https://github.com/sig-net/mpc/issues/977

This PR replaces `height.is_multiple_of(interval)` with a bucket-crossing check (`height / interval > prev / interval`) so that a checkpoint is created the first time a visible slot crosses any interval boundary, regardless of the exact slot number.

### Root cause

The Solana indexer uses a WebSocket subscription filtered to transactions that mention the MPC program (`logs_subscribe` with `Mentions(program_id)`). Unlike Ethereum which processes every finalized block, the Solana indexer only emits `ChainEvent::Block(slot)` for slots where the program is mentioned. With the previous exact-multiple check, a checkpoint was only created when the observed slot number was exactly divisible by the interval (120). On devnet with ~4 visible slots per minute, the probability of hitting an exact multiple was ~1/30, resulting in roughly one checkpoint every 30 minutes instead of the expected ~48 seconds.

### Caveat

This does not checkpoint every sparse request immediately. If the last processed height and the new height are in the same interval bucket, the request waits until the next observed height crosses a boundary. On restart/recovery, the node resumes from the latest confirmed checkpoint and replays only the post-checkpoint same-bucket tail.

## Test plan

- [x] `test_boundary_crossing_sparse_request_waits_until_next_bucket` — 480→500 same bucket (no checkpoint), 500→600 crosses (checkpoint)
- [x] `test_boundary_crossing_same_bucket_no_checkpoint` — 121 crosses, 130 stays in same bucket
- [x] `test_boundary_crossing_within_first_bucket_no_checkpoint` — first observed slot 50 stays in bucket 0
- [x] `test_boundary_crossing_exact_multiple_still_checkpoints` — 119→120 exact multiple still works
- [x] `test_boundary_crossing_first_observed_height_above_interval` — first ever slot 500 with no previous height
- [x] Existing `test_automatic_checkpoint_on_interval` and `test_automatic_checkpoint_solana_interval` still pass

### DIFF
--- a/chain-signatures/node/src/backlog/mod.rs
+++ b/chain-signatures/node/src/backlog/mod.rs
@@ -535,6 +535,7 @@ impl Backlog {
         interval: u64,
     ) -> Option<Checkpoint> {
         let mut pending = self.pending(&chain).write().await;
+        let prev = pending.processed_block_height().unwrap_or(0);
         pending.set_processed_block(height);
 
         tracing::trace!(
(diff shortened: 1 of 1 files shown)


jakmeier (commented):
Looks like this would work for our current setup in testnet. I'm not opposed to merging it as is, I think it would work to solve #977 right now.

But what if we don't have much traffic in mainnet, wouldn't we still have huge gaps in checkpoints? I wonder, could we instead force-process an empty slot at every interval threshold? That would guarantee we get the checkpoint regardless of traffic.

volovyks (approved):
@jakmeier, the issue is that on some chains we aren't processing blocks without signature requests. Not sure if that can be changed.
I'm ok with the current strategy, although we agreed with @ChaoticTempest on what we have now. @ChaoticTempest, please share your thoughts. The current approach is also deterministic. I do not have any concerns.

jakmeier:
> the issue is that on some chains we aren't processing blocks without signature requests. Not sure if that can be changed.

@volovyks I would expect we can somehow listen to latest block height on any chain, thus generate said "empty" events on a fixed interval

volovyks:
@jakmeier, that will require a separate stream of "latest block" events from indexing to the backlog layer. Not sure if it's worth the hassle
cc @isSerge

PR #985, opened 2026-07-09
refactor: remove redundant `eth_getBlockByNumber` re-fetch for finalized blocks

isSerge:
This PR removes the redundant per-block `eth_getBlockByNumber(Number)` re-fetch from the catchup path for blocks that were already finalized at fetch time. The original reorg check is kept for live processing (defense-in-depth) and for catchup blocks that were unfinalized at fetch time.

Main changes:

- Add `is_finalized` param for `process_block` and `emit_processed_block` (`true` for catchup blocks at or below the sampled finalized head, `false` otherwise; always `false` for live)
- Add `block_timestamp` field to `BlockAndRequests`
- Add `catchup_finalized_head: AtomicU64` field to `EthereumIndexer` and sample the finalized head once in `catchup_range`
- Add unit tests: 
  - `catchup_does_not_refetch_block_for_emission` - verifies zero `eth_getBlockByNumber(Number)` for catchup
  - `catchup_unfinalized_block_keeps_reorg_check` - verifies a catchup block above the finalized head re-fetches once and emits when hashes match
  - `live_process_refetches_block_for_reorg_check` - verifies the live path re-fetches once and emits when hashes match
  - `live_process_skips_emission_when_block_reorged` - verifies the live path skips emission when the re-fetched hash differs
  
Benchmark (100 blocks range):

- wall time: -5.9%
- blocks_per_sec: +6.2%
- total_rpc: -33%
- total_http: -48.5%

### DIFF
--- a/chain-signatures/chain-ethereum/src/indexer.rs
+++ b/chain-signatures/chain-ethereum/src/indexer.rs
@@ -32,7 +32,10 @@ fn live_blocks_channel() -> (mpsc::Sender<MaybeBlock>, mpsc::Receiver<MaybeBlock
 
 pub struct BlockAndRequests {
     block_number: u64,
-    /// Block timestamp captured from the block we already fetched
+    block_hash: alloy::primitives::B256,
+    /// Block timestamp captured from the block we already fetched, so
+    /// `emit_processed_block` doesn't need to re-fetch the block just to read
+    /// it back. Used for `ChainEvent::SignRequest`'s `block_timestamp` field.
     block_timestamp: u64,
     indexed_requests: Vec<IndexedSignRequest>,
     respond_logs: Vec<Log>,
@@ -42,6 +45,7 @@ pub struct BlockAndRequests {
 impl BlockAndRequests {
     fn new(
         block_number: u64,
+        block_hash: alloy::primitives::B256,
         block_timestamp: u64,
         indexed_requests: Vec<IndexedSignRequest>,
         respond_logs: Vec<Log>,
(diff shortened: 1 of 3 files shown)


chain-signatures/chain-ethereum/src/indexer.rs
```diff
@@ -180,11 +186,16 @@ impl<S: StateManager, T: ChainTelemetry> EthereumIndexer<S, T> {
     }
 
     /// Process the block and emit relevant ChainEvents from the block.
-    async fn process_block(&self, block: &Block) -> anyhow::Result<()> {
+    ///
+    /// `is_finalized` indicates the block is already finalized (true for the
+    /// catchup path over historical history). When set, `emit_processed_block`
```
ppca: catchup blocks can be unfinalized yet tho...like every time we restart, it starts from the last checkpoint, which is typically the last finalized block, and the blocks after that are not finalized yet...

isSerge: then we could fetch current finalized block once per catchup and skip re-fetch for `block_number <= finalized`

jakmeier (approved):
Nice optimization for long catchup ranges!

This makes me wonder though, why are we even fetching non-final blocks?

I believe the optimistic path at this point is largely a leftover from when we didn't have finality checks implemented at all. Not sure if we have any use for it now or in the future.

So, maybe we should look into removing the optimistic path. That would mean we can always batch-process blocks when an epoch finalizes. I imagine that would simplify the code and be more efficient.

Just an idea for a potential follow-up. WDYT @isSerge ?

isSerge:
@jakmeier I did not dive into why we need optimistic path yet, but removing it would definitely simplify a lot. Will put it on my list

PR #986, opened 2026-07-10
refactor: batch `eth_getBlockReceipts` during catchup

isSerge:
This PR replaces individual per-block `eth_getBlockReceipts` requests with a single batch:

- Introduce `CatchupItem` enum (`BatchBlock` | `Missing` | `LiveBlock`) to make sure batch block cannot exist without pre-fetched receipts
- Add new RPC method `get_block_receipts_batch`. `HeliosEthereumClient` is using `join_all` (same as `get_blocks`)
- Update `CatchupIter` to fetch both block and receipts batches and zip them
- Update `process_catchup` to use new `CatchupItem::BatchBlock` which provides both block and receipts without I/O. `CatchupItem::Missing` falls back to individual `get_block` and `get_block_receipts` requests
- Live path: 
  - `process()` matches on `LiveBlock` variant and does single `eth_getBlockReceipts` and passes receipts to `process_block()
  - `parse_block`, `process_block`, `collect_execution_confirmations` now all accept `receipts: &[TransactionReceipt]`
- New unit tests added:
  - `catchup_iter_receipt_batch_order_preservation` (`client.rs`)
  - `catchup_batch_receipts_single_batch_post` - assert exactly 1 `eth_getBlockReceipts(batch)` POST (not 32 singles)
  - `catchup_batch_receipts_preserves_order` - assert each block gets its own receipts
  - `catchup_missing_block_falls_back_to_single_receipt_fetch` - assert `process_catchup` does 1 `get_block` + 1 single `eth_getBlockReceipts` for `Missing` block
  - `catchup_batch_receipts_null_treated_as_empty` - assert the block emits `ChainEvent::Block` with empty vector if block is empty
  - `live_process_fetches_receipts_singly` - confirms live isn't accidentally batched
  
This reduces total number of HTTP requests from 105 to 9 (-91%) for 100 block range


ppca (approved):
Nice optimization!

Issue #987, opened 2026-07-13
eth indexer: checkpoint accumulate after catchup


ppca:
solana does not seem to have the same problem.

ppca:
Canton seeing same issue. 
The issue is that if the same request has any field that differs in [BacklogEntry](https://github.com/sig-net/mpc/blob/7a259e555994e07fac1e72060323e109df2e01ce/chain-signatures/node/src/backlog/mod.rs#L778) across nodes, then this will serialize to a different transaction in the checkpoint digest, hence cause different nodes to have different signIds for the same checkpoint. 

<img width="1432" height="681" alt="Image" src="https://github.com/user-attachments/assets/a1b97fda-ca95-43cc-b096-292c179657a7" />

This is particularly problematic when the nodes index requests at different timestamps. Eth typically suffer this almost always, because in IndexedSignRequest, the indexed timestamp field round to sec, and nodes can easily index same request at seconds away. 
This will cause no signature generated for the checkpoint forever, and the pending checkpoints will reach CAP locally and block any further blocks to be processed. 
Looking into a fix now.

ChaoticTempest:
good catch on the timestamps being included, im surprised we got any `respond_checkpoint` at all on our governance contract

ppca:
The checkpoints in dev near contract need to be cleared because after we reset checkpoint storage version in https://github.com/sig-net/mpc/pull/990, nodes can't find the contract's checkpoint from any peer and they are stuck forever

volovyks:
Closing — fixed by excluding node-local state from the checkpoint digest (#990) and making the digest depend only on source-chain-observable info (#1102). The dev-contract checkpoint cleanup mentioned in the last comment was a one-off ops task.

PR #990, opened 2026-07-13
fix: exclude node-local state from checkpoint consensus digest

ppca:
https://github.com/sig-net/mpc/issues/987
Checkpoint::digest() hashed the full CBOR-serialized BacklogEntry (including unix_timestamp_indexed and SignStatus), causing each node to produce a different sign_id for the same logical checkpoint when indexer lag among nodes is bigger than 1 sec.
Hash only sign_id.request_id so nodes agree on the pending request set regardless of local progress state. 

Recovery is unaffected: the near contract only stores the consensus digest (chain, height, hash), not the checkpoint payload. During recovery, the full Checkpoint (including PendingTx.transaction bytes) is loaded from local persistent storage or fetched from a peer — never from the contract. recover_by_checkpoint() then deserializes transaction bytes to restore BacklogEntry (SignArgs, SignStatus, execution watchers) exactly as before. The digest change only affects what nodes must agree on for consensus, not what is stored or transferred for recovery.

NOTE: need to run ```near call dev.sig-net.testnet reset_checkpoint '{"chains": ["Ethereum", "Solana", "Canton"]}' --accountId dev.sig-net.testnet``` once this merges.

### DIFF
--- a/chain-signatures/node/src/sign_bidirectional.rs
+++ b/chain-signatures/node/src/sign_bidirectional.rs
@@ -17,19 +17,6 @@ pub struct PublishState {
     pub is_proposer: bool,
 }
 
-impl PublishState {
-    fn digest_bytes(&self, tag: u8) -> Vec<u8> {
-        let mut bytes = vec![tag];
-        bytes.extend_from_slice(&self.signature.to_bytes());
-        bytes.extend_from_slice(&(self.participants.len() as u32).to_le_bytes());
-        for participant in &self.participants {
-            bytes.extend_from_slice(&participant.bytes());
-        }
-        bytes.push(u8::from(self.is_proposer));
-        bytes
-    }
-}
-
 #[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 pub enum SignStatus {
     PendingGeneration,
@@ -49,21 +36,6 @@ impl SignStatus {
 
     pub fn is_pending_execution(&self) -> bool {
         matches!(self, SignStatus::PendingExecution { .. })
-    }
-
-    pub fn digest_bytes(&self) -> Vec<u8> {
-        match self {
-            SignStatus::PendingGeneration => vec![0],
-            SignStatus::PendingPublish { publish } => publish.digest_bytes(1),
-            SignStatus::PendingExecution { tx } => {
-                let mut bytes = vec![2];
-                bytes.extend_from_slice(tx.id.0.as_slice());
-                bytes.extend_from_slice(&tx.target_chain.to_bytes());
-                bytes
-            }
-            SignStatus::PendingGenerationBidirectional => vec![3],
-            SignStatus::PendingPublishBidirectional { publish } => publish.digest_bytes(4),
-        }
     }
 
     pub fn checkpoint_consensus_bytes(&self) -> Vec<u8> {
--- a/chain-signatures/node/src/sign_bidirectional.rs
+++ b/chain-signatures/node/src/sign_bidirectional.rs
@@ -66,6 +66,21 @@ impl SignStatus {
         }
     }
 
+    pub fn checkpoint_consensus_bytes(&self) -> Vec<u8> {
+        match self {
+            SignStatus::PendingGeneration => vec![0],
+            SignStatus::PendingPublish { .. } => vec![1],
+            SignStatus::PendingExecution { tx } => {
+                let mut bytes = vec![2];
+                bytes.extend_from_slice(tx.id.0.as_slice());
+                bytes.extend_from_slice(&tx.target_chain.to_bytes());
+                bytes
+            }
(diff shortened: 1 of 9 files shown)


chain-signatures/primitives/src/backlog.rs
```diff
@@ -68,7 +68,6 @@ impl Checkpoint {
         hasher.update(self.block_height.to_le_bytes());
         for pending in &self.pending_requests {
             hasher.update(pending.sign_id.request_id);
-            hasher.update(&pending.transaction);
```
ChaoticTempest, line 71: There may an unlikely case where we fetch a checkpoint that have the same request but each node has observed two different status for it. Like node A may have already see it is in watching execution meanwhile node B is still awaiting the first signature response. In this case, they would appear the same. So we probably need to also hash the sign status with this

ChaoticTempest: This wouldn't be a concern if we have a guarantee in event ordering, but indexer code can be incorrect and we may end up missing events in some nodes. Like how we currently don't handle the case where solana disconnects and reconnects to live, missing a couple events in the mean time

ppca: seems like there's pros and cons to adding sign status: 
pro: we can be sure not to mistaken the state of a pending request
cons: it may happen a lot that 2 nodes disagree on the checkpoint because some requests are at different status due to timing. 
But I guess if we cannot reach consensus with same sign state then essentially there's not true consensus on the checkpoint.
I'll add sign status

chain-signatures/node/src/sign_bidirectional.rs
```diff
@@ -66,6 +66,21 @@ impl SignStatus {
         }
     }
 
+    pub fn checkpoint_consensus_bytes(&self) -> Vec<u8> {
```
ppca, line 41: @ChaoticTempest not directly using SignStatus::digest() because that contains proposer info and especially fields like `is_proposer`, which will definitely differ among nodes

chain-signatures/primitives/src/backlog.rs
```diff
@@ -23,6 +23,13 @@ pub struct PendingTx {
     pub sign_id: SignId,
     #[serde(with = "serde_bytes")]
     pub transaction: Vec<u8>,
+    /// Stable checkpoint status bytes for this request, used for consensus.
+    ///
+    /// Kept separate from `transaction` (the full CBOR-serialized entry) so
+    /// that nodes can agree on cross-node request progress without hashing
+    /// node-local fields such as indexing timestamps or proposer state.
+    #[serde(default, with = "serde_bytes")]
+    pub checkpoint_status: Vec<u8>,
```
ChaoticTempest: it's not ideal to store it in each PendingTx. How about we just store it as a combination of all sign status bytes within `Checkpoint` itself? It's only used for the checkpoint digest anyways

chain-signatures/node/src/sign_bidirectional.rs
ChaoticTempest: we can get rid of hashing the proposer info for sign status. I don't remember if there was a reason to have it in there

chain-signatures/primitives/src/backlog.rs
ppca: good call. I'll remove it from PendingTx, and do this:
```
pub struct PendingCheckpointRequest {                                                                                    
        pub pending: PendingTx,                                                                                              
        pub checkpoint_status: Vec<u8>,                                                                                      
    }                                                                                                                        
                                                                                                                             
    pub struct Checkpoint {                                                                                                  
        pub chain: Chain,                                                                                                    
        pub block_height: u64,                                                                                               
        pub pending_requests: Vec<PendingCheckpointRequest>,                                                                 
    }             
```

chain-signatures/node/src/sign_bidirectional.rs
ppca: removing the digest_bytes altogether since it's not used anywhere, and this new function has a name with a clear purpose

chain-signatures/primitives/src/backlog.rs
ChaoticTempest: I think we'd still have the same issue with this one. Can't we just do:
```
struct Checkpoint {
    pub chain: Chain,
    pub block_height: u64,
    pub pending_requests: Vec<PendingTx>,
    pub cumulative_digest: [u8; 32],
}
```

So on checkpoint creation, we can just pre-hash the sign status of each request and set the cumulative_digest

ppca: Changed to 
```
struct Checkpoint {
    pub chain: Chain,
    pub block_height: u64,
    pub pending_requests: Vec<PendingTx>,
    pub cumulative_digest: [u8; 32],
}
```

ChaoticTempest (approved):
this looks ready to go now, nice work

PR #996, opened 2026-07-13
Message channels refactoring

volovyks:
- docs
- renamings
- separated mod.rs into a couple of files


volovyks:
Restarted because of a flaky Ethereum test.
timeout waiting for two nodes to converge on a checkpoint after one comes back online (fixes are coming from other people)

chain-signatures/node/src/cli/mod.rs
```diff
@@ -236,6 +236,8 @@ pub async fn run(cmd: Cli) -> anyhow::Result<()> {
             );
             crate::metrics::nodes::CONFIGURATION_DIGEST.set(digest);
 
+            // SignCommand channel: chain sign events (requests, completions,
+            // chain aborts) from the indexers/streams into the SignatureSpawner.
             let (sign_tx, sign_rx) = mpsc::channel(16384);
```
isSerge, line 241: `16384` seems like a magic number, worth moving to a constant same as sibling channels (`MAX_MESSAGE_INCOMING`, `MAX_MESSAGE_OUTGOING`, etc.)

chain-signatures/node/src/protocol/message/crypto.rs
```diff
@@ -0,0 +1,384 @@
...
+        let from = Participant::from(0);
+        let batch = vec![
+            Message::Triple(TripleMessage {
+                id: 1,
+                epoch,
+                from,
+                data: vec![128u8; 1024],
+                timestamp: 1,
+            }),
+            Message::Triple(crate::protocol::message::TripleMessage {
```
isSerge, line 349: `TripleMessage` and `crate::protocol::message::TripleMessage` seem like the same type

chain-signatures/node/src/protocol/message/crypto.rs
```diff
@@ -0,0 +1,384 @@
...
+                id: 3,
+                epoch,
+                from,
+                data: vec![101u8; 1337],
+                timestamp: 3,
+            }),
+        ];
+
+        let batch_bytesize = batch.iter().map(|msg| msg.size()).sum::<usize>();
+        dbg!(batch_bytesize);
```
isSerge, line 366: probably should not keep `dbg!` in tests

chain-signatures/node/src/protocol/message/inbox.rs
```diff
@@ -0,0 +1,665 @@
...
+            &node_id,
+            root_sk.public_key().into_affine_point(),
+            2,
+            participants,
+        );
+        let (inbox, _outbox, channel) = MessageChannel::new();
+        let inbox = tokio::spawn(inbox.run(config_rx, contract_watcher));
+
+        // Case 1:
+        // Check that the inbox received our messages correctly:
```
isSerge, line 465: God test, probably should extract each case into a separate test

chain-signatures/node/src/protocol/message/outbox.rs
```diff
@@ -0,0 +1,264 @@
...
+    if !current_messages.is_empty() {
+        // Add the last partition
+        partitions.push(Partition {
+            messages: current_messages,
+            timestamp: earliest,
+        });
+    }
+
+    partitions
+}
```
isSerge, line 264: No tests, worth adding in the future

chain-signatures/node/src/protocol/message/inbox.rs
```diff
@@ -0,0 +1,665 @@
...
+        {
+            let batch = vec![
+                Message::Triple(TripleMessage {
+                    id: 1,
+                    epoch,
+                    from,
+                    data: vec![128u8; 1024],
+                    timestamp: 1,
+                }),
+                Message::Triple(crate::protocol::message::TripleMessage {
```
isSerge, line 475: Same as above, can drop fully-qualified path

chain-signatures/node/src/protocol/message/crypto.rs
```diff
@@ -0,0 +1,384 @@
...
+    /// The message with all it's related info.
+    #[serde(with = "serde_bytes")]
+    pub msg: Vec<u8>,
+    /// The signature used to verify the authenticity of the encrypted message.
+    pub sig: Signature,
+    /// From which particpant the message was sent.
+    pub from: Participant,
+}
+
+impl SignedMessage {
```
isSerge, line 26: Minor: we have three `impl` blocks here, probably can collapse into one

isSerge (commented):
Few things caught my eye, I understand all of these are carryovers, so feel free to ignore in this PR, but worth adding some TODO comments to address later

volovyks:
@isSerge thanks! Do you mind if I merge this as is and address all the comments right after that?
There are 2 PRs on top of this one, and I want to avoid conflicts

isSerge (approved):

PR #997, opened 2026-07-13
Collups subscribe/unsubscribe

volovyks:
Review after https://github.com/sig-net/mpc/pull/996 is merged


chain-signatures/node/src/protocol/message/mod.rs
```diff
@@ -176,241 +178,141 @@ impl MessageChannel {
...
-            tracing::warn!(id, "failed to subscribe for triple");
+    /// Subscribe, returning an already-closed receiver on failure so the
+    /// caller's recv loop ends instead of panicking.
+    async fn subscribe_or_closed<T: SubscriptionMessage>(
+        &self,
+        id: SubscribeId,
+        what: &'static str,
+    ) -> mpsc::Receiver<T> {
+        let Some(subscription) = self.subscribe(id).await else {
+            tracing::warn!(what, "failed to subscribe");
```
ChaoticTempest: shouldn't we be including id as well?

ChaoticTempest (approved):
looks good to me besides one minor thing with logging

PR #994, opened 2026-07-13
refactor: update Eth indexer to prefetch batches concurrently

isSerge:
- **Config**: add `EthConfig.catchup_concurrent_batches: usize`, with `DEFAULT_CATCHUP_CONCURRENT_BATCHES` (3) and `MAX_CATCHUP_CONCURRENT_BATCHES` (8)
- **Client**: delete `CatchupIter` struct and add `EthereumClient::catchup_batch_stream(&self, start, end, concurrency)`, which returns a stream of `CatchupItem`
- **Indexer**: `type Iter` changed to `futures_util::Stream`
- **Benchmark**: include `batch_wait_ms` metric for critical-path fetch wait

PR #999, opened 2026-07-14
feat: bloom-filtered logs and targeted tx receipts (Ethereum indexer)

isSerge:
Moving away from fetching `eth_getBlockReceipts` for every single block:

- **Introduce bloom filter gating**: indexer now completely skips fetching logs for blocks that do not contain `contract_address`
- **Switch to `eth_getLogs`**: Replaced the heavier `eth_getBlockReceipts` fetches with targeted `eth_getLogs` requests (based on positive bloom filter matches)
- **Targeted Execution Confirmations**: instead of fetching and processing all receipts, indexer only issues `eth_getTransactionReceipt` for specific pending transactions currently tracked by the `StateManager`

Benchmark changes (100 blocks range against local eRPC with warm cache):
catchup_progress: 0.68s [-81.5%]
blocks_per_sec: 146.4 [+438.2%]
total_rpc: 110 [-45.3%]
batch_fetch_ms: 590 [-83.0%]
process_ms: 4 [-85.7%]

### DIFF
--- a/chain-signatures/chain-ethereum/src/indexer.rs
+++ b/chain-signatures/chain-ethereum/src/indexer.rs
@@ -442,6 +442,8 @@ impl<S: StateManager, T: ChainTelemetry> EthereumIndexer<S, T> {
         // a failed extraction stays pending for retry, not flagged Failed.
         let mut observed_tx_ids = HashSet::new();
 
+        // TODO: submit as batch or parallelize the per-watcher `eth_getTransactionReceipt` calls once
+        // we have a way to track rate-limits and backoff for all requests (right now we just retry individually on failure)
         // Per-watcher `eth_getTransactionReceipt`
         for (tx_id, (sign_id, pending_tx)) in watchers {
             tracing::info!(?tx_id, ?sign_id, "querying receipt for bidirectional tx");
--- a/chain-signatures/chain-ethereum/src/indexer.rs
+++ b/chain-signatures/chain-ethereum/src/indexer.rs
@@ -229,9 +229,7 @@ impl<S: StateManager, T: ChainTelemetry> EthereumIndexer<S, T> {
         Ok(())
     }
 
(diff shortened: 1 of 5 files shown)


hilljessica8787-dot (approved):

chain-signatures/chain-ethereum/src/indexer.rs
```diff
@@ -468,22 +442,9 @@ impl<S: StateManager, T: ChainTelemetry> EthereumIndexer<S, T> {
         // a failed extraction stays pending for retry, not flagged Failed.
         let mut observed_tx_ids = HashSet::new();
 
+        // Per-watcher `eth_getTransactionReceipt`
```
ppca, line 447: when there are a lot of pending tx's,  we could be sending a lot of RPC calls for each block, is there a way to optimize it a bit?

ppca (commented):
Mostly LGTM. Just one comment

chain-signatures/chain-ethereum/src/indexer.rs
isSerge: by optimize you mean send all in parallel instead of sequentially to reduce waiting time?

ppca: or maybe they can be batched? Either sounds ok to me

isSerge: We can do that, but I am worried about our CU per sec budget on Alchemy, `eth_getTransactionReceipt` is relatively expensive. Currently it's sequential and we await and retry each request individually - it's slower, but safer. In order to make it batched or parallel we need to have some mechanism to track rate limits for all requests, so we don't thunder RPC. I planned to implement this in #994. Can add a TODO comment here if that works for you

ppca: yeah sounds good! let's add a todo

ppca (approved):

PR #1002, opened 2026-07-15
feat: implement nonce-gated execution confirmation polling

isSerge:
Re-architect how Ethereum indexer confirms cross-chain execution requests. 

Original solution involved polling `eth_getTransactionReceipt` for every pending watcher on every block. This was breaking with thousands of watchers (tested 5000 watchers against local eRPC, see new benchmark below).

New solution employs multi-tiered strategy:

1. **In-block transaction list check** (no RPC requests): For every incoming block check if any pending watchers transaction hash is included. Handles majority of successful executions
2. **Nonce-gated polling**: Secondary check for watchers not found in the current block.  `eth_getTransactionReceipt` is performed only if current sender nonce has advanced. Watchers are grouped by sender address ensuring only one `eth_getTransactionCount` RPC call per unique address. This is triggered for every 10th block (waiting for ~120 sec for sad path seems like an acceptable trade-off).
3. **Individual `eth_getTransactionReceipt`**: Receipts are only requested for transaction identified by Tier 1 or Tier 2. Fetching is executed concurrently (using `futures::try_join_all`)

This PR also introduces benchmark for watchers `examples/bench_watchers.rs`:
```bash
RPC_URL=http://localhost:4000/sepolia/evm/11155111 \
CONTRACT_ADDRESS=0x69C6b28Fdc74618817fa380De29a653060e14009 \
START=11214938 END=11215038 WATCHERS=5000 \
RUST_LOG=mpc_chain_ethereum::bench=info \
cargo run --example bench_watchers --features bench
```

### DIFF
--- a/chain-signatures/chain-ethereum/src/indexer.rs
+++ b/chain-signatures/chain-ethereum/src/indexer.rs
@@ -2234,4 +2234,503 @@ mod tests {
         nonce_mock.assert_async().await;
         receipt_mock.assert_async().await;
     }
+
+    #[tokio::test]
+    async fn new_watcher_resolves_at_appearance_block() {
+        let mut server = Server::new_async().await;
+
+        let from_address = address!("f39fd6e51aad88f6f4ce6ab8827279cfffb92266");
+        let tx_hash = b256!("4444444444444444444444444444444444444444444444444444444444444444");
+        let block_hash = b256!("6e4e53d1de650d5a5ebed19b38321db369ef1dc357904284ecf4d89b8834969c");
+        let receipt_response = json!({
+            "transactionHash": format!("{tx_hash:#x}"),
+            "blockHash": format!("{block_hash:#x}"),
+            "blockNumber": "0x3",
+            "transactionIndex": "0x0",
+            "from": format!("{from_address:#x}"),
+            "to": format!("{from_address:#x}"),
+            "gasUsed": "0x5208",
+            "effectiveGasPrice": "0x3a29f0f8",
+            "contractAddress": null,
+            "logsBloom": format!("0x{}", "0".repeat(512)),
+            "cumulativeGasUsed": "0x5208",
+            "type": "0x2",
+            "logs": [],
+            "status": "0x0"
+        });
+
+        // Nonce check pinned at block 5 (NOT a tick): one-shot
+        // appearance gate fires immediately for newly seen watchers
+        let nonce_mock = server
+            .mock("POST", "/")
+            .match_body(Matcher::PartialJson(json!({
+                "method": "eth_getTransactionCount",
+                "params": [format!("{from_address:#x}"), "0x5"]
+            })))
+            .with_status(200)
+            .with_header("content-type", "application/json")
+            .with_body(json!({ "jsonrpc": "2.0", "id": 1, "result": "0x1" }).to_string())
+            .expect(1)
+            .create_async()
+            .await;
+
+        // Receipt shows the tx already mined at block 3 (before registration)
+        let receipt_mock = server
+            .mock("POST", "/")
+            .match_body(Matcher::PartialJson(json!({
+                "method": "eth_getTransactionReceipt",
+                "params": [format!("{tx_hash:#x}")]
+            })))
+            .with_status(200)
+            .with_header("content-type", "application/json")
+            .with_body(json!({ "jsonrpc": "2.0", "id": 1, "result": receipt_response }).to_string())
+            .expect(1)
+            .create_async()
+            .await;
+
+        let builder = test_utils::TestIndexerBuilder::new(server.url());
+        builder
+            .state_manager
+            .watch_execution(
+                Chain::Ethereum,
+                SignId::new([4; 32]),
+                BidirectionalTx {
+                    id: BidirectionalTxId(tx_hash.0),
+                    sender: [0u8; 32],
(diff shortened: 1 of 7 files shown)


volovyks (commented):
This is a very interesting PR that needs a couple of questions to be answered first. 

Overall, I like the approach, and it allows us to clean up our watching list, but:
- Not sure if our backlog is ready for that. Maybe we want to keep things simple and "keep watchers forever". At least that was the plan. But I agree that nonce increment evidence is good enough to remove it from the backlog.
-  Looking back in time (blocks) can also be tricky for consensus. Also, we should design our state machine to disallow moving to the next block without all watching transactions in the backlog.
cc @ChaoticTempest

volovyks:
I thought about it a bit more.

There should be no harm in looking back in time if that is done deterministically (same block height and same depth on each node). That would also help us identify backlog problems (we should log when we find something "in the past").

Security-wise, I cannot come up with an attack vector, since the signature is unpredictable (?) by the caller and included (in the case of Ethereum) in the transaction. But this is something we need to think about. cc @scafuro

chain-signatures/chain-ethereum/src/indexer.rs
```diff
@@ -420,97 +421,143 @@ impl<S: StateManager, T: ChainTelemetry> EthereumIndexer<S, T> {
...
+                                tx_id,
+                                sign_id,
+                                &pending_tx,
+                                block_number,
+                            )
+                            .await?;
+                        Ok::<_, anyhow::Error>(outcome)
+                    });
+
+            future::try_join_all(futures)
```
ppca: this would fail the whole thing if just one future inside failed. Would it be better to switch to join_all with per item logging?

chain-signatures/chain-ethereum/src/indexer.rs
```diff
@@ -420,97 +421,143 @@ impl<S: StateManager, T: ChainTelemetry> EthereumIndexer<S, T> {
...
-                        ?tx_id,
-                        ?sign_id,
-                        ?err,
-                        "Failed to fetch nonce for bidirectional tx"
-                    );
-                    continue;
-                }
-            };
+        // Throttle the nonce check to every 10 blocks (Detects replaced/dropped txs)
+        if block_number % 10 == 0 && !unmined_watchers.is_empty() {
```
ppca: If some watcher failed for example due to rpc problems in this attempt, its next attempt is 10 blocks later?  When thinking about the checkpoint consensus among nodes, if we polled every block, divergence would be corrected in 1 block, with this change, it would be 10 blocks.
When a Eth chain reorg happens, nodes may observe the reorg at different block height, and that difference will also take 10 blocks to resolve.
This larger correction/convergence window would likely also mean  divergence among nodes happen more often in this larger window and it might be hard to reach consensus, and much more regression than desired because checkpoint differs more often.

chain-signatures/chain-ethereum/src/indexer.rs
isSerge: good point, adjusted

chain-signatures/chain-ethereum/src/indexer.rs
isSerge: Let me change this to retry failed RPC calls on the next block. That way the 10-block throttle only applies to the watcher-less nonce-consumption edge case (which should not happen in practice). Even if it does fire, ticks are block-aligned and nonce reads are pinned to the tick block's historical state, so all nodes compute the identical outcome. The only divergence source is RPC failures at the tick (next-block retry bounds it to ~1 block). Reorgs are unaffected: forked payloads are dropped at the finality stage and recomputed from the canonical block, and mined/replaced detection derives from block content, not the tick.

ppca: looks good then!

ppca (approved):

volovyks (approved):
I've talked with @isSerge and realized that my understanding was off. We are not removing requests from the backlog based on nonce (although we can in the future), and we are already checking for transaction hashes in the past. 
cc @ChaoticTempest , take a look please

chain-signatures/chain-ethereum/src/indexer.rs
```diff
@@ -412,101 +437,295 @@ impl<S: StateManager, T: ChainTelemetry> EthereumIndexer<S, T> {
...
+            .into_iter()
+            .filter(|(tx_id, _)| sweep_all || new_watchers.contains(tx_id) || retry.contains(tx_id))
+            .collect();
+
+        let (gated_events, gated_failed) = self.resolve_gated_watchers(gated, block_number).await;
+        events.extend(gated_events);
+        failed.extend(gated_failed);
+
+        // Failed watchers are retried on the next block.
+        self.watcher_gate.lock().unwrap().retry = failed;
```
ChaoticTempest: I'm a bit sketched out about the unwraps for these mutexes. Can't we just use tokio's async mutexes so we don't have this problem?

ChaoticTempest (commented):
Nice work, everything looks good to me besides the one thing I mentioned

chain-signatures/chain-ethereum/src/indexer.rs
isSerge: agreed. tokio would force all methods to be async, I added a helper for poison recovery using `std` primitives instead

PR #1018, opened 2026-07-19
refactor: Ethereum indexer migration to a new design

isSerge:
This PR focuses on migration of Ethereum indexer to a new trait design and creates shared supervision logic on the node side. 

Main changes:
- `ChainIndexer` trait: add default implementation for `run()` - returns `Err` so an unmigrated chain can't be silently
   supervised (After migration all default implementations will be removed)
- Create `stream/supervisor.rs` - includes all node-side logic: recovery, regression detection, watchdog, etc. (chain-specific logic has been moved to `run()` method - see  `EthereumIndexer` changes below). Should probably become part of `ChainPipeline` eventually, but for now keep separate, because original `ChainPipeline` is still used by Canton and Solana (TODO added)
- Update `cli/mod.rs` to use `supervisor::run_supervised` instead of `run_stream` for Ethereum
- Create `stream/recovery.rs` - extract backlog recovery logic from `stream/pipeline.rs`
- Update `EthereumIndexer`:
  - add `run` method: anchor retry-loop -> `catchup_blocks()` -> `CatchupCompleted` -> spawns `index_live_blocks` -> `process_live_retrying`. All retry loops are cancel-aware
  - add `AbortOnDrop` wrapper to prevent task leaking
  - remove `livestream/next/catchup_range/process_catchup/process/notify_catchup_completed` methods (currently covered by default trait implementations, but none is used)
  - remove struct fields: `catchup_complete`, `events_tx`, `live_blocks_rx`
  - deleted `EthereumStream` struct

### DIFF
--- a/chain-signatures/chain-ethereum/src/indexer.rs
+++ b/chain-signatures/chain-ethereum/src/indexer.rs
@@ -33,6 +33,7 @@ fn live_blocks_channel() -> (mpsc::Sender<MaybeBlock>, mpsc::Receiver<MaybeBlock
     mpsc::channel(MAX_LIVE_BLOCK_BUFFER)
 }
 
+// TODO: Move this to some common crate so it can be shared across workspace
 /// Aborts the wrapped task on drop, prevents leaking a background task
 /// if the indexer is dropped while the live block fetcher is still running.
 struct AbortOnDrop(tokio::task::JoinHandle<()>);
--- a/chain-signatures/chain-ethereum/src/indexer.rs
+++ b/chain-signatures/chain-ethereum/src/indexer.rs
@@ -634,7 +634,6 @@ impl<S: StateManager, T: ChainTelemetry> EthereumIndexer<S, T> {
     /// Catchup blocks in `[processed + 1, anchor_height)` fetched in batches.
     /// Samples the finalized head once at catchup start, so blocks at or below
     /// it can skip the per-block re-fetch + reorg hash check.
-
     pub async fn catchup_blocks(
         &self,
         anchor_height: u64,
--- a/chain-signatures/chain-ethereum/src/indexer.rs
+++ b/chain-signatures/chain-ethereum/src/indexer.rs
@@ -761,7 +761,8 @@ impl<S: StateManager, T: ChainTelemetry> EthereumIndexer<S, T> {
         Ok(())
     }
 
-    /// Shared body of the legacy `process` and the `run()` live loop.
+    /// Process a single live item and emit its events, keeping the reorg hash
+    /// check (live blocks are not yet finalized).
     async fn process_live_item(
         &self,
         events_tx: &mpsc::Sender<ChainEvent>,
--- a/chain-signatures/chain-ethereum/src/indexer.rs
+++ b/chain-signatures/chain-ethereum/src/indexer.rs
@@ -966,8 +966,13 @@ mod tests {
         BidirectionalTx, BidirectionalTxId, Chain, ChainEvent, ExecutionOutcome, SignId,
         LATEST_MPC_KEY_VERSION,
     };
+    use mpc_chain_integration_core::StateManager;
     use serde_json::json;
-    use std::sync::atomic::Ordering;
+    use std::sync::atomic::{AtomicU64, Ordering};
+    use std::sync::Arc;
+    use tokio::sync::mpsc;
+    use tokio::time::Duration;
+    use tokio_util::sync::CancellationToken;
 
     #[tokio::test]
     async fn missing_catchup_block_is_refetched() {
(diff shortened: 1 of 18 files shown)


volovyks (approved):

integration-tests/tests/cases/ethereum_stream.rs
```diff
@@ -376,50 +383,41 @@ async fn next_event_within<S: StateManager, T: ChainTelemetry>(
...
-    // Wait until the pipeline is live so the subscription and anchor are established
+) -> Result<StartedEthereumIndexer> {
+    let indexer =
+        EthereumIndexer::new(ctx.config(true), backlog.clone(), NoopChainTelemetry).await?;
+    let (events_tx, mut events_rx) = chain_event_channel();
+    let cancel = CancellationToken::new();
+    let run_task = tokio::spawn({
+        let cancel = cancel.clone();
+        async move { indexer.run(events_tx, cancel).await }
+    });
```
ChaoticTempest, line 403: nit: 
```suggestion
    let run_task = tokio::spawn(indexer.run(events_tx, cancel.clone()));
```

chain-signatures/chain-integration-core/src/indexer.rs
```diff
@@ -21,7 +23,11 @@ pub trait ChainIndexer: Send + 'static {
         Ok(())
     }
 
-    async fn catchup_range(&self, anchor_height: u64) -> Self::Iter;
+    // TODO: this is only used in the legacy ChainStream pipeline, remove it once possible
+    async fn catchup_range(&self, anchor_height: u64) -> Self::Iter {
+        let _ = anchor_height;
+        unimplemented!("catchup_range is only driven by the legacy ChainStream pipeline")
+    }
```
ChaoticTempest, line 30: one of the issues with having some stuff enabled and some disabled is that things start to get entangled, especially if implementations are separate from each other and could take time. I recommend we create an entirely separate ChainIndexer trait in a separate module so we don't have this issue. That way we can incrementally move to this new design and not have rough edges like this with `unimplemented!()`

chain-signatures/chain-ethereum/src/indexer.rs
```diff
@@ -33,6 +33,16 @@ fn live_blocks_channel() -> (mpsc::Sender<MaybeBlock>, mpsc::Receiver<MaybeBlock
...
+/// Aborts the wrapped task on drop, prevents leaking a background task
+/// if the indexer is dropped while the live block fetcher is still running.
+struct AbortOnDrop(tokio::task::JoinHandle<()>);
+
+impl Drop for AbortOnDrop {
+    fn drop(&mut self) {
+        self.0.abort();
+    }
+}
+
```
ChaoticTempest, line 38: this is a common pattern that @ppca also added in her PR, so we should probably put this in utils or primitives so we can reuse when needed

chain-signatures/node/src/stream/supervisor.rs
```diff
@@ -0,0 +1,438 @@
...
+                        RegressionOutcome::Recovery => {
+                            break Exit::Restart;
+                        }
+                        RegressionOutcome::Aligned => {}
+                        RegressionOutcome::Shutdown => break Exit::Shutdown,
+                    }
+                }
+                // Watchdog: restart when no `ChainEvent::Block` was observed within
+                // the per-chain timeout (other event kinds do not count).
+                _ = tokio::time::sleep_until(last_block_event + watchdog_timeout) => {
```
ChaoticTempest, line 116: Is this how we normally do watchdog and restart? Feels like we'd run into frequent restarts with no events happening on the chain

chain-signatures/node/src/stream/supervisor.rs
```diff
@@ -0,0 +1,438 @@
...
+                        // Err (or panic) is treated as a crash and restarted.
+                        break match (&mut run_handle).await {
+                            Ok(Ok(())) => Exit::Shutdown,
+                            Ok(Err(err)) => {
+                                tracing::warn!(?err, %chain, "chain run() failed; restarting");
+                                tokio::time::sleep(ERROR_RESTART_DELAY).await;
+                                Exit::Restart
+                            }
+                            Err(err) => {
+                                tracing::warn!(?err, %chain, "chain run() panicked; restarting");
```
ChaoticTempest: Does this branch need a sleep delay as well just like the other ok(err) branch? If so then we can collapse them to one single branch

chain-signatures/chain-ethereum/src/indexer.rs
```diff
@@ -216,6 +204,7 @@ impl<S: StateManager, T: ChainTelemetry> EthereumIndexer<S, T> {
     /// For the live path, `is_finalized` is always `false`
     async fn process_block(
         &self,
+        events_tx: &mpsc::Sender<ChainEvent>,
         block: &Block,
```
ChaoticTempest, line 201: Instead of passing the events_tx each time, we can make the chainindexer run return a events_rx such that each chain creates the events channel themselves and internalizes it

ChaoticTempest (commented):
Did a cursory look at it, will need to do another more in depth look later

integration-tests/tests/cases/ethereum_stream.rs
isSerge: clippy is not happy about this: `indexer` does not live long enough

chain-signatures/chain-ethereum/src/indexer.rs
isSerge: The plan is for supervisor to own the channel (mailbox), which is re-created fresh on restart, while the chain indexer just owns the loop. Also, `run` loops until cancelled, it cannot return `events_rx` immediately. 

Instead of passing the `events_tx` around we could update methods to just return events and do `events_ts.send(...)` in one place (probably in `run`), but I would prefer to do this refactor in a separate PR because it would require updating a lot of tests

chain-signatures/node/src/stream/supervisor.rs
isSerge: good point, updated

chain-signatures/node/src/stream/supervisor.rs
isSerge: we should not if we keep getting `ChainEvent::Block`, this is the same mechanism we had before

chain-signatures/chain-ethereum/src/indexer.rs
isSerge: I have added a TODO comment for this. There are some other helpers and wrappers I was planning to move to a common crate, probably need to create `mpc-utils` for that

chain-signatures/chain-integration-core/src/indexer.rs
isSerge: It's definitely not great, but the plan is to migrate fast - we just need 2-3 PRs to conclude and do final trait cleanup.

chain-signatures/node/src/stream/supervisor.rs
ChaoticTempest: I haven't seen the latest state of our code, but last I remember, we had optimizations for each of the indexers to only emit events only when the contract had an event for it. So only the corresponding `ChainEvent::Block` associated to those events would be emitted.

Just reading https://github.com/sig-net/mpc/pull/980 now, I think we still have this issue for solana. We are still only getting block events for events based on the contract. So this assumption might be incorrect. We can leave it for another issue/PR but if so, please make an issue and tag it here

isSerge: I believe this is no longer the case
Ethereum: `emit_processed_block` unconditionally sends `ChainEvent::Block` for every processed block
Solana: 
```rust
                            // block is not relevant to our program, skip but still
                            // emit block event for progress tracking
                            if let Err(err) = events_tx.send(ChainEvent::Block(slot)).await {
                                tracing::warn!(?err, "failed to send block event");
                            }
                            continue;
```

ChaoticTempest: for solana, that's what I'm talking about. The `ChainEvent::Block(slot)` here is sent for every filtered program_id transaction. If you take a look at how we have:
```
    let filter = RpcTransactionLogsFilter::Mentions(vec![program_id.to_string()]);
    let config = RpcTransactionLogsConfig {
        commitment: Some(CommitmentConfig::confirmed()),
    };

    let (mut stream, _unsubscriber) = pubsub_client.logs_subscribe(filter, config).await?;
```

isSerge: I will take a closer look when I start Solana migration, this PR only covers Ethereum

ChaoticTempest (approved):
Overall looks good to me, if we can move fast and none of this becomes much crutch then it's fine

PR #1019, opened 2026-07-20
refactor: adjust `eth_optimistic_requests` CLI flag

isSerge:
- Use `anvil` for integration tests, which sets `optimistic_requests: true`
- Update comment for `MPC_ETH_OPTIMISTIC_REQUESTS` / `eth_optimistic_requests` CLI args to clarify demo / dev use
- Add `warn!` if using optimistic path with `mainnet`


ppca:
i know Applications team sometimes ask us to set this to true for dev so they can do demos...so might be good to keep around

isSerge:
@ppca restored, added clarification and `warn!` if used with `mainnet` network

ppca (approved):

Issue #1007, opened 2026-07-17
Ensure we are able to reset checkpoints and restart from a set height


ChaoticTempest:
We need to add in the ability to have the checkpoints be reset on the NEAR contract. This will reset the checkpoint on both the NEAR contract and the node's storage. A height will be specified to tell all nodes to start their indexers at a specific height. This should remain as an admin function for now, where we will add voting for gradually. This is not a huge issue as this does not affect signing

ChaoticTempest:
resolved in https://github.com/sig-net/mpc/pull/1176

PR #1026, opened 2026-07-21
add workflow to reset checkpoint per chain on dev contract

ppca:
Can be used by team to reset checkpoints on contract when dev is down. Also save the need to pass the dev contract keys around.


volovyks:
Can be handy. But I want @ChaoticTempest to say whether this aligns with his strategy.

ChaoticTempest:
right now, reseting the checkpoint on the contract would still leave the nodes in limbo. We need to implement https://github.com/sig-net/mpc/issues/1007 first before we can do this pipeline

PR #1027, opened 2026-07-21
Verify message senders

volovyks:
At first, I wanted to move the `if from == me { continue; }` check into a single place. However, I then realized that we never verify whether an incoming message was actually sent by the claimed signer. This means an attacker could resend messages originating from other nodes.

cc @jakmeier — security-related.


jakmeier (approved):
Nice catch!

The inner `from`  needs to be verified, of course, but I guess we missed that and thought the envelop authentication checks are enough. 😅

PR #1029, opened 2026-07-21
Remove redundant sign posit channel

volovyks:
Now handle_posit pushes directly into the per-sign queue, which also serves as the pre-spawn buffer — one less buffer, one less task per request, no lossy hop, and memory tightens from 4096 messages per sign ID to one per participant


jakmeier (approved):
Oh wow, this simplification looks so obvious now :D The indirection seems utterly unnecessary.

PR #1031, opened 2026-07-21
Remove contract-state handling from SignTask

volovyks:
- Removed contract-state handling from SignTask. Such lightweight, short-lived tasks should not be aware of it. I'm also thinking about recreating the whole SignatureSpawner, not to track contract state there as well; happy to get your opinion.
- SignTasks are recreated on `not_running => running` and killed on `running => not_running`
- Added `next_governance()` to avoid triggering Spawner without a good reason

### DIFF
--- a/chain-signatures/node/src/protocol/request/mod.rs
+++ b/chain-signatures/node/src/protocol/request/mod.rs
@@ -11,7 +11,7 @@ use crate::protocol::contract::primitives::intersect_vec;
 use crate::protocol::message::{MessageChannel, PositMessage, PositProtocolId, Subscriber};
 use crate::protocol::posit::{PositAction, PositRejectReason, SinglePositCounter};
 use crate::protocol::presignature::PresignatureId;
-use crate::protocol::{Chain, ProtocolState};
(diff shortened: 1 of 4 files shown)


chain-signatures/node/src/protocol/request/mod.rs
```diff
@@ -148,6 +161,29 @@ impl SignatureSpawner {
...
+        cfg: ProtocolConfig,
+    ) {
+        let sign_id = request.id;
+        tracing::info!(?sign_id, "spawning signature task");
+
+        let is_proposer = self
+            .requests
+            .get(&sign_id)
+            .map(|entry| Arc::clone(&entry.is_proposer))
+            .unwrap_or_default();
```
jakmeier: This `unwrap_or_default` allocates a new Arc, which isn't a valid fallback.

Maybe `.expect()` instead? I don't think we should ever hit the None case here, so an explicit unwrap/panic might be better than staying silent if this bug occurs somehow.

jakmeier (approved):
Looks good to me, good refactor!

PR #1035, opened 2026-07-22
refactor: enhance `EthConfig`

isSerge:
- Stricter types: `EthConfig` now holds `PrivateKeySigner`, `reqwest::Url`, `Address` instead of `String`
- Drop all `.expect()`, `.unwrap()`, `format!("0x{}")` in `indexer.rs` and `publisher.rs`
- Validate config once at startup


volovyks (approved):
Nice!

PR #1046, opened 2026-07-22
feat: checkpoint voting instead of signing for consensus

### LINKED ISSUE #1042: Consider voting over signing for checkpoint consensus
right now, we use the MPC signing process to basically vote and reach consensus on a checkpoint. But this prioritizes signing for checkpoints over regular sign requests. This takes up additional resources such as triples/presignatures, and in the case it fails, we need to burn more of them. Voting for a specific checkpoint hash/digest is much simpler and does not burn such resources

ChaoticTempest:
Resolves https://github.com/sig-net/mpc/issues/1042.

This removes checkpoint signing through the MPC signing flow in favor of contract based threshold voting. This reduces the complexity of requiring the sign flow, while needing to prioritize checkpoint sign requests over regular sign requests. If we have a series of checkpoint votes that fails, we don't keep burning rounds and wasting presignatures with this voting.

One new great design with this PR is we now have a new migration flow for each of our production environments (Devnet, Testnet, Mainnet), where we keep tabs of what state each has. A test is added to check against what is on each network to make sure our migrate function works as intended.


chain-signatures/contract/src/lib.rs
```diff
@@ -832,52 +779,68 @@ impl VersionedMpcContract {
...
-            return Err(CheckpointError::InvalidSignature.into());
-        }
+    ) -> Result<bool, Error> {
+        let voter = self.voter()?;
+        let threshold = match self.state() {
+            ProtocolContractState::Running(state) => state.threshold,
+            _ => return Err(InvalidState::ProtocolStateNotRunning.into()),
+        };
 
         if let Some(existing) = self.checkpoints().get(&checkpoint.chain) {
```
volovyks: I would rename `checkpoints()` to `latest_checkpoints()`, it would be much easier to read

chain-signatures/contract/src/lib.rs
```diff
@@ -832,52 +779,68 @@ impl VersionedMpcContract {
...
-            if existing.checkpoint.height > checkpoint.height {
-                return Ok(());
+            if existing.height > checkpoint.height {
+                return Ok(true);
```
volovyks: The meaning of `Ok(true)` vs `Ok(false)` is not well defined. Here we return `Ok(true)`. Such a response tells the caller that "your proposed checkpoint is fine and accepted", but in reality, the contract does not know if that is so. That may create confusion in the future. I would consider returning an enum.

chain-signatures/contract/src/lib.rs
```diff
@@ -832,52 +779,68 @@ impl VersionedMpcContract {
...
-                checkpoint,
-                signature,
-            },
-        )]);
-        Ok(())
+        let vote_count = {
+            let checkpoint_votes = self.mutable_checkpoint_votes();
+            for (candidate, voters) in &mut checkpoint_votes.votes {
+                if candidate.chain == checkpoint.chain && candidate.height == checkpoint.height {
+                    voters.remove(&voter);
```
volovyks: Why is it necessary to do the cleanup?
If it is necessary, let's document why.

chain-signatures/contract/src/lib.rs
```diff
@@ -832,52 +779,68 @@ impl VersionedMpcContract {
...
+            voters.len()
+        };
+
+        if vote_count < threshold {
+            return Ok(false);
+        }
+
+        self.mutable_checkpoints()
+            .insert(checkpoint.chain, checkpoint);
+        self.mutable_checkpoint_votes()
```
volovyks: `vote_checkpoint is complex`. If all the logic is necessary, I would add docs or move some of it into separate functions.

chain-signatures/contract/src/lib.rs
```diff
@@ -1145,28 +1106,51 @@ mod tests {
         V0(OldMpcContractTest),
     }
 
+    #[derive(BorshDeserialize, BorshSerialize)]
```
volovyks: Would we remove this after the Mainnet release?

volovyks: Why is it called *Test?

volovyks (commented):
Overall, I'm not against moving away from MPC signatures to voting at the contract level. But I would not say that saving Ps is a good reason to make that move. Maybe race conditions are. WDYT? Would that make our lives easier?

As for migration logic, I do not think that awareness of our specific environments is something that needs to be in the repo. Why can't we use the same logic for all environments? I'm not sold on the change. And it deserves a separate PR.

ChaoticTempest:
> But I would not say that saving Ps is a good reason to make that move. Maybe race conditions are. WDYT? Would that make our lives easier?

It's just a plus that we're not burning as much resources as before, but the main take away for this is less moving pieces here where things will not as fail as easily. If we have a lot of sign requests, they won't stall our process for checkpoint consensus, and we won't likely run into lagging behind checkpoints waiting to reach consensus. Overall, less complexity with this.

> As for migration logic, I do not think that awareness of our specific environments is something that needs to be in the repo. Why can't we use the same logic for all environments? I'm not sold on the change. And it deserves a separate PR.

So we do need to be aware of what each network has, because we may have varying changes on each network in relation to the state. This PR changes the state on devnet but wasn't sure if testnet already had changes relating to checkpoint consensus. For these types of situations, we should really be aware as doing a deployment later and not knowing if we can call `migrate` or not would be a real pain. Any dev making changes to contract state must be aware going forward

volovyks:
> So we do need to be aware of what each network has, because we may have varying changes on each network in relation to the state. This PR changes the state on devnet but wasn't sure if testnet already had changes relating to checkpoint consensus. For these types of situations, we should really be aware as doing a deployment later and not knowing if we can call migrate or not would be a real pain. Any dev making changes to contract state must be aware going forward

I was sure that contract versioning is the answer to that. You always migrate from vX to vY.

chain-signatures/contract/src/lib.rs
ChaoticTempest: reasoning is to remove the voter for any competing checkpoint at the same height, but maybe not necessary operation

chain-signatures/node/src/rpc/mod.rs
```diff
@@ -765,10 +904,172 @@ mod tests {
...
+        let dispatch = tokio::spawn(async move {
+            RpcExecutor::dispatch_loop(&publishers, None, &mut rx).await;
+        });
+        tokio::time::sleep(Duration::from_millis(100)).await;
+
+        assert_eq!(call_count.load(Ordering::SeqCst), 0);
+
+        drop(tx);
+        dispatch.await.unwrap();
+    }
```
jakmeier: This test is flaky on my machine, it passes only around 50% of the runs.

I think there is a race between the cancellation and `execute_publish` in the `select!`. I don't think it's a problem in production code but for CI we should make at least the test deterministic.

But then I wonder, why do we even need a test asserting that cancellation overrules a pending publish action? I would suggest to remove this test, `executor_aborts_publish_on_chain_abort` seems sufficient.

chain-signatures/contract/src/lib.rs
ChaoticTempest: will remove some of these functions since we have the more direct poll from mainnet, testnet and devnet test

ChaoticTempest:
it's not the best, because then we would need to keep track of V0 to V5 to V9. We would still need to understand which version is associated to which network. What i have in this PR with per network is more explicit. In a separate PR, I'm probably going to get rid of VersionedMpcContract altogether in favor of explicit network tracking

chain-signatures/node/src/rpc/mod.rs
```diff
@@ -37,8 +38,14 @@ const UPDATE_INTERVAL: Duration = Duration::from_secs(10);
 const PUBLISH_MIN_DELAY: Duration = Duration::from_secs(5);
 const PUBLISH_MAX_DELAY: Duration = Duration::from_secs(60); // Cap to 1 min so backoff doesn't get too long for infinite retries
 
+#[allow(clippy::large_enum_variant)]
```
jakmeier, line 45: nit: consider adding a comment why we don't care about a large enum here.

Since we send it over mpsc channels, it will increase the memory usage, so I looked at it briefly.

 ```
PublishAction = 728 bytes
VoteCheckpoint  =  64 byts
AbortChain  =  1 bytes
 ```

Boxing `PublishAction` would reduce it from 728 to 72 bytes. And the eager allocation of mpsc channel is limited by `MAX_CONCURRENT_RPC_REQUESTS = 1024`. So the eager mpsc allocation uses < 1MB, so I guess that's fine.

Also, copying around these values is not on a hot path.

So in conclusion, the memory overhead is < 1MB and copying enum values around is not performance critical, thus allowing a large enum makes sense to me. (But it wasn't obvious to me, so maybe a comment would make sense.)

jakmeier (commented):
Yeah I guess signing checkpoints doesn't add much if we end up submitting them to a on-chain contract anyway. I didn't anticipate that differing checkpoints will be a problem, as **in theory**, it should never happen. Sadly, we need it to work in practice. too. 🙃 

This makes the process a bit more dependent on the governance contract, when previously it was only an optimization but we could have safely recovered from a checkpoint from any untrusted source, without querying any on-chain data.

Other than that, I like the direction this is going and I support this change.

Please take a look at the flaky test, though, as mentioned in my inline comment.

chain-signatures/node/src/rpc/mod.rs
ChaoticTempest: sounds good, I'll leave a comment. Thanks for digging through the large size

chain-signatures/node/src/rpc/mod.rs
ChaoticTempest: I removed it since it was redundant where the other tests should already be covering it

chain-signatures/contract/src/lib.rs
ChaoticTempest: removed

chain-signatures/contract/src/lib.rs
ChaoticTempest: documented it further

chain-signatures/contract/src/lib.rs
ChaoticTempest: Make it purely `Ok(true)` if accepted or already already accepted and `Ok(false)` if more votes are required. The rest of the cases are returned as errors

chain-signatures/contract/src/lib.rs
ChaoticTempest: done

chain-signatures/contract/src/lib.rs
```diff
@@ -1099,29 +1091,23 @@ impl VersionedMpcContract {
...
-        for (chain, signed_checkpoint) in checkpoints {
-            self.mutable_checkpoints().insert(chain, signed_checkpoint);
+            Self::V0(mpc_contract) => {
+                mpc_contract.latest_checkpoints.insert(chain, checkpoint);
+            }
         }
     }
 
-    #[private]
-    pub fn reset_checkpoint(&mut self, chains: Vec<Chain>) {
```
ppca, line 1122: why is reset_checkpoint removed?

chain-signatures/node/src/stream/supervisor.rs
```diff
@@ -205,6 +215,12 @@ async fn run_supervised_with_watchdog<I: ChainIndexer, T: ChainTelemetry>(
             }
         };
 
+        // On requiring restart of the supervisor, we should abort all RPC tasks where
+        // we are trying to vote the checkpoints into
+        if matches!(exit, Exit::Restart) {
```
ppca: seems like this means whenever an indexer gets stuck we will abort the checkpoints for that chain. I feel like this is alright since we checkpoint at least once every 5-6 mins for each chain, but just wondering why this is necessary?

volovyks (approved):
Decoupling signature generation and backlog makes total sense.
Please clean up the contract in the follow-up PR or split the decoupling and migration updates.

chain-signatures/contract/src/lib.rs
ChaoticTempest: removed on accident, added back

chain-signatures/node/src/stream/supervisor.rs
ChaoticTempest: I don't remember the reasoning, but probably better to move it to just on regression/recovery

ppca (approved):

Issue #876, opened 2026-06-11
Do not elect peer in signature generation when peer indexer is stalled


ChaoticTempest:
The issue that was fixed with https://github.com/sig-net/mpc/pull/870 exposed a weakness in our system where we will try to participate with nodes that are stalled on particular chains and cannot really participate in signature generation.

I propose we fix this by adding back the MeshState::stable where the proposer will only add peers to be a part of the signature generation if they are relatively close enough to their own block height. Additionally, proposers can only participate if they themselves are live and have not been stalled for a while. Nodes can self-report that they are stalled just based on how they are processing the block and if they have moved in height or not

volovyks:
I'm skeptical about adding any new mechanics unless they have a solid justification. I don't think this one does.

At a high level, we have a problem of choosing a set of participants for the protocol. It can be solved either by asking all of them whether they are ready to participate in a specific protocol or by collecting all the data required to make the decision ourselves.

I'm in favor of the first option (which is our current posit approach). I'm not even sure we need the active part.

Collecting other nodes' state to make a decision opens a Pandora's box of state synchronization problems.

We recently talked about keeping things simple and accepting the right trade-offs. I think this is exactly that kind of situation. Under normal circumstances, we don't have this issue. When there is a node that is stuck and is 10,000 blocks behind, it should be able to reject all posit requests, since they are lightweight (or at least they should be).

ChaoticTempest:
@volovyks well to be fair, your first option also adds a new mechanism where we need to confirm with the indexer. In terms of complexity, they are similar.

> Collecting other nodes' state to make a decision opens a Pandora's box of state synchronization problems.

How so? I would think that my proposal is fairly similar to us just checking against the active set. We just have extra conditions on top of the active set that needs to be met now with stable. This would also be a passive thing, so nodes don't need to relearn this info on every new sign request.

This also reduces our need to send out a message which I'd argue is less chance of things going wrong, because the peer would have to be able to respond or process it in some manner. So, I'd say this would be less complex than responding to the posit message.

> When there is a node that is stuck and is 10,000 blocks behind, it should be able to reject all posit requests, since they are lightweight (or at least they should be).

Not necessarily lightweight, because to process a posit message requires jumping at least one level of indirection (the message layer), into the actual processor (signature spawners). The spawner would then have to be coupled to the indexers to watch for stalls and then be able to bounce the posit message with a reject.

[...]

Issue #814, opened 2026-05-19
dev: multiple node stop generating signatures after 1 node restart.


ppca:
At 05-18 around 14:45, node 8 restarted, likely because of quickly decreasing memory.

<img width="553" height="266" alt="Image" src="https://github.com/user-attachments/assets/f03b9f11-1af1-489e-8252-6b67cc8d2b0d" />

Shortly after, ~14:50, node 0, 10, 11: the awaiting generation requests number stalled for both eth and solana

<img width="1164" height="667" alt="Image" src="https://github.com/user-attachments/assets/556b24c8-4057-4df3-8eac-fe1f34a98eae" />

signature generators: node 0 stalled at 05-18 15:01, node 10 stalled at 05-18 15:00, and node 11 stalled at 05-18 15:24.
presig generators: node 0 stalled at 05-18 21:28, node 10 stalled at 05-18 21:30, and node 11 stalled at 05-18 21:56.
triple generators: node 0 stalled at 05-18 21:28, node 10 stalled at 05-18 21:28, and node 11 stalled at 05-18 21:56.

Also only node 11 has been consuming its mine presignatures since 05-18 21:55 nonstop, linear decline, while it's not generating anything. 

<img width="546" height="294" alt="Image" src="https://github.com/user-attachments/assets/76283d71-ad15-4e33-934b-70b830f6bbb9" />

ppca:
I looked at "spawning signature task" log:
node 0: no more logs after 14:55
node 10: no more logs after 14:55
node 11: no more logs after 15:19

ppca:
one likely cause is because node 0, 10 and 11 finished signature generation that involved node 8 in a burst right before node 8 restarts, while other nodes have not completed. When node 8 restarts, the other nodes can still be generating these signatures and they will send and consume posit messages for these SignId's. But node 0, 10, and 11, since they have already completed this request, they will only be receiving these posit messages but not consuming, and they could accumulate to the 4096 limit. And once any signId accumulates to 4096 in SignatureSpawner's inbox, then handle_request() will never be called since  in handle_posit(), the inbox.send() can block forever. 

Not sure if this actually is the case, will do more investigation tmr. If this indeed is the problem, will write a component test for it.

ppca:
This is not only an issue when some node restarts, the investigation result explains why we see a lot of times nodes stop generating signatures a while after starting.
TLDR:
[...]

PR #1047, opened 2026-07-23
feat(signature): log awaited peers on generation abort

yvonneanne:
## What

When a signature generation aborts, log the set of participants we
haven't yet heard from ("awaited") — the peers the round was blocked on.

## Why

Both abort paths in `signature.rs` — the `poke()` protocol error and the
`recv()` timeout/close — logged `sign_id` and `presignature_id` but never
*which* participant the round was waiting on. Under all-participant
signing, one silently-slow node aborts every signature it's a part of,
and neither the logs nor the label-free `SIGNATURE_GENERATOR_FAILURES`
counter name it. That's the diagnostic blind spot behind the signing
stalls in #814 / #876: you can see failures pile up but can't attribute
them to a peer.

This is the cheapest telemetry item on the protocol roadmap ("emit the
`!seen` set at abort") and a prerequisite for deciding whether the more
invasive stall fixes are worth building.

## How

- Track the participants heard from during a run (`seen`, seeded with
  self), pushing each `msg.from` as messages arrive.
- On every abort, emit `awaited = participants \ seen` as a tracing field.
- Extract the set difference into a pure `awaited_peers()` helper with
  unit tests.

## Scope / safety

Pure observability. No protocol, wire-format, or metric change — nothing
that can affect safety (S1–S4) or liveness. Logged as a tracing field
rather than a metric label to avoid per-peer cardinality.

## Testing

- 3 unit tests on `awaited_peers` (subset, all-seen, none-seen).
- `cargo test -p mpc-node`, `cargo fmt --check`, `cargo clippy` all clean.

Refs #814, #876

### DIFF
--- a/chain-signatures/node/src/protocol/signature.rs
+++ b/chain-signatures/node/src/protocol/signature.rs
@@ -124,8 +124,15 @@ impl SignGenerator {
         })
     }
 
-    /// Receive the next protocol message, erroring out on timeout.
-    async fn recv(&mut self) -> Result<SignatureMessage, SignError> {
+    /// Participants (excluding those in `seen`) we are still waiting on: the peers
+    /// an abort is blocked behind. `seen` holds everyone we have already heard from.
+    fn awaited(&self, seen: &[Participant]) -> Vec<Participant> {
+        awaited_peers(&self.participants, seen)
+    }
+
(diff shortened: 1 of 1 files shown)


jakmeier (approved):
LGTM, knowing who to blame for aborted generation seems very useful

PR #1048, opened 2026-07-23
fix(request): stamp reject with rejector's own round

yvonneanne:
## What

When a deliberator rejects a `Propose` from a **behind** proposer (`state.round > peer_round`, `request/posit.rs`), it stamped the reject with the *sender's own* round echoed back instead of its own. So the behind proposer learned nothing about how far the round had advanced and climbed **one round per attempt**, each costing a full posit round.

This stamps the reject with `state.round`. The behind proposer now sees a future round, updates `highest_seen_round`, and catches up in a single `bump_round`. This matches how every other reject in the file already stamps `state.round`; this branch was the lone outlier.

Flagged by Jakob as likely a bug in our discussion this morning.

## Changes

- `request/posit.rs`: stamp the stale-round reject with `state.round` (the fix).
- `protocol/posit.rs`: add a `PositRejectReason::StaleRound` variant (appended, discriminant `4` — wire-compatible) so this cause is distinguishable from a genuine `InvalidRequest` in logs. No consumer branches on it; it's diagnostic only.
- `message/outbox.rs`: broaden the test-only `intercept_outgoing_messages` hook from `feature = "test-feature"` to `any(test, feature = "test-feature")` so the unit test can drain outgoing messages under a plain `cargo test` (CI runs without `test-feature`).
- Unit test asserting the reject carries the rejector's round (verified it fails against the old code: `left: 2, right: 5`).

## Behavior note / tradeoff

With the fix, the reject arrives at the behind proposer as a *future-round* message, so it is **buffered** (updating `highest_seen_round`) rather than counted toward `enough_rejects`. The proposer waits out its posit deadline **once**, then jumps straight to the rejector's round — versus today's one-timeout-*per-round* climb. Strictly fewer timeouts, though not instantaneous.

## Testing

- `cargo test -p mpc-node --lib` — new test + all request-module tests pass.
- Confirmed the test catches the bug by reverting the one-line fix.
- `cargo fmt --check` clean; changed files are clippy-clean.

### DIFF
--- a/chain-signatures/node/src/protocol/request/posit.rs
+++ b/chain-signatures/node/src/protocol/request/posit.rs
@@ -45,6 +45,10 @@ impl PositPhase {
                 // that higher round, or else any peer could force themselves to
                 // be the proposer every time.
                 if state.round > *peer_round {
+                    // Stamp the reject with our own (higher) round, not the
+                    // sender's. A behind proposer then sees a future round,
+                    // updates its `highest_seen_round`, and catches up in one
+                    // `bump_round` instead of climbing one round per attempt.
                     ctx.msg
                         .send(
                             ctx.governance.me,
@@ -53,11 +57,11 @@ impl PositPhase {
                                 id: PositProtocolId::Signature(
                                     sign_id,
                                     *presignature_id,
-                                    *peer_round,
+                                    state.round,
                                 ),
                                 from: ctx.governance.me,
                                 action: PositAction::RejectWithReason(
-                                    PositRejectReason::InvalidRequest,
+                                    PositRejectReason::StaleRound,
                                 ),
                             },
                         )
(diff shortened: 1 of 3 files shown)


chain-signatures/node/src/protocol/request/posit.rs
```diff
@@ -45,6 +45,10 @@ impl PositPhase {
                 // that higher round, or else any peer could force themselves to
                 // be the proposer every time.
                 if state.round > *peer_round {
+                    // Stamp the reject with our own (higher) round, not the
+                    // sender's. A behind proposer then sees a future round,
+                    // updates its `highest_seen_round`, and catches up in one
+                    // `bump_round` instead of climbing one round per attempt.
```
jakmeier: nit: To me, this reads like a comment that's only useful in this specific PR, as it compares the new with the old code. LLMs like adding those comments for reviews but I usually remove these comments before merging. But feel free to keep it if you think it's useful to state explicitly why the own round is chosen over the peer's round.

jakmeier (approved):
Yes this looks like a bug to me, I don't see why we would want to echo back the peer's round.

volovyks (approved):
Nice catch! That can make the consensus on proposer faster.

chain-signatures/node/src/protocol/request/posit.rs
yvonneanne:
```suggestion
```
You're right, thanks for the suggestion

Issue #907, opened 2026-06-23
signature generation: remove checking for active


ppca:
remove the need to check for active participants when selecting proposer because that makes proposer per round not deterministic. Also, posit messages would already tell active or not

volovyks:
I like this direction in our system design — reducing complexity and minimizing the need to know other nodes' state (I've mentioned that in [this related issue](https://github.com/sig-net/mpc/issues/876#issuecomment-4690871382)).

Simply removing `active` without making other changes to our design will not work, since we can select an offline proposer. As a result, the proportion of stuck requests would be equal to the proportion of the offline nodes. Or, if we implement retry on proposer rotation, these requests would be delayed by the rotation time.

We have 3 problems:
- select a proposer (this is the most problematic place)
- select a set of participants (this is where we can rely on Posits safely)
- proposer rotation in case of failure or inaction (connected to the first one)

Related: The outcome of our conversation with @ChaoticTempest and @jakmeier last week concluded the following:
> - Accept any proposer for a valid (indexed) request. Start the protocol with a proposer who reached us first.

Problem: even with [T > N/2](https://github.com/sig-net/mpc-private/issues/28) and not using `active` when choosing the proposer and a set of participants for the protocol, it is still possible that with 2 nodes trying to be proposers would get accepts from different nodes but none of them would reach T.

volovyks:
Problems on the top level:
- (A) Indexing: 
    - (1) Non-indexed requests (should never happen, out of scope (Checkpoints))
    - (2) Delayed requests on some or all nodes (indexer-level optimisations are out of scope; regardless, on the protocol layer, this issue is inevitable)
- (B) Proposer/Participant selection
    - (1) Choose a proposer
    - (2) Choose a set of participants
    - (3) Rotate and retry in case of a failure
    - (4) Awareness of the failure is also a problem because of the long finality time
- (C) Presignature availability (should be good and out of scope here)

My proposition: 
I want to focus on simplicity and guarantee of a response. I'm ok with sacrificing latencies (when some nodes are offline or indexing is slow) and redundant posit traffic.

- Rely on Posits for all indexing (A2) problems. If the node is not ready, it will reject. No `active`, `stable`, or `block height` info from other nodes. No need to sync on those.
- Each request is assigned a queue of proposers that never rotates (based on SignId). No need to sync on that.
[...]

Issue #690, opened 2026-03-05
Investigate: Proposer of previous generator not joining the next


jakmeier:
I observed the following in logs:

- Participant 1 was proposer for a signature, broadcasts START but then times out before the signature finishes.
- Later on, participant 3 becomes the proposer for another attempt with a new set of participants.
- Participant 1 is now deliberator and ACCEPTs the posit from participant 3.
- Participant 3 broadcast START and particpant 0 and 2 (both deliberators) process it within 30ms of each other. Nothing happens on participant 1.

Full scenario: https://github.com/sig-net/mpc/issues/681#issuecomment-4006062800

Task: Try to replicate the scenario with a test. If this can be replicated, debug and fix the issue.

volovyks:
Closing. Not reproducible against the rewritten posit flow (#1051, #1100, #1104, #1185).

Issue #228, opened 2025-02-14
Fault tolerant proposer election and sign subset creation


ChaoticTempest:
Currently, in `SignQueue`, we will first create a `subset` up to threshold number of participants from all `stable` participants, and then elect a `proposer` from this `subset`.

This is not currently fault tolerant where if one of those nodes hop offline, then that particular signature will not be completed unless that node hops back online. This is because when we go to retry the signature generation (for whatever reason like timeout) the same `subset` of participants is used.

Additionally, we also use `stable` to create this subset/proposer which is not ideal, because all nodes would have to agree on the `stable` set of participants. We should probably be used all participants in this case to be as deterministic as possible.

We should come up with a reasonable way to elect a new `proposer` and `subset` in such cases. We have access to `entropy` by all nodes and we can potentially use it along with `retry_count` to form the seed to elect proposer/subset.

@jakmeier let me know your thoughts on this

jakmeier:
First of all, what kind of fault tolerance are we aiming for? Just solve the problem of nodes being offline? Or full fault tolerance?

If you just want fault tolerance against the simple case of nodes being offline sometimes, yeah we can add a patchwork solution to fix it. Your suggestion to increase the `retry_count` and changing to a different set of viable participants sounds good to me. I guess we can do it after a timeout and rely on a stable list to optimize the choice from a pure random choice.

But this would still not be fault tolerant against other problems.

- Every single participant has what I would described as "veto power". A single node trying to boycott or censor a sign request will succeed in doing so. It can work normally otherwise, just not participate properly in the protocol, making it hard to spot.
- Nodes that are faulty in any way other than being completely offline will not be detected reliably. As long as their link to the proposer is up, they might be selected in the set. But they might have problems reaching other nodes, they might use a wrong version, or maybe they are just super slow in processing requests. All reasons that could lead to 100% failure rate with that participant. The current `stable` property is not helping with these issues.

To resolve such problems for good, byzantine fault tolerance would be the gold standard. There are different ways to achieve it. I would suggest we would select redundant participants and ensure the protocol can finish with only a subset of them responding. This would make things more robust (fault tolerant) and possibly improve performance, since we don't have to wait for the slowest participant.

[...]

Issue #182, opened 2025-02-10
List of participants in incoming protocols


volovyks:
### Description
Currently, we are using the list of "active" participants in all incoming protocols.
To eliminate inconsistencies, the list of participants for each protocol should be determined by propose and passed to other nodes.
cc @ChaoticTempest @jakmeier

ChaoticTempest:
the proposer determining the list of participants doesn't really matter since entropy is what determines this list. If even one node is offline, the list is different for a node, and then that means the signature cannot be completed due to consensus of participants being wrong. What we really need in that case is state sync which is much further away than what is currently available. Just so you know, this case is already an issue with or without utilizing entropy to determine the list

PR #1050, opened 2026-07-24
test(integration): demonstrate the divergent-active-views liveness stall

yvonneanne:
## What

A component test that demonstrates a liveness failure of the current proposer election, plus the fixture change needed to express it.

The test belongs to the class of stall logged in #690 ("replicate with a test"). Related: #182 (closed predecessor), #228, #907.

Also drops a stale claim on the sibling staggered-observation test that the system cannot handle an observation delay above the organize/posit timeout — measurement shows it can, on this branch and with the follow-up change.

### Scenario

* 5 nodes, threshold 4. 
* Every node permanently hides one peer (its successor) from its `MeshState`. 
* Each node's active set still holds `threshold` participants (itself plus three peers), so `wait_for_active_participants` never blocks and any stall is attributable to proposer election alone.

### Failure

Election scans the node's own active set for a proposer and moves `state.round` to whichever round that scan lands on. With views disagreeing, nodes land on different rounds and never co-locate: a proposer that jumped ahead broadcasts at its round, deliberators on a lower round buffer that Propose as a future-round message instead of accepting, then time out waiting for a proposal at their own round.

On this branch: no signature within 90s, nothing reaches Start, and zero rejects — the stall is round divergence, not rejection.

The test is `#[ignore]`d so CI stays green. Reproduce with:

```
cargo test -p integration-tests -- --ignored test_divergent_active_views_still_converge
```

It is un-ignored by the election change in the follow-up PR, where it settles.

## Fixture change

The mock network resolved a recipient's encryption key from *the sender's* active set, so shrinking a node's view made it unable to send at all (it panicked on `participants.get(to).expect(...)`). This implements that method's existing `// TODO: support sending to non-active participants` by falling back to the participant set the network started with. The live active set stays the primary lookup, so existing tests are unaffected; the fallback only fires where the old code panicked. Message delivery must not depend on the sender's view of liveness, or a divergent failure detector cannot be modelled at all.

### DIFF
--- a/integration-tests/src/mpc_fixture/fixture_tasks.rs
+++ b/integration-tests/src/mpc_fixture/fixture_tasks.rs
@@ -36,6 +36,10 @@ pub(super) fn test_mock_network(
 ) -> JoinHandle<()> {
     let msg_log = Arc::clone(&shared_output.msg_log);
     let rpc_actions = Arc::clone(&shared_output.rpc_actions);
+    // Participant info as of network start, consulted only for the recipient's
+    // encryption key. A node's active set may shrink during a test; delivery in
+    // the mock network must not depend on the sender's view of liveness.
+    let initial_participants = mesh.borrow().active().clone();
 
     tokio::spawn(async move {
         tracing::debug!(target: "mock_network", "Test message executor started");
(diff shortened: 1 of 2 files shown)


integration-tests/src/mpc_fixture/fixture_tasks.rs
```diff
@@ -36,6 +36,10 @@ pub(super) fn test_mock_network(
 ) -> JoinHandle<()> {
     let msg_log = Arc::clone(&shared_output.msg_log);
     let rpc_actions = Arc::clone(&shared_output.rpc_actions);
+    // Participant info as of network start, consulted only for the recipient's
+    // encryption key. A node's active set may shrink during a test; delivery in
+    // the mock network must not depend on the sender's view of liveness.
+    let initial_participants = mesh.borrow().active().clone();
```
yvonneanne:
```suggestion
    // Participant info as of network start, consulted for recipient's encryption key
    let initial_participants = mesh.borrow().active().clone();
```

volovyks (commented):
Nice test!

volovyks (approved):
Looks good!
Now I'm thinking if there is anything wrong with active. If we remove it, we may hide the issue.

PR #1051, opened 2026-07-24
feat(request): deterministic proposer rotation with scheduled round timeout

### LINKED ISSUE #907: signature generation: remove checking for active

yvonneanne:
Closes #907. Realizes the design in #228; original intent of #182. 

## What

- **Election** (`organize.rs`): drop the forward scan for a locally-active proposer, the entropy-seeded fallback, and the `state.round = selected_round` jump. `proposer(r)` is now a pure function of `(round, participants, entropy)`; `state.round` advances only through `bump_round`.
- **Schedule** (`mod.rs`, `state.rs`): `T(0)` keeps the generous inherited budget; `T(r ≥ 1)` starts at a short floor and grows by `23/20` up to a separate ceiling, so rotating past a dead proposer no longer costs a full round-0 wait.
- `wait_for_active_participants` is untouched — it gates supply, not election.
- `proposer_per_round` reduces both operands before adding: `round` derives from `highest_seen_round`, which a peer sets unbounded, so the bare sum overflows.
- The deliberator's post-Accept wait carried an unexplained `500ms` literal (`ACCEPT_POSIT_TIMEOUT + 500ms`); it is now `2 * ACCEPT_POSIT_TIMEOUT`, and the schedule's floor is asserted to exceed it.

## Why

- Nodes with differing liveness views elected different proposers *and advanced to different rounds*, so they never co-located and the request could not settle.
- This PR un-ignores the test from the stacked PR below: it stalls indefinitely without this change, settles with it.

## Cost, deliberately paid

- A dead proposer now burns `T(r)` before rotating, where the local filter used to skip it free. That is the price of peer-identical election; skipping needs the local knowledge being removed.
- `T(r ≥ 1)` being short is what buys the cost down. Rounds 1..18 are shorter than the old uniform 20s, so an alive-but-contended proposer (permit or presignature wait) has less time than before.
- `23/20` over doubling: most failed rounds are dead-or-unable proposers, where the cheapest early rounds waste least, and rotation changes proposer every round so doubling would penalise a healthy node for its predecessors. A round only doubles after five rotations. Multiplicative over additive, though: a fixed step clears a late node's indexing skew by one step however long it has been growing, leaving a window too short to complete a round in, and a ceiling below that skew leaves the two permanently offset since both advance at the same rate.

## Open question: per-chain `T(0)`?

- `T(0)` decides whether this change is free or a latency regression, and one global value serves budgets differing by three orders of magnitude.
- `Chain::expected_response_time_secs`: NEAR/Solana **11s**, Hydration 29, Canton 35, Ethereum 3605, Bitcoin ~4805. At `T(0) = 20s` a single dead round-0 proposer is over budget on NEAR and Solana; the regression fires with probability ≈ (nodes down)/`n`.
- `chain` is on-chain request data, so a per-chain `T(0)` stays peer-identical 
- Options: one global `T(0)` sized for the tightest chain / derive from `expected_response_time_secs` / a per-chain constant beside `expected_finality_time_secs`. Flagging, not choosing.

## Known exposure, not introduced here

`entropy` is requester-supplied, and `entropy[0]` alone fixes the rotation offset, so a requester can grind for a cycle that front-loads nodes it believes are down. Bounded (one round each) but repeatable per request. The local filter used to mask it.

Using an input fixed after submission works (block hash), which `IndexedSignRequest` does not carry. This is a separate issue, not addressed in this PR.

## Tests

- `round_timeout_schedule` — every round outlasts the post-Accept wait, rotation is cheaper than round 0, the cap is reached.
- `rotation_reaches_live_proposer_within_f_plus_1_rounds` — exhaustive over down-sets and start rounds for `n = 4, 5`. `n = 4` is load-bearing: a stride sharing a factor with `n` covers only half the participants, which an odd `n` masks.
- The now-active integration test from the stacked PR.

### DIFF
--- a/chain-signatures/node/src/protocol/request/mod.rs
+++ b/chain-signatures/node/src/protocol/request/mod.rs
@@ -64,25 +64,22 @@ const ROUND_TIMEOUT_FLOOR: Duration = ACCEPT_POSIT_TIMEOUT
     .saturating_mul(2)
     .saturating_add(Duration::from_secs(1));
 
-/// Per-round growth factor, as a 23/20 ratio.
+/// Per-round growth factor of 1.15 (=23/20).
 ///
-/// Multiplicative rather than additive. Nodes that index a request at different
-/// times enter the same round at different moments, and can only transact while
-/// both are inside it — so a round has to grow past that skew before the request
-/// can settle. An additive step overshoots the skew by one step however long it
-/// has been growing, leaving a window too short to complete a round; a ratio
-/// overshoots by a fraction of the skew, so the window scales with the problem.
-///
-/// Any ratio above 1 eventually crosses the skew, so the value is chosen for
-/// what it costs the other case: rotating past dead proposers wants rounds to
-/// stay short. At 23/20 a round only doubles after five rotations, so rotating
-/// past a run of dead proposers stays cheap for as long as it plausibly needs to.
+/// Nodes that index a request at different times enter the same round at
+/// different moments, and can only transact while both are inside it.
+/// Thus, a round length has to exceed that skew before the request
+/// can be processed.
+/// Any factor above 1 eventually crosses the skew, so the value is chosen
+/// for what it costs the other case: rotating past inactive proposers
+/// benefits from short rounds. At 1.15 the round length doubles after five rounds.
 const ROUND_TIMEOUT_GROWTH_NUM: u32 = 23;
 const ROUND_TIMEOUT_GROWTH_DEN: u32 = 20;
 
 /// Longest a round may be. High enough that any plausible indexing skew is
 /// crossed, low enough that a wedged request keeps rotating visibly instead of
 /// disappearing into a multi-hour round.
+/// tolerable skew < ROUND_TIMEOUT_CEILING - ROUND_TIMEOUT_FLOOR
 const ROUND_TIMEOUT_CEILING: Duration = Duration::from_secs(if cfg!(feature = "test-feature") {
     30
 } else {
@@ -677,22 +674,17 @@ mod tests {
 
     #[test]
     fn round_timeout_schedule() {
-        // Every round must outlast a deliberator's post-Accept wait, or the
-        // round is too short for any proposer to gather accepts and rotating
-        // through it accomplishes nothing.
-        let min_viable = 2 * ACCEPT_POSIT_TIMEOUT;
+        // Every round must outlast a propose broadcast (<1s) plus accept
+        // gathering inside.  For r >= 1 this holds by construction today;
+        // the test serves as a guard so the code can't silently drift.
         for r in 0..64usize {
             assert!(
-                round_timeout(r) > min_viable,
+                round_timeout(r) >= ROUND_TIMEOUT_FLOOR,
                 "round {r} is too short to complete"
             );
         }
 
-        // Rotating past a dead proposer must be cheaper than the round-0 wait,
-        // otherwise the schedule buys nothing.
-        assert!(round_timeout(1) < round_timeout(0));
-
-        // Growth is monotonic, so a request that keeps failing keeps getting
+        // Growth is monotonic from round 1, so a request that keeps failing gets
         // more time rather than retrying forever on equally short rounds.
         for r in 1..64usize {
             assert!(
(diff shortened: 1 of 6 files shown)


chain-signatures/node/src/protocol/request/organize.rs
```diff
@@ -80,21 +84,12 @@ impl OrganizingPhase {
...
-                            .unwrap(),
-                    )
-                });
-
-            state.round = selected_round;
+            // Elect from inputs every node shares — round, participants, entropy —
+            // and nothing local. Choosing among locally-active peers instead would
+            // let nodes with different views elect different proposers and advance
+            // to different rounds, so they never co-locate on one. A proposer that
+            // is down costs one silent round instead, bounded by `round_timeout`.
```
yvonneanne:
```suggestion
            // Elect from inputs every node shares (round, membership, entropy), no local-only information
```

chain-signatures/node/src/protocol/request/organize.rs
```diff
@@ -9,13 +9,17 @@ pub struct OrganizingPhase;
...
+    /// `highest_seen_round`, which a peer sets, so it can be arbitrarily large.
     fn proposer_per_round(
         round: usize,
         participants: &[Participant],
         entropy: &[u8; 32],
     ) -> Participant {
-        let index = entropy[0] as usize + round;
-        participants[index % participants.len()]
+        let n = participants.len();
+        let index = (entropy[0] as usize % n + round % n) % n;
```
yvonneanne, line 21:
```suggestion
        let index = (entropy[0] as usize  + round ) % n;
```

volovyks (approved):
Overall, I believe this is the right strategy not to use `active` when choosing a proposer.
At the same time, we may try to understand why https://github.com/sig-net/mpc/pull/1050 fails, since we rely on `active` in other places.

yvonneanne:
> Overall, I believe this is the right strategy not to use `active` when choosing a proposer. At the same time, we may try to understand why #1050 fails, since we rely on `active` in other places.

Thanks! I don't think [#1050](https://github.com/sig-net/mpc/pull/1050) points at a bug in  `active`: the test manufactures the divergence by construction (each node permanently hides its successor), so it's a fault model, not a symptom. The mechanism is accounted for: election filtered through a local-only set and wrote the scan's landing round into state.round, so views differing by one participant land on different rounds and never co-locate.

The fair concern underneath is that divergent views are normal in production. So the line here is: peer-identical decisions (proposer, round, deadline) derive only from (round, participants, entropy); local liveness is still fine for single-decider or local choices (mesh gate, presignature intersection, addressing).

Maybe we should make divergence observable (export active-set size + membership hash per node) so we can see how often views actually disagree?

volovyks:
@yvonneanne Agreed. Yes, we need to understand whether the active set actually differs in practice and, if it does, why.

We have an endpoint that shows the node's state, but a long that is triggered only on active set change is the simplest option.

integration-tests/tests/cases/mpc_with_stream.rs
ppca, line 164: Want to check my understanding: this implementation fixes the bug in proposer_per_round (!yay!), do not select round based on the round's proposer being active, and a new schedule for retries. So this makes sure 1) the proposer_per_round algorithm now has a complete and equal sweep over the participants set; 2) nodes will progress in the same line of (round, proposer) as all of it is deterministic. 

And this does not address the issue of no successful posit round when nodes index requests with considerable delay yet? Like mentioned in this TODO here. 

Just want to check that I did get it right. 🙏

ppca (approved):

integration-tests/tests/cases/mpc_with_stream.rs
yvonneanne: Yes to both, with some clarifications

A) yes it now iterates over all participants deterministically, regardless of what is in the active set.
The sweep was already complete before this PR — (entropy[0] + round) % n existed already so it always visited every participant. What's fixed inside proposer_per_round is the overflow, since round comes from peer-set highest_seen_round. Points 1 and 2 otherwise exactly right.

B) if there is considerable delay of indexing, then this PR does help, but it takes some time. 

Your question made me look at it more closely. With the linear schedule  capped at 20s I suggested in the PR first, two nodes whose indexing differs by more than 20s advance at the same rate and are never in the same round simultaneously; catch-up rested entirely on the highest_seen_round jump. I've pushed a follow-up commit growing rounds by a factor 1.5 to a much higher ceiling, so a 60s skew is crossed at round 8 with 34s of usable overlap instead of never. Worth a re-look since it changes the schedule you approved.

The TODO's full scope (raising the delayed test above ORGANIZE_POSIT_TIMEOUT) is something still worth addressing in addition to this PR.

chain-signatures/node/src/protocol/request/mod.rs
```diff
@@ -599,4 +647,51 @@ mod tests {
         assert!(spawner.test_posit_queues_contains(&sign_id));
         assert!(!spawner.test_dead_ids_contains(&sign_id));
     }
+
+    #[test]
+    fn round_timeout_schedule() {
+        // Every round must outlast a deliberator's post-Accept wait, or the
+        // round is too short for any proposer to gather accepts and rotating
+        // through it accomplishes nothing.
+        let min_viable = 2 * ACCEPT_POSIT_TIMEOUT;
```
ppca: wondering why this does not include the time for proposer to broadcast Propose as the ROUND_TIMEOUT_FLOOR? does? I think this might be an issue when we check ```round_timeout(crossing) - skew > min_viable```, because if the lagged node is actually the proposer for the round where the 2 nodes cross, then that lagged node would need at least ROUND_TIMEOUT_FLOOR to complete the round.

chain-signatures/node/src/protocol/request/mod.rs
```diff
@@ -61,6 +56,59 @@ const ORGANIZE_POSIT_TIMEOUT: Duration = Duration::from_secs(if cfg!(feature = "
...
+        return ORGANIZE_POSIT_TIMEOUT;
+    }
+    // Saturates at the ceiling within ~16 iterations, which bounds the loop:
+    // `round` derives from `highest_seen_round`, so a peer can make it huge.
+    let mut timeout = ROUND_TIMEOUT_FLOOR;
+    for _ in 1..round {
+        if timeout >= ROUND_TIMEOUT_CEILING {
+            return ROUND_TIMEOUT_CEILING;
+        }
+        timeout = timeout * ROUND_TIMEOUT_GROWTH_NUM / ROUND_TIMEOUT_GROWTH_DEN;
```
ppca, line 104: Love the geometric design!!!!

integration-tests/tests/cases/mpc_with_stream.rs
yvonneanne: Instead of 1.5 I pushed a factor of 1.15, the former was too aggressive for the integration suite.
A 60s indexer lag now requires 26 rather than 8 rounds, but early rotation is cheaper than the old linear schedule instead of more expensive.

chain-signatures/node/src/protocol/request/mod.rs
yvonneanne: You're absolutely right, thanks for catching this, I'll change it.

jakmeier (approved):
Overall great changes, as discussed in our call

ChaoticTempest (approved):
Overall looks good to me, but one thing: maybe not something for this PR, but since we are not validating round, an dishonest node can send `usize::MAX` for the round which should halt rotation, so something we should consider

yvonneanne:
> Overall looks good to me, but one thing: maybe not something for this PR, but since we are not validating round, an dishonest node can send `usize::MAX` for the round which should halt rotation, so something we should consider.

Yes, I agree with that. I'll prevent a halt with saturating add in this PR. The more serious concern that a malicious node can dictate the next proposer to consider. However, it is outside of the failure model currently adopted for liveness in the MPC code in general and therefore better address in a separate effort.

PR #1052, opened 2026-07-24
refactor(request): retire the pause_proposing fail-over

yvonneanne:
## What

Remove `pause_proposing_until` (its field, the `skip_proposing` read, the `enough_rejects -> pause` set, and the reset on entering generation). Keep the `AlreadyGenerating` reject reply and the `SignLimiter` permit.

## Why

`pause_proposing` only ever fired when free peers < threshold — i.e. when a *competing* generation was impossible anyway; it just stopped a re-elected proposer from spinning futilely. With one deterministic leader per round (the rotation PR this stacks on) and the `T(r)` schedule bounding re-election cost, that spin is already cheap, and the fast `AlreadyGenerating` reject still prevents duplicate generation. `num_peers_with_ongoing_generation` is retained (unit-tested; logged on reorganize) as diagnostic input.

### DIFF
--- a/chain-signatures/node/src/protocol/request/organize.rs
+++ b/chain-signatures/node/src/protocol/request/organize.rs
@@ -84,11 +84,7 @@ impl OrganizingPhase {
             // and a dead proposer costs one silent round bounded by `T(r)`.
             let proposer = Self::proposer_per_round(state.round, &participants, &entropy);
 
-            // If proposing is paused (generating already ongoing), we act as if we weren't a proposer.
-            let skip_proposing = state
-                .pause_proposing_until
-                .is_some_and(|until| until > std::time::Instant::now());
-            let is_proposer = proposer == me && !skip_proposing;
+            let is_proposer = proposer == me;
             ctx.is_proposer.store(is_proposer, Ordering::Relaxed);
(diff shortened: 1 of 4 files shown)


yvonneanne:
Closing: the premise this change rests on is wrong, and there is no upside to offset the regression.

**`pause_proposing` is not leader-election fail-over.** It was added in #832 to address #793 ("Prevent non-participant from blocking resources"), which is still open. The trigger is `AlreadyGenerating` rejects, and those come from nodes already in `GeneratingPhase` — i.e. from a generation the proposer was *excluded* from, not from competing proposers within a round. So it is non-participant resource protection.

**Deterministic rotation does not make it dead weight.** The proposer still starts generation with whoever accepted by `ACCEPT_POSIT_TIMEOUT`, so `k < n` and nodes left out of the Start list still rotate back into the proposer role later. That path is unchanged by the election change, so the condition `pause_proposing` guards persists exactly as before.

**It also fails a test that exists specifically to pin this behaviour**: `test_non_participants_pause_posits` asserts an excluded node stays quiet for the full `signature_timeout_ms` after receiving `AlreadyGenerating`.

PR #1055, opened 2026-07-24
Standardize reorg logs

volovyks:


chain-signatures/node/src/protocol/request/posit.rs
```diff
@@ -298,27 +287,26 @@ impl PositPhase {
                         }
 
                         if counter.enough_rejects(ctx.governance.threshold) {
```
volovyks, line 289: @yvonneanne  the logic inside of this `if` is questionable, as the delay concept (IMO), but I have not touched the logic in this PR.

yapignolet: Thanks, will look into it!

yapignolet (approved):
🧹 Nice effort, this will make evaluating efficiency gains wrt re-orgs easier!

yvonneanne (approved):

Issue #1034, opened 2026-07-22
Standardize and centralize all primitive types across all codebases


volovyks:
Talked about this with @BRBussy . Must do as we scale. Helps a lot with breaking changes.

volovyks:
@BRBussy suggested a proto-first approach (like https://github.com/meshtrade/api),
but it doesn't fit our situation.

### Why not proto

Our wire formats are chain-owned: NEAR takes JSON args (serde), Solana takes Borsh
instruction data, Ethereum takes ABI calldata. No chain accepts protobuf bytes, so
proto could only describe shapes while the real encoding lives elsewhere — a parallel
definition to sync manually, not a source of truth.

Proto as codegen-only IDL doesn't help either: the struct's shape determines the bytes
once serde/borsh runs over it, and proto can't express what we need — no `[u8; 32]`
(→ `Vec<u8>`, different Borsh layout), enums become `Option`-wrapped `oneof`s
(different JSON and bytes), no serde attributes.

### Rust as the schema instead

Split today's `primitives` into `mpc-primitives` (internal) and a published
`signet-primitives` — everything clients use:

```
signet-primitives
├── core                 # SignId, SignArgs, Signature, bidirectional — chain-agnostic
├── near   [feature]     # chain-specific types and wrappers
├── solana [feature]
...
```

I'm still not sure about chain-specific types, but we can start with creating and publishing `signet-primitives`. 

### TS clients

Types generated from the same crate with ts-rs (https://docs.rs/ts-rs); Can be generated and published on CI.

BRBussy:
Thanks for the feedback! I agree that protobuf is not the right way as the protobuf generated types are really focussed on serialisation before communication, so the generated types can be a bit messy for our use case. 

More generally it would be cool if we could follow the same sort of generic pattern of describing our types once in a language agnostic way and then generating for each language that uses them. Without all the serialisation and deserialisation stuff that protobuf does.

Using rust as the schema instead is perfect as it is the language of the MPC. Given a set of primitives there I don't think it would be too hard to use some existing rust language parser to go from rust to AST and then generate the necessary types for all the other smart contract types.

PR #1049, opened 2026-07-23
Split primitives

volovyks:
First step in https://github.com/sig-net/mpc/issues/1034
The list of types can be adjusted as we need them.

PR #1057, opened 2026-07-24
Add public crypto and primitive crates

volovyks:
Let's avoid implementing this logic from scratch on the client side.
First step in https://github.com/sig-net/mpc/issues/1034


volovyks:
Regarding https://github.com/sig-net/mpc/pull/1049#discussion_r3646177044, I like the idea!
I've moved the new public crates to the top level (no /rs folder for now, but that would be easy to add).
Not creating language- and chain-specific folders for now, as we do not have them yet.
cc @BRBussy

volovyks:
cc @ChaoticTempest @ppca @isSerge

signet-primitives/src/requests.rs
```diff
@@ -0,0 +1,56 @@
+use borsh::{BorshDeserialize, BorshSerialize};
```
BRBussy, line 1: Should we add the protocol types for the BidirectionalFlow here as well as well, or perhaps future PR:
i.e. these types. naming convention here taken from (../../chain-signatures/contract-sol/src/lib.rs)
- SignBidirectionalEvent
- SignatureRespondedEvent
- RespondBidirectionalEvent

would potentially require a formal look at naming these structures as SignatureRespondedEvent means something different in the chain signatures events primitive crate.

volovyks: @BRBussy, those are the fundamental types we need to have in our primitives, but they are not so easy to move. The problem of ownership still exists. Our node code does not use the same types you are using to send and receive sign requests. The information we receive is often extended with extra info, such as `entropy` or `sender`.

Our Integration and other tests can benefit from reusing the same types as our client libraries, but even there we have issues. For example, Solana types are owned by the contract. And that contract is not even a real contract (as we discussed).

That is why I want to do all this in multiple steps. Unifying the mentioned types would require additional work.

BRBussy: cool, that makes sense. We can move stuff bit by bit for sure.

BRBussy (approved):

PR #1058, opened 2026-07-24
feat(midnight-publisher): TypeScript sidecar for the respond path

Pessina:
The SGN2 Midnight respond path has to build circuit calls from compactc output, and compactc emits only JavaScript, so the publisher is a TypeScript service. It also keeps the Midnight ledger WASM dependency tree out of the mpc workspace and off the keyholding node.

Decoding is no longer part of it. Contract state and transactions are read natively by the Rust indexer against Midnight's own ledger crates, so this service is publish-only and the `/decode/*` routes are gone.

- Localhost HTTP seam: `GET /health` (ledger tags, network id) and `POST /respond` (validate, pinned reads, prove, balance, submit).
- Caller feedback: machine error codes with `stage` and cause-chain `detail`, `tx_id` and `block_hash` on success, per-stage deadlines that turn dead-dependency hangs into answers, and a `wallet_busy` gate for the single-dust-UTXO wallet.
- The raw chain fixtures stay here because this is where they are captured from, but their only readers are the Rust `chain-midnight` tests, which land with the indexer. `tests/fixtures/README.md` says so on the tin.

#### Diff size

Most of the added lines are `package-lock.json` (4k + lines) and fixtures.


isSerge:
How are going to run it? Do we need a Docker container?

Pessina:
> How are going to run it? Do we need a Docker container?

Yes, the plan it's run o Docker. 

cc: @isSerge

Pessina:
Hold a bit on the review here, I may have a found a way to do it in Rust while implementing the indexer side.

Issue #777, opened 2026-04-28
Lack of checkpoint should start indexers at signet contract genesis block


ChaoticTempest:
If we are to somehow lack a checkpoint in our persistent storage, and we are unable to select a checkpoint from other nodes, we should start the chain indexer at a specific genesis block height defined when we deployed the contract.

The issue is where do we define such values, in our governance config, hardcoded in the chain indexers themselves and how should we handle varying networks (mainnet, testnet, devnet, ...).

volovyks:
Ideally, the starting block height should be set once per chain (separately for testnet, mainnet, etc). It means that it should not be a configurable parameter (node CLI argument). If we go extreme, the starting block is a block when the signing contract was deployed to the network.

Such an approach will solve the issue of having different environments with different starting block heights, but determining a block height on which the contract was deployed becomes a separate challenge on each chain. A simple binary search (0 - latest block height range) can be a fairly universal solution here. I'm not sure about Canton; cc @Pessina.

I'm not a fan of keeping a list of supported chains in our governance contract, but that may be a simpler solution. It can be a map `Chain - block height` set during contract init.

ChaoticTempest:
yeah I would ideally like it to be a part of the contract of the equivalent chain, but that requires redeploying a bunch of contracts and having the contract be self-reflective which is not great when deploying newer contracts...

wait, I wonder if we can just do a lookup of what block the contract was initially created/deployed at. Maybe that would be simpler, but the issue is if all chains support such a feature

jakmeier:
Could we just hard-code these genesis values in Rust code? Or if you want to be fancy about it, put it in a file e.g. chains_config.json and `include_str!` it.

ChaoticTempest:
we could hardcode it, but we would still need to figure out if the node is running in which environment (dev, testnet or mainnet)

Pessina:
Yes, Canton is possible to lookup on previous blocks.

+1 on @jakmeier idea. 

I'm wondering, what we do when each contract upgrade changes the Contract Address? Will we track events that were emitted by legacy contracts? What we store as genesis in that case? 

On EVM we can use the Upgrade pattern so the Address it's stable. But in Canton that's not possible, a contract upgrade will change it's address.

jakmeier:
> we could hardcode it, but we would still need to figure out if the node is running in which environment (dev, testnet or mainnet)

@ChaoticTempest yes we could select this maybe via CLI or ENV argument.

[...]

PR #1059, opened 2026-07-26
feat(chain-midnight): index Midnight signature requests for the MPC

Pessina:
## Why

We are adding Midnight as a source chain for the MPC. This PR is the read path; the publish path lands separately. Stacked on #1075.

## What

- New `mpc-chain-midnight` crate: contract state decoded with Midnight's own ledger crates, the singleton's notification map diffed per finalized block, the carried ledger path walked to the caller's request map, declared-width record decode, request-id recompute, EIP-1559 assembly, epsilon conversion, catchup-then-live indexer under the node supervisor.
- Node wiring: CLI args, off unless supplied. `MidnightPublisher` bails so requests retry rather than settle falsely; the publish path lands separately.
- Capture-backed tests run real chain bytes through the pipeline, asserted against contract-pinned values; the goldens are regression pins.

## What this protects

- A caller only gets a signature over a payload it filed: the request id is recomputed from the record and must match the key it was filed under.
- A caller only gets a signature under its own key: epsilon derives from the address the record was read from, and `sender` must equal it.
- One bad entry cannot stall the chain: caller-attributable failures are counted drops, caller reads are budget-bounded, and only our-own-schema drift halts indexing, loudly.

## Known gaps

- Provenance is not enforced: `caller_address` is producer-supplied, so a third party can trigger a signature over a record another contract filed (content stays authenticated). TODO in `process_entry`, scoped as its own PR.
- Responses are not indexed, so requests never settle; deferred with the publisher.
- The notification map only grows, so per-block read cost rises with history.
- A fresh node anchors at the finalized head, so pre-boot requests are never seen (same as Solana and Ethereum, #777).
- The path encoding (full 32 bytes, lowercase hex) is decided node-side; the TS derivation reference, the fakenet, and the epsilon goldens move to the same rule before first deploy.


chain-signatures/chain-midnight/src/reader.rs
```diff
@@ -0,0 +1,767 @@
...
+/// `sender` through `calldata.no_words`; the calldata words begin here.
+const RECORD_HEAD_ATOMS: usize = 18;
+
+/// `caip2_id` and the two schemas.
+const RECORD_TAIL_ATOMS: usize = 3;
+
+pub type Node = StateValue<DefaultDB>;
+
+/// Resolve a flat ledger field index to its node in the state tree.
+pub fn signet_field_node(root: &Node, flat_index: usize) -> anyhow::Result<&Node> {
```
BRBussy: Might be worth adding a comment about what could break this in future here, if anything else can end up looking like a 15 entry array (i.e. full chunk size) other than a compiler chunk then this decoding will break down. Some Comment like:

/// Resolve a flat ledger field index to its node in the state tree.
///
/// Chunk detection assumes that only compiler chunks look like FULL 15 (chunk size) entry arrays.
/// That holds today: no ledger type compiles to an array node except `List`,
/// whose node is a three-slot linked list [element value, rest of list, count],
/// so not 15 length. If a future ledger type compiles to an 15 length array and
/// lands as a contract's last field, the walk counts it as a chunk and every
/// field index shifts. Re-verify when Midnight adds ledger types.

chain-signatures/chain-midnight/src/indexer.rs
```diff
@@ -0,0 +1,1720 @@
...
+                let parent_keys: HashSet<&AlignedValue> =
+                    parent.iter().map(|entry| &entry.key).collect();
+                entries
+                    .iter()
+                    .filter(|entry| !parent_keys.contains(&entry.key))
+                    .collect()
+            }
+            // No diff is possible, so this block emits nothing, giving up only the
+            // entries filed at this exact height.
+            None => {
```
BRBussy: I think this degraded read should throw an error instead of just logging. The only way that execution reaches here is when the central state decoded fine but the field walk failed, i.e. schema drift or a wrong central address. So it is deterministic how to get here and what is going wrong, but it is not transient. So from this block onwards we log this error on every block, but we keep emitting Block events, the checkpoint keeps advancing, and the notifications filed at these heights are dropped for good.
If we eventually fix the binary or the address, catchup restarts from a checkpoint that has already moved past everything we never evaluated, so the normal recovery re-walk cannot get them back. The indexer will look alive but just be logging errors while it progresses.
There is already precedent for treating this class of error as fatal: if the central state itself is undecodable, that becomes an Err in central_tree that halts ingestion in the retry loop until we ship a fix. An error here is the same class of problem (our own trusted contract not matching this binary).
On the caller side failures can drop, but on our data side we should halt I think.

chain-signatures/chain-midnight/src/indexer.rs
```diff
@@ -0,0 +1,1720 @@
...
+    /// `at_hash`, gate through `resolve_verified_record`, convert.
+    async fn process_entry<C: ChainSource>(
+        &self,
+        source: &C,
+        entry: &MapEntry,
+        at_hash: &str,
+        height: u64,
+        indexed_ts: u64,
+    ) -> anyhow::Result<Option<IndexedSignRequest>> {
+        let Some((_, rid)) = signet_map_key_rid(&entry.key) else {
```
BRBussy: I think this drop should be an error instead. The map key here is built by our own central contract (SignetMapKey{count, requestId} in signBidirectional), the caller only supplies the 32 byte requestId value and has no way to change the shape of the key. So if a key fails to parse it means this binary no longer matches our own contract's schema. That is deterministic, not a bad entry, and it will hit every entry in the map. Meanwhile we keep emitting Block events, the checkpoint keeps advancing, and all of these requests are dropped for good, with no way for the recovery re-walk to get them back after we ship a fix. Same argument as my comment on the parent None read.

chain-signatures/chain-midnight/src/indexer.rs
```diff
@@ -0,0 +1,1720 @@
...
+            return Ok(drop_entry(
+                "malformed-map-key",
+                height,
+                None,
+                &format!("{} atoms in a SignetMapKey", entry.key.value.0.len()),
+            ));
+        };
+        let notification = match decode_notification(&entry.value) {
```
BRBussy: Same argument as the map key above. The wire shape here is the typed struct {version: Uint<8>, payload: Bytes<128>}, enforced by the signBidirectional circuit's signature, so no caller can file an entry that breaks this decode. A failure here means the binary has drifted from our own contract, which is deterministic and will affect every entry while we keep progressing checkpoints.

chain-signatures/chain-midnight/src/reader.rs
```diff
@@ -0,0 +1,767 @@
...
+    Absent,
+    /// The entry exists and must not be signed.
+    Dropped {
+        reason: &'static str,
+        detail: String,
+    },
+}
+
+/// Decode the two-atom notification cell: version, then the 128-byte payload.
+pub fn decode_notification(node: &Node) -> anyhow::Result<SignBidirectionalEventNotification> {
```
BRBussy, line 241: An extra test may be good here to pin the decoding here to real bytes from the chain snapshots. The test can take the real notification entry that is already in the 1366 fixture and run it through decode_notification and unpack_notification_v1, asserting the known caller address and field index. The current tests build their cells with the same widths the decoder expects, so they can't catch the case where both are wrong about the actual contract.

chain-signatures/chain-midnight/src/indexer.rs
```diff
@@ -0,0 +1,1720 @@
...
+                    "notification-version",
+                    height,
+                    Some(rid),
+                    &format!("{err:#}"),
+                ));
+            }
+        };
+        let caller_hex = hex::encode(unpacked.caller_address);
+
+        // TODO: decide whether `caller_address` must be checked against the
```
BRBussy, line 344: I think we do need to do this confirmation since not confirming that the notification was added as a result of a cross contract call opens up some fringe attack cases such as where a contract stores a SignatureRequest but for whatever reason is not ready yet to request the signature. In that fringe scenario not checking here would allow an attacker to spoof a request to get a signature for that request (which may not be ready yet). Also an attacker could set the index location of the requests map to some other variable location in the supposed caller contract, which would be dangerous if (as an edge case example) the contract authors choose to store "spent" requests, or "cancelled requests" or something in a map. Using events could allow us to dodge this whole thing.

chain-signatures/chain-midnight/src/reader.rs
```diff
@@ -0,0 +1,767 @@
...
+        entries,
+        keys,
+    })
+}
+
+/// Decode a stored request record in one pass, refusing a cell whose declared widths
+/// are not a signet record's.
+pub fn decode_record(node: &Node) -> anyhow::Result<SignBidirectionalRecord> {
+    let cell = cell_of(node, "request record")?;
+    let widths = declared_widths(cell, "request record")?;
```
BRBussy, line 173: We could read tx_param_type (fixed atom index 7 in the cell) here to add a check to reject unsupported transaction parameter types before doing capacities count. Then if the compact types run ahead of the MPC the rejection error will be clear about the reason for failure (unsupported txn_param_type) instead of a capacity counting error. Then we also have the point set where we can switch on the txn_param_type for the other types we'll add.

chain-signatures/chain-midnight/src/reader.rs
```diff
@@ -0,0 +1,767 @@
...
+
+struct Capacities {
+    words: usize,
+    entries: usize,
+    keys: usize,
+}
+
+/// Each scaled vector's capacity, read off the declared widths. The tail is taken from
+/// the end because `caip2_id` and a storage key are both `Bytes<32>`.
+fn capacities(widths: &[u32]) -> anyhow::Result<Capacities> {
```
BRBussy: This function, its return type and the consts it uses could be renamed to clarify that they are EVM_TYPE2 specific:
- capacities() → evm_type2_capacities()
- Capacities → EvmType2Capacities
- REQUEST_FIXED_VALUE_ATOMS → EVM_TYPE2_FIXED_ATOMS
- RECORD_HEAD_ATOMS → EVM_TYPE2_HEAD_ATOMS

chain-signatures/chain-midnight/src/reader.rs
```diff
@@ -0,0 +1,767 @@
...
+    Ok(cell)
+}
+
+struct Capacities {
+    words: usize,
+    entries: usize,
+    keys: usize,
+}
+
+/// Each scaled vector's capacity, read off the declared widths. The tail is taken from
```
BRBussy: Understanding this function assumes a lot of prior knowledge of how the evm type2 SignBidirectionalEvent flattens. It may help future readers to say in this comment what the function is actually doing: recovering the capacity parameters the caller contract was compiled with (#maxCalldataWords, #maxAccessListEntries, #maxStorageKeysPerEntry) from the record's atom shape alone. Or include a diagram like this somewhere in the file, maybe above the three atom-count constants (), since they are its atom field numbers:

```
export struct SignBidirectionalEvent<TxParams, #LenOut, #LenResp> {

  FIXED HEAD -- 18 atoms, indices 0..17 (RECORD_HEAD_ATOMS = 18)
  +----+---------------------------------------------+-------------+
  |  0 | sender                                      | Bytes<32>   |
  |  1 | requestNonce                                | Uint<64>    |
  |  2 | keyVersion                                  | Uint<8>     |
  |  3 | path                                        | Bytes<32>   |
  |  4 | algo                                        | enum, 1B    |
  |  5 | dest                                        | enum, 1B    |
  |  6 | params                                      | Bytes<64>   |
  |  7 | txParamType                                 | enum, 1B    |
  |    | txParams: EvmType2TxParams<#W, #E, #K> {    |             |
  |  8 |   chainId                                   | Uint<64>    |
  |  9 |   nonce                                     | Uint<64>    |
  | 10 |   maxPriorityFeePerGas                      | Uint<128>   |
  | 11 |   maxFeePerGas                              | Uint<128>   |
  | 12 |   gasLimit                                  | Uint<64>    |
  | 13 |   to                                        | Bytes<20>   |
  | 14 |   value                                     | Uint<128>   |
  |    |   calldata: Maybe<EvmCalldata<#W>> {        |             |
  | 15 |     isSome                                  | Boolean, 1B |
  |    |     value: EvmCalldata<W> {                 |             |
  | 16 |       selector                              | Bytes<4>    |
  | 17 |       noWords                               | Uint<16>    |
  +----+---------------------------------------------+-------------+

  SCALED REGION -- W atoms, indices 18 .. 18+W-1, always at capacity
  +~~~~~~~~~~+-----------------------------------+----------------+
  | 18 ...   |   words: Vector<W, Bytes<32>>     | W x Bytes<32>  |
  +~~~~~~~~~~+-----------------------------------+----------------+
  |    |     } }                                 |             |

  FIXED ATOM, FLOATING POSITION
  +------+-------------------------------------------+-----------+
  | 18+W |   accessListEntryCount                    | Uint<8>   |
  +------+-------------------------------------------+-----------+

  SCALED REGION -- E x (2+K) atoms, each entry flattened in place
  +~~~~~~~~~~+---------------------------------------+-----------+
  | ...      |   accessList: Vector<E, Entry<K>>     |           |
  |          |     entry.address                     | Bytes<20> |
  |          |     entry.storageKeyCount             | Uint<8>   |
  |          |     entry.storageKeys                 | K x 32B   |
  +~~~~~~~~~~+---------------------------------------+-----------+
  |    | }                                           |             |

  FIXED TAIL -- 3 atoms, sliced off the END (RECORD_TAIL_ATOMS = 3)
  +-------+--------------------------------------+----------------+
  | end-3 | caip2Id                              | Bytes<32>      |
  | end-2 | outputDeserializationSchema          | Bytes<LenOut>  |
  | end-1 | respondSerializationSchema           | Bytes<LenResp> |
  +-------+--------------------------------------+----------------+
}

fixed atoms:  18 head + 1 count + 3 tail = 22   (REQUEST_FIXED_VALUE_ATOMS)
total atoms:  22 + W + E*(2+K)

why the tail anchors from the end: storage keys, calldata words and
caip2Id are ALL Bytes<32>, so no forward width-scan can find the
head/region/tail boundaries; only the tail's fixed count of 3 can.
```

chain-signatures/chain-midnight/src/reader.rs
```diff
@@ -0,0 +1,767 @@
...
+
+fn cell_of<'a>(node: &'a Node, what: &str) -> anyhow::Result<&'a AlignedValue> {
+    let StateValue::Cell(cell) = node else {
+        anyhow::bail!("{what} node is not a cell");
+    };
+    Ok(cell)
+}
+
+struct Capacities {
+    words: usize,
```
BRBussy: perhaps name these 3 fields the same as the dynamic sizing parameters from Signet.compact that they carry to help readers:
- maxCalldataWords
- maxAccessListEntries
- maxStorageKeysPerEntry

Named as they are now there could be confusion as well between them and the actual integer no. of used words and access list entries.

chain-signatures/chain-midnight/src/reader.rs
```diff
@@ -0,0 +1,767 @@
...
+}
+
+/// A two-variant Compact enum: one declared byte, and only 0 or 1 is in range.
+fn bounded_enum(cursor: &mut AtomCursor, what: &'static str) -> anyhow::Result<u8> {
+    let value = uint::<u8>(cursor, 1, what)?;
+    anyhow::ensure!(value <= 1, "{what}: {value} exceeds the maximum 1");
+    Ok(value)
+}
+
+fn decode_at(
```
BRBussy: could be worth adding a warning comment here that the field ordering is critical here so that the decoding happens in the right order, and that the field order must exactly match the field ordering in Signet.compact. Something like:

```ts
/// Field order here is critical: the struct literal's decode calls run in
/// textual order and each consumes the next atom, so this list must mirror,
/// field for field, the declaration order of `SignBidirectionalEvent` in
/// signet-midnight's Signet.compact. `decode_tx_params` likewise mirrors
/// `EvmType2TxParams`. A reorder between fields of different widths fails
/// loudly on the declared-width check. A reorder between same-width
/// neighbours (chain_id and nonce, the two fees, algo and dest) decodes
/// cleanly and surfaces only as an rid-mismatch drop on every record, which
/// reads like caller fault rather than our bug.
```

chain-signatures/chain-midnight/src/convert.rs
```diff
@@ -0,0 +1,240 @@
...
+    let algo = match record.algo {
+        ALGO_ECDSA => "ecdsa".to_string(),
+        reserved => anyhow::bail!("unsupported algo {reserved}: only ecdsa (0) is real"),
+    };
+    let dest = match record.dest {
+        DEST_UNUSED => String::new(),
+        reserved => anyhow::bail!("unsupported dest {reserved}: only unused (0) is real"),
+    };
+
+    let path = render_padded_ascii(&record.path, "path")?;
```
BRBussy: We need to support path being any arbitrary 32 byte array here. render_padded_ascii requires that the bytes, minus trailing zeros, are already valid utf8, which a real path like the erc20-vault's raw commitment hash essentially never is, so those requests get dropped and never signed. We cannot enforce printable bytes on the compact side (well we chose not to for performance), so the path must be treated as opaque bytes and hex encoded directly:

```rs
// No trimming first on the path (trimming would make 0xab..00 and 0xab.. derive the same key).
let path = hex::encode(record.path);
```

Params may eventually have the same problem, although we are enforcing that is blank for now, so possibly fine to ignore for now? Perhaps just a comment there, or asserting again that it is blank here.

chain-signatures/chain-midnight/src/indexer.rs
```diff
@@ -0,0 +1,1720 @@
...
+                    // re-anchor.
+                    None => anyhow::bail!("midnight finalized block producer terminated"),
+                },
+            };
+            if block.number <= last_processed {
+                continue;
+            }
+            // Close any finality gap so a lagging subscription cannot skip
+            // notifications; each height processes exactly like catchup.
+            for number in (last_processed + 1)..block.number {
```
BRBussy, line 647: The gap-fill here is a direct copy of the catchup walk above so to tidy this up the shared part could be put into a helper like walk_heights(range) that could be used by both? CatchupCompleted event emit could stay at the call site between the two helper calls. Definitely a nit though.

chain-signatures/chain-midnight/src/rpc.rs
```diff
@@ -0,0 +1,459 @@
...
+}
+
+fn hex_0x(hash: H256) -> String {
+    format!("0x{}", hex::encode(hash.as_bytes()))
+}
+
+/// True when `err` carries `contract_state`'s pruned-or-unknown-hash mapping (the
+/// outermost context survives `retry_rpc!`'s flattening, per the note on
+/// [`STATE_UNSERVABLE_MSG`]).
+pub(crate) fn is_state_unservable(err: &anyhow::Error) -> bool {
```
BRBussy: This classification is a substring match on error text that we generate and return ourselves in contract_state_over in rpc.rs. It would be more robust to return a typed error instead, something like a StateUnservable type we recover here with err.downcast_ref::<StateUnservable>(). If protects against someone rewording strings in that error string later.

chain-signatures/chain-midnight/src/indexer.rs
```diff
@@ -0,0 +1,1720 @@
...
+        // means joining each notification to the central call it came from through the
+        // claimed communication commitment, which the ledger validates for uniqueness
+        // and for corresponding to a real call.
+
+        // Authority: the caller's own ledger at the SAME finalized hash the
+        // notification was read at.
+        let caller_tree = match source
+            .contract_state_tree(&caller_hex, at_hash)
+            .await
+            .map_err(|err| note_unservable(err, height))?
```
BRBussy: An error reading the caller state here (one that survives the rpc retries) escalates to a block level retry. A caller whose contract state is too large to serve inside request_timeout triggers that, so one bloated contract could halt indexing. The checkpoint will freeze and the watchdog restart comes back to the same block. The node being down does not reach this line (the central read in process_block fails first), so a persistent error here is very likely specific to this caller's state. Perhaps after a bounded number of retries this should become a per-entry drop, charged to the caller like caller-state-undecodable?

chain-signatures/chain-midnight/src/indexer.rs
```diff
@@ -0,0 +1,1720 @@
...
+    /// Emits a block's requests, then its Block event, then records telemetry.
+    ///
+    /// The Block event is the whole progress report: the node advances the persisted
+    /// height in `process_block_event` after CONSUMING it, and only once caught up.
+    /// Writing the height here instead would claim blocks whose requests are still
+    /// queued, and a supervised restart allocates a fresh event channel and drops
+    /// whatever the old one still held, so those requests would be neither delivered
+    /// nor re-walked. Leaving the height to the consumer is what makes a restart
+    /// self-healing, and is what every other chain does.
+    async fn emit_block(
```
BRBussy, line 560: It might be worth a one liner here noting that blocks processed after the last consumer checkpoint re-emit their requests on a restart, i.e. a comment indicating that this side of the channel intentionally provides at least once delivery, and that that is intended behaviour, with dedup handled downstream by SignId in the node's request handling. So nobody should try to dedup emissions in this file.

chain-signatures/chain-midnight/src/reader.rs
Pessina: Superseded by 121416b9: the heuristic is gone. The notification now carries the caller's resolved ledger path and the walk follows it via the ledger's own `Array::get`, so there is no arity inference left to document.

chain-signatures/chain-midnight/src/indexer.rs
Pessina: Fixed in a98ecb43: central-field failures hard-error now, and the parent-diff degrade branch is deleted (same drift), so the checkpoint freezes loudly instead of advancing past dropped requests.

chain-signatures/chain-midnight/src/indexer.rs
Pessina: Fixed in a98ecb43: a key that does not parse as `SignetMapKey` is a hard error. The singleton mints the key from its own counter, so callers cannot produce this.

chain-signatures/chain-midnight/src/indexer.rs
Pessina: Fixed in a98ecb43, extended to the version byte: the circuit asserts `version == 1`, so a non-1 version is drift too. Only the caller-writable depth/path bytes remain a per-entry drop.

chain-signatures/chain-midnight/src/reader.rs
Pessina: Deferred to the fixture recapture: the committed 1366 capture predates the path payload (byte 32 is the old flat ordinal 0x04), so it cannot back this test. Queued first on the recapture checklist.

chain-signatures/chain-midnight/src/indexer.rs
Pessina: Agreed it is real, and the path payload widens it: a spoof can name any depth-4 node (e.g. a cancelled-requests archive), and the caller template's own `remove` circuits make file-then-cancel concrete. Scoping as its own PR since the state-diff path never sees transactions (the commitment join is new read surface); a98ecb43 removes the DoS edge meanwhile.

chain-signatures/chain-midnight/src/reader.rs
Pessina: Fixed in 2f4388c0: `decode_record` rejects a foreign `tx_param_type` by name (atom 7) before capacity recovery.

chain-signatures/chain-midnight/src/reader.rs
Pessina: Fixed in 2f4388c0: `EvmType2Capacities`, `evm_type2_capacities()`, `EVM_TYPE2_FIXED/HEAD/TAIL_ATOMS`.

chain-signatures/chain-midnight/src/reader.rs
Pessina: Adopted the prose half in 2f4388c0: the doc states the function recovers `#maxCalldataWords` / `#maxAccessListEntries` / `#maxStorageKeysPerEntry` from the declared widths, plus the tail-anchor rationale. Skipped the full diagram for comment density; can add it if the prose still reads thin.

chain-signatures/chain-midnight/src/reader.rs
Pessina: Fixed in 2f4388c0: `max_calldata_words`, `max_access_list_entries`, `max_storage_keys_per_entry`.

chain-signatures/chain-midnight/src/reader.rs
Pessina: Adopted in 2f4388c0, tightened: cites Signet.compact and names the same-width neighbour pairs.

chain-signatures/chain-midnight/src/convert.rs
Pessina: Fixed in e88a975f: full 32 bytes, lowercase hex, never trimmed; all-NUL renders as 64 zeros. Nothing is deployed, so the TS reference (epsilon-derivation.ts, fakenet `getPath`, vault tooling) and the epsilon goldens move to the same rule in lockstep before first deploy. `params` asserted blank per your side note.

chain-signatures/chain-midnight/src/indexer.rs
Pessina: Done in 121416b9: `index_height` shared by catchup and the live gap walk, `CatchupCompleted` emitted between the call sites.

chain-signatures/chain-midnight/src/rpc.rs
Pessina: Deferred: `retry_rpc!` rebuilds the error from its Display, severing the source chain, so the downcast returns `None` on exactly the exhausted-budget path. Matcher and attacher already share the one `STATE_UNSERVABLE_MSG` constant; the typed error needs a core `retry_rpc!` change, folded into the planned core consolidation.

chain-signatures/chain-midnight/src/indexer.rs
Pessina: Fixed in a98ecb43: caller-read failures that survive the transport budget (unservable included) drop per entry as `caller-state-unreadable`; block-level retries stay reserved for central reads. The transient-loss trade is as you stated, and right against a chain-wide halt.

chain-signatures/chain-midnight/src/indexer.rs
Pessina: Added in a98ecb43: `emit_block`'s doc states at-least-once delivery, SignId dedup downstream, never here.

chain-signatures/chain-midnight/src/reader.rs
Pessina: Done in dde376d3: fixtures recaptured on the field-path contract (blocks 63/64, caller state included), and the capture-backed tests now run the real notification, record, parent diff and epsilon derivation through the pipeline, asserted against contract-pinned values.

chain-signatures/chain-midnight/src/config.rs
```diff
@@ -0,0 +1,103 @@
...
+    pub central_address: String,
+    pub rpc: RpcConfig,
+    pub indexer: IndexerConfig,
+}
+
+impl MidnightConfig {
+    /// Rejects endpoints that cannot work, before anything dials them, so an unusable
+    /// config fails once at construction with the field named rather than forever at
+    /// runtime.
+    pub fn validate(&self) -> anyhow::Result<()> {
```
isSerge, line 89: Worth adding a unit test for `validate`

chain-signatures/chain-midnight/src/indexer.rs
isSerge, line 1: This module is pretty big and mixes a lot of concerns, would be better to extract `ChainSource`, map-key helpers and other non-indexer logic

isSerge:
Do you plan to add integration tests in this or following PR?

chain-signatures/chain-midnight/src/config.rs
```diff
@@ -0,0 +1,103 @@
...
+    /// How long the finalized-head subscription may go silent before `run()` returns
+    /// and lets the supervisor restart it
+    pub stall_timeout: Duration,
+}
+
+impl Default for IndexerConfig {
+    fn default() -> Self {
+        Self {
+            live_block_buffer: 16384,
+            stall_timeout: Duration::from_secs(60),
```
ppca, line 70: question: how often does midnight produce a block?

chain-signatures/chain-midnight/src/indexer.rs
```diff
@@ -0,0 +1,1911 @@
...
+                    height,
+                    Some(rid),
+                    &format!("{caller_hex}: {err:#}"),
+                ));
+            }
+            // The transport already spent its retry budget, and block-level retries
+            // cannot be bounded per entry, so escalating would let one caller whose
+            // state cannot be served (`caller_address` is producer-supplied) halt the
+            // chain. Charged to the entry instead, the unservable answer included.
+            Err(err) => {
```
ppca: is it still possible that this is purely RPC error for example unstable connection timeouts? If so, we shouldn't drop_entry. Can we differentiate the type of error and retry the ones that failed because of rpc?

chain-signatures/chain-midnight/src/indexer.rs
```diff
@@ -0,0 +1,1911 @@
...
+                continue;
+            }
+            // Close any finality gap so a lagging subscription cannot skip
+            // notifications; each height processes exactly like catchup.
+            for number in (last_processed + 1)..block.number {
+                match self
+                    .index_height(source, &mut cache, &events_tx, number, &cancel)
+                    .await?
+                {
+                    Indexed::Done => {}
```
ppca, line 652: can add last_processed = number, so the progress is more accurate

Pessina:
> Do you plan to add integration tests in this or following PR?

@isSerge  Yes, it will come on following PRs.

chain-signatures/chain-midnight/src/indexer.rs
```diff
@@ -0,0 +1,1911 @@
...
+}
+
+struct LiveSource {
+    rpc: MidnightRpc,
+}
+
+impl LiveSource {
+    async fn connect(config: &MidnightConfig) -> anyhow::Result<Self> {
+        Ok(Self {
+            rpc: MidnightRpc::connect(config).await?,
```
ppca: nit: this is correct by itself. should we retry the connect()? a network hiccup may not need to go through a supervised restart.

chain-signatures/chain-midnight/src/config.rs
```diff
@@ -0,0 +1,103 @@
...
+                max_delay: Duration::from_secs(10),
+                max_times: 5,
+                jitter: true,
+            },
+        }
+    }
+}
+
+/// Tuning for the indexing pipeline (catchup and the live finalized-head loop).
+#[derive(Clone, Debug, PartialEq)]
```
isSerge, line 57: Why does it have to be `PartialEq`? I could not find any place we need to check for equality.

chain-signatures/chain-midnight/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+midnight-serialize.workspace = true
+midnight-storage.workspace = true
+mpc-chain-integration-core.workspace = true
+mpc-crypto.workspace = true
+mpc-primitives.workspace = true
+# Crate-local rather than workspace so this manifest states its own features.
+# Cargo unifies features across this crate and `node` in a workspace build, so
+# nothing here may rely on `unstable-light-client` being absent. Keep the
+# version in lockstep with `chain-signatures/node/Cargo.toml`.
+subxt = { version = "0.44", default-features = false, features = ["jsonrpsee", "native"] }
```
isSerge: We already have `subxt` as dependency in `mpc-node` Cargo.toml. Worth centralizing at workspace level.

chain-signatures/chain-midnight/src/convert.rs
```diff
@@ -0,0 +1,259 @@
...
+mod tests {
+    use super::*;
+    use crate::records::SignBidirectionalRecord;
+    use mpc_primitives::{Chain, SignKind};
+
+    const READ_ADDRESS: [u8; 32] = [0xab; 32];
+    const REQUEST_ID: [u8; 32] = [0x77; 32];
+    const INDEXED_TS: u64 = 1_753_000_000;
+
+    fn ascii_padded<const N: usize>(text: &[u8]) -> [u8; N] {
```
isSerge: This util is defined multiple times

chain-signatures/chain-midnight/src/test_utils.rs
```diff
@@ -0,0 +1,253 @@
...
+use midnight_storage::storage::{Array, HashMap};
+use midnight_storage::DefaultDB;
+
+use crate::reader::Node;
+
+/// Wire-form atoms of a record, each field trimmed as the state layer stores it:
+/// trailing zeros dropped, a false Boolean the empty atom and a true one `[1]`,
+/// integers little-endian trimmed.
+pub(crate) fn atoms_from_record(record: &SignBidirectionalRecord) -> Vec<Vec<u8>> {
+    fn trim(bytes: &[u8]) -> Vec<u8> {
```
isSerge: duplicate, another one is defined in `indexer.rs`

isSerge (commented):
Couple of things caught my eye, but none are critical - feel free to add TODO comments and address later

chain-signatures/chain-midnight/src/config.rs
Pessina: Added in [`54cacc11`](https://github.com/sig-net/mpc/commit/54cacc118f5c9a9b4ea824b4a76aaf39f9dc10bd)

chain-signatures/chain-midnight/src/indexer.rs
Pessina: Fixed [`08f841b3`](https://github.com/sig-net/mpc/commit/08f841b3afb421b5a0bc3ba3a1eb9211bb50d0b6)

chain-signatures/chain-midnight/src/config.rs
Pessina: Now it's used here: [`into_str_args_round_trips_into_config`](https://github.com/sig-net/mpc/blob/287e352ba113522987cd5afd5cabac0f8f9d407b/chain-signatures/node/src/cli/args/midnight.rs#L130)

chain-signatures/chain-midnight/Cargo.toml
Pessina: Good point, moved: [`41a91f69`](https://github.com/sig-net/mpc/commit/41a91f69642981999c5db6b5e06c5854ee4fff4e)

chain-signatures/chain-midnight/src/convert.rs
Pessina: Fixed [`08f841b3`](https://github.com/sig-net/mpc/commit/08f841b3afb421b5a0bc3ba3a1eb9211bb50d0b6), [`287e352b`](https://github.com/sig-net/mpc/commit/287e352ba113522987cd5afd5cabac0f8f9d407b)

chain-signatures/chain-midnight/src/config.rs
Pessina: ~6s, added a comment in [`287e352b`](https://github.com/sig-net/mpc/commit/287e352ba113522987cd5afd5cabac0f8f9d407b)

chain-signatures/chain-midnight/src/test_utils.rs
Pessina: Fixed [`08f841b3`](https://github.com/sig-net/mpc/commit/08f841b3afb421b5a0bc3ba3a1eb9211bb50d0b6), [`287e352b`](https://github.com/sig-net/mpc/commit/287e352ba113522987cd5afd5cabac0f8f9d407b)

chain-signatures/chain-midnight/src/indexer.rs
Pessina: Added [`287e352b`](https://github.com/sig-net/mpc/commit/287e352ba113522987cd5afd5cabac0f8f9d407b)

chain-signatures/chain-midnight/src/indexer.rs
Pessina: Agree, I improved the err classifier and added clear logging so we can monitor after deployment on testnet. 

Commits [`287e352b`](https://github.com/sig-net/mpc/commit/287e352ba113522987cd5afd5cabac0f8f9d407b), [`42c23681`](https://github.com/sig-net/mpc/commit/42c23681a2fd812f46f74c5e6b658c553ad6fd78)

chain-signatures/chain-midnight/src/indexer.rs
Pessina: I get a rustc msg

```
value assigned to `last_processed` is never read
maybe it is overwritten before being read?
`#[warn(unused_assignments)]` (part of `#[warn(unused)]`) on by default
```

chain-signatures/chain-midnight/src/indexer.rs
```diff
@@ -0,0 +1,1975 @@
...
+        source: &C,
+        at_hash: &str,
+    ) -> anyhow::Result<Option<Node>> {
+        match source
+            .contract_state_tree(&self.config.central_address, at_hash)
+            .await?
+        {
+            ContractState::Tree(tree) => Ok(Some(tree)),
+            ContractState::Absent => Ok(None),
+            ContractState::Undecodable(err) => {
```
isSerge: Seems like we might enter infinite retry loop without recovery in case of this error

Pessina: Yes, that's correct and expected. If we can't decode it's a mismatch between the MPC version and Midnight Chain version so we halt and we need to update the code. 

This can happen because we are building against an rc branch. When we go mainnet things should be stable and this will only happen if they update on testnet first

I've improved a bit the err handling here: https://github.com/sig-net/mpc/pull/1059/commits/eec28abb59789b5479ad61564893387d3d5376c0

ppca (approved):

PR #1060, opened 2026-07-27
fix(rpc): deduplicate SignIds in publish batches

isSerge:
Currently nothing prevents publishing batches with the same `SignId` (can happen if node restarts), which may result in wasted gas and extra log noise. This PR implements a deduplication mechanism in `RpcExecutor` 

Changes:

- Add `dashmap` crate (already was a transitive dependency)
- Use `Arc<DashSet<SignId>>` to track in-flight publish requests in `RpcExecutor`
- Add `InFlightGuard` struct to release `SignId` from in-flight set on drop (so the slot is freed for re-publish)
- Add `executor_dedupes_concurrent_publishes_with_the_same_sign_id` test
- Update `make_indexed` and `make_publish_action` test utils to accept `SignId` param. Update tests to provide `SignId` explicitly

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7509,6 +7509,7 @@ dependencies = [
  "chrono",
  "ciborium",
  "clap",
+ "dashmap 6.1.0",
  "deadpool-redis",
  "ed25519-zebra",
  "enum-map",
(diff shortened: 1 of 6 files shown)


volovyks (approved):
Nice!

volovyks (approved):

PR #1061, opened 2026-07-27
Triple reservation

volovyks:
It appeared we never implemented this for presignature generation (triples).
Closes https://github.com/sig-net/mpc-private/issues/31

### DIFF
--- a/chain-signatures/node/src/protocol/presignature.rs
+++ b/chain-signatures/node/src/protocol/presignature.rs
@@ -7,8 +7,7 @@ use crate::protocol::contract::primitives::intersect_vec;
 use crate::protocol::posit::{PositInternalAction, PositRejectReason};
 use crate::protocol::MpcSignProtocol;
 use crate::storage::presignature_storage::{PresignatureSlot, PresignatureStorage};
-use crate::storage::protocol_storage::ProtocolArtifact;
-use crate::storage::triple_storage::{TriplesTaken, TriplesTakenDropper};
+use crate::storage::triple_storage::{TriplesReserved, TriplesTaken, TriplesTakenDropper};
 use crate::storage::TripleStorage;
 use crate::types::{PresignatureProtocol, SecretKeyShare};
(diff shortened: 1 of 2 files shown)


chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -483,33 +485,42 @@ impl PresignatureSpawner {
...
+            if participants.len() < self.threshold {
+                tracing::warn!(
+                    ?pair_id,
+                    ?active,
+                    ?participants,
+                    "intersection < threshold, skipping triple pair"
+                );
+                local_skip.push(pair_id);
+                // drop: in-memory reservation released, pair stays in storage
+                continue;
```
yvonneanne, line 514: iiuc, the analogous loop for the signature (`request/organize.rs:150`) has a timeout and and is sleeping between empty peeks. what is the reason for doing it differently here?

yvonneanne (approved):
LGTM with some open questions :)

1. Is there a deadlock opportunity when committing in protocol_storage.rs? Would scoping the guard so it drops before the away help and on a failed commit the pair becomes peekable again?
https://github.com/sig-net/mpc/pull/1072 

2. `peek_mine` has no `REDIS_LATENCY` observation** while `take_mine` does, so moving the triple path onto it silently zeroes `multichain_redis_operation_latency_ms{operation="take_mine"}` and leaves the new hot path unmeasured.

3.  does `take_mine` still has production caller? If not, can it be deleted?

4. take_mine` used `SPOP`, i.e. random, so head-of-line blocking was impossible, would that make sense here too?

chain-signatures/node/src/protocol/presignature.rs
yvonneanne: pushing the `holders ∩ active >= threshold` predicate into the Lua script would eliminate
round trips and skip list?

volovyks:
Thanks @yvonneanne! Merging and looking into all the mentioned issues.

PR #842, opened 2026-06-03
feat: upgrade near to latest

ChaoticTempest:
update to the latest near dependencies (protocol 85)

PR #1062, opened 2026-07-27
chore: build the contract with rust 1.93 instead of 1.81

Pessina:
`build-contract.sh` has pinned Rust 1.81.0 since before edition 2024 existed, and cargo 1.81 cannot parse an edition-2024 manifest anywhere in the workspace, which is what blocks the Midnight integration. This moves the toolchain to 1.93.0, matching #842, and leaves the rest of that modernization there: `near-sdk` stays at 5.7.0 and no contract source changes.

Any rustc past 1.81 emits bulk-memory opcodes that the sandbox `near-workspaces` 0.16.0 downloads refuses to deploy, so it goes to 0.21.0 with the sandbox pinned to 2.12.0, the first release on protocol 84. That exposed a near-fetch bug where a timed-out status query surfaces as a hard error instead of as pending, which `poll_status` now corrects.

### DIFF
--- a/.github/workflows/canton.yml
+++ b/.github/workflows/canton.yml
@@ -46,11 +46,11 @@ jobs:
           dpm install 3.5.1
           dpm --version
 
-      - name: Install Rust (1.81.0)
+      - name: Install Rust (1.93.0)
         uses: actions-rs/toolchain@v1
         with:
           profile: minimal
-          toolchain: 1.81.0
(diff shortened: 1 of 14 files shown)


ChaoticTempest (approved):
Nicely done, didn't think we could bump the rust version without doing all the near dependency changes so this is actually pretty good in the mean time before we upgrade all the deps

PR #1065, opened 2026-07-27
refactor: extract `mpc-chain-hydration`

isSerge:
This PR aims to decouple Hydration chain integration from `mpc-node` by moving into a separate crate:

- Create `mpc-chain-hydration` crate in the workspace, move config, indexer and publisher
- Update `HydrationIndexer` to implement `ChainIndexer` trait and listen to `CancellationToken`
- Update wiring in `cli/mod.rs` to use `run_supervised` (consistent with other chain integrations)
- Update imports in integration tests


ChaoticTempest (commented):
the changes for the most part look good to me, but I'm always skittish when we don't have any integration tests. Can we hold moving this to supervisor / chain indexer abstraction until we have a local hydration node testing the whole flow?

volovyks (approved):
@ChaoticTempest since we are not actively working on this integration, it should be fine to merge it with a risk of issues. We are not going to work on them for now.

PR #1070, opened 2026-07-28
fix(solana): guard CPI event slicing against too-short instruction data

ppca:


volovyks (approved):
Do we still need this?
`// TODO: enhance to handle malformed events and prevent panic`

ChaoticTempest (approved):
nice

PR #1072, opened 2026-07-28
fix(storage): release reserved guard before take_reserved_inner

yvonneanne:
ArtifactReservation::commit() bound the `reserved` write guard to a function-scoped local, so it was still held when calling take_reserved_inner(). That function re-acquires the same lock on both of its failure paths to roll `using` back, and tokio's RwLock is not reentrant — so any Redis failure during commit made the task wait on a guard it already owned and hang forever.

The only caller is the signature request task, so a single Redis hiccup wedged that task permanently, and with it every other task queued behind the `reserved` lock.

Scope the guard to a block. `using` is still set before the guard is released, which is what keeps remove_outdated() from pruning an artifact mid-commit; dropping it early additionally keeps the Redis round trip out of the critical section.

Adds a regression test that drives commit() against an unreachable Redis under a timeout — it hangs on the old code and passes on the fixed code.

See also https://github.com/sig-net/mpc/pull/1061#pullrequestreview-4794078106

### DIFF
--- a/chain-signatures/node/src/storage/protocol_storage.rs
+++ b/chain-signatures/node/src/storage/protocol_storage.rs
@@ -145,9 +145,15 @@ impl<A: ProtocolArtifact> ArtifactReservation<A> {
         drop(self);
 
         let me = storage.me().ok()?;
-        let mut guard = storage.reserved.write().await;
-        guard.pending.remove(&id);
-        guard.using.insert(id, true);
+        // The guard must be released before `take_reserved_inner`: that function
+        // re-acquires `reserved` on its failure paths, and `RwLock` is not
+        // reentrant — holding it here would deadlock the task. Dropping it early
+        // also keeps the Redis round trip out of the critical section.
+        {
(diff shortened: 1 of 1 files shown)


chain-signatures/node/src/storage/protocol_storage.rs
```diff
@@ -1051,3 +1061,106 @@ impl<A: ProtocolArtifact> ProtocolStorage<A> {
...
+    }
+
+    /// Regression test: `commit()` used to hold the `reserved` write guard across
+    /// `take_reserved_inner`, which re-acquires the same non-reentrant lock on its
+    /// failure paths. A Redis failure therefore hung the task forever, wedging every
+    /// other task waiting on `reserved`.
+    #[tokio::test]
+    async fn commit_does_not_deadlock_when_redis_is_unreachable() {
+        let storage = unreachable_storage();
+        let id = 42u64;
```
volovyks, line 1141: :)

volovyks (approved):
Wow, that is a nice catch!!

PR #1073, opened 2026-07-28
Remove take_mine

volovyks:
This should fix 2 and 3 from https://github.com/sig-net/mpc/pull/1061#pullrequestreview-4794078106


yvonneanne (approved):
LGTM

PR #1075, opened 2026-07-28
feat(primitives): add Chain::Midnight and its epsilon derivation

Pessina:
Carves `Chain::Midnight` and `derive_epsilon_midnight` out of #1059 so the one change in that stack that touches a wire format is reviewed on its own rather than inside a 7.5k-line diff. Adds the enum variant, the KDF wrapper and its re-export, the per-chain config arms, and epsilon plus `sender_string` routing on both bidirectional extension traits. The rest of #1059 rebases on top once this lands.

The chain id is `midnight:mainnet`. It is a chain-level constant here, not a per-environment one: `Chain::Ethereum` is `eip155:1` while tests run against 31337, and NEAR and Solana pin mainnet the same way. It is also the most stable of the candidates. Midnight has no registered CAIP-2 namespace at all (MIP-0008 proposes `midnight` but is still Draft, with the reference format undecided), and its illustrative registry keeps `midnight:mainnet` while flagging `midnight:testnet` as pending a core decision on whether `testnet` stays distinct from `preprod` and `preview`.

Follow-up required before anything derives keys against this: `@sig-net/midnight` still exports `MIDNIGHT_TESTNET_CHAIN_ID = "midnight:testnet"` as the default third argument of `deriveEpsilon`. The golden fixture here passes the chain id explicitly so its values stay anchored to that package, but the constant in midnight-integration has to move with this commit or integrators calling the default will derive different addresses than the node does.


signet-primitives/src/chain.rs
```diff
@@ -89,6 +93,7 @@ impl Chain {
             // ChainAgnostic/namespaces. "canton:global" follows the
             // namespace:reference format as a project-local identifier.
             Chain::Canton => "canton:global",
+            Chain::Midnight => "midnight:testnet",
```
ChaoticTempest: So midnight doesn't have a caip id either or is testnet actually expected for it?

signet-primitives/src/chain.rs
```diff
@@ -75,6 +78,7 @@ impl Chain {
             Chain::Bitcoin => "bip122:000000000019d6689c085ae165831e93",
             Chain::Hydration => "polkadot:2034",
             Chain::Canton => "canton:global",
+            Chain::Midnight => "midnight:testnet",
```
ppca: I think MPC only uses caip2_id for the mainnet of the network, it does not differentiate testnet and mainnet because one node will only connect to either testnet or mainnet at the same time.

signet-primitives/src/chain.rs
Pessina: They have a draft proposal for the caip2id: https://github.com/midnightntwrk/midnight-improvement-proposals/blob/main/mips/mip-0008-caip-2-network-identifiers.md

I would say we can use `midnight:mainnet` for now (per https://github.com/sig-net/mpc/pull/1075#discussion_r3671503392, cc @ppca)

Later when they settle we update it. Anyways we won't go live on mainnet before they release the new ledger version (expected to be released end of September)

signet-primitives/src/chain.rs
Pessina: See: https://github.com/sig-net/mpc/pull/1075#discussion_r3672699845

signet-primitives/src/chain.rs
Pessina: Updated: https://github.com/sig-net/mpc/pull/1075/commits/dca1a5e57144e80b96250e31fc692c7fd4b5fb31

ChaoticTempest (approved):

ppca (approved):

PR #1076, opened 2026-07-29
feat(ethereum): introduce shared finalized-head watcher

isSerge:
This PR introduces a single long-lived finalized-head watcher task which maintains a cached head (`watch::Sender<u64>`) on `refresh_finalized_interval` and replaces per block `eth_getBlockByNumber(Finalized)` requests in `wait_for_finalized_block`.

- Update `wait_for_finalized_block`: now instant for blocks the head already covers (no extra RPC requests), otherwise awaits the shared watcher
- Add `watch_finalized_head`: polls on cadence, publishes advances and retries forever (stream supervisor watchdog is the backstop)
- `catchup_finalized_head` + `is_finalized` plumbing removed, watcher head is a single source of truth for finality
- Add `spawn_finalized_head_watcher` method: used both in `run` method and benchmarks
- Update benchmarks to run in non-optimistic mode by default to mimic production environment

Benchmark changes (100 block range, against Alchemy RPC):
Finality RPCs (`getBlock(Finalized)`): 101 → 1
End-to-end catchup wall time: 6.66 s → 1.12 s
Blocks per sec: 15 → 89

Roughly ~6x improvement

### DIFF
--- a/chain-signatures/chain-ethereum/examples/helpers/mod.rs
+++ b/chain-signatures/chain-ethereum/examples/helpers/mod.rs
@@ -6,6 +6,7 @@ use mpc_chain_ethereum::{EthConfig, EthereumIndexer};
 use mpc_chain_integration_core::{
     utils::stream::chain_event_channel, MockStateManager, NoopChainTelemetry,
 };
+use tokio_util::sync::CancellationToken;
 
 /// Read a required environment variable, erroring if it's not set.
 pub fn opt_env(name: &str) -> anyhow::Result<String> {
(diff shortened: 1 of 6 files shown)


chain-signatures/chain-ethereum/examples/helpers/mod.rs
```diff
@@ -86,6 +90,11 @@ pub async fn run_catchup(
     let indexer = EthereumIndexer::new(config, state, NoopChainTelemetry).await?;
     let (events_tx, mut events_rx) = chain_event_channel();
 
+    // Maintain the finalized head the same way `run()` does, so OPTIMISTIC=0
+    // (the default) exercises the production finality-watch path.
+    let cancel = CancellationToken::new();
+    let _watcher = indexer.spawn_finalized_head_watcher(cancel.clone());
```
ppca, line 96: is it possible for the watcher spawned to silently die? And we can't recover if that happens?

ppca (approved):
mostly LGTM just one minor comment

chain-signatures/chain-ethereum/examples/helpers/mod.rs
isSerge: Good observation! Yes, but we recover: supervisor watchdog will fire, it will cancel and restart. But watchdog timeout is 35 min for Ethereum - we can probably do better. Let me think how to fail faster.

ppca: Thanks! 🙏

ppca: 35 min is alright altho slow since we will catchup eventually when we recover. So feel free to merge this and do that in another PR, not urgent and not like absolutely necessary.

isSerge: added TODO comment for this

PR #1077, opened 2026-07-29
fix: cap candidate join url length

ppca:
part 1 of fixing https://github.com/sig-net/mpc-private/issues/33
another PR with vote_remove_candidates that can restore our contract state and storage stake will come in a separate PR


volovyks (approved):
Looks good. I prefer this style:
```rust
#[error("Candidate URL is too long (max {} bytes).", crate::MAX_JOIN_URL_LEN)]
UrlTooLong,
```
But not a blocker

PR #1078, opened 2026-07-29
make join attack economic bad: require join deposit

ppca:
part 2 of https://github.com/sig-net/mpc-private/issues/33

### DIFF
--- a/chain-signatures/contract/src/lib.rs
+++ b/chain-signatures/contract/src/lib.rs
@@ -57,6 +57,9 @@ const UPDATE_CONFIG_GAS: Gas = Gas::from_tgas(5);
 // Maximum number of concurrent requests
 const MAX_CONCURRENT_REQUESTS: u32 = 128;
 
+/// Maximum accepted byte length for a candidate node URL.
+pub const MAX_JOIN_URL_LEN: usize = 2048;
+
 #[near_bindgen]
 #[derive(BorshDeserialize, BorshSerialize, Debug)]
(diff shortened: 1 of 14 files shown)


chain-signatures/contract/src/lib.rs
```diff
@@ -60,6 +60,9 @@ const MAX_CONCURRENT_REQUESTS: u32 = 128;
 /// Maximum accepted byte length for a candidate node URL.
 pub const MAX_JOIN_URL_LEN: usize = 2048;
 
+/// Fixed anti-spam deposit required to join as a candidate.
+pub const REQUIRED_JOIN_DEPOSIT: NearToken = NearToken::from_near(1);
```
volovyks, line 64: That may be too much. What is the math behind 1N?

ppca: I was thinking then it takes 2000N (4mb/2kb) to get to 4M on a state() read which makes mpc nodes unable to read state at all. Quite expensive for the attacker, and we could reset that what they did with threshold number of vote_remove_candidates call, very cheap.

But 1n was just trying to make it expensive. we can go down as long as it's still considerable.

ChaoticTempest: yeah 1N is not bad, just gotta remember for node providers that they need sufficient balance for this but their wallets should at least be around 5N anyways for all the checkpoint requests to the gov contract

ChaoticTempest (approved):

PR #1079, opened 2026-07-29
Collapses the two posit buffers + refactoring

volovyks:
This PR is mixing a couple of changes, but the goal is to have simpler, straightforward logic. I recommend reviewing commit by commit.

### [work queue -> mailbox renamings](https://github.com/sig-net/mpc/commit/2550fa61bed588d85def688edc75430dad348fbc)
Pure rename: `work_queue.rs` → `mailbox.rs`, `SignPositWorkQueue` → `PositMailbox`,
`SignTaskMessage` → `SignPositMessage`.

### [do not filter posit messages based on posit rounds in mailbox](https://github.com/sig-net/mpc/commit/a565cf86d2bdb1e66487c084e553054ed201e26d)
One slot per sender, last-arrived-wins (was: higher-round-wins). Rounds can
legitimately go down after a governance respawn, so the mailbox stays
round-agnostic and round judgment is the task's job.

### [carry the rejector's round inside StaleRound instead of the reject id](https://github.com/sig-net/mpc/commit/baa32d24389db629664c616202af326fbed5035d)
- The reject's id echoes the rejected message's round; the rejector's round
  rides in `StaleRound(usize)`. Receivers harvest it into `highest_seen_round`,
  keeping one-bump catch-up, and rejects feed `enough_rejects` naturally.
- A deliberator drops incoming Rejects: it has sent nothing that could be
  rejected, and answering one would ping-pong rejects between two nodes.

### [Reject messages for older rounds in advance()](https://github.com/sig-net/mpc/commit/b657a80410b6099626d7718ddb63b7eccbc44276)
The deliberation loop now also answers stale messages with `StaleRound` (was:
silent drop), so a behind peer catches up no matter which phase it hits.

### [drop future-round buffering, keep only the highest round](https://github.com/sig-net/mpc/commit/e3b6d15c29e28cc0d335e369ff3cbd7def00ce8c)
`SignState.buffered_messages` and its helpers are deleted: a future round is
remembered in `highest_seen_round` and the message dropped — `bump_round`
already jumps to the high-water. The high-water resets on governance respawn by
construction (`SignState` is rebuilt per task incarnation).

### DIFF
--- a/chain-signatures/node/src/protocol/request/mailbox.rs
+++ b/chain-signatures/node/src/protocol/request/mailbox.rs
@@ -86,7 +86,7 @@ mod tests {
         mailbox.push(posit(1, 0, PositAction::Propose));
         mailbox.push(posit(2, 3, PositAction::Accept));
 
-        let mut got = vec![
+        let mut got = [
             mailbox.try_recv().expect("sender 1"),
             mailbox.try_recv().expect("sender 2"),
         ];
--- a/chain-signatures/node/src/protocol/request/mailbox.rs
+++ b/chain-signatures/node/src/protocol/request/mailbox.rs
@@ -8,17 +8,15 @@ use std::sync::Arc;
 use cait_sith::protocol::Participant;
 use tokio::sync::Notify;
 
-/// Posit messages routed to a running signature task.
-pub(crate) enum SignTaskMessage {
-    PositMessage {
-        presignature_id: PresignatureId,
-        round: usize,
-        from: Participant,
-        action: PositAction,
-    },
+/// A posit message routed to a signature task.
+pub(crate) struct SignPositMessage {
+    pub presignature_id: PresignatureId,
+    pub round: usize,
+    pub from: Participant,
+    pub action: PositAction,
 }
 
-/// Work queue for posit messages, keyed by sending participant.
+/// Mailbox holding the latest posit message per sending participant.
 ///
 /// A message for round N from sender P replaces any earlier message from P.
 ///
@@ -29,13 +27,13 @@ pub(crate) enum SignTaskMessage {
 ///   here. But we can't get a `Start` message for a round that we haven't already
 ///   responded to a `Propose` message. Therefore, one message per round is all we
 ///   need to buffer.
-pub(crate) struct SignPositWorkQueue {
+pub(crate) struct PositMailbox {
     // using std Mutex here, do not hold across .await
-    messages: std::sync::Mutex<HashMap<Participant, SignTaskMessage>>,
+    messages: std::sync::Mutex<HashMap<Participant, SignPositMessage>>,
     notify: Notify,
 }
 
-impl SignPositWorkQueue {
+impl PositMailbox {
     pub(crate) fn new() -> Arc<Self> {
         Arc::new(Self {
             messages: std::sync::Mutex::new(HashMap::new()),
@@ -45,8 +43,8 @@ impl SignPositWorkQueue {
 
     /// Buffer `msg` in its sender's slot, overwriting when its round is `>=` the
     /// buffered one, then wake one consumer.
-    pub(crate) fn push(&self, msg: SignTaskMessage) {
-        let SignTaskMessage::PositMessage {
+    pub(crate) fn push(&self, msg: SignPositMessage) {
+        let SignPositMessage {
             from,
             round: new_round,
             ..
@@ -55,7 +53,7 @@ impl SignPositWorkQueue {
         let mut inserted = false;
         match guard.entry(from) {
             Entry::Occupied(mut occupied_entry) => {
-                let SignTaskMessage::PositMessage {
+                let SignPositMessage {
                     round: existing, ..
                 } = occupied_entry.get();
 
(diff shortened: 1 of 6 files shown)


volovyks:
Moved some changes (commit-based) into their own PRs: 
- https://github.com/sig-net/mpc/pull/1082
https://github.com/sig-net/mpc/pull/1081

Other things may be good, but depend on design decisions.

chain-signatures/node/src/protocol/request/mailbox.rs
```diff
@@ -0,0 +1,116 @@
...
+        })
+    }
+
+    /// Store `msg` in its sender's slot — last arrived wins — then wake one
+    /// consumer.
+    pub(crate) fn push(&self, msg: SignPositMessage) {
+        self.messages.lock().unwrap().insert(msg.from, msg);
+        // Wake up potential work consumers. (after releasing the lock)
+        self.notify.notify_one();
+    }
```
yvonneanne, line 43: We don't have any ordering guarantees from transport or sequence numbers, afaics. In that case letting the last msg override the slot is dangerous, isn't it? Maybe we can make the guard epoch-aware instead?
```suggestion
    /// Store `msg` in its sender's slot, wake up consumer:
    /// newer epoch always wins; 
    /// within the same epoch, never let a stale round clobber
    /// fresher messages, despite unordered delivery
pub(crate) fn push(&self, msg: SignPositMessage) {
    let mut guard = self.messages.lock().unwrap();
    match guard.entry(msg.from) {
        Entry::Occupied(mut slot) => {
            if (msg.epoch, msg.round) < (slot.get().epoch, slot.get().round) {
                return; // no wakeup needed
            }
            slot.insert(msg);
        }
        Entry::Vacant(slot) => { slot.insert(msg); }
    }
    drop(guard);
    self.notify.notify_one();
}
```

chain-signatures/node/src/protocol/request/posit.rs
```diff
@@ -304,22 +294,17 @@ impl PositPhase {
...
-                            "Storing message for future round",
+                            "Dropping message from future round",
                         );
-                        state.buffer_future_posit_message(SignPositMessage {
-                            presignature_id,
-                            round: peer_round,
-                            from,
-                            action,
-                        });
+                        state.highest_seen_round = state.highest_seen_round.max(peer_round);
```
jakmeier, line 307: I would prefer to keep the one-message buffer for the highest seen round, I think it has an important purpose.

For example, if a node drops the proposal for a future round, the node will never process the proposal, thus never accept and it will effectively not participate in that round.

chain-signatures/node/src/protocol/request/mailbox.rs
jakmeier: Agreed, I think replacing based only on time-of-arrival is wrong.

Using another number (e.g. epoch) combined with the round could work. Or maybe we can preserve the round number across governance updates, an idea @yvonneanne brought up to me a moment ago in call. That might be the simplest solution.

jakmeier (commented):
Many good changes. I really like the improved stale round handling. The renaming is also good.

For collapsing posit buffers and changing how messages are filtered, we should be careful. See my inline comments for my specific concerns.

yvonneanne:
With #1081 merged and commits 3 and 4 carried by #1082 and #1104, what remains here are the mailbox and buffering commits. I'd drop both.

**Mailbox round filtering:** its rationale (rounds go down after a respawn) is fixed at the root by #1100, which makes per-sender rounds monotonic. Last-arrived-wins would add a real hazard: the outbox retries each partition independently, so a delayed stale message can land after a fresher one and overwrite it, and the fresh one is never retransmitted; that costs a full round timeout.

**Future-round buffering:** the buffered Propose is what makes catch-up productive. A behind node that jumps to round N joins it only via that buffered Propose, since the proposer never retransmits; dropping it makes every catch-up jump cost an extra timeout, and #1104 makes such jumps more frequent. Safety is unaffected, so I'd defer this and delete the buffer only if `SIGN_REQUEST_LOOPS` shows it rarely fires once #1100 and #1051 are in.

chain-signatures/node/src/protocol/request/mailbox.rs
yvonneanne: Preserving the round across governance updates is now implemented in #1100: the round lives in the retained SignEntry and each task incarnation resumes from it, so per-sender rounds stay monotonic and the highest-round-wins slot stays correct. No epoch tag needed, which also avoids a wire change. Would appreciate your review there.

chain-signatures/node/src/protocol/request/posit.rs
yvonneanne: Agreed, the buffer should stay. I recommended dropping this commit in my PR-level comment: without the buffered Propose, a node that jumps to a learned round has nothing to accept (the proposer never retransmits), and #1104 makes such jumps more frequent. If simplification is still wanted later, SIGN_REQUEST_LOOPS can tell us whether the buffer actually fires once #1100 and #1051 are in.

yvonneanne:
This PR is now superseded by PRs #1082, #1100, #1104 and issue #1105

PR #1082, opened 2026-07-29
Carry the rejector's round inside StaleRound instead of the reject id

volovyks:
Reshapes the `StaleRound` reject introduced in #1048:

- The reject's id now echoes the rejected message's round (an address, not
  content); the rejector's current round rides inside `StaleRound(usize)`

### DIFF
--- a/chain-signatures/node/src/protocol/posit.rs
+++ b/chain-signatures/node/src/protocol/posit.rs
@@ -57,19 +57,19 @@ pub enum PositAction {
 
 #[derive(Debug, PartialEq, Serialize, Deserialize, Clone, Copy, Hash)]
 pub enum PositRejectReason {
-    Unknown = 0,
+    Unknown,
     /// The node is already participating in a generation, or has already
     /// finished generation.
-    AlreadyGenerating = 1,
+    AlreadyGenerating,
     /// The node cannot participate because it doesn't have the required
     /// artifact.
-    MissingArtifact = 2,
+    MissingArtifact,
     /// The posit message is invalid, usually because of bad timing leading to
     /// round / proposer mismatches.
-    InvalidRequest = 3,
-    /// The sender's round is behind the rejector's current round. The reject
-    /// carries the rejector's round so the sender can catch up in one bump.
-    StaleRound = 4,
+    InvalidRequest,
+    /// The message's round is behind the rejector's current round, carried
+    /// here so the sender can catch up in one bump.
+    StaleRound(usize),
 }
 
 impl PositAction {
--- a/chain-signatures/node/src/protocol/posit.rs
+++ b/chain-signatures/node/src/protocol/posit.rs
@@ -68,8 +68,8 @@ pub enum PositRejectReason {
     /// round / proposer mismatches.
     InvalidRequest,
     /// The message's round is behind the rejector's current round, carried
-    /// here so the sender can catch up in one bump.
-    StaleRound(usize),
+    /// in `PositMessage::stale_round` so the sender can catch up in one bump.
+    StaleRound,
 }
 
 impl PositAction {
--- a/chain-signatures/node/src/protocol/request/posit.rs
+++ b/chain-signatures/node/src/protocol/request/posit.rs
@@ -44,7 +44,7 @@ impl PositPhase {
                     Some(peer_current),
                 ) = (action, stale_round)
                 {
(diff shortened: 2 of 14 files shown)


yapignolet:
Cool, much better!

chain-signatures/node/src/protocol/request/posit.rs
```diff
@@ -36,6 +36,21 @@ impl PositPhase {
...
                 } = &task_msg;
 
+                // A StaleRound reject carries the rejector's current round;
+                // remember it so our next bump catches up in one step.
+                if let PositAction::RejectWithReason(PositRejectReason::StaleRound(peer_current)) =
+                    action
+                {
+                    state.highest_seen_round = state.highest_seen_round.max(*peer_current);
+                }
+
```
yapignolet: should this be tested too?

volovyks: Extended the tests

yvonneanne (approved):

chain-signatures/node/src/protocol/posit.rs
```diff
@@ -57,19 +57,19 @@ pub enum PositAction {
...
-    MissingArtifact = 2,
+    MissingArtifact,
     /// The posit message is invalid, usually because of bad timing leading to
     /// round / proposer mismatches.
-    InvalidRequest = 3,
-    /// The sender's round is behind the rejector's current round. The reject
-    /// carries the rejector's round so the sender can catch up in one bump.
-    StaleRound = 4,
+    InvalidRequest,
+    /// The message's round is behind the rejector's current round, carried
```
yvonneanne, line 70: Unit → tuple changes this variant's CBOR shape and breaks wire compat.

Externally tagged, StaleRound encodes as a bare text string (6a "StaleRound"); StaleRound(usize) encodes as a map (a1 6a "StaleRound" 05). A single message seems to decode fine only because ciborium::from_reader ignores trailing bytes — in the real Vec<Message> batch decode the leftover byte desyncs the stream and both directions fail.

PositRejectReason has no fallback (unlike Message::Unknown, message/types.rs:174-177), so inbox.rs drops the entire batch — including unrelated messages bundled alongside — with just a warn!. During a rolling upgrade, posit traffic between mixed-version peers silently disappears until every node is upgraded.

Can we keep the variant shape and carrying the round as an optional field instead?

yapignolet:
Pushed two commits addressing the wire-compat comment above:

**`c4a73677`** keeps `StaleRound` a unit variant and carries the rejector's round in a new optional `PositMessage.stale_round` field instead. Same behavior, but wire-compatible: the unit→tuple change altered the variant's CBOR shape and would desync the `Vec<Message>` batch decode between mixed-version nodes during a rolling restart, silently dropping whole batches. `test_posit_stale_round_field_is_wire_compatible` proves both directions against real encrypted batches.

**`f75937a8`** clears `buffered_messages` when a StaleRound reject raises `highest_seen_round`, as the invariant on `SignState::buffered_messages` requires.

yvonneanne:
Heads-up: #1104 adds a third `StaleRound` send site, in `PositPhase::advance()`'s deliberation loop (the `state.round > peer_round` branch in `request/posit.rs`). If this PR lands after #1104, the rebase should convert that site to the `StaleRound(usize)` convention as well, and adjust the two tests there that assert the round is stamped in the reject id (`advance_rejects_older_round_with_stale_round`, `behind_peer_harvests_rejectors_round`). Nothing breaks if the site is missed, a mixed-convention reject just degrades to the old one-round-per-attempt catch-up on the receiver, but the sites should stay uniform.

ppca (approved):

yvonneanne:
Correction to my heads-up above: since c4a73677 the round rides in `PositMessage.stale_round`, not in the variant. So on rebase the #1104 send site in `advance()` becomes `id: ...Signature(sign_id, presignature_id, peer_round)` plus `stale_round: Some(state.round)`, and its tests should assert the id carries the rejected round (2) with `stale_round == Some(5)`. A side effect worth noting: with the id echoing the rejected round, the reject is no longer a future-round message on the receiver. It is handled immediately via `record_peer_round` instead of landing in the one-slot mailbox, which removes the ordering dependency of #1104 on #1100 (that dependency assumed the reject plants a high-round mailbox entry). Also heads-up on scope: this branch predates the mailbox refactor (`posit_queue`/`SignTaskMessage`), so the rebase over develop plus #1100 and #1104 is not mechanical.

PR #1084, opened 2026-07-30
feat: add vote for a new threshold

ChaoticTempest:
add in the ability to vote for a new threshold. It needs to be at minimum `compute_threshold(participants.len())` and at maximum `participants.len()`. Once voted in via prior threshold, it will trigger resharing with the same participant set as the new participant set, just with different threshold.

Our devnet is currently at 5 threshold and 12 nodes, so we need to utilize this so we don't end up with split brain. So will need to set new threshold this way.


chain-signatures/contract/src/lib.rs
```diff
@@ -458,6 +460,60 @@ impl VersionedMpcContract {
...
+                public_key,
+                threshold_votes,
+                ..
+            }) => {
+                let participants_len = participants.len();
+                let min_threshold = compute_threshold(participants_len);
+                if new_threshold < min_threshold || new_threshold > participants_len {
+                    return Err(VoteError::ThresholdOutOfRange.into());
+                }
+                if new_threshold == *threshold {
```
ppca: can check this before doing  compute_threshold?

chain-signatures/contract/src/lib.rs
```diff
@@ -458,6 +460,60 @@ impl VersionedMpcContract {
         }
     }
 
+    /// Vote to change the running threshold without otherwise modifying the
+    /// participant set. Multiple competing thresholds can be voted on in
+    /// parallel; the first to reach the current `threshold` triggers a
+    /// resharing whose `new_threshold` is the proposed value.
+    #[handle_result]
+    pub fn vote_threshold(&mut self, new_threshold: usize) -> Result<bool, Error> {
```
ppca, line 502: question: let's say with 12 nodes, we voted threshold to be 7. And later, 2 nodes dropped out, left with 10 nodes, the current contract code says the resharing will bring threshold down to 6 in vote_leave. Is that what we desire? I think it's alright no issues in correctness, but want to check if that's what we want.

ChaoticTempest: yep, that's what's happen with our dynamic threshold when participants leave and join. We can increase it back to 7 we desire so. I don't know what threshold would be best, so maybe we could have it never decrease on people leaving unless it drops below threshold

chain-signatures/contract/src/lib.rs
ChaoticTempest: true, rearranged it

chain-signatures/contract/src/lib.rs
ppca: seems alright to leave it as is. seems more natural to have lower threshold when participants set is smaller.

ppca (approved):

chain-signatures/node/src/protocol/mod.rs
```diff
@@ -75,6 +75,11 @@ pub trait Governance {
         &self,
         public_key: &near_crypto::PublicKey,
     ) -> impl std::future::Future<Output = anyhow::Result<bool>> + Send;
+
+    fn vote_threshold(
```
yvonneanne, line 79: Who/what is calling this?

chain-signatures/node/src/protocol/consensus.rs
```diff
@@ -911,3 +911,187 @@ async fn resharing(
...
+        );
+        let contract_state = build_running_state(
+            /* epoch */ 2,
+            /* threshold */ 3, // matches new_threshold
+            participants,
+        );
+
+        // SAFETY: the `ProtocolState::Running` branch we are exercising never
+        // touches `_ctx` or `_gov`, so casting an uninitialised slot satisfies
+        // the trait bounds without producing UB on this path.
```
yvonneanne, line 1015: Nit: advance() never uses ctx (it's _ctx), so the test doesn't really need one either. If we
move the body into a helper that doesn't take ctx, the unsafe block goes away

chain-signatures/contract/src/lib.rs
```diff
@@ -458,6 +460,60 @@ impl VersionedMpcContract {
...
+                ..
+            }) => {
+                let participants_len = participants.len();
+                if new_threshold == *threshold {
+                    return Err(VoteError::ThresholdUnchanged.into());
+                }
+                let min_threshold = compute_threshold(participants_len);
+                if new_threshold < min_threshold || new_threshold > participants_len {
+                    return Err(VoteError::ThresholdOutOfRange.into());
+                }
```
yvonneanne, line 530: The upper bound permits new_threshold == participants_len, i.e., no fault tolerance. Let's exclude that too.
```suggestion
                let min_threshold = compute_threshold(participants_len);
                let max_threshold = participants_len.saturating_sub(1);
                if new_threshold < min_threshold || new_threshold > max_threshold {
                    return Err(VoteError::ThresholdOutOfRange.into());
                }
```

chain-signatures/contract/src/lib.rs
```diff
@@ -458,6 +460,60 @@ impl VersionedMpcContract {
...
+                let voted = threshold_votes.entry(new_threshold);
```
yvonneanne: Do I understand the code correctly that `threshold_votes` is keyed by proposed threshold, so a single participant can hold live votes for several thresholds simultaneously, and there is no way to withdraw one?

yvonneanne: The contract already establishes a possible cleanup pattern elsewhere: remove_candidacy at lib.rs:358 does join_votes.remove(&signer_account_id). 
So we could fix it by adding this in to ThresholdVotes (contract/src/primitives.rs:396):

```
/// Record `voter`'s support for `threshold`, dropping any vote they had
/// previously cast for a different threshold, and return the resulting
/// number of voters backing `threshold`. A participant backs at most one
/// proposed threshold at a time.
pub fn vote(&mut self, threshold: usize, voter: AccountId) -> usize {
    for voters in self.votes.values_mut() {
        voters.remove(&voter);
    }
    self.votes.retain(|_, voters| !voters.is_empty());
    let voters = self.votes.entry(threshold).or_default();
    voters.insert(voter);
    voters.len()
}
```
and the call site becomes:

`if threshold_votes.vote(new_threshold, voter) >= *threshold {`

The retain also keeps the map from accumulating empty entries as operators change their minds. Probably, worth a test that voting for 9 and then 10 leaves exactly one vote, under 10.

Wdyt?

jakmeier: I think what @yvonneanne suggests would make sense, a node should be able to change an already cast vote.

If that's what we implement, I wonder if retracting a vote entirely by voting to keep the current threshold should also be possible. So, instead of returning `ThresholdUnchanged`, the contract should remove an existing vote and not produce an error.

chain-signatures/contract/src/primitives.rs
```diff
@@ -361,6 +361,14 @@ pub struct PkVotes {
     pub votes: BTreeMap<PublicKey, HashSet<AccountId>>,
 }
 
+#[derive(BorshDeserialize, BorshSerialize, Serialize, Deserialize, Debug, Clone, PartialEq, Eq)]
+pub struct ThresholdVotes {
+    /// Maps each proposed new threshold to the set of participants that have
+    /// voted for it. Multiple thresholds can be voted on in parallel; the first
+    /// to reach the running threshold triggers a resharing.
+    pub votes: BTreeMap<usize, HashSet<AccountId>>,
```
jakmeier: If we end up enforcing that each node can only vote for one threshold at the time, `BTreeMap<usize, HashSet<AccountId>>,` is a bit awkward, as it can represent many invalid states that need to be enforced by code.

Perhaps `votes: BTreeMap<AccountId, usize>` could represent "each node has one vote" more cleanly.

Or even `votes: Vec<usize>` using the participant id as index could work.

chain-signatures/node/src/protocol/mod.rs
ChaoticTempest: currently, nodes will not call it automatically. I don't envision this to be automated in the future either. Only node providers will be calling this. The reason why its added to the interface here is so we can have testing on it

chain-signatures/node/src/protocol/consensus.rs
ChaoticTempest: good point, but requires more refactoring than I would like to add to this already big PR. Will save it for another time: https://github.com/sig-net/mpc/issues/1087

chain-signatures/contract/src/lib.rs
ChaoticTempest: yeah, I guess this is better than adding a new function to retract voting. Will make voting unique to voter. Not a bad idea to vote for the same threshold to retract your vote entirely.

chain-signatures/contract/src/primitives.rs
ChaoticTempest: quite the optimization with `votes: Vec<usize>` but would require hand rolling more code for us to manage, so went with `BTreeMap<AccountId, usize>` instead

PR #1085, opened 2026-07-30
feat: cap candidates set size

ppca:
part 3 of fixing https://github.com/sig-net/mpc-private/issues/33
Purpose:
To limit the possibility of join() attack, we've so far added url size cap and requiring non-refundable 1N deposit when calling join(). With those, it is still possible to bloat our payload by joining with 2000 accounts if the attacker is willing to spend money (2000N). 
This PR aims to cap the number of candidates to 10, and on a new join(), evict the oldest candidate if already at 10. This will make sure the payload cannot be bloated and genuine nodes can still join.
Now there's no way to bloat the contract state. But it is still possible for attacker to stop genuine nodes from joining participants by joining with new accounts to push the genuine nodes out. However they'd need to continuously do that to permanently stop a genuine node, which will be costly. But if they have some kind of bot out to get us, well, it's not good, but they spend 10x than us in this race.

I also thought about vote_remove_candidates and an admin method like clear_candidates. Latter is centralized and former basically ask node operators to be part of managing candidates, a bit weird too.

What changed:
1. candidates changed from btreemap to VecDeque so that it naturally preserves order. The lookup would now be O(candidates_set_size) but since it's capped at 10, it's quite small. Thought about adding an order: VecDeque<AccountId>, that's ugly; also maybe using an ordered map type, but seems to add unnecessary imports to the wasm.
2. evict oldest candidate in join() when candidates size already 10

Note even if old candidate have some votes, they will be evicted. Reason is our assumption is attacker haven't controlled >= T nodes, but we do not assume they control no nodes. So if they control some nodes already, but not >= T yet, then we'd still be able to have genuine nodes join.

### DIFF
--- a/chain-signatures/contract/src/lib.rs
+++ b/chain-signatures/contract/src/lib.rs
@@ -63,6 +63,9 @@ pub const MAX_JOIN_URL_LEN: usize = 2048;
 /// Fixed anti-spam deposit required to join as a candidate.
 pub const REQUIRED_JOIN_DEPOSIT: NearToken = NearToken::from_near(1);
 
+/// Maximum number of candidate applications retained at a time.
+pub const MAX_CANDIDATES: usize = 10;
+
 #[near_bindgen]
 #[derive(BorshDeserialize, BorshSerialize, Debug)]
 pub enum VersionedMpcContract {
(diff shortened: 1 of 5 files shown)


volovyks:
@ppca can I close it?

PR #1092, opened 2026-07-31
refactor(ethereum): decompose indexer

isSerge:
Refactor Ethereum indexer by extracting few largely independent concerns: 

1. Execution confirmation watching
2. Finalized-head tracking
3. Cancellation-aware retry

Main changes: 
- New shared primitives (`chain-integration-core::utils::task`)
  - Add `CancellationTokenExt` trait that implements `cancelled_within(dur)` method - a cancellation-aware sleep
  - Add `retry_until_ok(cancel, delay, label, process)` helper - retries async operation in the loop with backoff until Ok or cancel (re-used in both Ethereum and Solana crates)
- Create `ExecutionWatcher` (`chain-ethereum/src/execution_watcher.rs`) to isolate execution-confirmation logic
- Create `FinalizedHeadTracker` (`chain-ethereum/src/finalized_head.rs`) - owns the finalized-head `watch` channel, exposes `current()`, `wait_for(block)`, `spawn_watcher(client, cancel)` method. The optimistic-mode gate lives here now.

No behavior changes. `indexer.rs` went from 3177 to 1393 LOC


chain-signatures/chain-integration-core/src/utils/task.rs
```diff
@@ -7,3 +12,44 @@ impl Drop for AbortOnDrop {
...
+        tokio::select! {
+            _ = cancel.cancelled() => return,
+            result = process() => match result {
+                Ok(()) => return,
+                Err(err) => tracing::warn!(?err, "{label} failed; retrying"),
+            },
+        }
+        if cancel.cancelled_within(delay).await {
+            return;
+        }
```
ChaoticTempest, line 53: can't we just do:

```suggestion
        tokio::select! {
            cancelled = cancel.cancelled_within(delay) => if cancelled { return },
            result = process() => match result {
                Ok(()) => return,
                Err(err) => tracing::warn!(?err, "{label} failed; retrying"),
            },
        }
```

ChaoticTempest (approved):

chain-signatures/chain-integration-core/src/utils/task.rs
ChaoticTempest: actually, I guess the sleep would actually cancel the `process` function here. so nevermind

PR #1100, opened 2026-08-04
fix(request): carry posit round across respawns

yvonneanne:
## Problem

A respawned sign task (`spawn_tasks` on a governance change) rebuilds `SignState` and restarted at round 0, while its `SignEntry`, its mailbox, and every peer kept the old round. Two symptoms:

- Peers' mailboxes silently drop the fresh low-round messages: the `PositMailbox` slot keeps the highest round per sender, so a leftover pre-respawn message wins over a legitimate new round-0 Propose.
- Leftover high-round messages get harvested into `highest_seen_round`, sending nodes to a round nobody else is in.

## Fix

Carry the round in `SignEntry` (which already outlives incarnations, like `is_proposer`) via an `Arc<AtomicUsize>`, threaded through `SignTask` into `SignState::new`. The `round` field is now private; `set_round()` is the sole write path and keeps the carried value in step.

This makes per-sender rounds monotonic for the life of an entry — the invariant the mailbox's highest-round-wins slot already relies on: the outbox spawns an independent retry loop per encrypted partition (`outbox.rs`), so a stale message can arrive after a fresher one and arrival order alone is not trustworthy. This is also the reason **not** to make the mailbox round-agnostic (#1079, second commit): last-arrived-wins would let a retried stale message overwrite a fresh one, which is never retransmitted at the posit level.

## Tests

- `round_survives_a_respawn` (state.rs): a rebuilt `SignState` resumes from the carried value.
- Lifecycle test step 6 (mod.rs): the respawned incarnation shares the entry's round (Arc identity), rather than minting a fresh one.
- `one_slot_per_sender_highest_round_wins` (mailbox.rs): pins the guard, which previously had no test.

`cargo test -p mpc-node --lib` 159/0, clippy and fmt clean.

## Known limits (deliberate)

- A full reset (process restart, or retire + re-add after chain abort) still starts at 0; recovery there runs via StaleRound rejects, as today. The mailbox doc comment states this.
- `highest_seen_round` and the future-round buffer are intentionally not carried; a respawn costs at most one extra reorganize before re-harvesting.
- Respawn no longer acts as an accidental round re-sync across nodes. Round convergence rests on StaleRound harvesting (ahead nodes advance +1, behind nodes jump); the deterministic rotation + growing round timeouts in #1051 further break lockstep divergence.

### DIFF
--- a/chain-signatures/node/src/protocol/request/mailbox.rs
+++ b/chain-signatures/node/src/protocol/request/mailbox.rs
@@ -18,7 +18,7 @@ pub(crate) struct SignPositMessage {
 
 /// Mailbox holding the latest posit message per sending participant.
 ///
-/// A message for round N from sender P replaces messages from P for earlier rounds.
+/// A message for round N from sender P replaces messages from P for earlier or equal rounds.
 ///
 /// Only a single message per round and sender buffered. This is enough because:
 ///
--- a/chain-signatures/node/src/protocol/request/mailbox.rs
+++ b/chain-signatures/node/src/protocol/request/mailbox.rs
@@ -18,7 +18,12 @@ pub(crate) struct SignPositMessage {
 
 /// Mailbox holding the latest posit message per sending participant.
 ///
-/// A message for round N from sender P replaces any earlier message from P.
+/// Ordered by round, not by arrival: messages travel as independent HTTP
+/// requests with their own retries, so a stale one can land after a fresher
+/// one and must not overwrite it. Rounds normally only go up per sender (they
+/// are carried across task respawns); after the rare reset (process restart,
+/// retire and re-add) fresh low-round messages may be dropped while an old
+/// one sits unread — the task's StaleRound rejects drive catch-up instead.
 ///
 /// Only a single message per round and sender buffered. This is enough because:
 ///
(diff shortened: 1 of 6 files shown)


chain-signatures/node/src/protocol/request/mailbox.rs
```diff
@@ -18,7 +18,12 @@ pub(crate) struct SignPositMessage {
 
 /// Mailbox holding the latest posit message per sending participant.
 ///
-/// A message for round N from sender P replaces any earlier message from P.
+/// Ordered by round, not by arrival: messages travel as independent HTTP
+/// requests with their own retries, so a stale one can land after a fresher
+/// one and must not overwrite it. Rounds normally only go up per sender (they
+/// are carried across task respawns); after the rare reset (process restart,
+/// retire and re-add) fresh low-round messages may be dropped while an old
+/// one sits unread — the task's StaleRound rejects drive catch-up instead.
```
yvonneanne:
```suggestion
/// A message for round N from sender P replaces messages from P for earlier rounds.
```

ppca (approved):

ChaoticTempest:
should we also carry the round in the case that the task gets aborted via `AbortChain`? This happens when the chain for our node gets regressed back to a specific point, it proceeds to catchup to whatever is latest on the chain, but then decides to requeue the task again. My intuition says yes because other nodes more than likely did not regress and so are continuing on with the same sign request, so when we requeue it, we should carry the round here as well. But I don't like snowballing it, so we can leave it for another branch/PR

edit: ahh didn't read when you wrote `A full reset (process restart, or retire + re-add after chain abort) still starts at 0; recovery there runs via StaleRound rejects, as today. The mailbox doc comment states this.`, so I guess that's fine then.

yvonneanne:
Right, `AbortChain` -> `retire_task` drops the `SignEntry` (and its mailbox), so the re-indexed request gets a fresh entry at 0, deliberately out of scope here. Recovery is bounded though: the peer's one-slot mailbox may swallow a low-round message from us while its stale slot is still unconsumed, but the peer's task drains its mailbox continuously, so the next message gets through, is answered with a `StaleRound` reject (in every phase once #1104 lands), and we jump in a single `bump_round`. Costs roughly one extra round, not a one-per-attempt climb.

chain-signatures/node/src/protocol/request/mailbox.rs
```diff
@@ -18,7 +18,7 @@ pub(crate) struct SignPositMessage {
 
 /// Mailbox holding the latest posit message per sending participant.
 ///
-/// A message for round N from sender P replaces any earlier message from P.
+/// A message for round N from sender P replaces messages from P for earlier rounds.
```
jakmeier: super nit: if you already update this comment, might as well make it precise

```suggestion
/// A message for round N from sender P replaces messages from P for earlier or equal rounds.
```

jakmeier (approved):
Nice, I'm glad you found a good way to solve the problem with governance contract state updates while keeping the mailbox rules intact.

Issue #1094, opened 2026-08-03
indexer: an eth request that somehow only got indexed by 4 nodes


ppca:
SignId("9612195bd73b1bb45d0c69a7a1e2d6b83ee4c57dc7bdd92f894b8e972dce220c")  only got indexed by 4 nodes ever on dev

isSerge:
There was a reorg at that block height [11400622](https://sepolia.etherscan.io/block/11400622/f) only 4 nodes parsed the canonical block, others did not. The indexer parses live blocks before finality but can't recover when a fetched block is reorged

PR #1101, opened 2026-08-04
fix(ethereum): prevent live path from processing block pre-finality

isSerge:
This is a local fix for #1094 

Main changes:
- Rename `index_live_blocks` to `track_chain_tip` - now sends live block numbers instead of eagerly fetching blocks
- Rename `process_live_item` to `process_finalized_block` - waits for finality, fetches canonical block, parses, emits events
- Extracted `emit_block_events` (emit only trail)
- Remove `CatchupItem::LiveBlock` in `client.rs` - only keep `BatchBlock {..}` and `Missing(block_id)` variants

Long-term solution will require a larger refactor: deferring *all* block fetching to post-finality (both catchup and live), eliminating reorg hash-check (dead code in production), etc. Will do in a separate PR


ppca (approved):
thanks! let's do the refactor too

PR #1102, opened 2026-08-05
checkpoint consensus: digest only rely on source chain observable info

### LINKED ISSUE #1095: checkpoint: nodes' checkpoints still differ even if the pending tx set is the same
[log](https://console.cloud.google.com/logs/query;query=%0AjsonPayload.chain%3D%22Ethereum%22%0AjsonPayload.block%3D%2211400820%22%0AjsonPayload.message%3D%22created%20checkpoint%22%0ASEARCH%2528%229612195bd73b1bb45d0c69a7a1e2d6b83ee4c57dc7bdd92f894b8e972dce220c%22%2529;lfeCustomFields=jsonPayload%252Ftarget;cursorTimestamp=2026-08-02T03:58:34.883057302Z;startTime=2026-08-02T01:07:19.599Z;endTime=2026-08-02T19:31:19.599Z?project=near-cs-dev)

ppca:
Fixes #1095

  ## Problem

  Nodes at the same block height with identical pending requests computed different
  `cumulative_digest` values, so they could not agree on a checkpoint. The digest
  committed to each request's full `SignStatus`, including `PendingGeneration` vs
  `PendingPublish` — local progress toward an on-chain event, where only a signature generation's participants can proceed to `PendingPublish`. And when there are a lot of pending requests, having threshold nodes agree on all requests' status, i.e. they need to all be or not be participants for all the requests, is almost impossible.

  ## Change

  have a minimal checkpoint_consensus_bytes() function:
```
match self {
    PendingGeneration | PendingPublish { .. } => vec![0],
    PendingExecution { .. }
    | PendingGenerationBidirectional
    | PendingPublishBidirectional { .. } => vec![1],
}
```

### DIFF
--- a/chain-signatures/node/src/sign_bidirectional.rs
+++ b/chain-signatures/node/src/sign_bidirectional.rs
@@ -38,6 +38,25 @@ impl SignStatus {
         matches!(self, SignStatus::PendingExecution { .. })
     }
 
+    /// Project this status onto what is observable at a checkpoint's own chain height.
+    ///
+    /// Generation and publication are local attempts to reach an on-chain event: only
+    /// a signature's participants advance to publishing, so nodes cannot be required
+    /// to agree on which of the two a request is in. They collapse to one value.
+    ///
+    /// The two values are "awaiting the initial response on this chain" and "past it".
+    /// The remaining boundary — target-chain execution versus the final response — is
+    /// gated on target-chain progress, which a checkpoint does not commit to, so it
+    /// collapses as well.
+    pub fn checkpoint_consensus_byte(&self) -> u8 {
+        match self {
+            SignStatus::PendingGeneration | SignStatus::PendingPublish { .. } => 0,
+            SignStatus::PendingExecution { .. }
+            | SignStatus::PendingGenerationBidirectional
+            | SignStatus::PendingPublishBidirectional { .. } => 1,
+        }
+    }
+
     pub fn execution_tx(&self) -> Option<&BidirectionalTx> {
         match self {
             SignStatus::PendingExecution { tx } => Some(tx),
--- a/chain-signatures/node/src/sign_bidirectional.rs
+++ b/chain-signatures/node/src/sign_bidirectional.rs
@@ -38,21 +38,6 @@ impl SignStatus {
         matches!(self, SignStatus::PendingExecution { .. })
     }
 
-    pub fn checkpoint_consensus_bytes(&self) -> Vec<u8> {
-        match self {
-            SignStatus::PendingGeneration => vec![0],
-            SignStatus::PendingPublish { .. } => vec![1],
-            SignStatus::PendingExecution { tx } => {
-                let mut bytes = vec![2];
-                bytes.extend_from_slice(tx.id.0.as_slice());
-                bytes.extend_from_slice(&tx.target_chain.to_bytes());
-                bytes
-            }
-            SignStatus::PendingGenerationBidirectional => vec![3],
-            SignStatus::PendingPublishBidirectional { .. } => vec![4],
-        }
-    }
-
     pub fn execution_tx(&self) -> Option<&BidirectionalTx> {
         match self {
             SignStatus::PendingExecution { tx } => Some(tx),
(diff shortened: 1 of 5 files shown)


ChaoticTempest:
im not sure why we need all this change to how we do the consensus bytes. I feel like we're adding more complexity than needed here, when all we need to do is have a minimal checkpoint_consensus_bytes() function:
```
match self {
    PendingGeneration | PendingPublish { .. } => vec![0],
    PendingExecution { .. }
    | PendingGenerationBidirectional
    | PendingPublishBidirectional { .. } => vec![1],
}
```

proceed with migration code to convert the previously stored checkpoint over to this new format.

ChaoticTempest:
I think we need to separate out node-local work vs chain verifiable work in a clean manner. We can do something like a `WorkJournal` which is a node local only way of tracking what phase of the work is being done, so we can have it be separate from `SignStatus` which tracks chain/indexer related events

ppca:
> im not sure why we need all this change to how we do the consensus bytes. I feel like we're adding more complexity than needed here, when all we need to do is have a minimal checkpoint_consensus_bytes() function:
> 
> ```
> match self {
>     PendingGeneration | PendingPublish { .. } => vec![0],
>     PendingExecution { .. }
>     | PendingGenerationBidirectional
>     | PendingPublishBidirectional { .. } => vec![1],
> }
> ```
> 
> proceed with migration code to convert the previously stored checkpoint over to this new format.

was mostly thinking plain sign request actually do not need any status as all of them would be carrying 0, and once they see respond they will be removed from pending tx. But yeah, what you said might be easier.

and migration is not necessary as long as we don't wipe or change version of redis since whatever in contract already would be able to match the old checkpoint in redis. 
dev always restart simulteneously, so all of them would start using new bytes at same time.

ppca:
> I think we need to separate out node-local work vs chain verifiable work in a clean manner. We can do something like a `WorkJournal` which is a node local only way of tracking what phase of the work is being done, so we can have it be separate from `SignStatus` which tracks chain/indexer related events

Let's do that in a followup and keep this one's scope small

ppca:
The `transition_to_bidirectional_response` function is needed now 
because `process_execution_confirmed` set the request kind and the status in two separately-locked
  steps, so in between an entry reads `(RespondBidirectional, PendingExecution)`. Execution
  confirmations run on the target chain's stream while checkpoints are cut on the source
  chain's, so a checkpoint can land in that gap and serialize the half-transitioned entry as
  its recovery payload.

  On recovery that entry only matches `pending_executions()` — it gets an execution watcher
  for an event that already fired — and never matches `take_requeueable_requests()` or
  `publishable_requests()`. So it is never signed, published, or removed, and the request
  stalls permanently.

  This used to be masked: under the old encoding the intermediate state hashed differently
  from the settled one, so such a checkpoint never reached quorum and the node regressed to a
  peer's clean payload. With the status-only encoding both hash to `1`, so the bad checkpoint
  now confirms and gets persisted. `transition_to_bidirectional_response` makes the two steps
  atomic and closes the window.

chain-signatures/node/src/sign_bidirectional.rs
```diff
@@ -38,18 +38,22 @@ impl SignStatus {
...
-                bytes.extend_from_slice(tx.id.0.as_slice());
-                bytes.extend_from_slice(&tx.target_chain.to_bytes());
-                bytes
-            }
-            SignStatus::PendingGenerationBidirectional => vec![3],
-            SignStatus::PendingPublishBidirectional { .. } => vec![4],
+            SignStatus::PendingGeneration | SignStatus::PendingPublish { .. } => 0,
+            SignStatus::PendingExecution { .. }
+            | SignStatus::PendingGenerationBidirectional
+            | SignStatus::PendingPublishBidirectional { .. } => 1,
```
ChaoticTempest: we should still have `PendingExecution` be a separate byte entry because it's tracked on the source chain as well via the first respond event we receive for bidirectional requests, so we should make this `1` and the next two `2`

ChaoticTempest (approved):
besides the one thing I mentioned, everything else looks good

PR #1104, opened 2026-08-05
fix(request): reject stale rounds in advance()

yvonneanne:
## Problem

The posit deliberation loop silently dropped messages for older rounds. A behind peer therefore only learned it was behind when it happened to hit `wait_for_propose`; in any other phase it burned its full timeout in silence, climbing one round per attempt instead of catching up.

## Fix

Answer older-round messages in `advance()` with a `StaleRound` reject stamped with our round, exactly as `wait_for_propose` already does. The behind peer harvests the stamped round into `highest_seen_round` and catches up in a single `bump_round`.

No ping-pong: rejects only answer strictly older rounds and carry the rejector's higher round, so the receiver classifies them as future-round messages (buffered and harvested, never answered).

## Tests

- `advance_rejects_older_round_with_stale_round`: the send side. A proposer at round 5 answers a round-2 straggler with `StaleRound` stamped 5, then exits to Organizing on its deadline.
- `behind_peer_harvests_rejectors_round`: the receive side. A deliberator at round 2 receiving a reject stamped 5 buffers it without answering (pinning the no-ping-pong property) and reorganizes straight to round 5 in one bump.
- The pre-existing `reject_of_older_round_carries_rejectors_round` is unchanged in what it asserts; its setup moved into shared `test_task`/`test_request`/`sent_posit` helpers (dropping a `Participants` object it built but never used).

`cargo test -p mpc-node --lib` green, clippy and fmt clean.

## Ordering

Land **after #1100** (carry posit round across respawns). The reject plants a high-round entry in the peer's one-slot mailbox for us; without #1100 a respawn resets us to round 0 underneath it, and that entry silently swallows our fresh messages until we climb back past it. #1100 makes rounds monotonic per sender, closing that window. (Independent of #1082: the reject uses the existing wire shape.)

### DIFF
--- a/chain-signatures/node/src/protocol/request/posit.rs
+++ b/chain-signatures/node/src/protocol/request/posit.rs
@@ -687,4 +687,45 @@ mod tests {
         assert_eq!(stale_round, Some(5));
     }
 
+    /// The receive side inside `advance()`: a behind proposer harvests the
+    /// round carried by a StaleRound reject without answering it (even one
+    /// echoing an older round), then reorganizes straight to that round on
+    /// its deadline.
+    #[tokio::test]
+    async fn advance_harvests_rejectors_round() {
+        let proposer = Participant::from(0);
+        let rejector = Participant::from(1);
+        // Threshold 2 keeps the proposer's own accept from starting generation.
+        let mut t = setup(proposer, rejector, 2);
+
+        // We propose at round 3; a reject echoing our earlier round 2 arrives,
+        // carrying the rejector's round 5. Short budget so the loop exits via
+        // its deadline.
+        t.state.set_round(3);
+        t.state.budget.reset(Duration::from_millis(200));
+
+        let mailbox = PositMailbox::new();
+        mailbox.push(SignPositMessage {
+            presignature_id: 42,
+            round: 2,
+            from: rejector,
+            action: PositAction::RejectWithReason(PositRejectReason::StaleRound),
+            stale_round: Some(5),
+        });
+
+        let mut phase = PositPhase {
+            proposer,
+            active: [proposer, rejector].into_iter().collect(),
+            presignature_id: 42,
+            presignature: None,
+        };
+        let next = phase.advance(&mut t.ctx, &mut t.state, &mailbox).await;
+        assert!(matches!(next, SignPhase::Organizing(_)));
+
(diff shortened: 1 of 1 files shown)


chain-signatures/node/src/protocol/request/posit.rs
```diff
@@ -491,23 +535,90 @@ mod tests {
...
+        mailbox.push(SignPositMessage {
+            presignature_id: 42,
+            round: 5,
+            from: proposer,
+            action: PositAction::RejectWithReason(PositRejectReason::StaleRound),
+        });
+
+        // The reject is a future-round message: buffered, never answered. The
+        // timeout then reorganizes straight to the harvested round 5.
+        let phase = PositPhase::wait_for_propose(&mut ctx, &mut state, &mailbox, proposer).await;
```
ppca: not correctness issue, but this test should already hold for pre-PR because wait_for_propose already harvest the peer round before this PR?
I think we should probably add a test to cover harvesting peer round in advance?

yvonneanne: Correct, this test goes through `wait_for_propose` and would pass pre-PR; it pins the no-ping-pong property of the new send site rather than new receive code. Added `advance_harvests_rejectors_round` in 480e646c to cover the receive side in `advance()`: a behind proposer at round 2 receives a reject stamped 5, buffers it without answering, and reorganizes straight to round 5 on its deadline.

jakmeier (approved):
LGTM, thanks for finding and closing this gap

yvonneanne:
Rebased over #1100 and #1082. The advance() reject site now follows #1082's convention: the id echoes the rejected round and our round rides in `stale_round`. One consequence needed a small piece of new code: under the echo convention a reject can arrive as an older-round message on a peer that has already moved on, and advance() would have answered it with another reject, ping-ponging. advance() now never answers rejects (mirroring wait_for_propose's guard), while still harvesting `stale_round` first. Test changes: `behind_peer_harvests_rejectors_round` is dropped as redundant with #1082's `received_stale_round_reject_catches_up_in_one_bump`, and `advance_harvests_rejectors_round` now also pins the no-answer guard (a reject echoing an older round is harvested but never answered). 168 lib tests, clippy and fmt clean.

PR #1106, opened 2026-08-05
refactor(ethereum): introduce unified finalized-head-gated mechanism for catchup and live phases

isSerge:
Long-term fix for #1094 I previously mentioned in #1101 (fixed the live phase)

This PR makes both catchup and live phases use the same finalized-head-gated mechanism ensuring finality for all blocks.

Main changes:

- Indexer: 
  - Add new methods: 
    - `sample_anchor` (returns tip + 1)
    - `wait_processable_bound` (waits till ceiling is ready, used to gate processable range) 
    - `process_range` (batch fetch and process)
  - Update `run` method: both catchup and live loops now use the same primitives (`wait_processable_bound` + `process_range`). Difference between the two is trivial: catchup is bounded, live is uncapped. `ChainEvent::CatchupCompleted` emission is preserved. 
  - Transition to a single mechanism for both phases enabled a significant cleanup:
    - reorg hash-check (dead code in production)
    - per-block finality wait - `wait_processable_bound` guarantees all blocks are finalized
    - internal `mpsc` channel to track tip of the chain
    - `process_finalized_block` and `emit_processed_block` methods
    - `block_hash` field in `BlockAndRequests`
    - `live_block_buffer` field in `IndexerConfig`
  - Add more tracing to log how far behind it is at startup and when it transitions to the live tail
  
- `FinalizedHeadTracker` now encapsulates optimistic mode: watcher polls `eth_getBlockByNumber(Latest)` for optimistic mode and `eth_getBlockByNumber(Finalized)` in production. Indexer becomes mode-agnostic.
- Minor readability changes:
  - `parse_block`: replace `partition`+`filter`+`is_some_and` chain with two `topic0()` filters
  - `process_catchup_item`: replace `_block` and `_logs` lifetime workaround
- New shared helper and new unit tests in `chain-integration-core/src/utils/task.rs`:
  -  `retry_until_some` - similar to existing `retry_until_ok`, but returns `Option`, used for `sample_anchor` method
  - 8 unit tests covering: `retry_until_ok`, `retry_until_some`, `cancelled_within`, `AbortOnDrop`

### DIFF
--- a/chain-signatures/chain-integration-core/src/utils/task.rs
+++ b/chain-signatures/chain-integration-core/src/utils/task.rs
@@ -89,16 +89,6 @@ mod tests {
     };
 
     #[tokio::test]
-    async fn retry_until_some_returns_value_when_ready() {
-        let cancel = CancellationToken::new();
-        let val = retry_until_some(&cancel, Duration::from_millis(1), "test", || async {
-            Ok::<_, anyhow::Error>(Some(42u64))
-        })
-        .await;
-        assert_eq!(val, Some(42));
-    }
-
-    #[tokio::test]
     async fn retry_until_some_retries_on_none_then_yields_value() {
         let cancel = CancellationToken::new();
         let count = Arc::new(AtomicU32::new(0));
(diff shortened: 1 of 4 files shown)


chain-signatures/chain-integration-core/src/utils/task.rs
```diff
@@ -53,3 +53,188 @@ pub async fn retry_until_ok<F, Fut>(
...
+pub async fn retry_until_some<T, F, Fut>(
+    cancel: &CancellationToken,
+    delay: Duration,
+    label: &str,
+    poll: F,
+) -> Option<T>
+where
+    F: Fn() -> Fut,
+    Fut: Future<Output = anyhow::Result<Option<T>>>,
+{
```
jakmeier: Would it make sense to delegate from `retry_until_ok` to this new function to avoid some (minor) code duplication?

 ```rust                                                                                                                                                  
   pub async fn retry_until_ok<F, Fut>(                                                                                                                   
       cancel: &CancellationToken,                                                                                                                        
       delay: Duration,                                                                                                                                   
       label: &str,                                                                                                                                       
       process: F,                                                                                                                                        
   ) where                                                                                                                                                
       F: Fn() -> Fut,                                                                                                                                    
       Fut: Future<Output = anyhow::Result<()>>,                                                                                                          
   {                                                                                                                                                      
       retry_until_some(cancel, delay, label, || async { process().await.map(|_| Some(())) })                                                             
           .await;                                                                                                                                        
   }                                                                                                                                                      
 ```

Feel free to keep it as is though, this is a subjective style question.

jakmeier (approved):
Nice refactor, unifying these two paths looks much cleaner to me.

PR #1107, opened 2026-08-06
feat: add versioning to /checkpoint

ChaoticTempest:
we need to ease the pain of changing the behavior of checkpoints between nodes, so with this I'm introducing a new `version` field to the `CheckpointResponse` for `/checkpoint`. This is similar to how we have a version field for protocol messages via `/msg` endpoint. 

Changes
-  /checkpoint  now returns a versioned envelope  ` CheckpointResponse { version, checkpoints }`
-  Mismatches in checkpoint versions skips the checkpoint entirely for that peer. This makes it so that nodes will loop until they upgrade to the newest version or if they're able to communicate with one of the other old nodes.


chain-signatures/node/src/web/mod.rs
```diff
@@ -241,6 +241,13 @@ async fn state(Extension(web): Extension<Arc<AxumState>>) -> Result<Json<StateVi
     result
 }
 
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct CheckpointResponse {
+    #[serde(default)]
+    pub version: u64,
```
jakmeier, line 247: Is this intended to be backwards-compatible as version 0 with the current format? I don't think it is, since previously we sent `HashMap` and the new code expects `{ checkpoints: HashMap }`.

jakmeier (approved):
Yeah, versioned checkpoints make a lot of sense.

I think  you chose the right approach by adding a version field and only allowing an exact match. Let's keep it simple for now.

Long-term, we should aim to allow gradual node updates. I like the pattern where the caller requests a specific version, e.g. specified in the path (`/checkpoint/v13`). A node can then support the latest and one older version for outdated peers. And when requesting the latest node from an outdated peer, they respond with an HTTP 404 and the requester tries an older version, if that's compatible.
(Just a thought, no action needed right now.)

chain-signatures/node/src/web/mod.rs
ChaoticTempest: yeah that's true, alright I set the version to `0`, since the old format just errors out with conversion error anyways. So it's not going to be backwards compatible with the previous /checkpoint endpoint which is fine since prior to changes on this branch, we're handling errors fine on /checkpoint by skipping the peer with errors

PR #1113, opened 2026-08-10
feat(chain-midnight): wire Rust response publisher

Pessina:
Completes the Rust Midnight response publisher on top of #1143.

- Builds and submits responder intents through the TypeScript child process.
- Pins finalized request state and keeps publication idempotent.
- Returns ambiguous submits as retryable errors; the retry's pinned finalized read settles a landed post without reposting.
- Wires publisher configuration through the node CLI and Docker runtime.

Stack: #1129 → #1142 → #1143 → #1113 → #1134


chain-signatures/chain-midnight/src/rpc.rs
```diff
@@ -95,7 +103,63 @@ fn hex_0x(hash: H256) -> String {
...
+async fn connect_bounded(
+    config: &MidnightConfig,
+) -> anyhow::Result<(OnlineClient<SubstrateConfig>, RpcClient, Liveness)> {
+    let connect_timeout = config.rpc.connect_timeout;
+    let url = config.node_ws_url.as_str();
+
+    // `RpcClient::from_url` with the one knob it does not expose, the response
+    // cap (`RpcConfig::max_response_size`): the same transport, client and
+    // subscription buffer, through subxt's own jsonrpsee re-export.
+    let ws_client = tokio::time::timeout(connect_timeout, async {
```
ppca: recently mpc team is planning to move solana from websocket push to pulling latest blocks, reason being websocket has a few complications like reconnect that has been error prone. Ethereum was moved to pulling a while ago and has been much more stable.
Is it worth changing this to not use websocket too? @isSerge

chain-signatures/chain-midnight/src/publisher.rs
```diff
@@ -1,15 +1,1051 @@
...
+            ?sign_id,
+            circuit = call.circuit.wire_name(),
+            central = %self.central_address,
+            request_id = %hex::encode(call.request_id),
+            elapsed = ?action.timestamp.elapsed(),
+            "midnight: publishing signature"
+        );
+
+        // One head for every read below it.
+        let at_hash = self.reads.finalized_head().await?;
```
ppca, line 176: so if the previous publish attempt is not yet finalized here, we would publish again. since we allow duplicate publish, should be good i guess.

chain-signatures/chain-midnight/src/publisher.rs
```diff
@@ -1,15 +1,1051 @@
...
+                if stored_request_id == call.request_id
+                    && stored_signature == call.stored_signature =>
+            {
+                return Ok(true);
+            }
+            Ok(_) => {}
+            Err(err) => tracing::warn!(
+                circuit = call.circuit.wire_name(),
+                request_id = %hex::encode(call.request_id),
+                "midnight finalized response entry ignored: {err:#}"
```
ppca: if `decode_response_entry` errs out, that means schema drift, should we return an error for `response_present` instead of ignoring and continue?

Pessina: We cant err here because the contract `respond` circuit is permissionless, so an attacker could halt the publisher with an invalid signature

Its safe to skip because an entry is only deemed published when the `requestId + signature` match what the node holds, right? That's my understanding of the MPC policy

Schema drift still checked on: 

- decode_contract_state
- central_map

chain-signatures/chain-midnight/src/publisher.rs
Pessina: Yeah, that's the design, duplicate publishes are safe

I was trying to avoid it but the err categorization + retry policy was getting very complicated and also increasing the time to try again due to Midnight tx TTL

Following this rule, I will also be able to simplify the TS/Rust wire on a later PR

chain-signatures/chain-midnight/src/publisher.rs
ppca: ah I see. sounds good!

Pessina: Actually, you were right. 

We now err on decoding failures and skip invalid signature: https://github.com/sig-net/mpc/pull/1113/commits/999ce2ca44553b3e94fd7c3936db03863a63bd11

chain-signatures/chain-midnight/src/rpc.rs
isSerge: probably worth changing - WebSocket makes sense if we really care about sub-second latency, otherwise HTTP is easier and more stable, plus normally we can re-use same primitives we use for catchup

chain-signatures/chain-midnight/src/publisher.rs
```diff
@@ -1,15 +1,1065 @@
...
+        Ok(())
+    }
+}
 
 #[async_trait]
 impl ChainPublisher for MidnightPublisher {
-    async fn publish_signature(&self, _action: &PublishAction) -> anyhow::Result<()> {
-        anyhow::bail!("midnight publisher not implemented")
+    async fn publish_signature(&self, action: &PublishAction) -> anyhow::Result<()> {
+        self.publish_inner(action).await.inspect_err(|e| {
```
isSerge: Can we just implement everything in `ChainPublisher::publish_signature` without `publish_inner`?

chain-signatures/chain-midnight/src/publisher.rs
isSerge, line 1: Does Midnight support batch publishing?

Pessina: Yes, each Midnight tx can carry multiple sub-tx, but I would defer it as it will add complexity. 

Other option it's change the contract to support an array of Signature

What do you think?

isSerge: Let's just leave a TODO comment for now, we'll consider options when this becomes a bottleneck

Pessina: Added: https://github.com/sig-net/mpc/pull/1113/commits/8e784cd2dc600da5ec1e727e850252646169b9dd

chain-signatures/chain-midnight/src/rpc.rs
ppca: @Pessina do you mind changing this to not use websocket then? This can help avoid a lot of bugs with indexer in the future. 🙏

Pessina: Sure, I'm on it. Will ping for review when I'm done :)

Pessina: Done here: https://github.com/sig-net/mpc/pull/1113/commits/bafc52ebd199bc45ee0578062763a1122ec5c622

isSerge (approved):

PR #1118, opened 2026-08-11
fix(solana): replay catchup slots in order

### LINKED ISSUE #1119: solana indexer: catchup will miss respond events for a sign request if they are split across pages
currently during catchup, we fetch pages starting from anchor all the way back to checkpoint.
Within each page, we fetch blocks in ascending order, but pages are ordered from new to old. This means the final blocks are not in right order.

Example: 
older page = [8,7,6]
newer page = [10,9]
request in block 7, respond in block 9. 
but with current logic, the blocks are processed in this order: [9,10,6,7,8]. The sign request is processed after respond event, so this request will never receive its respond event.

Also since each node is restarted at a slightly different time, the anchor height can be a bit different across nodes, thus the pages fetched, especially page boundaries can be different. This would mean what requests are pending at each height during catchup will be different, and then we fail to reach consensus on checkpoints, checkpoints accumulate, Solana indexer stall.

ppca:
## Summary

  Fix Solana catch-up replay ordering across paginated signature results.

  Solana returns signature pages newest-first, while slots were only sorted within each page. This could replay a response
  before its request and leave nodes with divergent pending-request sets.

  The fix globally sorts and deduplicates discovered slots before fetching blocks in bounded, sequential chunks. It also adds
  pagination observability and deterministic multi-page ordering tests.

fix https://github.com/sig-net/mpc/issues/1119

### DIFF
--- a/chain-signatures/chain-solana/src/indexer.rs
+++ b/chain-signatures/chain-solana/src/indexer.rs
@@ -171,18 +171,15 @@ impl<S: StateManager, T: ChainTelemetry> SolanaIndexer<S, T> {
     fn chunk_slots(slots: BTreeSet<u64>, chunk_size: usize) -> Vec<BTreeSet<u64>> {
         debug_assert!(chunk_size > 0, "catchup slot chunk size must be positive");
 
-        let mut chunks = Vec::new();
-        let mut chunk = BTreeSet::new();
-        for slot in slots {
-            chunk.insert(slot);
-            if chunk.len() == chunk_size {
-                chunks.push(std::mem::take(&mut chunk));
+        slots.into_iter().fold(Vec::new(), |mut chunks, slot| {
+            match chunks.last_mut() {
+                Some(chunk) if chunk.len() < chunk_size => {
+                    chunk.insert(slot);
+                }
+                _ => chunks.push(BTreeSet::from([slot])),
             }
-        }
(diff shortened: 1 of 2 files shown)


chain-signatures/chain-solana/src/indexer.rs
```diff
@@ -109,9 +109,80 @@ impl<S: StateManager, T: ChainTelemetry> SolanaIndexer<S, T> {
...
+        for slot in slots {
+            chunk.insert(slot);
+            if chunk.len() == chunk_size {
+                chunks.push(std::mem::take(&mut chunk));
+            }
+        }
+        if !chunk.is_empty() {
+            chunks.push(chunk);
+        }
+        chunks
```
isSerge: Would you consider a more functional way?

```suggestion
        slots.into_iter().fold(Vec::new(), |mut chunks, slot| {
            match chunks.last_mut() {
                Some(chunk) if chunk.len() < chunk_size => {
                    chunk.insert(slot);
                }
                _ => chunks.push(BTreeSet::from([slot])),
            }
            chunks
        })
```

isSerge (approved):

chain-signatures/chain-solana/src/indexer.rs
ppca: done!

PR #1122, opened 2026-08-11
governance contract: move candidates out of state

ppca:
fix https://github.com/sig-net/mpc-private/issues/33.

What's included:
- Move candidate info out of STATE into a IterableMap: avoids gas costs for other writes + removes it from state() view calls.
- Create a new method to read candidate info by account id: so that node can use it


chain-signatures/contract/src/lib.rs
```diff
@@ -85,6 +87,13 @@ pub struct MpcContract {
     config: Config,
     latest_checkpoints: IterableMap<Chain, ConsensusCheckpointDigest>,
     checkpoint_votes: CheckpointVotes,
+    /// Candidate registry (open-`join()` pool + genesis founding set), each
+    /// entry folding in the participants that voted to admit it. Stored as an
+    /// `IterableMap` so its entries live under their own storage keys rather
+    /// than inline in the `STATE` blob — `sign`/`vote_checkpoint` and every
+    /// other `&mut` call would otherwise re-serialize the whole (unbounded) set
+    /// on each invocation.
```
jakmeier: nit: I don't think we need such a long comment here

chain-signatures/contract/src/lib.rs
```diff
@@ -373,25 +391,16 @@ impl VersionedMpcContract {
     #[handle_result]
     pub fn remove_candidacy(&mut self) -> Result<(), Error> {
         log!("remove_candidacy: signer={}", env::signer_account_id());
-        let protocol_state = self.mutable_state();
+        let signer_account_id = env::signer_account_id();
```
jakmeier: pre-existing, but we really shouldn't use the signer id, we should use the predecessor.

Partners may want to use some kind of multi-sig contract to control the account that is added as participant. Or they might use a relayer to submit their transactions, maybe to pay gas costs in a different currency than NEAR.

In such cases, the original transaction signer is an irrelevant party and the predecessor is the real account we care about.

chain-signatures/contract/src/state.rs
```diff
@@ -65,3 +68,58 @@ impl ProtocolContractState {
...
 }
+
+/// A `Running` state as exposed to external callers: only the fields a node
+/// actually consumes (`epoch`, `participants`, `threshold`, `public_key`). The
+/// stored `RunningContractState` additionally holds the vote maps
+/// (`candidates`, `join_votes`, `leave_votes`, `threshold_votes`); those are
+/// omitted here — the attacker-writable `candidates`/`join_votes` could inflate
+/// this payload past the RPC view-execution limit, and no node reads any vote
+/// map from the polled state. A joining node reads its own candidacy through
+/// the `candidate_info` view instead.
```
jakmeier: I think this comment is quite specific to this PR, I would largely remove it.

chain-signatures/contract/src/state.rs
```diff
@@ -6,21 +6,24 @@ use near_sdk::{AccountId, PublicKey};
 
 use crate::primitives::{Candidates, Participants, PkVotes, ThresholdVotes, Votes};
 
+/// Stored `Initializing` state. Candidates are held in the top-level
+/// `MpcContract::candidates` map, not here (see [`InitializingContractStateView`]
+/// for what external callers receive).
 #[derive(BorshDeserialize, BorshSerialize, Serialize, Deserialize, Debug, Clone)]
 pub struct InitializingContractState {
-    pub candidates: Candidates,
```
jakmeier, line 11: Hm, I don't exactly see why this PR moves `candidates` out of the enum variant. The IterableMap should be possible to still live in here.

This leads to quite some refactoring and conceptual changes to the contract. Especially on this critical code, I really value keeping the changes as small as possible.

@ChaoticTempest Maybe you have an opinion on how data should be represented in the contract? I think you worked the most on this.

jakmeier (commented):
This is an important change and I think this PR fixes the underlying problem well.

However, I would like to see comments cleaned up a bit. And I would take another look if the change could be smaller by keeping the candidates field where it was, just as an `IterableMap`.

chain-signatures/contract/src/state.rs
ppca: yeah agree in the contract the candidates field should stay where it was and just change type, that would already achieve purpose of moving candidates out of STATE key. And also keep join_votes where it was.

jakmeier (approved):
Thanks for the refactoring @ppca , I like this new version better. It seems cleaner.

chain-signatures/contract/src/migration.rs
```diff
@@ -53,6 +85,71 @@ fn upgrade_protocol_state(old: OldProtocolContractState) -> ProtocolContractStat
...
+                leave_votes: state.leave_votes,
+                threshold_votes: state.threshold_votes,
+            })
+        }
+        InlineProtocolContractState::Resharing(state) => ProtocolContractState::Resharing(state),
+    }
+}
+
+#[derive(BorshDeserialize)]
+pub(crate) struct PreviousInline {
```
ChaoticTempest: why do we have this and what is it trying to represent exactly? We shouldn't have anything else besides previous `devnet`, `testnet` and `mainnet` variants here. The purpose of this migration is to only adjust each of these as running versions of each network. So adding this variant here would probably override the test entirely for each of the networks which breaks our migration for each

ppca: removed unnecessary new formats.

chain-signatures/contract/src/migration.rs
```diff
@@ -53,9 +85,51 @@ fn upgrade_protocol_state(old: OldProtocolContractState) -> ProtocolContractStat
     }
 }
 
+#[derive(BorshDeserialize)]
+struct DevnetRunningContractState {
```
ChaoticTempest: wait, we don't need any of these new migraiton code. There's nothing you changed about the serialization format for the smart contract at all, so let's just revert all these changes

integration-tests/src/commands.rs
```diff
@@ -133,15 +133,17 @@ pub fn init_running_command(
 }
 
 pub fn dummy_candidates() -> Candidates {
-    let mut candidates = Candidates::new();
+    let mut candidates = Candidates {
+        candidates: Default::default(),
+    };
```
ChaoticTempest, line 138: nit: `Candidates::default` should be implemented in such a case.

ChaoticTempest (approved):

PR #1127, opened 2026-08-12
feat: durable pending checkpoints

### LINKED ISSUE #1008: Ensure that checkpoints are durable
Currently, pending or unconfirmed checkpoints exist only in memory but redis persistence happens only after the contract watcher observes confirmation. If all nodes that have this pending checkpoint restart after on-chain publish but before that observation on the node, the contract retains only the digest and catchup can loop forever looking for the checkpoint none of the peer would have.

This could especially be the case in our devnet where we restart all nodes at once. Or the unlikely chance that all node providers upgrade at once.

We should either be persisting pending checkpoints locally and then promote to confirmed (or latest), or we need to have the ability to walk up to that same pending checkpoint again and then save it as latest. Both are complex in their own way

ChaoticTempest:
implements and fixes https://github.com/sig-net/mpc/issues/1008.

All pending checkpoints are now stored. Most of the behavior is the same here. The reason to have these pending checkpoints be stored is because a node can create and vote for checkpoint H, restart before consensus confirms it, then lose the exact checkpoint body needed to. So this change with durable pending checkpoints preserve that in-flight pending checkpoint state across restarts.

Changes:
- refactored and move all checkpoint related logic to `backog/checkpoints.rs`.
- new `Checkpoints` struct that manages all checkpoints (both pending and consensus/confirmed/latest).


chain-signatures/node/src/storage/checkpoint_storage.rs
```diff
@@ -58,7 +72,164 @@ impl CheckpointStorage {
...
+                            redis.call('HDEL', KEYS[2], entries[i])
+                        end
+                    end
+                    return 1
+                "#;
+                let promoted: i32 = redis::Script::new(PROMOTE)
+                    .key(self.checkpoint_key(chain))
+                    .key(self.pending_checkpoint_key(chain))
+                    .arg(height)
+                    .arg(height)
```
isSerge, line 276: Second `arg` is probably redundant

chain-signatures/node/src/backlog/mod.rs
```diff
@@ -644,75 +595,49 @@ impl Backlog {
...
+        self.checkpoints.load_local(chain).await
+    }
 
-        tracing::info!(
-            ?chain,
-            height = checkpoint.block_height,
-            "consensus checkpoint confirmed and persisted"
-        );
+    /// Replace the local backlog with a consensus checkpoint after divergence.
+    pub(crate) async fn regress(&self, checkpoint: Checkpoint) -> anyhow::Result<()> {
```
isSerge: no need for `pub(crate)` here

chain-signatures/node/src/backlog/checkpoints.rs
```diff
@@ -0,0 +1,506 @@
...
+use mpc_primitives::{Chain, Checkpoint, PendingTx};
+use sha3::Digest;
+use std::collections::{BTreeMap, HashMap};
+use std::sync::Arc;
+use tokio::sync::RwLock;
+
+#[derive(Debug, Clone)]
+pub(super) struct Checkpoints {
+    storage: CheckpointStorage,
+    pending: Arc<HashMap<Chain, RwLock<BTreeMap<u64, Checkpoint>>>>,
```
isSerge: This mirrors what's already stored in `CheckpointStorage` and about a third of `Checkpoints` public methods exist because of this mirror. Perhaps there is a way we can reduce duplication and public surface

chain-signatures/node/src/backlog/mod.rs
```diff
@@ -644,75 +595,49 @@ impl Backlog {
...
+        self.checkpoints.regress(&checkpoint).await?;
+        self.recover_by_checkpoint(checkpoint).await
```
yvonneanne: These two lines can lead to a inconsistent local state, afaics. The first commits durably (SET latest, DEL all pending); the second can fail, because from_checkpoint decodes the peer's CBOR. If it fails, storage says we regressed and memory says we did not.

And I don't think we retry: the next detect_regression matches the stored latest against the consensus digest, so confirm returns true and we look aligned. Maybe we should decode first and write durably only once that succeeds?

chain-signatures/node/src/backlog/checkpoints.rs
```diff
@@ -0,0 +1,506 @@
...
+                    ?digest,
+                    "pending checkpoint disappeared before promotion"
+                );
+                return false;
+            }
+            Err(err) => {
+                tracing::warn!(?chain, %err, "failed to promote consensus checkpoint");
+                return false;
+            }
+        }
```
yvonneanne, line 183: A storage failure here returns false, which is the same value confirm returns when we genuinely do not have the checkpoint, and detect_regression turns that into a full regression (cancel run(), AbortChain, refetch, roll back). Should those two cases be distinguishable, so a transient Redis error retries with backoff instead of rolling the backlog back?

chain-signatures/node/src/backlog/checkpoints.rs
```diff
@@ -0,0 +1,506 @@
...
+        let latest_height = latest.as_ref().map(|checkpoint| checkpoint.block_height);
+        let pending = self
+            .storage
+            .load_pending(chain)
+            .await?
+            .into_iter()
+            .filter(|checkpoint| {
+                latest_height.is_none_or(|height| checkpoint.block_height > height)
+            })
+            .collect::<Vec<_>>();
```
yvonneanne: Question about restart behaviour here. Iiuc, this rehydrates the whole pending set, including when it was at MAX_PENDING_CHECKPOINTS. So a process restart no longer clears it the way it did when the map was in-memory only, and the node may come back still refusing events with the watchdog restarting run() every 5 min?

Is persisting such a stall across restarts the intent? In this case deleting the Redis key is the only way out, isn't it? If it is intended, would an alert make sense? If not, would hydrating only the newest items still give #1008 what it needs?

chain-signatures/node/src/backlog/checkpoints.rs
```diff
@@ -0,0 +1,506 @@
...
+        {
+            return Some(CheckpointKind::Pending(checkpoint));
+        }
+        self.storage
+            .load_latest(chain)
+            .await
+            .ok()
+            .flatten()
+            .filter(|checkpoint| checkpoint.digest() == digest)
+            .map(CheckpointKind::Latest)
```
yvonneanne: Should there be a third lookup here, falling back to storage.load_pending? As written this searches the in-memory mirror and then load_latest, and the mirror is only filled by load_local, which runs after mesh::wait_threshold_active. Since the web server is spawned independently of the supervisors, I think that means after an all-node restart a node answers "not found" on /checkpoint for the pending digest while the body is in its Redis, which is the case #1008 is about. Am I missing a path that hydrates earlier?

chain-signatures/node/src/backlog/checkpoints.rs
```diff
@@ -0,0 +1,506 @@
...
+            local.extend(
+                pending
+                    .iter()
+                    .cloned()
+                    .map(|checkpoint| (checkpoint.block_height, checkpoint)),
+            );
+            local.len()
+        };
+        self.observe(chain, len);
+        Ok(pending.into_iter().next_back().or(latest))
```
yvonneanne, line 240: Iiuc, returning the newest pending checkpoint means restart resumes there rather than at the confirmed one, so we no longer re-derive the blocks below it. It looks safe to me, since align_backlog_with_consensus runs straight after and would regress us on a digest mismatch, and it restores the state we had before the crash. Was skipping the re-derivation the intent? Then maybe it's worth a comment?

chain-signatures/node/src/backlog/checkpoints.rs
ChaoticTempest: Made it distinguishable by returning `Result<bool, CheckpointError>`

chain-signatures/node/src/backlog/mod.rs
ChaoticTempest: good point, I moved the `PendingRequests::from_checkpoint` call to be before the two calls

chain-signatures/node/src/backlog/checkpoints.rs
ChaoticTempest: not sure I follow with the public method mirroring, because the only thing that we mirror is `Backlog` methods with `Checkpoints` methods. `Checkpoints` don't really mirror any methods form `CheckpointStorage`. The only mirroring is that we have an in-memory cache vs a persisted pending checkpoint

chain-signatures/node/src/backlog/checkpoints.rs
ChaoticTempest: The idea is yes, we should still be stalling across restarts. It's not ideal, but there's not really anyway to resolve a stalled pending set like this unless we reach agreement either from our current pending set or the other threshold nodes reached a separate agreement. In such cases, we do want to be stalled because we need to figure out what's going wrong and not waste resources going forward in the timeline.

We already have `crate::metrics::requests::PENDING_CHECKPOINTS` which is already being sent to grafana, so we can already add alerts with that. cc @auto-mausx

chain-signatures/node/src/backlog/checkpoints.rs
ChaoticTempest: good eye, yep that's the intent, will leave a comment

chain-signatures/node/src/backlog/checkpoints.rs
ChaoticTempest: good catch again, in addition to the fallback you mention, we need to move the load_local to be earlier before wait_threshold since loading from local doesn't depend on it

chain-signatures/node/src/backlog/checkpoints.rs
isSerge: yeah, mirroring was an overstatement, never mind

chain-signatures/node/src/backlog/checkpoints.rs
```diff
@@ -0,0 +1,631 @@
...
+        match self
+            .storage
+            .promote_pending(chain, checkpoint.block_height)
+            .await
+        {
+            Ok(true) => {}
+            Ok(false) => {
+                tracing::warn!(
+                    ?chain,
+                    ?digest,
```
ppca, line 174: nit: can be helpful for debugging to add the block height here too

chain-signatures/node/src/backlog/checkpoints.rs
```diff
@@ -0,0 +1,631 @@
...
+            }
+        }
+
+        // The in-memory mirror is only hydrated by `load_local` once the mesh is
+        // active, but the web server serves `/checkpoint` independently. Look up
+        // durable pending storage directly so a body persisted before restart can
+        // still be handed out to peers, applying the same "above confirmed height"
+        // filter `load_local` uses so stale entries are ignored.
+        let latest_height = latest.as_ref().map(|checkpoint| checkpoint.block_height);
+        match self.storage.find_pending(chain, digest).await {
```
ppca, line 318: I'm wondering is it actually cleaner and less error prone to have checkpoints only exist in redis, and do not keep them in memory at all? Like triple and presigs do.
I understand that having them in memory is faster, but we'd need to make sure memory and storage are consistent in every operation, and is more error prone.

chain-signatures/node/src/backlog/checkpoints.rs
```diff
@@ -0,0 +1,631 @@
...
+    /// checkpoint (restoring the pre-crash in-flight state and skipping
+    /// re-derivation of lower blocks), or the confirmed latest when no pending
+    /// checkpoint remains.
+    pub(super) async fn load_local(&self, chain: Chain) -> anyhow::Result<Option<Checkpoint>> {
+        let latest = self.storage.load_latest(chain).await?;
+        let latest_height = latest.as_ref().map(|checkpoint| checkpoint.block_height);
+        let pending = self
+            .storage
+            .load_pending(chain)
+            .await?
```
ppca: so if we err out here we'd have returned without returning latest. Should we return latest if this errs?

chain-signatures/node/src/backlog/checkpoints.rs
ChaoticTempest: I deferred doing this for later, but it's not as bad as something as triples/presignatures to keep in two places. Checkpoints are immutable, and here in this PR, we only add them in-memory when persistence succeeds. So we can guarantee that they'll at least exist in storage.

chain-signatures/node/src/backlog/checkpoints.rs
ChaoticTempest: good point, returned latest on error

ppca (approved):

chain-signatures/node/src/backlog/checkpoints.rs
```diff
@@ -0,0 +1,635 @@
...
+        }
+    }
+
+    /// Returns the pending-checkpoint map for `chain`.
+    fn pending(&self, chain: Chain) -> &RwLock<BTreeMap<u64, Checkpoint>> {
+        &self.pending[chain]
+    }
+
+    /// Updates the pending-checkpoint metric for `chain`.
+    fn observe(&self, chain: Chain, len: usize) {
```
volovyks, line 57: Why do we pass len if we have `pending` inside of the `struct Checkpoints` ?

chain-signatures/node/src/backlog/checkpoints.rs
```diff
@@ -0,0 +1,635 @@
...
+        crate::metrics::requests::PENDING_CHECKPOINTS
+            .with_label_values(&[chain.as_str()])
+            .set(len as i64);
+    }
+
+    /// Durably records an unconfirmed checkpoint and adds it to the local cache.
+    pub(super) async fn persist_pending(
```
volovyks, line 64: Redis is an in-memory DB with semi-instant persistence mode. I would avoid having the same info in memory and stick to a single source of truth. We can have an in-memory implementation for testing.

volovyks (commented):
The related issue had two possible paths to fix it. We should always prefer recovering from on-chain history rather than from local state. But persistence may be simpler and not harmful.

Not related to this PR, but we need to establish a design-first framework for feature design and decision-making.

volovyks:
PS. Not blocking the merge, but it would be great to address my comments and design suggestions here or in the follow up.

yapignolet:
Two small things on the logging while this PR is moving the affecte lines anyway.

tracing::info!(?chain, height, tx_count, ?checkpoint, "creating checkpoint") serializes the entire pending set into the log line. On dev that's 41k entries right now, so several MB per checkpoint, per chain, per node. tx_count is already there, so dropping ?checkpoint loses nothing, afaics.

Same in stream/ops.rs ("created checkpoint"), which logs ?checkpoint next to ?checkpoint_digest, so that one is covered too.

chain-signatures/node/src/backlog/checkpoints.rs
ChaoticTempest: it's to avoid lock re-acquisition

chain-signatures/node/src/backlog/checkpoints.rs
ChaoticTempest: yeah, I mentioned it already here, but the point is to keep checkpoints fast for now, because nodes can fetch other nodes' checkpoints and end up stalling the redis due to large payload being read. Redis is largely single threaded, so a 100mb checkpoint payload would stall the whole entire thing for a while

ChaoticTempest:
@yapignolet so, logging `?checkpoint` actually gives us a list of `[PendingTx { sign_id }]`, where the `sign_id` is pretty useful to debug whether our sign request was saved somewhere or not. I think having it take a buffer of 1-10mb for just that log is fine since we'd like trace a sign request end-to-end. But we can evaluate it further in a separate PR. We can optimize that log a bit by not including `PendingTx { }` bytes for each of the sign request and instead just have the `sign_id`

ChaoticTempest:
> The related issue had two possible paths to fix it. We should always prefer recovering from on-chain history rather than from local state. But persistence may be simpler and not harmful.

@volovyks so recovering from on-chain history is actually much more complex than you would think. I already tried, it requires covering a lot of edge cases which just makes it horribly complex

ppca:
> The related issue had two possible paths to fix it. We should always prefer recovering from on-chain history rather than from local state. But persistence may be simpler and not harmful.
> 
> Not related to this PR, but we need to establish a design-first framework for feature design and decision-making.

I think this PR is more about making sure peers can materialize the latest checkpoint in contract given that now we query peers for such checkpoint, contract only has digest. 

If we want to recover solely from on-chain history, and if that means not relying on what nodes have locally,  I guess the following data needs to be in the contract's checkpoint.
- start_block: the lowest block height of all pending requests' block where they are mined on sourced chain. A respond() can only happen on or after that block for any requests. This should be consensus on as it only anchors on source chain height and should be able to achieve consensus among nodes.

When we restore a checkpoint for chain A . For chain A, scan all blocks in [start_block, checkpoint_chain_A's height]. While scanning, we'd find all sign(), and we'd find all the respond(),  and with the signature in respond(), we can reconstruct the tx hash of bidirectional request's target tx.  We will remove a request if respond() has shown up in the block range. For all the target chains involved in the bidirectional requests, with the late-watcher refill mechanism implemented, we can just add these to watcher and the target chain indexer will be able to check whether the target chain tx has been executed and move on from there.

To detect regression, we compare the contract's latest checkpoint digest with a node's local checkpoint, if regression is detected, we use on-chain history as above to recover. Then we don't need to worry about nodes losing a checkpoint locally at restart or by regressing, basically https://github.com/sig-net/mpc/issues/1147 is not an issue any more.

https://github.com/sig-net/mpc/issues/1148 needs to be fixed first for this to work. Especially the late-watcher refill mechanism.

The assumption here is the checkpoint voted into contract always represent what's actually on chain. For example, if 9 out of 12 nodes all missed a request, then the node that restored from on chain will differ from these 9 nodes

@yvonneanne @ChaoticTempest

PR #1128, opened 2026-08-12
feat(ethereum): add isolated Anvil-backed integration tests for publishing

isSerge:
This PR adds isolated Anvil-backed integration tests setup and tests covering publishing. Primary motivation is to fill the gap between two existing test tiers: in-crate mockito-based unit tests and slower cluster integration tests. New middle tier provides realistic EVM environment (contract bytecode, gas estimation, nonce management, etc.) without heavy setup.

Changes:

- Consolidate test helpers in `mpc-chain-ethereum` - used by both new Anvil-based tests and existing tests in `integration-tests`, gated behind the `contract-test` feature
- Add new `isolated-anvil-test` job in `integration.yml`
- Add tests:
  - `publishes_single_response` - submit single sign request, publish single signature, observe response
  - `publishes_batched_responses` - 3 responses within the same flush window, single respond transaction
  - `publishes_across_multiple_batches` - 6 responses in 2 batches (2 transactions)


chain-signatures/chain-ethereum/tests/common/mod.rs
```diff
@@ -0,0 +1,132 @@
...
+/// the responder address and the transaction hash it was mined in.
+pub async fn wait_for_responded(
+    env: &EthTestEnv,
+    request_id: B256,
+    timeout: Duration,
+) -> Result<(Address, B256)> {
+    let filter = Filter::new()
+        .address(env.contract_address)
+        .event_signature(ChainSignatures::SignatureResponded::SIGNATURE_HASH)
+        .topic1(request_id);
```
yvonneanne, line 115: iiuc, wait_for_responded builds Filter::new() with no block range, and alloy omits fromBlock/toBlock when unset, which makes anvil default both to latest. Therefore a log one block behind the head may come back empty. Today the three-batch test seems to have roughly 300ms of slack, so it passes, but since this helper will likely get reused, would it be worth adding .from_block(0u64) so it doesn't depend on where the head happens to be?

chain-signatures/chain-ethereum/src/utils/test.rs
```diff
@@ -0,0 +1,130 @@
...
+    deployer: Address,
+    mpc_address: Address,
+    signature_deposit: U256,
+) -> Result<Address>
+where
+    P: Provider + Clone + 'static,
+{
+    let artifact: Value = serde_json::from_slice(include_bytes!(
+        "../../../contract-eth/artifacts/contracts/ChainSignatures.sol/ChainSignatures.json"
+    ))?;
```
yvonneanne, line 28: pub mod utils; in lib.rs:19 is ungated, and utils/test.rs include_bytes!s the Hardhat artifact, which is gitignored. Does that mean cargo build -p mpc-node now needs npx hardhat compile first? CI is fine because every workflow runs it and Docker copies the artifacts in, but I wondered about real deployments. Would #[cfg(any(test, feature = "integration"))] or a test-utils feature that integration-tests enables be worth it, also to keep the deploy bytecode out of the shipped crate?

isSerge: good point, I have added `contract-test` feature for this. `test-utils` would collide with the existing `test_utils` module (used for in-crate unit tests)

chain-signatures/chain-ethereum/tests/common/mod.rs
isSerge: good catch, updated

chain-signatures/chain-ethereum/tests/publisher_integration.rs
```diff
@@ -0,0 +1,137 @@
...
+use mpc_chain_integration_core::{ChainPublisher, NoopPublisherTelemetry};
+use mpc_primitives::{Chain, SignId, SignKind};
+
+#[tokio::test]
+async fn publishes_single_response() {
+    let env = EthTestEnv::new().await.expect("anvil env");
+
+    let request_id = submit_sign_request(&env.contract(), 7)
+        .await
+        .expect("submit sign request");
```
yvonneanne, line 22: iiuc, respond in ChainSignatures.sol doesn't validate or authorize anything, it just re-emits the requestId it's given. Would these tests still pass if you handed make_publish_action a random B256 and skipped the sign call entirely? If so, would it be worth feeding the emitted SignatureRequested log through parse_filtered_logs to get the SignId, so the tier also covers event-shape vs parser agreement?

chain-signatures/chain-ethereum/tests/publisher_integration.rs
```diff
@@ -0,0 +1,137 @@
...
+            .await
+            .expect("SignatureResponded observed");
+        assert_eq!(responder, env.signer.address());
+        by_tx.entry(tx_hash).or_default().push(*rid);
+    }
+    assert_eq!(
+        by_tx.len(),
+        2,
+        "expected exactly 2 batched respond txs, got {by_tx:?}"
+    );
```
yvonneanne: Would asserting "every tx has at most max_batch_size, every id responded once" be more robust than the exact split?

chain-signatures/chain-ethereum/src/event_parsing.rs
```diff
@@ -296,7 +311,43 @@ mod tests {
...
+                &parsed.params,
+            ),
+            expected
+        );
+    }
+
+    #[test]
+    fn generate_request_id_matches_contract_value() {
+        // Pinned to the value the ChainSignatures contract derives (legacy
+        // ethabi golden, ported from integration-tests).
```
yvonneanne: ChainSignatures.sol says the request id "must be calculated off-chain" and sign never computes one, so is "the value the contract derives" the right description?

yvonneanne (approved):
LGTM, thanks, Serge!

chain-signatures/chain-ethereum/tests/publisher_integration.rs
isSerge: yes, but I plan to do it after #1144 (packing should become more deterministic), added TODO comment

chain-signatures/chain-ethereum/tests/publisher_integration.rs
isSerge: yes, it would pass with any ID, updated `submit_sign_request` to pin the full cycle

chain-signatures/chain-ethereum/src/event_parsing.rs
isSerge: No, it's wrong. Adjusted comment and renamed test

PR #1129, opened 2026-08-12
feat(midnight-publisher-ts): add the 0.15.0 stdio publisher

Pessina:
Adds the Midnight publisher child used by the Rust node to build and submit response intents against `@sig-net/midnight-contract` 0.15.0.

- `build` executes `respond` or `respondBidirectional`, validates deployed verifier keys, and returns a serialized ledger intent with canonical contract key locations.
- `submit` proves the intent, balances DUST only, signs it, and submits through the tracked wallet facade.
- This boundary is TypeScript because Compact's generated contract bindings and executor, plus the Midnight wallet and proving SDKs, are TypeScript APIs. Rust retains orchestration and communicates with the child through a small line-delimited stdio protocol.
- The diff is large because it adds the complete standalone package: roughly two thirds is the pinned npm lockfile, with the remainder covering the protocol, runtime adapters, CI, fixtures and devtools, and boundary-focused tests.


chain-signatures/midnight-publisher-ts/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "midnight-publisher",
```
BRBussy, line 2: A minimal readme.md in this director explaining where/how this midnight-publisher-ts is intended to run would be good to help understand where this fits into the MPC network and node setup. Just something like: This is an MPC node sidecar to be started and run long running alongside an MPC node to be by the node for interacting with the Sig network singleton on Midnight. Required as there is no production Rust SDK for Midnight. Node communicates with this binary via STD In (i.e. cli like communication). Perhaps something about publishing as well, whether docker or as a binary on github etc. It could also explain some of the decisions around how the binary works such as that it processes requests 1 at a time. Also perhaps an explanation of why build and submit is separated here, I think it is so that the state already read and verified by the MPC can be used for txn building instead of the SDK wrapper that fetches its own state.

chain-signatures/midnight-publisher-ts/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "version": "0.1.0",
```
BRBussy, line 3: Typescript linting and formatting should be setup in this project, can be copied from either the midnight-integration repository + run in ci

chain-signatures/midnight-publisher-ts/src/submit.ts
```diff
@@ -0,0 +1,144 @@
...
+  // The gate follows the WORK, not the answer: an abandoned submit is still spending.
+  const release = (): void => {
+    inFlightId = undefined;
+  };
+  void work.then(release, release);
+
+  // Nothing below the wallet is cancellable, so a submit past its deadline keeps
+  // spending; the caller has to learn its post may land anyway.
+  return withDeadline(
+    work,
```
BRBussy: If work takes longer than 6 minutes then this times out and an `ambiguous_submit` error is returned to the caller. Since the wallet is opened and sycned lazily during the first call to submit, the timeout could be either due to the wallet not syncing in time or some other submit issue. Since it is possible to unambiguosly differentiate between the 2, and since it is extremely likely that wallet syncing will experience issues, it is worth separating them to return at least 2 different errors to the caller such as `wallet_not_ready` and `ambiguous_submit` which would help a lot with debugging.

That is a significant as syncing a wallet on first startup on mainnet for midnight is notoriously buggy. It is widely documented as fraught with performance issues and there are a bunch of ways it can go wrong (see [this article](https://gist.github.com/wilsonhoe/69be3c44a15e2447e0ccf2d9b893135d#file-tutorial-md) for example). So it would be worthwhile to create better separation between submission and wallet initialisation/startup to make it easier to debug what is going wrong.

See comment in wallet.ts about how adding a requireReady function to the FundingWallet could help with this.

chain-signatures/midnight-publisher-ts/src/wallet.ts
```diff
@@ -0,0 +1,77 @@
...
+
+// Also the dust intent's TTL: how long a submit dying between finalize and submit strands the coin.
+export const RECIPE_TTL_MS = 5 * 60 * 1000;
+
+async function openFacade(keys: AccountKeys, config: MidnightNodeConfig): Promise<FundingWallet> {
+  const facade = await initialiseWalletFacade(keys, config);
+
+  try {
+    await facade.start(keys.shieldedSecretKeys, keys.dustSecretKey);
+    await facade.waitForSyncedState();
```
BRBussy: This unbounded `waitForSyncedState` without any timeout is known to often hang forever or run for a very long time on mainnet and could become the root of a few problems, also as used here its return value is thrown away so dust generation and balance are not checked before the application uses the wallet. Different separately identifiable error scenarios in the way the wallet is used also end up behind the same `ambiguous_submit` error making it harder for a caller to debug what is going wrong.

A suggested change is as follows:

1. remove this `waitForSyncedState` from the openFacade function entirely. Open can just call start and nothing else. A timeout here will result in stop being called in the catch throwing away any synced progress, resulting in the whole process starting again on the next call. So it is best to separate "starting" the syncing from "waiting for it to finish". So removing the await here on waitForSyncedState allows syncing (which starts with the start() call) to carry on in the background across separate failed attempts to use the wallet.

2. Add a wallet readiness helper method to the funding wallet returned that isolates the `waitForSyncedState` such that the "wait for sync" is retriable and bounded by using the `withDeadline` function (that needs to move to a separate file to prevent cicular imports between wallet.ts and submit.ts):

The interface update:
```ts
export interface FundingWallet {
  // ADD THIS METHOD
  requireReady(deadlineMs: number, minimumDust?: bigint): Promise<void>;
  balanceTx(tx: UnboundTransaction): Promise<BalancingRecipe>;
  // ...
}
```

The function, inside the openFacade function, after the start:
```ts
const requireReady = async(
  facade: WalletFacade,
  deadlineMs: number,
  minimumDust = 1n,
): Promise<void> {
  const state = await withDeadline(facade.waitForSyncedState(), deadlineMs,
    () => new PublisherError("wallet_unsynced", "wallet not caught up with the chain in time; nothing was posted, retry later"));
  const night = state.unshielded.availableCoins;
  if (night.length === 0)
    throw new PublisherError("wallet_unfunded", "funding wallet holds no NIGHT");
  if (night.every((coin) => !coin.meta.registeredForDustGeneration))
    throw new PublisherError("wallet_unfunded", "NIGHT is not registered for dust generation");
  if (state.dust.balance(new Date()) < minimumDust)
    throw new PublisherError("wallet_unfunded", "not enough spendable DUST; a fresh registration takes a few blocks to start generating");
}

// add to the funding wallet:
return {
  requireReady,
  async balanceTx(tx) {
    // ADD HERE to ensure ready before balancing
    await requireReady(BALANCE_READY_TIMEOUT_MS);
    // ... existing body
  },
  // ...
};
```
Having the timeout in the `requireReady` function allows throwing without stopping the facade, so failing is cheaper and retriable. The different error messages also help a caller better understand what has gone wrong.


3. in the submit.ts file, gate the handleSubmit function with this, before the inFlightId gate:
```ts
export async function handleSubmit(config: Config, id: number, intent: Uint8Array): Promise<Landed> {
  const decoded = decodeIntent(intent);          // bad_request still fails instantly
  const ready = await publisher(config);         // fast now: open no longer waits for sync
  await ready.wallet.requireReady(READY_TIMEOUT_MS);
  
// ...
}
```
This splits "wallet never became ready" from ambiguous_submit, and since a not-ready wallet never takes the gate, a hung sync can no longer wedge the process into returning wallet_busy forever. The requireReady inside balanceTx covers the remaining gap: by the time balancing runs, the handleSubmit check is around 35s stale (proving sits in between), and balance is the one step that selects coins from live wallet state. Once synced, both re-checks resolve near-instantly.

chain-signatures/midnight-publisher-ts/src/prover.ts
```diff
@@ -0,0 +1,109 @@
...
+  if (hashVerifierKey(verifierKey) !== verifierKeyHash) {
+    throw new PublisherError(
+      "contract_mismatch",
+      `the packaged verifier key for circuit \`${circuit}\` does not match the intent`,
+    );
+  }
+  return { proverKey, verifierKey, ir };
+}
+
+async function post(proofServerUrl: string, path: string, body: Uint8Array): Promise<Uint8Array> {
```
BRBussy: A timeout on this proof server call could be useful for debugging/retries. Then the caller could get something like a "PROVING_TIMEOUT" which is an error class that separate to an `ambiguous_submit` error. Either `signal: AbortSignal.timeout(...)` argument to the fetch or timeout from above.

chain-signatures/midnight-publisher-ts/src/protocol.ts
```diff
@@ -0,0 +1,138 @@
...
+}
+
+// Absent `op` is `build`: the request predates the discriminator and the Rust client
+// sends none. An unknown op hits `default` before either schema can blame a missing field.
+async function answer(config: Config, body: Record<string, unknown>): Promise<Response> {
+  switch (body["op"]) {
+    case "ready": {
+      const parsed = ReadySchema.safeParse(body);
+      if (!parsed.success) throw toBadRequest(parsed.error);
+      return {
```
BRBussy, line 102: might be worth adding a protocol version in this ready call to allow rust client to align on version

chain-signatures/midnight-publisher-ts/src/submit.ts
Pessina: Split the absolute deadline into wallet_unsynced, proving_timeout, and post-wallet ambiguous_submit; wallet startup remains memoized across retries (3d765d2f).

chain-signatures/midnight-publisher-ts/src/prover.ts
Pessina: Added one aggregate proof budget with an owned AbortController across /check, /prove, response bodies, and transaction.prove; expiry returns proving_timeout before wallet work (3d765d2f).

chain-signatures/midnight-publisher-ts/src/wallet.ts
Pessina: Startup now returns while sync continues. requireReady uses a cancellable DUST-only RxJS wait; NIGHT/1n checks are omitted because only DUST is balanced and exact sufficiency belongs to wallet balancing (3d765d2f).

chain-signatures/midnight-publisher-ts/package.json
Pessina: Added a README covering the sidecar role, stdio protocol, build/submit split, sequential processing, configuration, and deployment seam (3d765d2f).

chain-signatures/midnight-publisher-ts/src/protocol.ts
Pessina: Ready now requires and echoes exact protocolVersion: 1; missing or mismatched versions return bad_request (3d765d2f).

chain-signatures/midnight-publisher-ts/package.json
Pessina: Added package-local ESLint/Prettier scripts and CI gates for formatting, lint, typecheck, and tests (3d765d2f).

chain-signatures/midnight-publisher-ts/src/submit.ts
```diff
@@ -0,0 +1,204 @@
...
+  // Balance, finalize and submit are not cancellable, so work past the deadline keeps
+  // spending; the caller has to learn its post may land anyway.
+  const ambiguousSubmit = new PublisherError(
+    "ambiguous_submit",
+    `submit exceeded the ${SUBMIT_TIMEOUT_MS} ms deadline; if it had reached submit the transaction ` +
+      `may still land, so check the chain for request ${id} before retrying`,
+  );
+  const answerBudgetMs = remaining();
+  if (answerBudgetMs <= 0) throw ambiguousSubmit;
+  return withDeadline(work, answerBudgetMs, ambiguousSubmit);
```
ppca, line 203: so is it like if any of the calls, including balanceTx, finalizeTx or submitTx somehow hangs, then the inflightId won't clear? 
Wondering is it possible for such calls to hang forever? if so, how do we recover from the hang?

chain-signatures/midnight-publisher-ts/src/wallet.ts
```diff
@@ -0,0 +1,104 @@
...
+export interface FundingWallet {
+  requireReady(timeoutMs: number): Promise<void>;
+  balanceTx(tx: UnboundTransaction): Promise<BalancingRecipe>;
+  finalizeTx(recipe: BalancingRecipe): Promise<FinalizedTransaction>;
+  submitTx(tx: FinalizedTransaction): Promise<Landed>;
+  close(): Promise<void>;
+}
+
+// Also the dust intent's TTL: how long a submit dying between finalize and submit strands the coin.
+export const RECIPE_TTL_MS = 5 * 60 * 1000;
```
ppca, line 33: why is this 5 min while the submit deadline is 6 min?

ppca (commented):
mostly LGTM, except 2 comments. 
I'm no typescript expert, mostly reviewing in terms of the logic.

chain-signatures/midnight-publisher-ts/src/submit.ts
Pessina: Yeah, that’s a known issue in midnight SDK: https://github.com/midnightntwrk/midnight-node/issues/1971

The fix will be on the Rust side, kill the child process when we receive `ambiguous_submit` and re-submit after confirming the tx TTL has passed.

The issue it's wallet sync, scanning the chain again to rebuild the wallet state. I’m thinking about persisting the full wallet state so restarting the child is faster.

chain-signatures/midnight-publisher-ts/src/wallet.ts
Pessina: They are different things, 6 min is the timeout for `handleSubmit` and 5 min is the balanceTx TTL. 

On Midnight TXs have TTLs, on our case 2: 

- balanceTx
- contractTx

The TX can execute up to min(all TTls on tx). 

The 6 min bound the `handleSubmit` starting from handleSubmit first line. The 5 min TTL starts when the `balanceTx` is called

So they start on different moments and bounds different things

ppca (approved):

PR #1130, opened 2026-08-12
docs: add the protocol properties document

yvonneanne:
doc/protocol_properties.md states what must hold in the MPC coordination layer and why: safety (S1-S4), liveness (L1-L4), efficiency targets, and the design invariants behind them. The appendix explains for each property how the protocol enforces it.

Once we agree on the main properties and how to enforce them we can work on changing the code accordingly.

### DIFF
--- a/doc/protocol_properties.md
+++ b/doc/protocol_properties.md
@@ -1,0 +1,196 @@
+# **Protocol Properties** 
+
+Draft, last updated 2026, August 12, 15:31 UTC
+
+Goals:
+
+* capture *what must hold* in the MPC node coordination layer and *why* to be able to reason about simplifications that improve robustness
+
+Protocol properties defined below are separated into
+
+* safety (nothing bad ever happens: no key leak, no unrequested signature)  
+* liveness (every request eventually completes), and  
+* efficiency (we try not to waste artifacts or rounds doing it). Some waste is acceptable as long as it stays bounded.
+
+Not all of the below is implemented. Where a property or invariant is only partly enforced today, the gap is named where it is claimed.
+
+## 1\. Request lifecycle and its objects
+
+* Sign request flow  
+  1. request is finalized on-chain 
+  2. indexed by each node's own indexer; the request\_id is a deterministic function of the request, so every node derives the same one  
+  3. a per-node task organizes rounds (attempts) until one generation instance produces a signature  
+  4. the successful generation's proposer submits it on chain  
+  5. settlement is chain-specific: NEAR verifies the signature and resumes the yielded promise, emitting no event of its own; the EVM contract keeps no request state and emits an event for any submitter without checking anything  
+  6. downstream process is responsible for event verification and exactly once semantics 
+
+Both artifacts below have a single owner, the node that coordinated their generation, which simplifies statesync and coordination, and a set of holders, the nodes storing their shares.
+
+* Triple pair (artifact): two secret-shared random triples, produced in the background and stored, owned and consumed as a single artifact under one id, spent to build one presignature.
+
+* Presignature (artifact): pre-computed, secret-shared ECDSA nonce (big\_r, and shares k, sigma).
+
+* Posit: prepare phase leading up to an instance
+  * proposer sends Propose message to each peer  
+  * peers answer Accept/Reject  
+  * proposer sends Start carrying the final participant list to each accepter  
+  * the proposer reserves its artifact when proposing and commits it, removing it from the pool, when generation starts; aborting before the commit releases the reservation, aborting after does not  
+
+* Instance: one attempt to run the protocol to completion over a specific artifact, named by that artifact
+  * Signature: (sign\_id, presignature\_id) (the code's SignId wraps the request\_id above, same 32 bytes)  
+  * Triple pair/presignature: identified by artifact id. A triple pair's id is drawn at random by its proposer; a presignature's id is derived from the pair it consumes (hash of the pair id), not chosen.  
+
+  A request's round tries to establish a signature instance; since the artifact tentatively picked for this round can differ round to round, rounds are usually different instances. If a round is not successful, a new proposer is chosen deterministically (rotating over the nodes). Only posit messages name the round; generation messages name only (sign\_id, presignature\_id), so two rounds that re-pick the same presignature, possible when the first aborted before committing it, share one name.
+
+## 2\. System model
+
+* On-chain information
+  * Entities: n nodes; membership, threshold t, and epoch are fixed by the NEAR governance contract. 
+    * Together with the public keys and whether the contract is `Running`, they form the **committee** a node operates under (`GovernanceInfo` in code).
+    * Each node holds its own local, read-only snapshot of it, refreshed on contract update. Throughout, *committee* and *committee member* are contract-scoped, while *participant* and *participant list* always refer to one instance.
+  * Request and response (per supported chain): NEAR holds pending requests, verifies signatures for them and settles them. The others mostly emit an event for whoever submits, which is why exactly-once is the consumer's job (§1).
+  
+  (i) all nodes observe the same finalized sequence of requests and committee states, possibly at different times;  
+  (ii) nodes act only on finalized state (finalization definition may differ per chain).  
+* Hybrid fault model.  
+  * Liveness is argued under crash-recovery: correct nodes may halt and restart (otherwise never deviating). After restart the node's durable storage is assumed to be intact.  
+  * Safety of S1–S4 must hold against f arbitrarily corrupted nodes.
+    Two conditions bound f. 
+    * f \< t keeps the adversary from assembling a signing quorum out of its own key shares; with f ≥ t it holds one and all bets are off. 
+    * f \< 2t − n makes any two participant lists share at least one honest member, which is what stops one artifact being consumed by two otherwise disjoint lists. The contract sizes t at ⌊2n/3⌋ \+ 1 for n ≥ 5 and thus works for every f up to n − t; a threshold vote can only raise t further, since the contract clamps a proposed threshold to \[⌊2n/3⌋ \+ 1, n − 1\], and both conditions get easier as t grows, so that floor is what keeps the model intact. Below n \= 5 it falls back to a simple majority (⌊n/2⌋ \+ 1), which does not: at n \= 3 that gives 2t − n \= 1, so only f \= 0 is inside the model. The floor is applied on resharing and by threshold votes, not to the value passed at contract initialization, which is checked only against n, so a network can also be deployed outside the model.  
+* Network  
+  * Safety is guaranteed under asynchrony (no bounds on message delivery), while liveness is guaranteed for the periods where the network is synchronous (messages between honest nodes arrive within some bound δ, not assumed known) for long enough.  
+    * Formally (see e.g., Shoup, DISC 2024): the network is δ-synchronous over \[a, b+δ\] if every message an honest node sends at time T in \[a, b\] reaches its honest recipient by T+δ ; liveness needs such an interval lasting longer than the current timeout. With unknown δ, the classic technique is to increase timeouts until progress is observed (DLS 1988, used as view-timeout doubling in PBFT and as a linear per-round schedule in Tendermint; the Simplex family instead assumes a known bound and fixes ∆timeout ≥ 3δ per slot). This implementation sits in between: the per-round schedule `round_timeout(r)` grows only up to a hard ceiling, which amounts to assuming a whole round fits inside that ceiling: a Propose, Accept and Start exchange, the generation protocol, and the indexing skew between two nodes. Liveness is therefore argued for δ up to a fraction of the ceiling, not for arbitrary finite δ (more details in L1)  
+  * Links are fair-lossy (may drop messages), which every phase compensates for by timeout-and-retry, so during a synchronous interval, communication between live nodes is effectively reliable and timely.
+  * Channels are authenticated and encrypted. 
+* Each node runs its own chain indexers and eventually observes every finalized request (assumption; liveness depends on it, safety does not).  
+* Chain state reaches a node only through its RPC provider, which sits outside the fault model above: a provider that reports a committee or a request the chain never finalized breaks S3 and S2 at that node, and nothing in this layer detects it. Verifying responses against a light client is what would remove the assumption; until then it is one, and it covers the indexers as well.  
+* The mesh active set is a local, unreliable failure detector: each node's own guess at which members are currently reachable and up-to-date (active), a subset of the committee. It may be wrong, and no two nodes ever need the same guess; §6 D2 governs what may be derived from it. A reachable peer is additionally kept *out* of the active set while an initial or post-reconnect state sync runs (a transient Syncing state), so "active" is strictly narrower than "reachable".
+
+## 3\. Safety properties
+
+* **S1. One-shot artifacts are consumed at most once:** No presignature is used in signature shares for more than one sign request, and no triple pair for more than one presignature.
+* **S2. Only indexed requests are signed:** An honest node contributes a signature share only to requests its own indexer delivered from a finalized chain state.  
+* **S3. Membership and epochs change only on-chain:** No honest node adopts a committee other than from finalized contract state, and no node processes requests under a non-Running state. 
+* **S4. Instance-local agreement:** Any two honest participants that reach the generating phase of the same instance use the same participant list or they abort.
+
+Violating any of these is unacceptable in any execution within §2's fault bound. See Appendix for more details, including enforcing mechanisms.
+
+## 4\. Liveness
+
+Progress is guaranteed during a sufficiently long synchronous interval, never at a fixed time. The precise bounds and the enforcing mechanisms are in the Appendix.
+
+* **L1. Signature progress:** during a long-enough δ-synchronous interval with ≥ t correct committee members online, a usable presignature (owner and holders online), and ≥ t of that presignature's holders having indexed the request, the request produces a signature within bounded time (O(f \+ p)·∆timeout \+ one generation round, where p counts the members that are offline or own no usable presignature). That such a presignature keeps existing is L2.  
+* **L2. Artifact supply:** during a long-enough synchronous interval with ≥ t members (itself included) in the node's active set, a node below its artifact floor and under the network-wide cap eventually completes a generation, given its inputs (none for a triple pair, one owned triple pair for a presignature).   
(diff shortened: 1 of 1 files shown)


doc/protocol_properties.md
```diff
@@ -0,0 +1,196 @@
+# **Protocol Properties** 
+
+Draft, last updated 2026, August 12, 15:31 UTC
+
```
yvonneanne:
```suggestion
```

doc/protocol_properties.md
```diff
@@ -0,0 +1,194 @@
...
+For NEAR-originated requests this requires settling before the yielded promise's bounded lifetime expires (a hard deadline of 200 blocks \~= 200s)  
+
+L1 and L3 draw on that one deadline, and the rounds spend it first: round 0 costs 20 s and later rounds start at 2 s and grow by 1.15, so about 19 further rounds fit inside 200 s. A NEAR request that needs more rounds than that cannot settle at all, however the settlement path is built, which is the constraint any republish or failover schedule has to fit into.
+
+*Enforcement*: The owner republishes on a timer, retrying indefinitely with backoff capped at 60 s.
+Currently not guaranteed, for two reasons. Only the successful instance's proposer publishes, and no other participant takes over, even though every participant holds the finished signature. And the retry itself lives in the publishing process, so what survives a restart is only what reached a confirmed checkpoint: on the checkpointed chains that includes the pending publish and its signature, while NEAR and Bitcoin produce no checkpoints at all and keep it in memory alone. On NEAR the 200 s yield deadline can also expire while the publisher is still backing off.
```
jakmeier, line 186: Hm, I'm not sure this model for NEAR works well with the no-timeout guarantee we want in bidirectional requests. I guess if we ever implement that for NEAR as the source chain, that will need to work without the yield/resume mechanism.

doc/protocol_properties.md
```diff
@@ -0,0 +1,194 @@
...
+  * Links are fair-lossy (may drop messages), which every phase compensates for by timeout-and-retry, so during a synchronous interval, communication between live nodes is effectively reliable and timely.
+  * Channels are authenticated and encrypted. 
+* Each node runs its own chain indexers and eventually observes every finalized request (assumption; liveness depends on it, safety does not).  
+* Chain state reaches a node only through its RPC provider, which sits outside the fault model above: a provider that reports a committee or a request the chain never finalized breaks S3 and S2 at that node, and nothing in this layer detects it. Verifying responses against a light client is what would remove the assumption; until then it is one, and it covers the indexers as well.
```
jakmeier, line 65: This is correct. S2 and S3 fully reliy on a trustworthy RPC provider.

@ppca @and @isSerge How much work would it be to use at lest 2 different RPC providers per chain?

jakmeier (approved):
Thank you for writing this down! It's great to have it all summarized in a document in the repo, especially in the time of LLMs.

In my opinion, we can merge it pretty much as-is. But I would like to see a follow-up that also describes the bidirectional flow and its properties. I think this isn't covered, yet. (Which makes sense, we are still figuring out what to do exactly with duplicate requests.)

It might make sense to complement this document with a table somewhere that lists what features and what security guarantees we have per chain. At least I always get confused about which property currently hold simply because it may be different on each chain.

doc/protocol_properties.md
yvonneanne: it doesnt. Afaics, this is today's status.

yvonneanne:
> Thank you for writing this down! It's great to have it all summarized in a document in the repo, especially in the time of LLMs.
> 
> In my opinion, we can merge it pretty much as-is. But I would like to see a follow-up that also describes the bidirectional flow and its properties. I think this isn't covered, yet. (Which makes sense, we are still figuring out what to do exactly with duplicate requests.)
> 
> It might make sense to complement this document with a table somewhere that lists what features and what security guarantees we have per chain. At least I always get confused about which property currently hold simply because it may be different on each chain.

I agree, bidirectional flows and chain-specific guarantees and assumptions need to be documented as well (follow-ups to this PR.)

doc/protocol_properties.md
jakmeier: Yeah it is today's status but we also don't really care about signature requests on NEAR. Something to keep in mind if traffic on NEAR becomes important later.

doc/protocol_properties.md
```diff
@@ -0,0 +1,194 @@
...
+
+Not all of the below is implemented. Where a property or invariant is only partly enforced today, the gap is named where it is claimed.
+
+## 1\. Request lifecycle and its objects
+
+* Sign request flow  
+  1. request is finalized on-chain 
+  2. indexed by each node's own indexer; the request\_id is a deterministic function of the request, so every node derives the same one  
+  3. a per-node task organizes rounds (attempts) until one generation instance produces a signature  
+  4. the successful generation's proposer submits it on chain
```
volovyks, line 21: We may add that "in case of a failure, the signature is published by another generation participant".

doc/protocol_properties.md
```diff
@@ -0,0 +1,194 @@
...
+Not all of the below is implemented. Where a property or invariant is only partly enforced today, the gap is named where it is claimed.
+
+## 1\. Request lifecycle and its objects
+
+* Sign request flow  
+  1. request is finalized on-chain 
+  2. indexed by each node's own indexer; the request\_id is a deterministic function of the request, so every node derives the same one  
+  3. a per-node task organizes rounds (attempts) until one generation instance produces a signature  
+  4. the successful generation's proposer submits it on chain  
+  5. settlement is chain-specific: NEAR verifies the signature and resumes the yielded promise, emitting no event of its own; the EVM contract keeps no request state and emits an event for any submitter without checking anything
```
volovyks, line 22: I would not even mention NEAR. It is not something we released. It is only used for tests, and we intend to have event-based design everywhere for consistency and simplicity.

doc/protocol_properties.md
```diff
@@ -0,0 +1,194 @@
...
+
+## 2\. System model
+
+* On-chain information
+  * Entities: n nodes; membership, threshold t, and epoch are fixed by the NEAR governance contract. 
+    * Together with the public keys and whether the contract is `Running`, they form the **committee** a node operates under (`GovernanceInfo` in code).
+    * Each node holds its own local, read-only snapshot of it, refreshed on contract update. Throughout, *committee* and *committee member* are contract-scoped, while *participant* and *participant list* always refer to one instance.
+  * Request and response (per supported chain): NEAR holds pending requests, verifies signatures for them and settles them. The others mostly emit an event for whoever submits, which is why exactly-once is the consumer's job (§1).
```
volovyks: Again, I would not focus on NEAR as much.

doc/protocol_properties.md
```diff
@@ -0,0 +1,194 @@
...
+* Chain state reaches a node only through its RPC provider, which sits outside the fault model above: a provider that reports a committee or a request the chain never finalized breaks S3 and S2 at that node, and nothing in this layer detects it. Verifying responses against a light client is what would remove the assumption; until then it is one, and it covers the indexers as well.  
+* The mesh active set is a local, unreliable failure detector: each node's own guess at which members are currently reachable and up-to-date (active), a subset of the committee. It may be wrong, and no two nodes ever need the same guess; §6 D2 governs what may be derived from it. A reachable peer is additionally kept *out* of the active set while an initial or post-reconnect state sync runs (a transient Syncing state), so "active" is strictly narrower than "reachable".
+
+## 3\. Safety properties
+
+* **S1. One-shot artifacts are consumed at most once:** No presignature is used in signature shares for more than one sign request, and no triple pair for more than one presignature.
```
volovyks, line 70: I would replace "sign request" with "signature generation". We want to ensure node providers do not see 2+ generations as well.

doc/protocol_properties.md
```diff
@@ -0,0 +1,194 @@
...
+    * Together with the public keys and whether the contract is `Running`, they form the **committee** a node operates under (`GovernanceInfo` in code).
+    * Each node holds its own local, read-only snapshot of it, refreshed on contract update. Throughout, *committee* and *committee member* are contract-scoped, while *participant* and *participant list* always refer to one instance.
+  * Request and response (per supported chain): NEAR holds pending requests, verifies signatures for them and settles them. The others mostly emit an event for whoever submits, which is why exactly-once is the consumer's job (§1).
+  
+  (i) all nodes observe the same finalized sequence of requests and committee states, possibly at different times;  
+  (ii) nodes act only on finalized state (finalization definition may differ per chain).  
+* Hybrid fault model.  
+  * Liveness is argued under crash-recovery: correct nodes may halt and restart (otherwise never deviating). After restart the node's durable storage is assumed to be intact.  
+  * Safety of S1–S4 must hold against f arbitrarily corrupted nodes.
```
volovyks, line 55: Let's define "Safety properties" before referencing them to avoid confusion.

doc/protocol_properties.md
volovyks: Note: the current security model (without a light client) is to use different RPC providers on different nodes. NR < T, where NR is a number of nodes running on R RPC provider.

doc/protocol_properties.md
```diff
@@ -0,0 +1,194 @@
...
+  * Channels are authenticated and encrypted. 
+* Each node runs its own chain indexers and eventually observes every finalized request (assumption; liveness depends on it, safety does not).  
+* Chain state reaches a node only through its RPC provider, which sits outside the fault model above: a provider that reports a committee or a request the chain never finalized breaks S3 and S2 at that node, and nothing in this layer detects it. Verifying responses against a light client is what would remove the assumption; until then it is one, and it covers the indexers as well.  
+* The mesh active set is a local, unreliable failure detector: each node's own guess at which members are currently reachable and up-to-date (active), a subset of the committee. It may be wrong, and no two nodes ever need the same guess; §6 D2 governs what may be derived from it. A reachable peer is additionally kept *out* of the active set while an initial or post-reconnect state sync runs (a transient Syncing state), so "active" is strictly narrower than "reachable".
```
volovyks, line 66: This is all true, but it may need simplification, since we are not using `active` when selecting protocol participants anymore.

doc/protocol_properties.md
```diff
@@ -0,0 +1,194 @@
...
+
+Violating any of these is unacceptable in any execution within §2's fault bound. See Appendix for more details, including enforcing mechanisms.
+
+## 4\. Liveness
+
+Progress is guaranteed during a sufficiently long synchronous interval, never at a fixed time. The precise bounds and the enforcing mechanisms are in the Appendix.
+
+* **L1. Signature progress:** during a long-enough δ-synchronous interval with ≥ t correct committee members online, a usable presignature (owner and holders online), and ≥ t of that presignature's holders having indexed the request, the request produces a signature within bounded time (O(f \+ p)·∆timeout \+ one generation round, where p counts the members that are offline or own no usable presignature). That such a presignature keeps existing is L2.  
+* **L2. Artifact supply:** during a long-enough synchronous interval with ≥ t members (itself included) in the node's active set, a node below its artifact floor and under the network-wide cap eventually completes a generation, given its inputs (none for a triple pair, one owned triple pair for a presignature).
```
volovyks, line 82: Do we still want to use `active` in our descriptions?

doc/protocol_properties.md
```diff
@@ -0,0 +1,194 @@
...
+ 
+* D1. The chain is the sole arbiter. Request settlement and membership live on-chain. The off-chain layer must remain safe under arbitrary duplication, reordering, loss, and retry of its own actions. *Serves S2, S3 (safety-critical agreement is delegated to the chain) and is necessary for D4.*  
+* D2. Anything two nodes must compute identically (shared input, e.g. "who proposes in round r") may depend only on contract state, request data (e.g. the entropy, which is chain-fixed per request, the transaction hash on Ethereum), and the round number, never on the local failure detector or local clocks.  Rationale: local views differ by design; deriving shared roles from them silently reintroduces the agreement problem this layer avoids. Two caveats: shared is not the same as unbiasable, since a requester who controls the transaction controls the entropy and with it the first proposer; and "not local clocks" constrains the mapping from shared inputs to roles, not the round number itself, which each node advances on its own timeouts (nodes disagreeing about the current round is expected and is repaired by buffering future-round messages and by stale-round rejects). *Serves L1.*
```
volovyks, line 101: The requester does not control or provide the entropy. It originates from the chain.
But the current implementation of entropy selection is chain-specific and has questionable logic. Related: https://github.com/sig-net/mpc/pull/724

doc/protocol_properties.md
```diff
@@ -0,0 +1,194 @@
...
+
+This holds even against a Byzantine proposer: A malicious proposer can send divergent lists to different honest accepters, but each honest node uses the list it was told and different nodes were told different lists, so they never reconstruct a valid signature: the combine fails and the instance aborts. The worst outcome is an abort, never a bad signature. 
+
+### L1. Signature progress.
+
+*Property.* During a long-enough δ-synchronous interval with ≥ t correct committee members online, a usable presignature (owner and holders online), and ≥ t of that presignature's holders having indexed the request, the request produces a signature within bounded time (O(f \+ p)·∆timeout \+ one generation round, where p counts the members that are offline or own no usable presignature). That such a presignature keeps existing is L2.
```
volovyks, line 148: I would ask AI to look for duplications in this doc; a couple of things are repeated at least 2 times.

doc/protocol_properties.md
```diff
@@ -0,0 +1,194 @@
...
+
+*Rationale.* This is the property a user actually observes: a request that never reaches a signature is indistinguishable from the service being down.
+
+*Enforcement*: 
+The proposer starts as soon as every invitee has answered (Accept or Reject) and at least t have accepted, so with everyone honest and up-and-running the signature is produced at network speed in one generation round, independent of ∆timeout. Silent invitees do not hold the round hostage: 500 ms after it sends Propose the proposer goes ahead with whoever accepted, as long as that is still t. An alternative proposer gets a chance after ∆timeout.
+
+Proposer election is a pure function of shared inputs (`proposer_per_round` over round, membership, entropy) and the deadline `round_timeout(r)` depends only on the round `r`. Therefore peers in the same round agree on the proposer and the deadline (D2). A round advance is only ever triggered locally (deadline, enough rejects, abort), but the round it advances to is the highest a peer has shown us, so a node that falls behind catches up in one bump by learning the rejector's current round rather than climbing one round per attempt.
```
volovyks, line 159: I believe it is better to have a single description of each layer. Looks like posits will have their own doc: https://github.com/sig-net/mpc/pull/1152
Would it be a good idea to move some of the posit descriptions there and then link that doc here?

doc/protocol_properties.md
```diff
@@ -0,0 +1,194 @@
...
+A node re-proposes every 100 ms while it is below its floor (`min_triples` / `min_presignatures`) and the network-wide potential is under the cap (`max_triples` / `max_presignatures`); concurrency counters free themselves as generation tasks finish and as proposals expire.  
+Generation is skipped entirely while \< t nodes are in the active set, so the failure detector gates supply though never correctness.
+
+### L3. Settlement
+
+*Property.* Once a signature is produced with a correct, online owner, it is eventually accepted on-chain (on NEAR this must happen before the request's yield deadline).
+
+*Rationale.* A signature that never reaches the chain is worth no more to the user than no signature at all, while the presignature spent producing it is gone either way.
+
+For NEAR-originated requests this requires settling before the yielded promise's bounded lifetime expires (a hard deadline of 200 blocks \~= 200s)
```
volovyks, line 181: I'm not highlighting every NEAR occurrence, but I think we need to remove all the mentions, since they are not important and only create mental load.

doc/protocol_properties.md
volovyks: If we ever have traffic on NEAR, we can switch to an event-based model.

volovyks (approved):
Great doc!

doc/protocol_properties.md
yvonneanne:
```suggestion
  4. the signature is submitted on-chain
```
Thinking a bit more about it, I think it's better not to have too much detail here. We haven't introduced proposer etc yet.

doc/protocol_properties.md
yvonneanne:
```suggestion
  5. the contract emits an event for the submission without checking anything  
```
better?

doc/protocol_properties.md
yvonneanne:
```suggestion
  * Request and response (per supported chain): 
```

doc/protocol_properties.md
yvonneanne:
```suggestion
  * Safety must hold against f arbitrarily corrupted nodes.
```

doc/protocol_properties.md
yvonneanne:
```suggestion
* Chain state reaches a node only through its RPC provider. Different nodes can use different RPC providers. The current security model assumes that the number of nodes using a specific RPC provider is below t.  
```

Did I interpret your statement correctly.

doc/protocol_properties.md
yvonneanne: We still use it to determine a "good" presignature to propose.

doc/protocol_properties.md
yvonneanne:
```suggestion
* **S1. One-shot artifacts are consumed at most once:** No presignature is used in signature shares for more than one signature generation, and no triple pair for more than one presignature.
```
true, thanks for catching this!

doc/protocol_properties.md
yvonneanne: The current implementation doesn't guarantee it if there are fewer in the active set.
We need some description of the status of the nodes in the subnet. We can use a different definition, but I don't think it helps us here.

doc/protocol_properties.md
yvonneanne: Each property is stated in the main part and then repeated in the appendix. I will go over the doc to avoid additional duplication.

doc/protocol_properties.md
yvonneanne: Yeah, that's a good idea. I'll do that in a follow-up PR.

doc/protocol_properties.md
yvonneanne: that's not entirely correct, the transaction hash is can be influenced with nonce and calldata.. to do it better we could rely on the chain's native per-block randomness (currently droppped by indexer).

doc/protocol_properties.md
yvonneanne: I like that :) Done

doc/protocol_properties.md
yvonneanne: I removed all mentions of NEAR

doc/protocol_properties.md
yvonneanne: Thanks for the clarification, I rephrased it accordingly

PR #1133, opened 2026-08-13
doc: add spec for backlogs/checkpoints

ChaoticTempest:
This adds the spec for backlogs, checkpoints, and requests incoming from chains. This is quite minimal for now

### DIFF
--- a/doc/backlog.md
+++ b/doc/backlog.md
@@ -1,0 +1,69 @@
+# Backlog
+
+The backlog is the node's per-source-chain record of signing work that has been indexed but is not yet terminal. It makes indexing, signing, publishing, and restart recovery converge on the same ordered chain history.
+
+## Scope And Invariants
+
+- A request is identified by `SignId` and belongs to one source chain.
+- Index a request before signing it. Do not start signing before the indexer completes catchup and the request meets that chain's finality policy.
+- Store a request before enqueueing it. Duplicate observations must not create duplicate active signing work.
+- A completed source-chain response removes ordinary work. A bidirectional response first moves work to target-execution watching.
+- Never reuse MPC triples or pre-signatures after a protocol attempt has begun; recoverable signing failures retry in a later round.
+- The backlog is recoverable from durable checkpoints. Live stream state and target-chain watchers are reconstructed after recovery.
+
+## Request Lifecycle
+
+1. The source signing contract emits a request. The indexer converts it to a normalized chain event.
+2. During catchup, requests enter the backlog but are not sent to signing. `CatchupCompleted` re-enqueues pending-generation requests in a stable order and resumes pending publications.
+3. The signing protocol deduplicates active `SignId`s. On success, only the proposer publishes the response and the request becomes pending publication.
+4. Publication is deduplicated and retried with backoff. A validated source response completes ordinary requests.
+5. A bidirectional request instead progresses through: `PendingGeneration -> PendingPublish -> PendingExecution -> PendingGenerationBidirectional -> PendingPublishBidirectional -> complete`. Target execution creates the final response-signing request.
+
+An empty bidirectional serialized transaction is rejected before storage. Target-chain decoding and transaction validation happen later, while processing the source response. Duplicate target execution confirmations are intended to be idempotent.
+
+## Checkpoints
+
+A checkpoint contains the chain, processed height, pending request recovery data, and a digest of the consensus-relevant request IDs and coarse phases. The serialized recovery data is local state and is intentionally not part of the consensus digest.
+
+- Create checkpoints only after catchup, when an observed height crosses a configured interval bucket. Sparse block/slot/offset observations are valid.
+- Persist the checkpoint as pending, then vote for its digest through governance.
+- After threshold confirmation, atomically promote it to the latest checkpoint and prune older pending checkpoints.
+- Retain at most 32 unconfirmed checkpoints per chain. At the limit, stop consuming stream events for that chain so bounded backpressure reaches the indexer rather than losing work.
+- On restart, restore the newest durable pending checkpoint when present, otherwise the latest confirmed checkpoint, then align it with governance.
+
+If pending-checkpoint storage fails, the in-memory processed height has already advanced and the current implementation does not retry that interval. A later interval crossing can checkpoint again; a crash before then replays from the last durable checkpoint. This is an accepted current behavior, not a settled durability policy.
+
+Default intervals are Ethereum 20 blocks, Solana 120 slots, Hydration 240 blocks, and Canton 50 offsets. NEAR and Bitcoin currently are not implemented.
+
+## Regression And Recovery
+
+A regression is a local checkpoint that does not match governance consensus.
+
+The intended recovery is:
+
+1. Abort source-chain signing tasks and checkpoint voting.
+2. Stop the indexer and obtain the consensus checkpoint from a peer by digest.
+3. Replace local durable and in-memory backlog state, rebuild execution watchers, and catch up from the restored height.
+
+The supervisor also restarts an indexer after an error or chain-specific stream stall. When governance has no checkpoint, no alignment occurs. When governance does have a checkpoint that is absent locally, including fresh bootstrap, the node fetches a digest-matching checkpoint from a peer and regresses to it.
+
+## Chains And Indexers
+
+| Chain | Current behavior | Backlog status |
+| --- | --- | --- |
+| Ethereum | Finalized-head indexing by default; optimistic mode is development/test only. Ordinary requests and responses are indexed. | Supervised, checkpointed. First catchup without a checkpoint starts at the startup anchor, not a deployment/genesis height. |
+| Solana | Historical catchup plus a pre-started websocket live stream; uses `confirmed` commitment. Ordinary and bidirectional requests are indexed. | Supervised, checkpointed. Live buffering can overflow during slow catchup. |
+| Canton | Resumes from ledger offset, catches up to an anchor, then streams; source requests are bidirectional. | Supervised, checkpointed. Finality semantics and lifecycle tests are incomplete. |
+| Hydration | Subscribes to finalized blocks and verifies event-storage proofs before parsing requests and responses. | Does not use the common stream/checkpoint event pipeline. Disconnect-window backfill and checkpoint advancement are open. |
+| NEAR | Polls contract pending requests with an in-memory, one-hour seen cache. | Integration-test-only path; bypasses backlog, checkpoint, and regression supervision. |
+| Bitcoin, Midnight | Chain identifiers/configuration exist. | TBD |
+
+Ethereum is the only implemented target-execution watcher found. Bidirectional
+execution monitoring for Solana, Canton, and Hydration is open.
+
+## Failure Rules
+
+- Indexer and transient RPC failures retry or cause supervised restart; Recognized terminal HTTP client errors (400, 401, 403, 404, 405) are not retried; 408, 429, timeouts, and other errors are retried according to the RPC policy.
+- Publication must survive restart while a request is pending publication.
+- A stalled checkpoint consensus intentionally backpressures that chain. Its operational behavior must be monitored because it can also trigger watchdog restart loops.
+- Do not treat provider data as cryptographically verified unless the indexer verifies it. Hydration currently verifies a state proof; the other indexers rely on their providers.


doc/backlog.md
```diff
@@ -0,0 +1,69 @@
...
+- Store a request before enqueueing it. Duplicate observations must not create duplicate active signing work.
+- A completed source-chain response removes ordinary work. A bidirectional response first moves work to target-execution watching.
+- Never reuse MPC triples or pre-signatures after a protocol attempt has begun; recoverable signing failures retry in a later round.
+- The backlog is recoverable from durable checkpoints. Live stream state and target-chain watchers are reconstructed after recovery.
+
+## Request Lifecycle
+
+1. The source signing contract emits a request. The indexer converts it to a normalized chain event.
+2. During catchup, requests enter the backlog but are not sent to signing. `CatchupCompleted` re-enqueues pending-generation requests in a stable order and resumes pending publications.
+3. The signing protocol deduplicates active `SignId`s. On success, only the proposer publishes the response and the request becomes pending publication.
```
jakmeier, line 18: Not a blocker on this spec document but a more general discussion:

The word **active** carries quite some weight here. The same `SignId` can enter as a new request later and I don't think we have a way to deduplicate that. So, we could have a request entering the backlog that already has a previously generated signature published on chain. It's a bit ambiguous what state that request is in.

This gets particularly tricky for the bidirectional flow.  I'm not sure what our strategy should be here.

I feel like we need a way to reject a SignId that was already processed, even when it is not in the active set. But I don't have a good idea on how to implement that.

doc/backlog.md
```diff
@@ -0,0 +1,69 @@
...
+4. Publication is deduplicated and retried with backoff. A validated source response completes ordinary requests.
```
jakmeier, line 19: What deduplication do you mean here?

jakmeier, line 19: What deduplication do you mean here?

doc/backlog.md
```diff
@@ -0,0 +1,69 @@
...
+## Scope And Invariants
+
+- A request is identified by `SignId` and belongs to one source chain.
+- Index a request before signing it. Do not start signing before the indexer completes catchup and the request meets that chain's finality policy.
+- Store a request before enqueueing it. Duplicate observations must not create duplicate active signing work.
+- A completed source-chain response removes ordinary work. A bidirectional response first moves work to target-execution watching.
+- Never reuse MPC triples or pre-signatures after a protocol attempt has begun; recoverable signing failures retry in a later round.
+- The backlog is recoverable from durable checkpoints. Live stream state and target-chain watchers are reconstructed after recovery.
+
+## Request Lifecycle
```
jakmeier, line 14: The lifecycle of a request seems to be a hot topic right now, specifically for bidirectional flows. I heard @DavidM-D is working on writing a document on in and @yvonneanne is also looking into it. You three should probably coordinate. :)

@ChaoticTempest has a great summary here for what our code does *today*. 

@yvonneanne also has a section on this in https://github.com/sig-net/mpc/pull/1130 describing it from a different angle.

Both descriptions seem to assume `SignId` is globally unique, but that's not really the case. We might need yet another perspective and define what really identifies a unique request from a user perspective. If it is `SignId`, then we need to define what we do if we see an already served `SignId` again later.

jakmeier (approved):
Nice to see this in a spec!

I like the focused document on just the backlog. 

https://github.com/sig-net/mpc/pull/1130 has some overlap but I think it serves a different purpose.

Someone at some point should probably clean up the files under `doc/` and come up with a coherent strategy on how we write documentation on different levels of abstraction. Otherwise, we just end up with a bunch of unorganized documents that are hard to maintain and they will quickly lose their value.

doc/backlog.md
```diff
@@ -0,0 +1,69 @@
+# Backlog
+
+The backlog is the node's per-source-chain record of signing work that has been indexed but is not yet terminal. It makes indexing, signing, publishing, and restart recovery converge on the same ordered chain history.
+
+## Scope And Invariants
+
+- A request is identified by `SignId` and belongs to one source chain.
+- Index a request before signing it. Do not start signing before the indexer completes catchup and the request meets that chain's finality policy.
```
yvonneanne, line 8: afaics in the current code base, this doesn't hold for Hydration, which hardcodes ctx.caught_up = true. Is that on purpose?

Also "that chain's finality policy" quietly covers Solana's confirmed, which is not really finality. 
Actually, what speaks agains moving from 1 confirmation to real finality for Solana?

doc/backlog.md
```diff
@@ -0,0 +1,69 @@
...
+- Store a request before enqueueing it. Duplicate observations must not create duplicate active signing work.
+- A completed source-chain response removes ordinary work. A bidirectional response first moves work to target-execution watching.
+- Never reuse MPC triples or pre-signatures after a protocol attempt has begun; recoverable signing failures retry in a later round.
+- The backlog is recoverable from durable checkpoints. Live stream state and target-chain watchers are reconstructed after recovery.
```
yvonneanne, line 12: Does this hold for all supported chains?

doc/backlog.md
yvonneanne: We should absolutely coordinate, this seems to be a crucial topic indeed!

iiuc, if a node sees a SignId for which a response has been published already, then it must not engage in another effort to create a signature.

does everyone agree with that statement? if not, how should we phrase it instead? where does it impact today's code?

doc/backlog.md
yvonneanne: Are you worried about the size of the list of processed SignIds? Or what makes this hard here?

at 32Bytes per SignId, 1 request per second we could store all SignIDs of one year in <1GB, can't we?

doc/backlog.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Request Lifecycle
+
+1. The source signing contract emits a request. The indexer converts it to a normalized chain event.
+2. During catchup, requests enter the backlog but are not sent to signing. `CatchupCompleted` re-enqueues pending-generation requests in a stable order and resumes pending publications.
+3. The signing protocol deduplicates active `SignId`s. On success, only the proposer publishes the response and the request becomes pending publication.
+4. Publication is deduplicated and retried with backoff. A validated source response completes ordinary requests.
+5. A bidirectional request instead progresses through: `PendingGeneration -> PendingPublish -> PendingExecution -> PendingGenerationBidirectional -> PendingPublishBidirectional -> complete`. Target execution creates the final response-signing request.
+
+An empty bidirectional serialized transaction is rejected before storage. Target-chain decoding and transaction validation happen later, while processing the source response. Duplicate target execution confirmations are intended to be idempotent.
```
yvonneanne, line 22: Does the second sentence have anything to do with the first or is it supposed to be completly independent?
Regarding the idempotency, do I understand correctly that this means that it is ok for mpc nodes to sign the execution of a transaction twice and publish both on the source chain? So it's the receiving contract's problem to avoid acting on them? iiuc, that's exactly on of the things @DavidM-D wants to prevent?
 
Regarding the phrasing, is there a gap in the implementation today or is there another reason to keep the intention in the text?

doc/backlog.md
```diff
@@ -0,0 +1,69 @@
...
+
+## Checkpoints
+
+A checkpoint contains the chain, processed height, pending request recovery data, and a digest of the consensus-relevant request IDs and coarse phases. The serialized recovery data is local state and is intentionally not part of the consensus digest.
+
+- Create checkpoints only after catchup, when an observed height crosses a configured interval bucket. Sparse block/slot/offset observations are valid.
```
yvonneanne, line 28: what do you mean by sparse here?

doc/backlog.md
```diff
@@ -0,0 +1,69 @@
...
+- Persist the checkpoint as pending, then vote for its digest through governance.
+- After threshold confirmation, atomically promote it to the latest checkpoint and prune older pending checkpoints.
+- Retain at most 32 unconfirmed checkpoints per chain. At the limit, stop consuming stream events for that chain so bounded backpressure reaches the indexer rather than losing work.
+- On restart, restore the newest durable pending checkpoint when present, otherwise the latest confirmed checkpoint, then align it with governance.
```
yvonneanne, line 32: what exactly do you mean by "align it with governance"? check if there is a newer confirmed checkpoint and use that. otherwise check if you already voted for your pending checkpoint?

doc/backlog.md
```diff
@@ -0,0 +1,69 @@
...
+Default intervals are Ethereum 20 blocks, Solana 120 slots, Hydration 240 blocks, and Canton 50 offsets. NEAR and Bitcoin currently are not implemented.
+
+## Regression And Recovery
+
+A regression is a local checkpoint that does not match governance consensus.
+
+The intended recovery is:
+
+1. Abort source-chain signing tasks and checkpoint voting.
+2. Stop the indexer and obtain the consensus checkpoint from a peer by digest.
```
yvonneanne, line 45:
```suggestion
2. Stop the indexer and obtain the newest checkpoint with threshold confirmation from a peer by digest.
```

is that what you mean by consensus checkpoint?

doc/backlog.md
```diff
@@ -0,0 +1,69 @@
...
+3. Replace local durable and in-memory backlog state, rebuild execution watchers, and catch up from the restored height.
+
+The supervisor also restarts an indexer after an error or chain-specific stream stall. When governance has no checkpoint, no alignment occurs. When governance does have a checkpoint that is absent locally, including fresh bootstrap, the node fetches a digest-matching checkpoint from a peer and regresses to it.
```
yvonneanne, line 48: is backwards regression intended? And does "the intended recovery is" (L42) mean implemented or planned?

yvonneanne (commented):
Great to have such a doc summarizing what the code is supposed to do in easy to understand sentences! This is very helpful and much appreciated!

I added some comments where I'm probably missing context or the text is ambiguous or different from the code, afaics.

doc/backlog.md
```diff
@@ -0,0 +1,69 @@
...
+
```
yvonneanne, line 49: add a sentence about checking the contents of the delivered checkpoint wrt to the digest before applying it?

doc/backlog.md
volovyks: Related: https://github.com/sig-net/mpc/pull/1169

doc/backlog.md
```diff
@@ -0,0 +1,69 @@
...
+The backlog is the node's per-source-chain record of signing work that has been indexed but is not yet terminal. It makes indexing, signing, publishing, and restart recovery converge on the same ordered chain history.
+
+## Scope And Invariants
+
+- A request is identified by `SignId` and belongs to one source chain.
+- Index a request before signing it. Do not start signing before the indexer completes catchup and the request meets that chain's finality policy.
+- Store a request before enqueueing it. Duplicate observations must not create duplicate active signing work.
+- A completed source-chain response removes ordinary work. A bidirectional response first moves work to target-execution watching.
+- Never reuse MPC triples or pre-signatures after a protocol attempt has begun; recoverable signing failures retry in a later round.
```
volovyks, line 11: I would say it belongs in the protocol spec, not in the backlog.

doc/backlog.md
```diff
@@ -0,0 +1,69 @@
...
+- The backlog is recoverable from durable checkpoints. Live stream state and target-chain watchers are reconstructed after recovery.
+
```
volovyks, line 13: This list lacks a recovery strategy in the event of a lost checkpoint (e.g., starting from the genesis block or from a provided block during a checkpoint reset).

doc/backlog.md
volovyks: I would move this discussion to the protocols spec. In the context of the backlog, the only thing that is required is consistency between different nodes. Now the request with the same Id would be ignored (deduplicated) if it is still in the backlog. If it is not, it would be added and processed again. @ChaoticTempest tell me if I'm wrong.

doc/backlog.md
volovyks, line 18: This can also be dropped from the backlog spec.

doc/backlog.md
```diff
@@ -0,0 +1,69 @@
...
+3. The signing protocol deduplicates active `SignId`s. On success, only the proposer publishes the response and the request becomes pending publication.
+4. Publication is deduplicated and retried with backoff. A validated source response completes ordinary requests.
+5. A bidirectional request instead progresses through: `PendingGeneration -> PendingPublish -> PendingExecution -> PendingGenerationBidirectional -> PendingPublishBidirectional -> complete`. Target execution creates the final response-signing request.
+
+An empty bidirectional serialized transaction is rejected before storage. Target-chain decoding and transaction validation happen later, while processing the source response. Duplicate target execution confirmations are intended to be idempotent.
+
+## Checkpoints
+
+A checkpoint contains the chain, processed height, pending request recovery data, and a digest of the consensus-relevant request IDs and coarse phases. The serialized recovery data is local state and is intentionally not part of the consensus digest.
```
volovyks, line 26: Should we add a starting block?
Not a starting block for this checkpoint, but a starting block overall.

doc/backlog.md
```diff
@@ -0,0 +1,69 @@
...
+
+- Create checkpoints only after catchup, when an observed height crosses a configured interval bucket. Sparse block/slot/offset observations are valid.
+- Persist the checkpoint as pending, then vote for its digest through governance.
+- After threshold confirmation, atomically promote it to the latest checkpoint and prune older pending checkpoints.
+- Retain at most 32 unconfirmed checkpoints per chain. At the limit, stop consuming stream events for that chain so bounded backpressure reaches the indexer rather than losing work.
```
volovyks, line 31: 32 to the future from `latest`?

doc/backlog.md
```diff
@@ -0,0 +1,69 @@
...
+- On restart, restore the newest durable pending checkpoint when present, otherwise the latest confirmed checkpoint, then align it with governance.
+
+If pending-checkpoint storage fails, the in-memory processed height has already advanced and the current implementation does not retry that interval. A later interval crossing can checkpoint again; a crash before then replays from the last durable checkpoint. This is an accepted current behavior, not a settled durability policy.
```
volovyks, line 34: I do not understand this paragraph.

doc/backlog.md
volovyks, line 48: We need to add a section on recovery in case all the checkpoints are lost (reset at the contract level).

doc/backlog.md
```diff
@@ -0,0 +1,69 @@
...
+2. Stop the indexer and obtain the consensus checkpoint from a peer by digest.
+3. Replace local durable and in-memory backlog state, rebuild execution watchers, and catch up from the restored height.
+
+The supervisor also restarts an indexer after an error or chain-specific stream stall. When governance has no checkpoint, no alignment occurs. When governance does have a checkpoint that is absent locally, including fresh bootstrap, the node fetches a digest-matching checkpoint from a peer and regresses to it.
+
+## Chains And Indexers
+
+| Chain | Current behavior | Backlog status |
+| --- | --- | --- |
+| Ethereum | Finalized-head indexing by default; optimistic mode is development/test only. Ordinary requests and responses are indexed. | Supervised, checkpointed. First catchup without a checkpoint starts at the startup anchor, not a deployment/genesis height. |
```
volovyks, line 54: Where is the startup anchor refined?

doc/backlog.md
```diff
@@ -0,0 +1,69 @@
...
+
+The intended recovery is:
+
+1. Abort source-chain signing tasks and checkpoint voting.
+2. Stop the indexer and obtain the consensus checkpoint from a peer by digest.
+3. Replace local durable and in-memory backlog state, rebuild execution watchers, and catch up from the restored height.
+
+The supervisor also restarts an indexer after an error or chain-specific stream stall. When governance has no checkpoint, no alignment occurs. When governance does have a checkpoint that is absent locally, including fresh bootstrap, the node fetches a digest-matching checkpoint from a peer and regresses to it.
+
+## Chains And Indexers
```
volovyks, line 50: I think there should be no differences between chains when defining backlog properties, since they all use the same interface. What can be defined instead are the expectations for chain implementations from the backlog’s point of view.

doc/backlog.md
```diff
@@ -0,0 +1,69 @@
...
+
+| Chain | Current behavior | Backlog status |
+| --- | --- | --- |
+| Ethereum | Finalized-head indexing by default; optimistic mode is development/test only. Ordinary requests and responses are indexed. | Supervised, checkpointed. First catchup without a checkpoint starts at the startup anchor, not a deployment/genesis height. |
+| Solana | Historical catchup plus a pre-started websocket live stream; uses `confirmed` commitment. Ordinary and bidirectional requests are indexed. | Supervised, checkpointed. Live buffering can overflow during slow catchup. |
```
volovyks, line 55: Websocket info is outdated now: https://github.com/sig-net/mpc/pull/1169

doc/backlog.md
```diff
@@ -0,0 +1,69 @@
...
+| Hydration | Subscribes to finalized blocks and verifies event-storage proofs before parsing requests and responses. | Does not use the common stream/checkpoint event pipeline. Disconnect-window backfill and checkpoint advancement are open. |
+| NEAR | Polls contract pending requests with an in-memory, one-hour seen cache. | Integration-test-only path; bypasses backlog, checkpoint, and regression supervision. |
+| Bitcoin, Midnight | Chain identifiers/configuration exist. | TBD |
+
+Ethereum is the only implemented target-execution watcher found. Bidirectional
+execution monitoring for Solana, Canton, and Hydration is open.
+
+## Failure Rules
+
+- Indexer and transient RPC failures retry or cause supervised restart; Recognized terminal HTTP client errors (400, 401, 403, 404, 405) are not retried; 408, 429, timeouts, and other errors are retried according to the RPC policy.
```
volovyks, line 66: Should be a part of the indexer specification.

doc/backlog.md
```diff
@@ -0,0 +1,69 @@
...
+- Publication must survive restart while a request is pending publication.
+- A stalled checkpoint consensus intentionally backpressures that chain. Its operational behavior must be monitored because it can also trigger watchdog restart loops.
```
volovyks, line 68: Can you please elaborate on this one?

doc/backlog.md
```diff
@@ -0,0 +1,69 @@
...
+- Do not treat provider data as cryptographically verified unless the indexer verifies it. Hydration currently verifies a state proof; the other indexers rely on their providers.
```
volovyks, line 69: Not sure I understand this either. But seams like an indexer thing.
Any info that is reaching the backlog must be final and verified.

doc/backlog.md
jakmeier: I think you are right @yvonneanne, if we really want to deduplicate these, we could simply keep all SignId stored for like a year. A bloom filter can make it quick to check for duplicates. That settles some of my implementation concerns, assuming we want `SignId` based deduplication.

The open questions are, whether we need/want this deduplication and if the current `SignId` is the correct choice for it.

For example, with bidirectional requests, I think there are legitimate use case to request the same signature more than once. Maybe the user had the `respond_serialization_schema` wrong in the initial request, so they send it again with a correction, hoping to get the correct response deliver to the source chain. If we deduplicate the request based on `SignId`, that wouldn't work, since `respond_serialization_schema` doesn't affect it right now.

PR #1134, opened 2026-08-14
test(midnight): add real-stack response lifecycle

Pessina:
Adds a live Midnight response-lifecycle test on top of #1113 and runs it for every pull request.

- Starts pinned Midnight node, indexer, and proof-server services.
- Exercises a Midnight → Ethereum → Midnight request and response lifecycle.
- Uses the real Rust publisher and compiled TypeScript child process.
- Verifies both response circuits, signed EVM execution, caller cleanup, and finalized checkpoints.

Stack: #1129 → #1142 → #1143 → #1113 → #1134


chain-signatures/midnight-publisher-ts/devtools/real-stack/caller.compact
```diff
@@ -0,0 +1,110 @@
+// Test-only real caller for the Rust-owned integration stack. Its structure is the
+// minimal subset of sig-net/midnight-integration's v0.15.0 test caller (commit
+// b270880f71a229ae0724c60d32d8215aeb0f59ed), using the same Signet.compact API.
+pragma language_version >= 0.25;
```
BRBussy, line 4: Perhaps we can publish @sig-net/midnight-contract-test-caller and @sig-net/midnight-contract-test-caller-deploy to avoid this duplication and minimise drift on the stack. Could be a follow up issue.

ppca (approved):

.github/workflows/midnight.yml
```diff
@@ -0,0 +1,129 @@
+name: Midnight Integration Tests
+
+on:
+  push:
+    branches:
+      - develop
+    paths:
```
BRBussy, line 7: Since these tests are really long, might be worth making a more detailed filter here to prevent run on minimal unrelated changes - optional.

integration-tests/Cargo.toml
```diff
@@ -73,6 +73,7 @@ mpc-chain-canton.workspace = true
 mpc-chain-solana.workspace = true
 mpc-chain-ethereum = { workspace = true, features = ["contract-test"] }
 mpc-chain-near.workspace = true
+mpc-chain-midnight = { workspace = true, features = ["sandbox"] }
```
BRBussy, line 76: Update integration testing readme integration-tests/README.md to have documentation of the new midnight tests + how to run them

Pessina: Added: https://github.com/sig-net/mpc/pull/1134/commits/761fcb646c33efff6744fd3a3da6e384e08efe1c

chain-signatures/midnight-publisher-ts/devtools/real-stack/caller.compact
Pessina: Agree, let's do it on a follow up issue

.github/workflows/midnight.yml
Pessina: I've run some benchmarks here with GPT and with filtering we would be skipping ~11-12% of the PRs. I'm afraid about adding fragile filtering rules and skip genuine necessary runs. 

I know it's slow, but I would prefer to run it on every PR, as other chains, and be sure that it's green

isSerge: You can leave a TODO comment, I will do it for all workflows https://github.com/sig-net/mpc/issues/1165

ChaoticTempest (approved):
might need to pull latest develop to fix that migration unit test

.github/workflows/midnight.yml
Pessina: Added: https://github.com/sig-net/mpc/pull/1134/commits/eb61348f41dfb285ad382a034d2a104453c51603

PR #1135, opened 2026-08-14
fix: wait on latest ethereum final block instead of latest optimistic for anchor

### LINKED ISSUE #1132: Sign request stuck in reorganize loop (dev cluster)
Request ID: `0a7346cfa924cc0f9c48e7620dc89fbf4188dd3211c14334257f7292ca0c6f0e` (Ethereum)
Received: 2026-08-12 13:47:14 UTC
Still unresolved after 2 hrs and 40+ reorganize rounds.

The request can't complete because the network cannot reach 9/12 threshold. Every proposer stalls at 8 accepts.

### LINKED ISSUE #1136: eth: indexed requests not go into sign queue
Image
Image
Image

ChaoticTempest:
Fixes #1132.

This fixes the ethereum catch-up, so it anchors at the first initialized finalized head instead of the latest unfinalized tip.

The indexer previously:
1. Sampled the anchor with latest tip at startup.
2. Used latest + 1 as the catch-up boundary.
3. Waited for the finalized head before processing each block.

Because Ethereum finality trails the tip by roughly two epochs, catch-up could wait 6–13 minutes for its startup boundary to become processable. If the indexer restarted during that period (i.e. from new consensus checkpoint), it sampled a newer tip and moved the boundary forward again.

This could prevent CatchupCompleted from being emitted and keep restored requests from entering the signing pipeline for some nodes.

So this PR just makes it so we get the anchor from the latest finalized instead of the latest optimistic

### DIFF
--- a/chain-signatures/chain-ethereum/src/indexer.rs
+++ b/chain-signatures/chain-ethereum/src/indexer.rs
@@ -2,7 +2,7 @@ use crate::abi::ChainSignatures;
 use crate::client::{block_may_contain_logs, CatchupItem, CatchupIter, EthereumClient};
 use crate::event_parsing::{emit_respond_events, parse_filtered_logs};
 use crate::execution_watcher::{ExecutionWatcher, WatcherGateState};
-use crate::finalized_head::FinalizedHeadTracker;
+use crate::finalized_head::{FinalizedHeadTracker, FinalizedHeadWatcher};
 use crate::EthConfig;
 use alloy::primitives::Address;
 use alloy::rpc::types::{Block, Log};
@@ -12,7 +12,7 @@ use async_trait::async_trait;
 use futures_util::{stream, Stream};
 use mpc_chain_integration_core::{ChainIndexer, ChainTelemetry, StateManager};
 use mpc_primitives::{Chain, ChainEvent, IndexedSignRequest};
-use mpc_utils::task::{retry_until_ok, retry_until_some, AbortOnDrop};
+use mpc_utils::task::retry_until_ok;
 use std::sync::{Arc, Mutex};
 use tokio::sync::mpsc;
 use tokio::time::Duration;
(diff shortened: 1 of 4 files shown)


isSerge:
I just realized that `TestIndexerBuilder` is built with `optimistic_requests: true` and `RunFixture` is still polling `Latest`. Tests would have caught the original issue if I haven't overlooked this before. I think it would make sense to adjust these in the context of this PR. I can submit a PR or push changes to this branch. What do you think?

chain-signatures/chain-ethereum/src/indexer.rs
```diff
@@ -315,32 +315,39 @@ impl<S: StateManager, T: ChainTelemetry> EthereumIndexer<S, T> {
...
+        cancel: &CancellationToken,
+    ) -> anyhow::Result<Option<u64>> {
+        tokio::select! {
+            _ = cancel.cancelled() => Ok(None),
+            head = watcher.wait_initialized() => match head {
+                Ok(head) => Ok(Some(head.saturating_add(1))),
+                Err(_err) if cancel.is_cancelled() => Ok(None),
+                Err(err) => Err(anyhow::anyhow!("head watcher exited: {err:?}")),
+            },
+        }
```
isSerge: This would work too, smaller change - just fetch `finalized` instead of latest block number:

```suggestion
        retry_until_some(
            cancel,
            Self::RETRY_DELAY,
            "ethereum finalized block",
            || async {
                let block = self
                    .client
                    .get_block(BlockId::Number(BlockNumberOrTag::Finalized))
                    .await?;
                Ok(block.map(|block| block.header.number))
            },
        )
        .await
        .map(|finalized| finalized.saturating_add(1))
```

isSerge (commented):
Fix makes sense, but having both `FinalizedHeadTracker` and `FinalizedHeadWatcher` is confusing. I think we can merge these.

chain-signatures/chain-ethereum/src/indexer.rs
ChaoticTempest: simplified it

isSerge (approved):

PR #1142, opened 2026-08-17
feat(chain-midnight): add Rust intent protocol client

Pessina:
Adds the Rust half of the versioned stdio publisher protocol on top of #1129, without registering a chain publisher yet.

- `IntentGen` drives ready/build/submit over the line-delimited JSON protocol: explicit v1 handshake, JS-safe request ids, one persistent child replaced on any doubt (one respawn for build, never for submit; ambiguous submits are typed for reconciliation), and deadline ordering (recipe TTL < child submit timeout < Rust backstop).
- Rust owns publisher config: `PublisherConfig` with validation (seed, URLs, network id, argv) and a seed-redacting `Debug`; five new `--midnight-*` flags / `MPC_MIDNIGHT_*` env vars, all-or-nothing with `--midnight-node-ws-url` (indexer-only Midnight config is no longer accepted); TS `config.ts` becomes a plain env→SDK map.
- Adds the ignored Rust↔TS seam tests (real client drives `dist/main.js`, decoded with the Rust ledger crate), a `midnight-publisher-runtime` image stage with a ready-handshake smoke, and runs both in the midnight-publisher-ts workflow. Pins the midnight-ledger dev-deps and moves the node-builder image to `rust:1.93-bookworm`.

Stack: #1129 → this PR → #1143 → #1113 → #1134.


chain-signatures/chain-midnight/src/intent_gen.rs
```diff
@@ -0,0 +1,1485 @@
...
+pub struct IntentGen {
+    config: MidnightConfig,
+    network_id: String,
+    /// Two interleaved requests on one pipe have no correct reading, so the lock is the
+    /// concurrency story: callers queue.
+    session: Mutex<Option<Session>>,
+}
+
+impl IntentGen {
+    /// Starts the builder eagerly: a command that cannot run fails here, not on the first signature.
```
ppca: what command are you referring to?

chain-signatures/chain-midnight/src/intent_gen.rs
```diff
@@ -0,0 +1,1485 @@
...
+}
+
+/// Starts the builder under the restart budget.
+async fn spawn_session(config: &MidnightConfig, network_id: &str) -> anyhow::Result<Session> {
+    retry_rpc!(
+        config.publisher.request_timeout,
+        config.publisher.restart_backoff,
+        "midnight_intent_gen_spawn",
+        {
+            let session = spawn_child(config, network_id)?;
```
ppca, line 368: nit: some deterministic stuff like intent_gen_command will also be retried many times. But no correctness issue, and it's easiest to write correctly this way

chain-signatures/node/src/cli/args/midnight.rs
```diff
@@ -18,35 +26,113 @@ pub struct MidnightArgs {
...
-        let mut args = Vec::with_capacity(4);
+        let mut args = Vec::with_capacity(14);
         if let Some(v) = self.midnight_node_ws_url {
             args.extend(["--midnight-node-ws-url".to_string(), v]);
         }
         if let Some(v) = self.midnight_central_address {
             args.extend(["--midnight-central-address".to_string(), v]);
         }
+        if let Some(v) = self.midnight_funding_seed {
+            args.extend(["--midnight-funding-seed".to_string(), v]);
```
ppca: do we need to refrain from printing this directly?

chain-signatures/chain-midnight/src/intent_gen.rs
```diff
@@ -0,0 +1,1485 @@
...
+                anyhow::anyhow!("midnight {} timed out after {deadline:?}", operation.name()),
+                operation,
+            ),
+        },
+    };
+    match decode_reply(&reply, id) {
+        // The child's verdict passes through; an unpairable reply leaves the doubt a dead pipe does.
+        answered @ Exchange::Answered(_) => answered,
+        Exchange::Broken { error, retry } => Exchange::Broken {
+            error: unanswered(error, operation),
```
ppca: unanswered showing up 3 times, we can probably refactor:
```
 async fn exchange<T: Serialize>(
      session: &mut Session,
      request: &T,
      operation: Operation,
      config: &PublisherConfig,
      cancel: &CancellationToken,
  ) -> Exchange {
      let id = take_wire_id(&mut session.next_id);

      let line = match encode_request(request, id, operation) {
          Ok(line) => line,
          // Definitely pre-write, so do not mark a submit ambiguous.
          Err(error) => return Exchange::Answered(Err(error)),
      };

      let outcome =
          exchange_inner(session, &line, id, operation, config, cancel).await;

      // Once encoding succeeded and exchange_inner started, a submit may have
      // reached the child. Apply that policy in one place.
      match outcome {
          Exchange::Broken { error, retry } => Exchange::Broken {
              error: unanswered(error, operation),
              retry,
          },
          answered => answered,
      }
  }

  async fn exchange_inner(
      session: &mut Session,
      line: &str,
      id: u64,
      operation: Operation,
      config: &PublisherConfig,
      cancel: &CancellationToken,
  ) -> Exchange {
      let deadline = operation.deadline(config);

      let reply = tokio::select! {
          _ = cancel.cancelled() => {
              return Exchange::give_up(anyhow::anyhow!(
                  "midnight {} cancelled",
                  operation.name(),
              ));
          }

          result = tokio::time::timeout(deadline, round_trip(session, line)) => {
              match result {
                  Ok(Ok(reply)) => reply,

                  Ok(Err(error)) => {
                      return Exchange::lost(error, operation);
                  }

                  Err(_) => {
                      return Exchange::lost(
                          anyhow::anyhow!(
                              "midnight {} timed out after {deadline:?}",
                              operation.name(),
                          ),
                          operation,
                      );
                  }
              }
          }
      };

      decode_reply(&reply, id)
  }

  Exchange::lost() would no longer call unanswered():

  fn lost(error: anyhow::Error, operation: Operation) -> Self {
      Self::Broken {
          error,
          retry: operation.is_repeatable(),
      }
  }
```

Dockerfile
```diff
@@ -7,7 +7,34 @@ RUN npm install
...
+      MIDNIGHT_PUB_PROOF_SERVER_URL=http://127.0.0.1:6300 \
+      MIDNIGHT_PUB_INDEXER_URL=http://127.0.0.1:8088/api/v3/graphql \
+      MIDNIGHT_PUB_INDEXER_WS_URL=ws://127.0.0.1:8088/api/v3/graphql/ws \
+      MIDNIGHT_PUB_FUNDING_SEED=abababababababababababababababababababababababababababababababab \
+      midnight-publisher 2>/dev/null)" = \
+      '{"id":0,"ok":true,"ready":true,"protocolVersion":1,"submitTimeoutMs":360000,"recipeTtlMs":300000}'
+
+FROM rust:1.93-bookworm AS node-builder
 RUN rustc --version --verbose
 WORKDIR /usr/src/app
```
ppca, line 39: noticed the publisher is not copied into final runtime with node, is there a reason why?

chain-signatures/chain-midnight/src/intent_gen.rs
isSerge, line 1: Not a blocker, but would nice to split this into couple of modules: wire codec, child, etc.

Cargo.toml
```diff
@@ -87,6 +87,8 @@ midnight-base-crypto = "1.1.0"
 midnight-onchain-state = "4.0.0"
 midnight-serialize = "1.2.0"
 midnight-storage = "2.0.1"
+midnight-transient-crypto = "3.0.0"
+midnight-ledger-v9 = { version = "1.0.0", default-features = false }
```
isSerge: these two only used in `chain-midnight` as dev dependencies, do you think any other workspace crate might depend on these in the future?

isSerge (commented):
Two minor comments

Cargo.toml
Pessina: No, I've checked my next PRs here. Updated: https://github.com/sig-net/mpc/pull/1142/commits/128f4a24450669ed3aef45d69d9692144fafbeb5

chain-signatures/chain-midnight/src/intent_gen.rs
Pessina: I like the idea, but I would propose to defer it a bit as we need to finish this release by end of Aug.

chain-signatures/node/src/cli/args/midnight.rs
Pessina: Good point, fixed in: https://github.com/sig-net/mpc/pull/1142/commits/128f4a24450669ed3aef45d69d9692144fafbeb5

chain-signatures/chain-midnight/src/intent_gen.rs
Pessina: Good point, fixed in: https://github.com/sig-net/mpc/pull/1142/commits/128f4a24450669ed3aef45d69d9692144fafbeb5

Dockerfile
Pessina: Yeah, it's coming on a future PR. For now we only need it on the tests, so I didn't added here.

chain-signatures/chain-midnight/src/intent_gen.rs
Pessina: Agree, but I would leave as is. Adding the branching to fail-fast will add extra lines for a small benefit. 

The core fail fast it's already validated on the config validation. What do you think?

isSerge (approved):

ppca (approved):

PR #1143, opened 2026-08-17
feat(chain-midnight): index response lifecycle

Pessina:
Adds finalized response indexing on top of #1142, without introducing publisher or node wiring.

- Decodes supported response entries from both central response maps.
- Diffs finalized state and emits Respond and RespondBidirectional lifecycle events before each block marker.
- Keeps malformed entries isolated and pins the ordering and replay behavior with deterministic tests.

Stack: #1129 → #1142 → this PR → #1113 → #1134.


chain-signatures/chain-midnight/src/indexer.rs
```diff
@@ -57,11 +86,11 @@ impl<S: StateManager, T: ChainTelemetry> MidnightIndexer<S, T> {
...
-fn unix_now() -> u64 {
-    SystemTime::now()
+pub(crate) fn unix_now() -> anyhow::Result<u64> {
+    Ok(SystemTime::now()
         .duration_since(UNIX_EPOCH)
-        .map(|d| d.as_secs())
-        .unwrap_or(0)
+        .context("the system clock is before the unix epoch")?
+        .as_secs())
 }
```
isSerge: We now have `current_unix_timestamp` helper in `mpc_utils::time`

isSerge (approved):

chain-signatures/chain-midnight/src/indexer.rs
Pessina: Replaced: https://github.com/sig-net/mpc/pull/1143/commits/20d0317646ce7196d404d0c2acb9afe992092a39

PR #1144, opened 2026-08-17
refactor(EthClient): enhance batch processing and flush logic 

isSerge:
This PR addresses some of the issues in `run_batch_respond` (publishing flow): 

- Idle idle polling every 100 ms
- Task leak on shutdown
- Structural cap of one action per tick (~10/sec) (not critical at the moment, but would become an issue if load increases)

Main changes:

- Extract `BatchPublisher` from `EthClient` (now a thin wrapper). `BatchPublisher` holds all publishing machinery. `EthClient` builds and spawns `BatchPublisher` on initialization. This helps with batch flushing on shutdown (spawned task owns no sender)
- Replace 100ms polling with a `tokio::select!` event loop
- Use `actions_rx.recv_many(...)`, which parks on the channel while idle
- Introduce flush deadline - partial batch flushes exactly `batch_flush_interval` after first action
- Added unit tests:
  - `test_run_batch_respond_flushes_full_batches_immediately` - verify full batch is flushed immediately
  - `test_run_batch_respond_flushes_partial_batch_after_interval` - verify partial batch is flushed after `batch_flush_interval`
  - `test_run_batch_respond_flushes_on_channel_close` - verify leftover batch is flushed on channel close

### DIFF
--- a/chain-signatures/chain-ethereum/src/publisher.rs
+++ b/chain-signatures/chain-ethereum/src/publisher.rs
@@ -49,43 +49,56 @@ impl From<&Signature> for ChainSignatures::Signature {
 /// This client is separate from the client used by the indexer (separation of concerns: read vs write).
 #[derive(Clone)]
 pub struct EthClient {
-    /// The contract instance for interacting with the ChainSignatures contract
-    contract: ChainSignatures::ChainSignaturesInstance<EthContractFillProvider>,
     /// Channel used to send actions to the background batching task
     batch_tx: mpsc::Sender<PublishAction>,
+}
+
+/// Publishing machinery for the background batching task.
+///
+/// Kept separate from `EthClient` so the spawned task owns no channel sender —
+/// otherwise the channel could never close and the batching loop could never
+/// shut down. Once every `EthClient` clone is dropped, the channel closes and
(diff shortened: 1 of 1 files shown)


chain-signatures/chain-ethereum/src/publisher.rs
```diff
@@ -49,79 +49,122 @@ impl From<&Signature> for ChainSignatures::Signature {
...
+        );
         let signatures: HashMap<SignId, Signature> = actions
             .iter()
             .map(|action| (action.request.id, action.signature))
             .collect();
 
         let res = retry_rpc!(
             Duration::MAX, // Prevent from timing out
-            self.publisher.batch_publish_retry,
+            self.config.batch_publish_retry,
```
yvonneanne, line 167: if one request in a batch reverts (say another proposer already responded), does this loop retry that same batch forever and never publish anything else?

isSerge: yes, it will retry forever the same batch, but I think currently both proposers publishing the same request will succeed, there is no revert, duplicate event is absorbed, some gas is wasted - probably can address this separately

ppca (approved):

Issue #1145, opened 2026-08-18
Improve memory usage for very large backlogs


ChaoticTempest:
## Issue

Right now, we have large unnecessary memory usage for large backlogs (+40k requests). We need to improve this such that we can have as many in the backlog as possible.

The memory footprint here is due in part to a couple components:
1. backlog requests are mirrored on the signature spawner side as sign tasks.
2. pending checkpoints create duplicate sign requests across multiple checkpoints.

We currently have issues with catch up leading to an OOM issue due to the bulk requeuing of sign requests. So 2 isn't what is causing this issue, but more so 1. 

### Possible Solutions

For 1, in the short term, we can try and limit the sign requests from being queued up for the tokio task until we actually need it to be queued. But this would likely only remedy it a little, since we are still cloning the entire sign request to be sent over to the `SignatureSpawner`. So longer term, we need to refactor the handling of backlog requests and the signature spawner itself.

This is all happening on our very low memory devnet machines anyways which had ~1.2gb of ram. Bumping it remedied it, so I recommend we go the longer term pursuit.

So for 1, the longer term approach here is to refactor the `Backlog` such that we have the `SignatureSpawner` react to changes on the backlog itself, and only get a reference to the backlog's request (like via an `Arc`). This gives us the opportunity to treat the backlog as the single source of sign requests completely, with no duplicated sign request state. This way the `SignatureSpawner` merely only has to spawn tasks, where the task handle is stored directly in the backlog request itself. This has the added advantage of getting rid of abort logic in the `SignatureSpawner` itself because on a `respond` event, we simply just drop the request which would auto-abort the task.

So no duplicate requests, and only spawn sign tasks on getting a posit message

ChaoticTempest:
2 is also important since there's a huge amount of memory it duplicates when we have over a million requests sitting idle. Each pending checkpoint can have the set of requests from the previous pending checkpoint potentially. This balloons when we start to account for the pending checkpoint limit.

I suggest we create a sign request pool for which each pending checkpoint can reference (either an Arc or directly via the hash of a SignId). And then on serialization, we just need to reference the specific request. The same can be done for the storage of the requests after https://github.com/sig-net/mpc/pull/1127 lands in.

ppca:
for 1 agree that a lazy and bounded way to spawn tasks is the key. Using backlog as sole source of spawner is reasonable, it is also cleaner to debug honestly. 
[...]

PR #1151, opened 2026-08-19
feat: reduce cloning of IndexedSignRequest

ChaoticTempest:
This is an easy win and first step for https://github.com/sig-net/mpc/issues/1145 but doesn't fix it completely. This just makes it so we don't clone the `IndexedSignRequest` around too much by wrapping it in an `Arc`.

---

Memory savings:

| Request Type | Previous (4–5 Owned Copies) | Current (`Arc<IndexedSignRequest>`) | **Net Memory Saved** |
| :--- | :--- | :--- | :--- |
| **100k Sign Requests** | `100,000 × 5 × 475 B` = **`~237 MB`** | `100,000 × (475 B + 4 × 8 B)` = **`~50 MB`** | **`~187 MB` (~79% reduction)** |
| **100k Bidirectional Requests** *(e.g. ~1 KB tx payload)* | `100,000 × 5 × 1.5 KB` = **`~750 MB`** | `100,000 × 1.5 KB` = **`~150 MB`** | **`~600 MB` (~80% reduction)** |

### DIFF
--- a/chain-signatures/node/src/backlog/mod.rs
+++ b/chain-signatures/node/src/backlog/mod.rs
@@ -281,11 +281,7 @@ impl Backlog {
     }
 
     /// Insert a new Sign request into the backlog for the specified chain.
-    pub async fn insert(
-        &self,
-        request: impl Into<Arc<IndexedSignRequest>>,
-    ) -> Option<BacklogEntry> {
-        let request: Arc<IndexedSignRequest> = request.into();
+    pub async fn insert(&self, request: Arc<IndexedSignRequest>) -> Option<BacklogEntry> {
         let chain = request.chain;
         let id = request.id;
         let entry = BacklogEntry::new(request);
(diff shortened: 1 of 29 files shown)


chain-signatures/node/src/backlog/mod.rs
```diff
@@ -839,23 +839,23 @@ pub enum BacklogError {
 
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
 pub struct BacklogEntry {
-    pub request: IndexedSignRequest,
+    pub request: Arc<IndexedSignRequest>,
```
isSerge, line 753: Not critical, but maybe we should explicitly enable `rc` feature for `serde` to make sure serialization with `Arc` does not break. It probably works now because some dependency enables it

chain-signatures/node/src/stream/ops.rs
isSerge, line 310: Not sure how big `pending_tx` is, but can also avoid this clone if hoist `pending_tx.source_chain`

isSerge (approved):
Two minor comments

chain-signatures/node/src/stream/ops.rs
ChaoticTempest: I'll leave this for another PR where I trace down all the other unnecessary clones

PR #1152, opened 2026-08-19
docs: posit state-machine analysis

yvonneanne:
## What

A reverse-engineered model of the signing request lifecycle in `chain-signatures/node/src/protocol/request/`, as `doc/posit_state_machine.md`:

- a top-level state machine matching the `SignPhase` enum, plus one detail diagram per phase (Organizing, Posit split into its proposer and deliberator machines, Generating), each state annotated with its numbered steps
- the round mechanism: the three-way round filter, the two round-surviving variables, and the `round_timeout(r)` schedule
- the per-node state needed to run the machines, including what survives a round bump vs a task respawn
- invariants, and observations the model surfaces (no terminal failure state, `Done` firing before on-chain confirmation and on failed reconstruction, the one-slot-per-sender buffers assuming an ordering HTTP does not provide, `pause_proposing_until` idling whole rounds)
- 48h of log evidence from devnet/testnet/mainnet backing or refuting the observations, and a short list of proposals against the round churn

## Why

The posit layer had no written model; reviewing changes to it meant re-deriving the machine from four files each time. The doc is deliberately claim-dense so it can be reviewed by attacking individual statements.


doc/posit_state_machine.md
```diff
@@ -0,0 +1,464 @@
...
+is why it exists — it waits for the node to reach `highest_seen_round`), and
+`pause_proposing_until`. A respawn keeps less: only `r` survives, via
+`SignEntry.round`; `highest_seen_round`, the buffer, and the pause all start
+empty again. The proposer's ACCEPT/REJECT tally is transient, rebuilt inside each
+`Propose sent`.
+
+Not per request: the mesh's active set (a shared `watch`) and the governance
+membership and threshold; the third election input, entropy, comes with the
+request itself. Per node across requests:
+an LRU of finished `sign_id`s, so a late posit for a done request is dropped
```
ppca: should be a bounded LRU?

doc/posit_state_machine.md
```diff
@@ -0,0 +1,464 @@
...
+instead of starting a fresh machine.
```
ppca: precisely it's not starting a fresh machine, it only adds to an inbox?

doc/posit_state_machine.md
```diff
@@ -0,0 +1,464 @@
...
+
+Sending `PROPOSE` to every member would help, cheaply: the tally is keyed
+on `SinglePositCounter::participants` and `process_action` drops senders outside
+that set, so extra replies cannot move `enough_rejects` or `meets_totality`. It
+would be a notification, not a vote.
+
+It only half-solves the problem, though. A member holding the presignature but
+absent from the active set at reservation time would `ACCEPT`, go uncounted, and
+wait for a `START` that never comes. And the excluded member still cannot tell
+"round `r` succeeded without me" from "round `r` still running", which is what it
```
ppca: is the aim of knowing "round `r` succeeded without me" that we can stop this request from holding the 1 out of 4 slot for signature generation for until respond() is finalized?

doc/posit_state_machine.md
```diff
@@ -0,0 +1,464 @@
...
+    Generating --> Done: signature produced
+
+    Organizing --> Organizing: new round
+    Posit --> Organizing: new round
+    Generating --> Organizing: new round
+
+    Done --> [*]
+```
+
+A task enters `Organizing` when `SignatureSpawner` spawns it for a newly indexed
```
ppca: not necessarily newly indexed, it could be spawned from backlog at restart or catchup

doc/posit_state_machine.md
```diff
@@ -0,0 +1,464 @@
...
+  devnet against exactly zero on testnet; the midday baseline hour had 3815
+  reorganizes across only 79 distinct sign_ids. ~83% of reorganize reasons are
+  "deliberator timeout waiting for Propose": rounds ending because no PROPOSE
+  ever arrived.
+- **The delayed watcher fired for ~3.1k distinct requests on devnet and ~0.7k
+  on testnet (§8.1);** on testnet each node logs the same delayed request, so
+  the per-node uniform count is one request counted nine times.
+- **No proposer.** Analysis 2: 71.8% of rotations are "deliberator timeout
+  waiting for Propose". Nodes differ widely in how many requests they are live
+  on, so election keeps picking nodes that are not working on the request:
```
ppca: what does "live on" refer to?

doc/posit_state_machine.md
```diff
@@ -0,0 +1,464 @@
...
+  log line. They do prove the channel is not exactly-once, which is why §8.4
+  is worth acting on, though reordering itself stays unobserved (zero REJECT
+  duplicates, START-from-non-proposer, or conflicting-proposer anywhere).
+
+
+Improvement proposals, in recommended order of attack. Requests never expire
+by design, so the rotation has to be ended by information, not time.
+
+1. **Answer posits for completed requests.** `dead_ids` silently drops them
+   today; replying REJECT Completed lets a straggler end its task on the next
```
ppca: I guess this would then assume a completed signature generation would be followed by successful publish. If this assumption can break under any condition, the nodes receiving this Reply Completed will never be able to join generation and generation may take more rounds as less participants are available.
This might mean publishing failover need to be implemented first.

We also need to make sure we don't use dead_ids as is to send Reject Completed, as the aborted requests should not send Reject Completed.

ppca: also we need to consider what condition should we send Reject Completed and whether the node should be able to re-spawn a task for generating signature for a request after it received f+1 Reject Completed. 
Because it might happen that nodes regressed in block height due to checkpoint divergence, yet they have completed signature generation but not published the signature successfully yet. But they may have sent Reject Completed to other nodes before they regress. Now after regressing, since respond() is not observed, nodes should generate signature again, but other nodes may not be able to join because they've got enough f+1 Reject Completed for this request.

ppca (commented):
.

doc/posit_state_machine.md
```diff
@@ -0,0 +1,464 @@
...
+
+    Organizing --> Organizing: new round
+    Posit --> Organizing: new round
+    Generating --> Organizing: new round
+
+    Done --> [*]
+```
+
+A task enters `Organizing` when `SignatureSpawner` spawns it for a newly indexed
+request, at round 0. If governance is not `Running`, the request is retained and
```
volovyks: DESIGN: What do you think about removing governance updates from this state machine completely? Even while it is still here, I would not focus on it too much. Probably move into a separate section. Gov update is a very rare ocation.

I would consider removing state updates from the Spawner and kill/recreate it based on governance changes.

(not a high priority)

doc/posit_state_machine.md
```diff
@@ -0,0 +1,464 @@
...
+Running these machines needs little memory. Per in-flight request a node holds:
+
+- **the current phase** — where it is in the diagrams above;
+- **`r`, the round** — the one real driver: it fixes the proposer and the round
+  timeout, and every node agrees on it without messaging (§5);
+- **`highest_seen_round`** — the largest round any peer has mentioned, so one
+  bump can jump straight to it;
+- **the round's timeout clock**, and, for a proposer, its concurrency permit;
+- **a one-slot-per-sender buffer** of messages for a future round (§5);
+- optionally **`pause_proposing_until`**, a deadline until which it declines to
```
volovyks, line 53: DESIGN: What if we remove this completely? It adds compexity and:
- the node goes to the ORG phase, and it is expected that after `round++` it will not be a proposer anymore
- as a result, it will wait for a proposal, and probably never get it, since the generation would be successful
- even if it sees itself as a proposer again, it will get rejected again, not a big deal
- and that should not happen often, since if the node is online and has a presignature, it should always participate in signature generation

volovyks: Instead, we can have a more generic 1s (or any other) delay when entering reorganization for any reason.

doc/posit_state_machine.md
```diff
@@ -0,0 +1,464 @@
...
+  propose.
+
+A round bump resets only the timeout clock and releases the permit. Everything
```
volovyks: Let's specify why these two flows are different and what is triggering each one.

doc/posit_state_machine.md
```diff
@@ -0,0 +1,464 @@
...
+else in the list carries across it: `r`, `highest_seen_round`, the buffer (which
+is why it exists — it waits for the node to reach `highest_seen_round`), and
+`pause_proposing_until`. A respawn keeps less: only `r` survives, via
+`SignEntry.round`; `highest_seen_round`, the buffer, and the pause all start
+empty again. The proposer's ACCEPT/REJECT tally is transient, rebuilt inside each
+`Propose sent`.
+
+Not per request: the mesh's active set (a shared `watch`) and the governance
```
volovyks: This paragraph is a bit hard to read. Rephrasing may be a good idea.

doc/posit_state_machine.md
```diff
@@ -0,0 +1,464 @@
...
+request itself. Per node across requests:
+an LRU of finished `sign_id`s, so a late posit for a done request is dropped
+instead of starting a fresh machine.
+
+## 2. Inside Organizing
+
+```mermaid
+stateDiagram-v2
+    state "<b>Waiting for participants</b><br/>1. round timeout starts ticking<br/>2. wait for t active peers<br/>3. determine role" as WaitingForParticipants
+    state "<b>Reserving</b><br/>1. take one of 4 permits<br/>2. reserve a presignature with at least t active holders<br/>3. send PROPOSE to those active holders" as Reserving
```
volovyks, line 78: DESIGN: We removed `active` in one place, but there are multiple usages of it in the previous steps. What if we stop checking for active and rely on Posits (only the final check)?

volovyks: I'm hesitant about this one. On one hand, `active` is a pure optimization that adds complexity. On the other hand, it saves us from calling offline nodes.

doc/posit_state_machine.md
```diff
@@ -0,0 +1,464 @@
...
+```
+
+### 3b. Deliberator
+
+```mermaid
+stateDiagram-v2
+    state "Organizing" as OrgIn
+
+    state "<b>Waiting for Propose</b>
+    1. wait for PROPOSE from the elected proposer
```
volovyks, line 138: DESIGN: Yes, we still have `if from == &proposer`. But we agreed to get rid of it, correct?

doc/posit_state_machine.md
```diff
@@ -0,0 +1,464 @@
...
+    2. reject any other proposer for r" as WaitingForPropose
```
volovyks, line 139: From what I see it is rejected with
```rust
action: PositAction::RejectWithReason(
    PositRejectReason::InvalidRequest,
),
```

doc/posit_state_machine.md
```diff
@@ -0,0 +1,464 @@
...
+
+    state "<b>Waiting for Start</b>
+    1. wait for START from the proposer" as WaitingForStart
+
+    state "Generating" as GenOut
+
+    OrgIn --> WaitingForPropose: deliberator
+    WaitingForPropose --> ProposeReceived: PROPOSE from elected proposer
+    ProposeReceived --> WaitingForStart: ACCEPT sent
+    ProposeReceived --> WaitingForPropose: presignature missing, REJECT MissingArtifact sent
```
volovyks, line 153: DESIGN: unrelated to this PR, but this may be an actual trigger for State sync. Not `inactive` -> `active`.

doc/posit_state_machine.md
```diff
@@ -0,0 +1,464 @@
...
+by design, so the rotation has to be ended by information, not time.
+
+1. **Answer posits for completed requests.** `dead_ids` silently drops them
+   today; replying REJECT Completed lets a straggler end its task on the next
+   message it sends. To be Byzantine fault tolerant, f+1 such rejects should
+   be seen before moving on. One new reject reason, no new message type. It is
+   the minimal form of the round-outcome signal (§3). It covers the subset of
+   "no proposer" where the elected node already completed and retired the
+   request.
+2. **Prune the pool.** Record `MissingArtifact` in the holder set. A `REJECT
```
volovyks: `MissingArtifact` should never occur under normal circumstances. Every node executes State Sync logic on its start. If we see this error, we may have a bigger problem - a ton of artifacts are not in place.

I would rather triger the state sync logic for all artifacts instead of doing that on `inactive` -> `active` change. Taht make little sense to me.

doc/posit_state_machine.md
```diff
@@ -0,0 +1,464 @@
...
+| `START` | the accepters only, a subset of the `PROPOSE` set |
+| `REJECT` | whoever sent the message being rejected |
+
+A member outside the `PROPOSE` set is never told that round `r` is running. It
+sits in `Waiting for Propose` until its timeout expires and bumps to `r+1`,
+where it may be elected proposer and reserve a *second* presignature for a
+request that is already being signed. That is the waste the `pause_proposing_until`
+flag exists to limit after the fact.
+
+Sending `PROPOSE` to every member would help, cheaply: the tally is keyed
```
volovyks: I like the simplification.

doc/posit_state_machine.md
```diff
@@ -0,0 +1,464 @@
...
+   catches up in one bump instead of one round at a time.
+3. A REJECT is never answered with a REJECT, otherwise two nodes ping-pong.
+
+Buffering keeps one slot per sender, and a message above `highest_seen_round`
+clears the whole buffer and raises `highest_seen_round` to `r'`; one below it is
+dropped. On replay the node takes senders in arbitrary order, and only once
+`highest_seen_round == r`, in `Waiting for Propose`.
+
+One slot per sender means a second message from the same sender for the same
+round overwrites the first, in arrival order. Two layers do this: the ingress
```
volovyks, line 277: DESIGN: I still think that a single layer should be enough. Probably with additional filtering.

doc/posit_state_machine.md
```diff
@@ -0,0 +1,464 @@
...
+1. **No way to leave the state machine except by generating.** No round cap, no
+   cross-round deadline, and `round_timeout` saturates at 600s. The one failure
+   terminal (`Complete(Err(Aborted))`) needs a closed proposer semaphore, which
+   nothing closes, so no request ever fails from inside the machine. This is
+   the desired behavior.
+2. **A late accepter burns a full round.** A node that catches up mid-round
+   sends `ACCEPT` after the proposer already sent `START`. The proposer is
+   in `Generating` and drops `ACCEPT` there, so the late node waits out its
+   timeout in `Waiting for Start` and only rejoins at `r+1`.
+3. **`REJECT MissingArtifact` costs a whole round**, for the reason given under §3.
```
volovyks: TODO: Is it? From a single deliberator?

doc/posit_state_machine.md
```diff
@@ -0,0 +1,464 @@
...
+4. **One slot per sender assumes an ordering which is not guaranteed.** Both
+   buffers overwrite on arrival, including for an equal round, so if a sender's
+   `PROPOSE` and `START` for one round arrive out of order the later arrival
```
volovyks: `r` is basically a nonce, but a couple of messages can be sent with the same nonce. If we add a couple of rules, like START cannot override PROPOSE, we would guarantee ordering. But in reality, START can not be sent by an honest node without receiving ACCEPT first. Maybe the ordering problem is not real?

doc/posit_state_machine.md
```diff
@@ -0,0 +1,464 @@
...
+  silent-success branch has never been observed firing; fixing it is insurance
+  against a latent path, not an active leak.
+- **Round desync within expectations.** Across 179 `(sign_id, round)`
+  pairs on all 12 pods, pods enter the same round within ~200 ms and
+  `received Propose from non-proposer` is 0 all day. The correction traffic is
+  trivial next to ~3250 rotations/hour: 1153 StaleRound rejects (median gap 3)
+  and 4919 future-round buffers (median gap 1, 89% of them gap 1).
+
+Not so great
+- **Frequent reorganization on devnet.** ~100k reorganizes in 48h on
```
volovyks: I'm surprized that Testnet does not have it, since Posits are taking a lot of time there as well.

doc/posit_state_machine.md
volovyks: Probably how many requests we are processing atm (the number of Sign Tasks).

volovyks: We may want to investigate why that is happening. This is the intersection of Backlog and Request processing.

doc/posit_state_machine.md
```diff
@@ -0,0 +1,464 @@
...
+2. **Prune the pool.** Record `MissingArtifact` in the holder set. A `REJECT
+   MissingArtifact` is a peer stating it does not hold that share. Feed it to
+   `remove_holder_and_prune`, which sync already uses, so the holder set in
+   Redis reflects it and later reservations skip that peer. Do not remove
+   holders for active-set absence: that is transient, and deleting on it would
+   discard usable presignatures during a network blip. Pruning stays keyed on
+   the holder count falling below `t`, never on liveness. Ranked here for
+   recovery value after storage incidents rather than for steady-state churn,
+   which the sign-path reject counts above do not support.
+3. **SKIP.** A proposer that cannot propose (no presignature, no permit,
```
volovyks: We may think about this once we make sure that when everybody is online and has presignatures works as expected.

volovyks (commented):
Nice doc!

Comments tagged with `DESIGN` are not strictly related to this PR. But I think this is a good chance for a discussion.

doc/posit_state_machine.md
yvonneanne: I agree with your concerns! Each of these improvement proposals would require a proper design doc and analysis. They are not as simple as they sound.

doc/posit_state_machine.md
yapignolet: Thanks, I rephrased it. It now says the request is usually newly indexed but not always, and names requeue_pending_sign_requests as the other path. Checked against the code: it fires on ChainEvent::CatchupCompleted and filters on is_pending_generation().

doc/posit_state_machine.md
yapignolet: Yes, bounded LRU at 4096. The doc now says "a bounded record of the 4096 most recently retired requests" rather than an unqualified LRU.

doc/posit_state_machine.md
yapignolet: Correct, and I had it wrong 😊 . A late posit never starts a machine; it only reaches the mailbox map and allocates an inbox that nothing drains. Rewritten to say that.

doc/posit_state_machine.md
yapignolet: yes, I added this explicitly below now

doc/posit_state_machine.md
yvonneanne: yes, Serhii guessed it correctly, I'll rephrase accordingly.

doc/posit_state_machine.md
yvonneanne: Added as open question 2 at the end of the doc

doc/posit_state_machine.md
yvonneanne: Added as open question 3 at the end of the doc

doc/posit_state_machine.md
yvonneanne: Added as open question 4

doc/posit_state_machine.md
yvonneanne: Added as open question 4

doc/posit_state_machine.md
yvonneanne: added as open question 5

doc/posit_state_machine.md
yvonneanne: yeah, that should help. It' related to open question 3, I'll add a forward reference.

doc/posit_state_machine.md
```diff
@@ -0,0 +1,551 @@
...
+A member outside the `PROPOSE` set is never told that round `r` is running. It
+sits in `Waiting for Propose` until its timeout expires and bumps to `r+1`,
+where it may be elected proposer and reserve a *second* presignature for a
+request that is already being signed. That is the waste the `pause_proposing_until`
+flag exists to limit after the fact.
+
+Sending `PROPOSE` to every member would help, cheaply: the tally is keyed
+on `SinglePositCounter::participants` and `process_action` drops senders outside
+that set, so extra replies cannot move `enough_rejects` or `meets_totality`. It
+would be a notification, not a vote.
```
yvonneanne:
```suggestion
would be a notification, not a vote. See also open questions on "active" checks below.
```

doc/posit_state_machine.md
yvonneanne: added as open question 1 at the end of the doc.

doc/posit_state_machine.md
yvonneanne: I rewrote it like this:

```
Three things reset this state.

A **new round** starts when a recoverable failure sends the task back to the
organizing phase while it keeps running. It resets the timeout clock and
releases the permit; `r`, `highest_seen_round`, the proposing pause and the
buffer all carry across.

A **respawn** is a governance change: the committee this node signs with
changed in membership, threshold, epoch, or by leaving the running state.
Every in-flight task is aborted and rebuilt, but `r` survives.

A **node restart** loses everything: the state lives only in memory and the
request re-enters from the backlog at 0, so a node that restarts alone is
behind its peers.
```

doc/posit_state_machine.md
yvonneanne: you're absolutely right. I rephrased it as follows, please speak up if it's still unclear

"Two things are held per node rather than kept per request: (i) the mesh's active set, used to select pre-signatures with shares held by reachable and up-to-date nodes, and (ii) dead_ids, a bounded record of the 4096 most recently retired requests. A posit arriving for a dead id is dropped. Without it, a late message would create an inbox that no task will ever read.

doc/posit_state_machine.md
```diff
@@ -0,0 +1,464 @@
...
+
+It only half-solves the problem, though. A member holding the presignature but
+absent from the active set at reservation time would `ACCEPT`, go uncounted, and
+wait for a `START` that never comes. And the excluded member still cannot tell
+"round `r` succeeded without me" from "round `r` still running", which is what it
+needs. The narrow addressing is a symptom; the missing round-outcome signal is
+the cause.
+
```
yvonneanne:
```suggestion
When the excluded member receives that signal it could stop. Not stopping costs three things. It burns the round for every peer still waiting. In the rounds where it is the elected proposer it takes one of the four concurrency permits, and reserves a presignature that it returns to the pool when the round times out. Since only a proposer takes a permit, the slot is held intermittently rather than for the whole wait, so the cost is the wasted rounds and the repeated reservations.

```

doc/posit_state_machine.md
yvonneanne: Recorded as §10.6, with what I think is the real obstacle. The two layers look alike but do different jobs: the ingress one exists before any task does, so a posit arriving before the request is spawned is not lost, and it survives a respawn deliberately; the per-round one lives in the task's state and holds messages whose round is ahead of this node's. A single layer has to be task-independent and round-aware at once. Plus §8.4: it must survive PROPOSE and START arriving out of order.

doc/posit_state_machine.md
```diff
@@ -0,0 +1,553 @@
...
+   edge. Every node already syncs on start, so a `REJECT MissingArtifact` in
+   steady state says something larger than one missing share is wrong. Two
+   candidates, not exclusive: sync all artifacts rather than keying on a
+   liveness edge, and treat the reject itself as a trigger. Overlaps proposal
+   2 above, which uses the same signal for a different purpose.
+6. **Is one buffering layer enough?** §5 describes two, the ingress
+   `PositMailbox` and the per-round buffer, both keeping one slot per sender
+   and both overwriting on arrival. Filtering rather than overwriting at
+   ingress may remove the need for the second. §8.4 is the constraint:
+   whatever remains has to survive PROPOSE and START arriving out of order.
```
yvonneanne:
```suggestion
6. **Is one buffering layer enough?** §5 describes two, both keeping one slot per sender and both overwriting on arrival, but they do different jobs. The ingress mailbox exists before any task does, which is how a posit that arrives before the request is spawned is not lost, and it survives a respawn on purpose. The per-round buffer lives inside the task's state and holds a message whose round is ahead of this node's. Collapsing them means the survivor has to do both: exist without a task, and know the node's current round. §8.4 is the other constraint, that whatever remains has to survive PROPOSE and START arriving out of order.
```

doc/posit_state_machine.md
yvonneanne: Only at a threshold-sized set. The rule is rejects > (size of the selected set − threshold), where the selected set is the presignature's active holders including the proposer itself. At exactly threshold there is no slack: the proposer pre-accepts itself and all t−1 others must accept, so one reject ends the round. With k spare members it takes k+1. Reworded to say that.

```suggestion
3. **`REJECT MissingArtifact` can cost a whole round.** The proposer selects the presignature's holders that are active, itself included, and that set is only guaranteed to reach t. It abandons the round once the rejects exceed the set's size minus t. So at exactly threshold size it has no slack at all: every one of the other t−1 members must accept, and one reject ends the round. With k members of slack it takes k+1.
```

doc/posit_state_machine.md
yvonneanne: iiuc, you're right for one incarnation. if we consider nodes crashing and recovering it could in theory happen. when looking at logs I couldn't observe any re-orderings, so I don't think it's a problem to worry about now.

I rephrased the text to make it more clear.

doc/posit_state_machine.md
```diff
@@ -0,0 +1,549 @@
...
+   timeout in `Waiting for Start` and only rejoins at `r+1`.
+3. **`REJECT MissingArtifact` costs a whole round**, for the reason given under §3.
+4. **One slot per sender assumes an ordering which is not guaranteed.** Both
+   buffers overwrite on arrival, including for an equal round, so if a sender's
+   `PROPOSE` and `START` for one round arrive out of order the later arrival
+   wins and the other is dropped silently, costing that node the round.
+   Causality makes it unlikely, not impossible: `START` follows the node's own
+   `ACCEPT`, which follows `PROPOSE`, so it takes a `PROPOSE` delayed past a
+   full round trip. This is the constraint to revisit before sending more
+   messages per sender per round (§3).
```
yvonneanne:
```suggestion
   messages for one round arrive out of order the later arrival wins and the other is dropped silently, costing that node the round.
   Currently, causality makes this impossible (`START` follows the node's own  `ACCEPT`, which follows `PROPOSE`), it is a constraint to revisit before sending more messages per sender per round though(§3).
```

doc/posit_state_machine.md
```diff
@@ -0,0 +1,567 @@
...
+- optionally **`pause_proposing_until`**, a deadline until which it declines to
+  propose.
+
+A round bump resets only the timeout clock and releases the permit. Everything
+else in the list carries across it: `r`, `highest_seen_round`, the buffer (which
+is why it exists — it waits for the node to reach `highest_seen_round`), and
+`pause_proposing_until`. A respawn keeps less: only `r` survives, via
+`SignEntry.round`; `highest_seen_round`, the buffer, and the pause all start
+empty again. The proposer's ACCEPT/REJECT tally is transient, rebuilt inside each
+`Propose sent`.
```
yvonneanne:
```suggestion
Three things reset this state.

A **new round** starts when a recoverable failure sends the task back to the
organizing phase while it keeps running. It resets the timeout clock and
releases the permit; `r`, `highest_seen_round`, the proposing pause and the
buffer all carry across.

A **respawn** is a governance change: the committee this node signs with
changed in membership, threshold, epoch, or by leaving the running state.
Every in-flight task is aborted and rebuilt, but `r` survives.

A **node restart** loses everything: the state lives only in memory and the
request re-enters from the backlog at 0, so a node that restarts alone is
behind its peers.
```

doc/posit_state_machine.md
```diff
@@ -0,0 +1,567 @@
...
+
+Not per request: the mesh's active set (a shared `watch`) and the governance
+membership and threshold; the third election input, entropy, comes with the
+request itself. Per node across requests:
+`dead_ids`, an LRU of retired `sign_id`s bounded at `MAX_DEAD_IDS` (4096). A
+late posit message for a retired request is dropped by it. Without this check,
+it would allocate an inbox that nothing ever drains.
```
yvonneanne:
```suggestion
Two things are held per node rather than kept per request: (i) the mesh's active set, used to select pre-signatures with shares held by reachable and up-to-date nodes, and (ii) dead_ids, a bounded record of the 4096 most recently retired requests. A posit arriving for a dead id is dropped. Without it, a late message would create an inbox that no task will ever read.
```

doc/posit_state_machine.md
```diff
@@ -0,0 +1,567 @@
...
+   the holder count falling below `t`, never on liveness. Ranked here for
+   recovery value after storage incidents rather than for steady-state churn,
+   which the sign-path reject counts above do not support.
+3. **SKIP.** A proposer that cannot propose (no presignature, no permit,
+   paused) says so; receivers end round `r` at once instead of timing out on
+   silence. Silent rounds dominate the churn. One class of them is now
+   accounted for, a proposer that already completed and retired the request,
+   which 1 covers; SKIP covers the rest and sharpens the diagnosis, since a
+   round that stays silent after SKIP has an absent proposer, not an unwilling
+   one.
```
yvonneanne:
```suggestion
   one. This should only be worked on if data shows it will help .
```

doc/posit_state_machine.md
yvonneanne: added to the text

doc/posit_state_machine.md
yvonneanne: Agreed, and the numbers back you up. Over a full day the sign-path reject ("deliberator does not have access to proposed presignature") fired 0 times on devnet and 7 on testnet.

The high counts we had before were the presignature path ("presignature required triples are not known"), which replies with the same MissingArtifact reason but is about triples rather than presignature shares: 60 on devnet and 3000+ on testnet over that same day. Conflating the two is what made this look worse than it is.

I added "Ranked here for
   recovery value after storage incidents rather than for steady-state churn,
   which the sign-path reject counts above do not support." and added the state-sync question to the next section.

doc/posit_state_machine.md
yvonneanne: I added some more text to cover your input, speak up if it doesn't fit.

doc/posit_state_machine.md
yvonneanne: Because of an older version my first query didn't see it... I updated the text

doc/posit_state_machine.md
volovyks: Thanks
Or, this can be a trigger for active-> inactive. Which then triggers the state sync.

yvonneanne:
I separated the logs and metrics analysis from and the design questions from the PR, so the PR is purely a description of the status quo. I moved them + some thoughts of mine to https://docs.google.com/document/d/1i2dXrm1UL3Ok_6oqvAaFydwmnCdyR6ghjenAFww_L80/edit?tab=t.0

and I found an issue: https://github.com/sig-net/mpc/issues/1178

volovyks (approved):
Looks good! I think it is ready to be merged and updated as we go. Thank you!

PR #1153, opened 2026-08-19
fix(audit): patch h2 0.4 line, ignore advisory for pinned 0.3 line

yvonneanne:
## What

RUSTSEC-2026-0258 (h2 accepts unbounded empty DATA frames, low-severity DoS, published Aug 17) entered the advisory DB on Aug 18, after develop's last green Audit run — every PR fails the Audit check since, regardless of content (first seen on #1152, a docs-only PR).

Two h2 versions are in the lockfile:

- `h2 0.4.14` → bumped to the patched **0.4.16** (lockfile-only, `cargo update -p h2@0.4.14 --precise 0.4.16`)
- `h2 0.3.26` → **ignored** in `.cargo/audit.toml` with the usual blocked-upstream note: the 0.3 line has no patched release, and it is pinned via hyper 0.14 / reqwest 0.11

## Why ignore is safe enough

Low severity, DoS-class only (unbounded queuing of empty DATA frames if streams are not drained), same treatment as the other pinned-transitive advisories already in the ignore list. Drops out once the hyper 0.14 / reqwest 0.11 consumers are gone.

### DIFF
--- a/.cargo/audit.toml
+++ b/.cargo/audit.toml
@@ -27,4 +27,6 @@ ignore = [
 	# wasmtime stores can mix up type indices between engines; only reachable via near-sdk's `unit-testing`
 	# dev-only dependency (near-vm-runner 0.37.2 pins wasmtime ^45, which has no patched release; blocked upstream).
 	"RUSTSEC-2026-0222",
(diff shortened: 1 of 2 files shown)


yvonneanne:
duplicate PR #1150

PR #1155, opened 2026-08-20
feat: optimize checkpoint memory usage

ChaoticTempest:
This optimizes memory usage across checkpoints (both latest consensus and pending) and execution watchers by eliminating redundant cloning of sign requests and transaction payloads. This does what we did earlier in https://github.com/sig-net/mpc/pull/1151 but for checkpoints this time by having `Checkpoint`s use `Arc<IndexedSignRequest>` by using `BacklogEntry`s directly which already had it.

* had to move `Checkpoint` from `mpc-primitives` to `mpc-node` since `BacklogEntry` has `SignStatus` that needs to be serialized but references `cait_sith::Participant`. That type cannot be put in `mpc-primitives` due to wasm compilation issues.
* Wrapped `BidirectionalTx` in an `Arc` in `SignStatus::PendingExecution`, `Backlog`, `StateManager`, and `CompletedTx`. This ensures immutable bidirectional transaction data is created once and shared across backlog transitions, watcher polling, and checkpoint restores.
* No more `PendingTx` so this is a serialization breaking change: checkpoint version bumped


chain-signatures/node/src/lib.rs
```diff
@@ -4,7 +4,7 @@ pub const PROTOCOL_VERSION: u64 = 1;
 /// Version of the checkpoint data that the node can work with.
 pub const CHECKPOINT_VERSION: u64 = 0;
```
isSerge, line 5: PR description says `checkpoint version bumped`, but this remains the same. Did you mean `CHECKPOINT_STORAGE_VERSION` or this should be bumped too?

isSerge (approved):

chain-signatures/node/src/node_client.rs
```diff
@@ -94,6 +94,7 @@ impl NodeClient {
     pub fn new(options: &Options) -> Self {
         Self {
             http: reqwest::Client::builder()
+                .no_proxy()
```
volovyks: Why do we need this change?

volovyks (approved):
Looks good!

chain-signatures/node/src/lib.rs
ChaoticTempest: sorry, should've mentioned storage version change. The `CHECKPOINT_VERSION` is also meant for the serialization being different underneath as well when over to other nodes, but it's not necessary to bump until we release on testnet

chain-signatures/node/src/node_client.rs
ChaoticTempest: shouldn't be necessary, it was just easier for mockito tests but it's fine without. Removed

PR #1156, opened 2026-08-20
fix(integration): align test tiers, rename jobs, update `fixture` profile, etc.

isSerge:
This PR includes few enhancements for integration tests:

- **Reclassify `cases::checkpoint` into the fixture tier**. These tests do not need any of the cluster infra. Previously were in cluster tier due to filtering (not `mpc`, `state_sync` or `mpc_with_stream`). 
- **Drop `--features helios` from fixture job.** Redundant for fixture tests
- **Rename `sequential-integration-test` to `cluster-test`.** Stale name, it's not sequential for a while
- **Rename `component-test` to `fixture-test`.** Name was misleading, aligned with `nextest` profile

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -3,17 +3,18 @@ cluster = { max-threads = "num-cpus" }
 fixture = { max-threads = "num-cpus" }
 
 [[profile.default.overrides]]
-filter = 'test(cases::mpc) | test(cases::state_sync) | test(cases::mpc_with_stream)'
+filter = 'test(cases::mpc) | test(cases::state_sync) | test(cases::mpc_with_stream) | test(cases::checkpoint)'
 test-group = 'fixture'
 
 [[profile.default.overrides]]
-filter = 'not (test(cases::mpc) | test(cases::state_sync) | test(cases::mpc_with_stream))'
+filter = 'not (test(cases::mpc) | test(cases::state_sync) | test(cases::mpc_with_stream) | test(cases::checkpoint))'
 test-group = 'cluster'
 
 # Usage: cargo nextest run -p integration-tests --profile fixture
 [profile.fixture]
-default-filter = 'test(cases::mpc) | test(cases::state_sync) | test(cases::mpc_with_stream)'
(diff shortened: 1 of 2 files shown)


.config/nextest.toml
```diff
@@ -3,17 +3,19 @@ cluster = { max-threads = "num-cpus" }
...
 [[profile.default.overrides]]
-filter = 'not (test(cases::mpc) | test(cases::state_sync) | test(cases::mpc_with_stream))'
+filter = 'not (test(cases::mpc) | test(cases::state_sync) | test(cases::mpc_with_stream) | test(cases::checkpoint))'
 test-group = 'cluster'
 
 # Usage: cargo nextest run -p integration-tests --profile fixture
 [profile.fixture]
-default-filter = 'test(cases::mpc) | test(cases::state_sync) | test(cases::mpc_with_stream)'
+default-filter = 'test(cases::mpc) | test(cases::state_sync) | test(cases::mpc_with_stream) | test(cases::checkpoint)'
+retries = 2
```
volovyks: Are we adding this because of instabilities? Does not look like a good practice.

isSerge: Right now if flaky tests fails, we have to re-run manually and wait for the whole set. With retries enabled `nextest` will retry a single failing test automatically and if it succeeds it will be marked as `FLAKY` (similar to `SLOW`, `LEAKY`, etc.). Seems like a better alternative to manual retries we have to do occasionally

volovyks: How is that represented in the PR?
Currently, we have a task that adds a comment that the tests were restarted.

isSerge: Right now it's not visible in PR, can add a task that will comment summary with `FLAKY` tests. That would be more informative than existing comment about tests restarted attempt

volovyks: Let's do that. But if that is hard, let's just get rid of it. It is always better to double down on real test stability.

P.S. Please prioritize this PR, because it blocks all other merges.

isSerge: removing retries - can't find any elegant way to extract these from the summary report and we seldom have flaky tests recently anyway

volovyks (approved):

PR #1160, opened 2026-08-22
test(solana): add missing integration tests

isSerge:
Close component tier gaps for Solana:

- Respond round trip: Index sign request in Live phase -> publish signature (using `SolanaClient`) -> indexer parses CPI `SignatureResponded` back (`test_solana_respond_round_trip`)
- Failed transaction skipped, stream survives (`test_solana_stream_failed_tx_skipped`)
- Gap free resume when validator restarts (`test_solana_validator_restart_gap_free_resume`)

### DIFF
--- a/integration-tests/tests/cases/solana_stream.rs
+++ b/integration-tests/tests/cases/solana_stream.rs
@@ -659,7 +659,7 @@ async fn test_solana_stream_failed_tx_skipped() -> Result<()> {
 }
 
 #[test_log::test(tokio::test)]
-async fn test_solana_validator_restart_gap_free_resume() -> Result<()> {
+async fn test_solana_stream_resumes_gap_free_after_outage() -> Result<()> {
     let mut solana = solana_sandbox().await?;
     let program_address = solana.program_keypair.pubkey().to_string();
     let config = solana.get_config(program_address);
--- a/integration-tests/tests/cases/solana_stream.rs
+++ b/integration-tests/tests/cases/solana_stream.rs
@@ -77,6 +77,22 @@ async fn wait_for_sign_request(
     }
 }
 
+/// Wait for a SignCommand::Request to arrive on the sign channel
+async fn next_sign_command(
+    sign_rx: &mut mpsc::Receiver<SignCommand>,
+) -> Result<Arc<IndexedSignRequest>> {
+    loop {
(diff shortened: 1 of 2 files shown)


isSerge:
Had to restart integration tests because some tests occasionally hang

integration-tests/tests/cases/solana_stream.rs
```diff
@@ -537,3 +554,262 @@ async fn test_solana_stream_republishes_pending_publish_after_checkpoint_recover
...
+        request.args.payload,
+        Scalar::from_bytes(valid_payload).unwrap(),
+        "stream must keep parsing sign requests after a failed tx"
+    );
+
+    Ok(())
+}
+
+#[test_log::test(tokio::test)]
+async fn test_solana_validator_restart_gap_free_resume() -> Result<()> {
```
volovyks: Why do we need to test validator restarts?
At first, I thought this test was about restarting the indexer when submitting a request and ensuring that the indexed set is "gap-free".

volovyks (commented):
Great addition to our coverage!
`test_solana_stream_failed_tx_skipped` is a must-have

While reviewing, I had an idea for a test that compares live and catch-up processing of the same series of blocks and compares the backlog.

integration-tests/tests/cases/solana_stream.rs
isSerge: we're not testing validator here, let me adjust the name

isSerge: test idea is interesting, but it will be less valuable as we transition to live polling - live phase will use the same mechanism catchup is using currently

volovyks: In the context of `test_solana_stream_resumes_gap_free_after_outage`, I'm asking if `// Kill the validator and submit a request while it's down` section is what we intended to do.
I think I understand now. You want to see if there are no gaps after the lost connection. And what I was expecting is that there are gaps after our indexer restart.

isSerge: I think I understand now, you mean a test that verifies stream backfills requests missed during our own downtime (not chain downtime)

volovyks: yes, exactly

volovyks (approved):

PR #1162, opened 2026-08-22
feat(midnight): migrate MPC to singleton events

Pessina:
## Summary

- index Signet requests and responses from guaranteed singleton event transcripts in finalized `TxApplied` transactions; warn and skip unsupported statuses or fallible singleton calls
- move the publisher child to `@sig-net/midnight-contract@0.20.0-rc.1` and remove the map-based response pre-check, so an ambiguous retry may post a duplicate
- carry raw block, header, body, and system-event inputs for a future verifier and port the real-stack event harness

This is intentionally one runnable migration: the contract flip, read path, publisher child, fixtures, and harness move together.

## Verification

- `npm test` (41 passed)
- `cargo nextest run -p mpc-chain-midnight` (114 passed, 4 skipped)
- Rust-to-TypeScript intent seam (3 passed)
- `cargo nextest run -p mpc-node midnight` (9 passed)
- exact Midnight real-stack acceptance test (1 passed)


chain-signatures/chain-midnight/src/emissions.rs
```diff
@@ -0,0 +1,526 @@
...
+    singleton: &[u8; 32],
+) -> anyhow::Result<Vec<SingletonCallEmissions>> {
+    tx.calls()
+        .enumerate()
+        .filter(|(_, (_, call))| call.address.0 .0 == *singleton)
+        .map(|(call_index, (_, call))| {
+            let call_index = u32::try_from(call_index)
+                .context("transaction contains more calls than a u32 locator can represent")?;
+            if call.fallible_transcript.is_some() {
+                return Err(anyhow::Error::new(UnsupportedFallibleCall { call_index }));
```
ppca, line 175: but if other calls are not fallible_transcript we miss all of those?

Pessina: Yes, there are a lot of edge cases to properly handle the Fallible segments on the MPC side. 

So for v0 we will add to our docs how integrators can avoid fallible segments when building the tx on the client side and implementing the contract. 

But I will deep dive into this to improve the MPC and be able to handle those on our side, and later clean up the docs.

ppca (approved):

PR #1168, opened 2026-08-25
feat: singular DelayMonitor for watching delay instead of multiple tasks

ChaoticTempest:
This makes it so we have a singular task for all of delay watching. We have a lot of scheduler contention if we end up with two tasks per sign request and additional memory overhead. The introduced singular `DelayMonitor` uses a `DelayQueue` which is much easier on the scheduler and easier on the memory overhead per task just for delay monitoring. 

We add new entries for the `DelayMonitor` to watch over an unbounded command channel. The SignatureSpawner will already bound the amount of work that can be done, so we don't need to know limitations of how much we can send here. The added tests now individually test against this delay monitoring.

### DIFF
--- a/chain-signatures/node/Cargo.toml
+++ b/chain-signatures/node/Cargo.toml
@@ -64,7 +64,7 @@ serde_json.workspace = true
 sha3.workspace = true
 thiserror.workspace = true
 tokio.workspace = true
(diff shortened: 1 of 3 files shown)


volovyks (approved):
I like the design

PR #1169, opened 2026-08-25
feat(solana): replace WS live sub with finalized interval polling

### LINKED ISSUE #1157: Solana: ChainEvent::Block is not a barrier, so checkpoints can capture partial state and diverge
The Solana live indexer emits ChainEvent::Block(slot) once per transaction notification (indexer.rs:836, :876), so a slot with N relevant transactions emits it N times, interleaved with that slot's own events.

Block(H) triggers checkpoint creation on interval crossing (set_processed_block → snapshot), and snapshot captures whatever is in memory at that instant. The node has not necessarily received all events for heights ≤ H, so the checkpoint is a partial view of H.

Checkpoint::digest() hashes block height and sign_ids, so partial views yield different digests → consensus mismatch → backlog regress, which fires SignCommand::AbortChain and kills every in-flight sign task on that chain.

Observed on devnet via /checkpoint, all at slot 485498296: dev-6 has 5 pending requests, dev-9/dev-7 have 3, dev-0 has 2. All statuses PendingGeneration, so this is set membership, not the tag coarsening addressed in #1154. Ethereum is unaffected — identical counts at every height.

Fix proposal: emit Block(slot) once per slot, after all its events — hold the pending slot and flush when a strictly higher slot arrives, plus a flush on the existing stall timer. The continue paths (already-seen signature, parse failure) must still advance the watermark.

isSerge:
This PR removes WebSocket live phase. Solana indexer now run a single loop forever using existing catchup mechanism. This should resolve #1157

**What mirrors Ethereum indexer:**

- Finalized commitment everywhere (previously `Confirmed` commitment level was used), this currently introduces ~13s additional latency
- Unified live/catchup. Single loop to serve both phases. 
- `SolIndexerConfig` is nested in `SolConfig` (mirrors `EthConfig.indexer`)
- Added `RunFixture` helper struct for tests (driving `run()` against a mockito RPC server)

**What is different from Ethereum indexer:**

- No optimistic mode. `solana-test-validator` does block finalization (unlike Anvil)
- Slot-stall watchdog for frozen RPC node so indexer can self-diagnose and bail (Ethereum does not have this problem)
- Fixed interval polling instead of finalized head watcher. Fits Solana cadence better. Default is 1 sec, but tunable via config. 
- `PollState` struct to track loop state with transition methods (replaces mutable variables in `run` method)

**Important notes:**

- Sign-request detection latency rises from ~1-2 sec to ~12 sec + poll_interval (worst case). Alpenglow Solana update should decrease it dramatically.
- New CLI argument: `--sol-poll-interval-ms` (default 1000)

### DIFF
--- a/chain-signatures/chain-solana/src/indexer.rs
+++ b/chain-signatures/chain-solana/src/indexer.rs
@@ -1465,44 +1465,39 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn poll_state_drained_advances_watermark_and_monotonic_heartbeat() {
+    async fn poll_state_drained_advances_next_start_monotonically() {
         let sm = MockStateManager::new();
         let state = PollState::resumed(&sm, 10).await;
 
-        // anchor 10 drains `[next_start, 10)`: watermark advances to 10,
-        // heartbeat covers slot 9, catchup gate flips.
-        let (state, heartbeat) = state.drained(10);
+        // anchor 10 drains `[next_start, 10)`: the next tick starts at 10
+        // and the catchup gate flips.
+        let state = state.drained(10);
         assert_eq!(state.next_start, 10);
-        assert_eq!(heartbeat, Some(9));
-        assert_eq!(state.heartbeat_floor, 9);
         assert!(state.caught_up);
 
-        // Heartbeat advances with the anchor: 30 - 1 = 29 beats the floor of 9.
-        let (state, heartbeat) = state.drained(30);
+        let state = state.drained(30);
         assert_eq!(state.next_start, 30);
-        assert_eq!(heartbeat, Some(29));
-        assert_eq!(state.heartbeat_floor, 29);
 
-        // A regressing anchor cannot regress the heartbeat: the floor holds at 29
-        let (_state, heartbeat) = state.drained(5);
-        assert_eq!(heartbeat, Some(29));
+        // A regressing anchor (lagging replica behind an LB) cannot rewind
+        // the next drain range.
+        let state = state.drained(5);
+        assert_eq!(state.next_start, 30);
     }
 
     #[tokio::test]
-    async fn run_emits_heartbeat_and_single_catchup_completed_when_idle() {
-        // Frozen anchor: an idle chain whose tip never moves. The indexer
-        // must still emit a heartbeat for the covered tip and a single
-        // CatchupCompleted event.
+    async fn run_emits_single_catchup_completed_when_anchor_frozen() {
+        // Frozen anchor: the range [10, 10) covers no new slots, so the only
+        // event is exactly one CatchupCompleted — no Block markers.
         let mut f = RunFixture::spawn(&[10], Some(9), None).await;
 
-        // First tick: heartbeat for the covered tip, then exactly one
-        // CatchupCompleted. Second tick: heartbeat again — same height.
-        assert!(matches!(f.next_event().await, Some(ChainEvent::Block(9))));
         assert!(matches!(
             f.next_event().await,
             Some(ChainEvent::CatchupCompleted)
         ));
-        assert!(matches!(f.next_event().await, Some(ChainEvent::Block(9))));
+        assert!(
+            f.events_rx.try_recv().is_err(),
+            "a frozen anchor covers no new slots, so no Block markers may be emitted"
+        );
 
         f.cancel_and_join().await;
     }
@@ -1510,8 +1505,8 @@ mod tests {
     #[tokio::test]
     async fn run_bails_when_observed_slot_frozen() {
         // A frozen RPC node: getSlot keeps returning the same slot forever.
-        // The heartbeat keeps Block events flowing, so only the indexer-side
-        // slot-stall watchdog can catch this.
+        // Dense markers cover only drained slots, so a frozen anchor stops
+        // Block flow; the indexer-side stall watchdog trips first.
         let f = RunFixture::spawn(
             &[10],
             Some(9),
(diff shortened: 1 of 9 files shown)


volovyks (commented):
If we are changing the definition of finality (expected time), we need to update the value in `ChainConfig::expected_finality_time_secs`, since it is used in the expected request delay calculation.

chain-signatures/chain-solana/src/indexer.rs
```diff
@@ -9,27 +9,117 @@ use async_trait::async_trait;
...
+                slot_last_advanced_at: Instant::now(),
+                ..self
+            }),
+        }
+    }
+
+    /// Update the heartbeat floor and next start slot after a tick has drained to the anchor.
+    fn drained(self, anchor: u64) -> (Self, Option<u64>) {
+        // The heartbeat is the highest slot that has been fully drained and emitted as a `Block` event.
+        let heartbeat = anchor.saturating_sub(1).max(self.heartbeat_floor);
```
yvonneanne: anchor is this node's own getSlot(finalized) at its own poll instant, so Block(anchor - 1) is a node-local height. That height becomes the checkpoint height in process_block_event, Checkpoint::digest() hashes block_height, and the contract keys its vote map on the whole (chain, height, digest) tuple. So two nodes with identical request sets but different heights never vote alike.

would quantizing help?

chain-signatures/chain-solana/src/indexer.rs
```diff
@@ -40,41 +130,54 @@ impl<S: StateManager, T: ChainTelemetry> SolanaIndexer<S, T> {
...
     /// `fetch_slots` failures are propagated so the supervisor can restart.
     async fn catchup_blocks(
         &self,
         anchor_height: u64,
+        start_slot: u64,
     ) -> anyhow::Result<
         Pin<Box<dyn Stream<Item = anyhow::Result<CatchupBlockItem>> + Send + 'static>>,
     > {
-        let Some((start_slot, end_slot)) = self.catchup_range(anchor_height).await else {
+        let end_slot = anchor_height.saturating_sub(1);
```
yvonneanne, line 167: Minor: end_slot comes from getSlot and the range is drained via getSignaturesForAddress, but those are two separate calls on one RpcClient with no session affinity. Behind a load balancer they can hit different backends, and if the getSlot backend is ahead of the one serving signatures, the top slots of the range come back empty, next_start advances past them, and they're never queried again. Silent loss, no error, no retry. 

Catchup had the same exposure but only at startup; now it's the steady-state path and may be worth tightening. E.g., bound next_start by the highest slot the signature query actually returned rather than by getSlot.

chain-signatures/chain-solana/src/indexer.rs
isSerge: Few questions on this:

1. My understanding of severity: Fragmentation happens if either node is stalled, catching up or there is an RPC lag. Two live nodes observe the same finalized step and should vote identically, fragmented votes are drained at the next checkpoint. So quantization is about fixing catchup-time divergence, correct?
2. Currently this seems like a Solana-specific problem (probably will change as we add more chains). If we implement quantizing within `chain-solana` indexer it will force indexer to know about checkpoint policy, which seems like a layering smell. Alternatively, we could implement it in the Backlog (policy owner, checkpoint construction already lives there). What do you think?

yvonneanne: haha, two newbies trying to figure out what is going on and what should be going on :) I hope someone more experienced will chime in too, but here are my thoughts anyways.

1. Not mostly catchup, mostly live. The crossing is whichever height first passes the boundary, and the per-tick heartbeat anchor - 1 fires every second while program slots are sparse, so the heartbeat is usually what crosses, carrying a node-local number. Unsynchronised poll phase alone is worth a couple of slots at 400ms.
Ethereum isn't exempt because of cadence, it's exempt structurally: it emits Block(n) for every finalized block, gapless, so all nodes emit an identical height sequence and poll timing changes only when a node votes, never what it votes for.
Also, fragmented votes aren't drained at the next checkpoint. The contract prunes stale votes only when one settles (lib.rs:990), and locally the 32 pending cap gates dispatch (supervisor.rs:167) and is restored at its previous size across restarts by design (checkpoints.rs:236).

2. Agreed the interval belongs in the Backlog. 
"Every slot it processes" is already what process_block does today; the fix is emitting Block(s) for every slot in the drained range, including the ones with no program txs. Do that and the crossing lands on k * interval for every node with no quantization anywhere.
And quantizing the height alone wouldn't be enough: snapshot membership is anchor-dependent too (Checkpoints::snapshot has no height filter), so two nodes can still produce different digests at the same height.

chain-signatures/chain-solana/src/indexer.rs
isSerge: Good catch, but the highest returned slot is an activity slot, not a verified bound. If the program is idle, the query returns only old activity, so `next_start` and heartbeat will freeze and block-event timeout will trigger restart. 

What if we shift both bounds back by a configurable margin (enough to cover realistic load-balancer replication lag)? `next_start` becomes `anchor - M`, heartbeat - `anchor - 1 - M`, resume starts at `watermark + 1 - M`. Re-emitted events dedup, re-emitted blocks below watermark are no-op.

chain-signatures/chain-solana/src/indexer.rs
isSerge: I have removed heartbeats, made every slot emit exactly one block in order and added a couple of tests. I believe this should now mirror Ethereum dense block emission

ChaoticTempest (approved):
looks good to me for the most part. The LB lag could be a thing we tackle another time if we want to get this in first

chain-signatures/primitives/src/chain.rs
```diff
@@ -50,7 +50,10 @@ impl ChainConfig for Chain {
         match self {
             Chain::NEAR => 3,
             Chain::Ethereum => 30 * 60,
-            Chain::Solana => 3,
+            // The indexer waits for finalized (supermajority-epoch) blocks: the
+            // finalized pointer advances in ~32-slot batches, trailing the tip
+            // by up to ~13s.
+            Chain::Solana => 15,
```
volovyks, line 56:
```rust
fn expected_response_time_secs(&self) -> u64 {
        // finality time * 2 = finality time of sign/sign_bidirectional event + finality time of respond event
        self.expected_finality_time_secs() * 2 + 5 // + Buffer time
}
```
I expect the stability to increase a lot :)

But that is fine. We can focus on latencies later.

integration-tests/tests/cases/solana_stream.rs
```diff
@@ -32,6 +32,9 @@ use tokio::time::Instant;
 
 use std::time::Duration;
 
+/// Timeout for waiting for a finalized event (SignRequest or Block)
+const FINALIZED_EVENT_TIMEOUT: Duration = Duration::from_secs(45);
```
volovyks, line 36: Is this a function of `expected_finality_time_secs()`?

volovyks (approved):
Looks good!

chain-signatures/chain-solana/src/indexer.rs
isSerge: added a TODO comment for this

integration-tests/tests/cases/solana_stream.rs
isSerge: they are related, but test budget is dominated by CI variance not just finality

PR #1170, opened 2026-08-25
refactor(tests): optimize waiting for slow `mpc_with_stream` tests

isSerge:
`mpc_with_stream` suite currently has a few slow tests (`test_respond_event_prevents_clog`, `test_missed_respond_event_does_not_clog_inbox`) each taking more than 100 sec to execute due to fixed sleep timeouts and manual polling loops. This PR addresses this bottleneck by introducing Notify-driven helpers to `MpcFixture` and `SignatureTracker` structs

- Add new Notify-driven helper methods to `MpcFixture`:
  - `wait_for_actions_matching` - similar to existing `wait_for_actions`, but counts actions matching predicate
  - `assert_actions_matching` - similar to `assert_actions` + predicate
  - `assert_publish_actions` - same as `assert_actions_matching`, but ignores `kind: Checkpoint`
- Add `wait_for_delivered_posit` method and `posit_delivered: Arc<Notify>` field to `SignatureTracker`
- Update `mpc_with_stream.rs` to replace manual polling loops with fixed timeouts with new methods

Benefits:

- Tests are easier to read 
- Decrease `mpc_with_stream` suite time from ~105 sec to ~12 sec (total time for `fixture` profile tests should decrease too)

### DIFF
--- a/integration-tests/src/mpc_fixture/fixture_interface.rs
+++ b/integration-tests/src/mpc_fixture/fixture_interface.rs
@@ -264,10 +264,10 @@ impl MpcFixture {
         result.unwrap_or_else(|_| panic!("should produce {threshold} matching actions"))
     }
 
-    /// Wait for `threshold` actions, ignoring checkpoint-related ones.
+    /// Wait for `threshold` `RpcAction::Publish` actions.
     pub async fn assert_publish_actions(&self, threshold: usize, timeout: Duration) -> Vec<String> {
         self.assert_actions_matching(threshold, timeout, |action| {
-            !action.contains("kind: Checkpoint")
+            action.contains("RpcAction::Publish")
         })
         .await
     }
--- a/integration-tests/src/mpc_fixture/fixture_interface.rs
+++ b/integration-tests/src/mpc_fixture/fixture_interface.rs
@@ -190,6 +190,32 @@ impl MpcFixture {
         }
     }
 
+    /// Like [`Self::wait_for_actions`], but only actions matching `predicate`
+    /// count towards `threshold`.
+    pub async fn wait_for_actions_matching(
+        &self,
+        threshold: usize,
+        predicate: impl Fn(&str) -> bool,
+    ) -> Vec<String> {
+        loop {
+            let matching: Vec<String> = self
+                .output
+                .rpc_actions
+                .lock()
+                .await
+                .iter()
+                .filter(|action| predicate(action.as_str()))
+                .cloned()
+                .collect();
+
+            if matching.len() >= threshold {
+                return matching;
+            }
+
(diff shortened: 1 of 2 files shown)


integration-tests/src/mpc_fixture/fixture_interface.rs
```diff
@@ -219,6 +245,33 @@ impl MpcFixture {
...
+        if result.is_err() {
+            self.print_actions().await;
+        }
+        result.unwrap_or_else(|_| panic!("should produce {threshold} matching actions"))
+    }
+
+    /// Wait for `threshold` actions, ignoring checkpoint-related ones.
+    pub async fn assert_publish_actions(&self, threshold: usize, timeout: Duration) -> Vec<String> {
+        self.assert_actions_matching(threshold, timeout, |action| {
+            !action.contains("kind: Checkpoint")
```
yvonneanne: Nit: The name promises publish actions but the predicate only excludes checkpoint publishes?

```suggestion
                 action.contains("RpcAction::Publish") && !action.contains("kind: Checkpoint")
```

integration-tests/tests/cases/mpc_with_stream.rs
```diff
@@ -396,7 +383,23 @@ async fn run_stale_task_test(drop_respond_event: bool) {
...
+            && !wait_for_delivered_posit(
+                &tracker_counts,
+                &posit_delivered,
+                node_2,
+                bad_sign_id,
+                stale_posit_timeout,
+            )
+            .await
+        {
+            tracing::warn!(
```
yvonneanne: is a warning the right call here?

integration-tests/src/mpc_fixture/fixture_interface.rs
isSerge: `action.contains("RpcAction::Publish")` alone should be sufficient

integration-tests/tests/cases/mpc_with_stream.rs
isSerge: do you think we should fail? I thought delays are tolerable as long as we get correct results in the end

yvonneanne: iiuc, node 2 going stale is the precondition of this test, not a slow path to the result. If the posit never shows up, everything downstream still passes: completed == 20 holds trivially and n2_bad.signature == 0 is satisfied equally by "node 2 was excluded and 0+1 are dropping its stale posits" and "node 2 never did anything". Same reasoning as the note in test_divergent_active_views_still_converge above.

isSerge: updated, thanks for clarification

ChaoticTempest (approved):

PR #1171, opened 2026-08-25
feat(storage): add retry with exponential backoff for Redis connection

isSerge:
Integration tests frequently fails on both CI and locally due to stale Redis connection (`connect()` is a single attempt), this PR introduces retry mechanism to mitigate this (shouldn't harm in production too).

- Add retry with backoff using `backon` for Redis connection in `protocol_storage.rs`
- Make Redis readiness check for integration tests (`containers.rs`) more reliable
- Add `slow-timeout = { period = "60s", terminate-after = 4 }` for nextest config, so stuck tests fail instead of blocking workflow

### DIFF
--- a/.config/nextest.toml
+++ b/.config/nextest.toml
@@ -1,3 +1,6 @@
+[profile.default]
+slow-timeout = { period = "30s", terminate-after = 4 }
+
 [test-groups]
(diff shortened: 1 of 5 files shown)


volovyks (approved):
Nice!

PR #1172, opened 2026-08-26
fix(tests): prevent cross-process port collisions

isSerge:
Occasionally different integration tests hang due to port collision (`failed to bind web server … AddrInUse` in logs). This happens because nextest runs tests in parallel as separate processes and few tests can be handed the same port. This PR implements port allocation from a per-process disjoint slice of the port range, which is used in `pick_unused_port` helper. Harness only change.


volovyks (approved):

isSerge:
Had to re-run tests a couple of times to make sure issue was fixed

PR #1173, opened 2026-08-26
feat: contract reset checkpoint now resets local checkpoint

ChaoticTempest:
Adds in the ability to reset checkpoints based on the governance contract `reset_checkpoint([(Chain, Height)])`.

* `Contract::latest_checkpoint` now returns `Option<CheckpointDirective>` where we have:
   ```rs
   enum CheckpointDirective {
       Restart(Height),
       Consensus(ConsensusCheckpointDigest),
   }
   ```
* Contract serialization breaking change with the above, so we need a new migration for current devnet
* Contract `vote_checkpoint` treats `Restart(h)` as a floor: digests below h are rejected (CheckpointBehind), anything ≥ h proceeds normally; once a digest settles it replaces the restart marker.
* A reset/restart follows somewhat of the same code path as regression.
* In-flight checkpoint-vote RPCs and sign tasks are aborted on reset, which is the same as divergence/regression.

### DIFF
--- a/chain-signatures/contract/src/lib.rs
+++ b/chain-signatures/contract/src/lib.rs
@@ -16,7 +16,7 @@ use mpc_crypto::{
     derive_epsilon_near, derive_key, kdf::check_ec_signature, near_public_key_to_affine_point,
     ScalarExt as _,
 };
-use mpc_primitives::ConsensusCheckpointDigest;
+use mpc_primitives::{CheckpointDirective, ConsensusCheckpointDigest};
 use near_sdk::borsh::{self, BorshDeserialize, BorshSerialize};
 use near_sdk::env::panic_str;
 use near_sdk::json_types::U128;
@@ -85,7 +85,7 @@ pub struct MpcContract {
     pending_requests: IterableMap<SignId, PendingRequest>,
     proposed_updates: ProposedUpdates,
     config: Config,
-    latest_checkpoints: IterableMap<Chain, ConsensusCheckpointDigest>,
+    latest_checkpoints: IterableMap<Chain, CheckpointDirective>,
     checkpoint_votes: CheckpointVotes,
 }
 
@@ -131,7 +131,7 @@ impl MpcContract {
             pending_requests: IterableMap::new(StorageKey::PendingRequests),
             proposed_updates: ProposedUpdates::default(),
             config: config.unwrap_or_default(),
-            latest_checkpoints: IterableMap::new(StorageKey::LatestCheckpointDigests),
+            latest_checkpoints: IterableMap::new(StorageKey::CheckpointDirectives),
             checkpoint_votes: CheckpointVotes::new(),
         }
     }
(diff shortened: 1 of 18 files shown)


yvonneanne:
I'm very happy we're making this more deterministic!

Before a line-by-line pass: there may be a smaller version of this that drops the migration stuff entirely.

`Checkpoint::digest()` is a pure function of (chain, height, request ids, cumulative digest), so "backlog empty at height h" has a digest the contract and every node can derive independently. 

Have `reset_checkpoints` write that as an ordinary settled checkpoint instead of a marker variant: `latest_checkpoints`
keeps its type (no layout change, no `interpret`/fingerprint code), the vote floor falls out of the existing `existing.height > checkpoint.height` guard, and the node side becomes one branch in `align_backlog_with_consensus` to synthesize the checkpoint locally rather than fetch it from peers, then `regress`.

Why I'd prefer it: if the node infers "have I applied this reset?" from local state, it can fail in several ways. 
A node with no confirmed checkpoint never applies it.
A second reset while the first is outstanding is dropped everywhere.
`recover_backlog` re-applies on every supervisor restart, so an RPC error mid-catchup rewinds the cursor again. 
Through `regress` it is idempotent by construction, and `apply_checkpoint` also clears `PendingRequests`, which
`apply_reset` skips (the digest depends on the pending set, so nodes holding different sets won't agree post-reset either).

Either way: rewinding replays already-responded requests and `AbortChain` clears `self.requests`, the only dedup, with `h` unbounded. What stops a second response?

yvonneanne:
> I'm very happy we're making this more deterministic!
> 
> Before a line-by-line pass: there may be a smaller version of this that drops the migration stuff entirely.
> 
> `Checkpoint::digest()` is a pure function of (chain, height, request ids, cumulative digest), so "backlog empty at height h" has a digest the contract and every node can derive independently.
> 
> Have `reset_checkpoints` write that as an ordinary settled checkpoint instead of a marker variant: `latest_checkpoints` keeps its type (no layout change, no `interpret`/fingerprint code), the vote floor falls out of the existing `existing.height > checkpoint.height` guard, and the node side becomes one branch in `align_backlog_with_consensus` to synthesize the checkpoint locally rather than fetch it from peers, then `regress`.
> 
> Why I'd prefer it: if the node infers "have I applied this reset?" from local state, it can fail in several ways. A node with no confirmed checkpoint never applies it. A second reset while the first is outstanding is dropped everywhere. `recover_backlog` re-applies on every supervisor restart, so an RPC error mid-catchup rewinds the cursor again. Through `regress` it is idempotent by construction, and `apply_checkpoint` also clears `PendingRequests`, which `apply_reset` skips (the digest depends on the pending set, so nodes holding different sets won't agree post-reset either).
> 
> Either way: rewinding replays already-responded requests and `AbortChain` clears `self.requests`, the only dedup, with `h` unbounded. What stops a second response?

Tried it out on [yap/reset-via-canonical-checkpoint](https://github.com/sig-net/mpc/compare/yap/reset-via-canonical-checkpoint). The contract settles checkpoint_digest(chain, height, [], empty_cumulative_digest) as an ordinary ConsensusCheckpointDigest. The node side is one branch in align_backlog_with_consensus that rebuilds the checkpoint locally and falls through to regress.

WDYT?

ChaoticTempest:
@yvonneanne actually yeah, that is simpler. Let's just close this PR and open your branch as a new PR then

Issue #1063, opened 2026-07-27
Signature publish has no failover: proposer going offline stalls the request indefinitely


volovyks:
### Problem

Only the round's proposer publishes a completed signature
([signature.rs#L306-L313](https://github.com/sig-net/mpc/blob/12abdd3a7021984cb20db87ee59c80b0bf74edda/chain-signatures/node/src/protocol/signature.rs#L306-L313)),
with infinite in-memory retry
([rpc/mod.rs#L498-L533](https://github.com/sig-net/mpc/blob/12abdd3a7021984cb20db87ee59c80b0bf74edda/chain-signatures/node/src/rpc/mod.rs#L498-L533))
that does not survive a restart. If the proposer goes offline before the respond transaction lands, no other node takes over — even though **every participant holds the complete signature** in its backlog
([`PendingPublish`, signature.rs#L283-L303](https://github.com/sig-net/mpc/blob/12abdd3a7021984cb20db87ee59c80b0bf74edda/chain-signatures/node/src/protocol/signature.rs#L283-L303)).

Healthy participants can't rescue the request:

- `PendingPublish` entries are never re-queued for signing — [`take_requeueable_requests`](https://github.com/sig-net/mpc/blob/12abdd3a7021984cb20db87ee59c80b0bf74edda/chain-signatures/node/src/backlog/mod.rs#L349-L366) only takes `PendingGeneration`;
- republish on catchup skips non-proposers — [`resume_pending_publish_requests`](https://github.com/sig-net/mpc/blob/12abdd3a7021984cb20db87ee59c80b0bf74edda/chain-signatures/node/src/stream/ops.rs#L59-L61);
- re-proposal posits are dropped via [`dead_ids`](https://github.com/sig-net/mpc/blob/12abdd3a7021984cb20db87ee59c80b0bf74edda/chain-signatures/node/src/protocol/request/mod.rs#L239-L243).

The request stalls until the proposer returns or the user resubmits.

**Edge case, permanent stall:** a node regressing to a peer's checkpoint adopts the peer's entries verbatim, including `is_proposer: false`
([consensus.rs#L46-L67](https://github.com/sig-net/mpc/blob/12abdd3a7021984cb20db87ee59c80b0bf74edda/chain-signatures/node/src/backlog/consensus.rs#L46-L67)).
If the original proposer regresses this way, no node considers itself publisher anymore and the request can never recover internally.

### Proposed fix

[...]

PR #1175, opened 2026-08-26
feat(sign): publish failover

yvonneanne:
**Problem.** Only the proposer publishes a completed signature, so a proposer that dies after generation leaves it unpublished for good, even though every participant holds it. #1063.

**Solution.** The randomized schedule from #1063: each node draws a delay uniformly from `[L, L + m·L/d]` and publishes any entry still in pending-publish when it expires, giving one response on the happy path and `1 + d` when the proposer is silent. `L` is the lag until a published response is observed, chain finality plus a 15s margin; `m` is the number of participants; `d` is the duplicate responses accepted per failover, 2 here.

The sweep runs on block arrival rather than a timer, because a node can only conclude a response is missing if it has been watching the chain: one whose stream is down stays quiet instead of guessing. A restarting proposer republishes from the catchup resume, which records what it sent so the next sweep does not repeat it.

**History.** Reworked since the first review round, which rightly noted that the original design (a waiting task parked inside each sign generator) pinned the generator for the whole window and did not survive restarts. A second round moved the sweep off its own once-a-second task onto the stream's block arm, and dropped the proposer exclusion. Builds on #1186, which carries the independent hardening and cleanups this rework surfaced.

### DIFF
--- a/chain-signatures/node/src/protocol/signature.rs
+++ b/chain-signatures/node/src/protocol/signature.rs
@@ -353,11 +353,7 @@ fn build_publish_state(
         request.args.payload,
     )
     .ok()?;
-    let publish = PublishState {
-        signature,
-        participants: participants.to_vec(),
-        is_proposer,
-    };
+    let publish = PublishState::new(signature, participants.to_vec(), is_proposer);
 
     Some(Arc::new(publish))
 }
--- a/chain-signatures/node/src/stream/ops.rs
+++ b/chain-signatures/node/src/stream/ops.rs
@@ -1,8 +1,11 @@
+use std::collections::HashSet;
 use std::sync::Arc;
 
 use anyhow::Context;
 
+use crate::protocol::deliberator_publish::publish_deadline;
 use crate::respond_bidirectional::CompletedTx;
+use crate::rpc::PublishKind;
 use crate::sign_bidirectional::SignBidirectionalEventExt;
 use crate::stream::StreamContext;
 use mpc_chain_integration_core::ChainTelemetry;
@@ -23,10 +26,9 @@ pub(crate) async fn process_sign_request(
 
     // Reject malformed bidirectional requests at ingestion, running the same
     // deterministic derivations the respond event will need later. A request
-    // admitted here but failing there is worse than one never admitted: its entry
-    // sticks in pending-publish forever, and every backup's sweep fires into a
-    // duplicate-publish retry loop that no cancellation ever ends, because the
-    // failing respond processing is where the cancel lives.
+    // admitted here but failing there can never advance: its entry sticks in
+    // pending-publish forever, and every deliberator publishes a leg-1 response
+    // whose second leg will never come.
     if let SignKind::SignBidirectional(event) = &sign_request.kind {
         event.validate().with_context(|| {
             format!("rejecting bidirectional sign request {:?}", sign_request.id)
@@ -73,6 +75,63 @@ pub(crate) async fn resume_pending_publish_requests(ctx: &StreamContext, source_
         let sign_id = sign_request.id;
         ctx.rpc.publish_with_state(sign_request, &publish);
         tracing::info!(?sign_id, %source_chain, "resumed pending publish request after catchup");
+    }
+}
+
+/// Publish entries whose proposer stayed silent past this node's deadline.
+///
+/// Runs on every block rather than on a timer: a chain observed through block N
+/// is exactly the state in which "the response did not land" is a conclusion
+/// rather than a guess, so a stream that is not delivering blocks holds fire
+/// instead of publishing blind. Only one of the `m` deliberators needs a healthy
+/// stream for failover to happen.
+pub(crate) async fn deliberator_publish_due(ctx: &mut StreamContext, chain: Chain) {
+    if !ctx.caught_up {
+        return;
+    }
+
+    let publishable = ctx.backlog.publishable_requests(chain).await;
+    if publishable.is_empty() {
+        ctx.deliberator_published.clear();
+        return;
+    }
+
+    // Only worth building when something has fired: an entry that left the backlog
+    // and came back is marked publishing afresh and must be scheduled afresh.
+    if !ctx.deliberator_published.is_empty() {
+        let live: HashSet<(SignId, PublishKind)> = publishable
+            .iter()
+            .map(|(request, _)| (request.id, PublishKind::of(&request.kind)))
+            .collect();
+        ctx.deliberator_published.retain(|key| live.contains(key));
+    }
+
+    let me = ctx.contract_watcher.account_id().clone();
+    let now = mpc_utils::time::current_unix_timestamp();
+    for (request, publish) in &publishable {
+        if publish.is_proposer {
+            continue;
+        }
+        let key = (request.id, PublishKind::of(&request.kind));
+        if ctx.deliberator_published.contains(&key) {
+            continue;
+        }
+        let Some(deadline) = publish_deadline(&request.id, publish, &me, chain, ctx.observe_lag)
+        else {
+            continue;
+        };
+        if now < deadline {
+            continue;
+        }
+
+        ctx.deliberator_published.insert(key);
+        let sign_id = request.id;
+        tracing::warn!(
+            ?sign_id,
+            %chain,
+            "proposer response not observed in time; publishing as deliberator"
+        );
+        ctx.rpc.publish_with_state(Arc::clone(request), publish);
     }
 }
 
@@ -137,7 +196,7 @@ async fn advance_bidirectional_to_execution(
     // without passing admission (checkpoint recovery restores them wholesale). One
     // that fails here fails identically on every node and on every replay, so it
     // can never advance: leaving it would park it in pending-publish forever, with
-    // every backup's sweep republishing a response that is already on chain.
+    // every deliberator publishing a response that is already on chain.
     // Removing it is deterministic across the network, so checkpoints stay aligned.
     if let Err(err) = event.validate() {
         tracing::error!(
--- a/chain-signatures/node/src/stream/ops.rs
+++ b/chain-signatures/node/src/stream/ops.rs
@@ -3,7 +3,7 @@ use std::sync::Arc;
 
 use anyhow::Context;
 
-use crate::protocol::deliberator_publish::publish_deadline;
+use crate::protocol::publish_failover::publish_deadline;
 use crate::respond_bidirectional::CompletedTx;
 use crate::rpc::PublishKind;
 use crate::sign_bidirectional::SignBidirectionalEventExt;
@@ -27,7 +27,7 @@ pub(crate) async fn process_sign_request(
     // Reject malformed bidirectional requests at ingestion, running the same
     // deterministic derivations the respond event will need later. A request
     // admitted here but failing there can never advance: its entry sticks in
-    // pending-publish forever, and every deliberator publishes a leg-1 response
+    // pending-publish forever, and every node publishes a leg-1 response
     // whose second leg will never come.
     if let SignKind::SignBidirectional(event) = &sign_request.kind {
         event.validate().with_context(|| {
@@ -66,13 +66,18 @@ pub(crate) async fn requeue_pending_sign_requests(
     Ok(())
 }
 
-pub(crate) async fn resume_pending_publish_requests(ctx: &StreamContext, source_chain: Chain) {
+pub(crate) async fn resume_pending_publish_requests(ctx: &mut StreamContext, source_chain: Chain) {
     for (sign_request, publish) in ctx.backlog.publishable_requests(source_chain).await {
         if !publish.is_proposer {
             continue;
         }
 
         let sign_id = sign_request.id;
+        // Recorded so the failover sweep does not publish this entry a second time
+        // on the next block: it is already dispatched, and its deadline, anchored
+        // on a stamp from before the restart, is long past.
+        ctx.published
+            .insert((sign_id, PublishKind::of(&sign_request.kind)));
         ctx.rpc.publish_with_state(sign_request, &publish);
         tracing::info!(?sign_id, %source_chain, "resumed pending publish request after catchup");
     }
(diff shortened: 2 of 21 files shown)


chain-signatures/node/src/protocol/signature.rs
```diff
@@ -300,6 +300,28 @@ impl SignGenerator {
...
+                                "proposer response not seen; publishing as backup"
+                            );
+                            crate::metrics::requests::SIGN_BACKUP_PUBLISH
+                                .with_label_values(&[self.request.chain.as_str()])
+                                .inc();
+                            ctx.rpc.publish_with_state(
+                                ctx.governance.public_key,
+                                Arc::clone(&self.request),
+                                &publish,
+                            );
```
jakmeier: Optional: 

The way the metric description is written (`"Number of responses published by a non-proposer participant because the proposer's response was not observed in time, by chain."`) I would expect that we only count successful publish attempts. I think this counts all attempts, even if it fails for some reason. So if for example all nodes are unable to submit transactions to the chain right for some time, we would overcount this.

Would it make sense to rephrase the metric or perhaps increase the metric only after a publishing succeeds?

integration-tests/tests/cases/mpc_with_stream.rs
```diff
@@ -68,6 +68,77 @@ async fn test_sign() {
...
+        network.publishes(),
+        1,
+        "a backup published despite the response landing"
+    );
+}
+
+/// Three nodes signing on Solana. With `drop_respond`, the respond event that
+/// the mock chain derives from a publish is withheld from every node, which is
+/// what a proposer dying before its response lands looks like to the others.
+async fn backup_fixture(drop_respond: bool) -> integration_tests::mpc_fixture::MpcFixture {
```
jakmeier: nit: the name `backup_fixture` is a bit ambiguous. Maybe rename it to `backup_publish_fixture`. Or move this helper and the tests in a separate mod for publishing tests, then the mod could have the `publish` information in the name.

jakmeier (approved):
Nice improvement!

integration-tests/tests/cases/mpc_with_stream.rs
```diff
@@ -68,6 +68,77 @@ async fn test_sign() {
     );
 }
 
+/// Solana's `L` is 3s and the fixture signs with 3 participants, so a backup
+/// draws its delay from `L + U[0, m*L/d]`, at most 6s.
+const MAX_BACKUP_DELAY: Duration = Duration::from_secs(6);
```
volovyks: The definition of finality was changed in https://github.com/sig-net/mpc/pull/1169
Ideally, we should do the actual math specifying and pass `expected_finality_time_secs()`.
We also have `expected_response_time_secs()`, but it will probably not be useful in this case.

integration-tests/tests/cases/mpc_with_stream.rs
```diff
@@ -68,6 +68,77 @@ async fn test_sign() {
...
+
+/// The proposer's response never reaches the chain, so no node ever sees a
+/// completion: a participant other than the proposer has to publish instead, or
+/// the request is answered by nobody at all.
+#[test(tokio::test(flavor = "multi_thread"))]
+async fn test_backup_publishes_when_the_response_never_lands() {
+    let network = backup_fixture(true).await;
```
volovyks: It is unclear what this boolean change does (I understand now, but it took a while to notice).

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -430,6 +477,40 @@ mod tests {
         Participant::from(i)
     }
 
+    const ALL_CHAINS: [Chain; 7] = [
```
volovyks: Have you tried using something like this?
```rust
for chain in (0..Chain::LENGTH).map(Chain::from_usize) { ... }
```

The number of chains is expected to grow. People will forget to update this list.

chain-signatures/node/src/protocol/signature.rs
```diff
@@ -300,6 +300,28 @@ impl SignGenerator {
                                 "failed to mark publishing for sign request"
                             );
                         }
+
+                        // Stand in for a proposer that never publishes. The sleep is
+                        // cancelled by the task abort that a completion event triggers,
+                        // so in the happy path this never runs.
+                        if !is_proposer {
```
volovyks: The logic looks good, but it doesn't belong here.
Also, the sleep pins the whole SignGenerator for the delay. Not a huge problem for Solana, but on networks like Bitcoin, that will take forever.

volovyks (commented):
Looks good overall

chain-signatures/node/src/protocol/signature.rs
yvonneanne: You're right. With the re-design the wait is no longer in the generator: a per-node task polls the backlog once a second and publishes entries stuck in pending-publish past their deadline. Nothing is parked and nothing pins a generator, and since the jitter is a pure function of sign id and account id, the schedule survives restarts.

Please have another read.

chain-signatures/node/src/protocol/signature.rs
yvonneanne: Thanks, this is gone with the rework: the poll loop and its tests iterate Chain::iter()

integration-tests/tests/cases/mpc_with_stream.rs
yvonneanne: The schedule now derives from expected_finality_time_secs() plus an explicit observe margin (ObserveLag::PRODUCTION), so it tracked #1169's change automatically. The fixtures pin ObserveLag::Fixed(5s) instead of deriving from chain constants, so test timing no longer moves when a finality constant does.

integration-tests/tests/cases/mpc_with_stream.rs
yvonneanne: Replaced with a small enum: deliberator_publish_fixture(RespondEvents::Dropped) vs (RespondEvents::Delivered).

chain-signatures/node/src/protocol/signature.rs
yvonneanne: Reworked: the description now says it counts attempts, at most one per entry while it stays in the backlog (the failover fires once; the dispatched publish then retries until it lands or the entry leaves the backlog). Counting only landed responses would need confirmation inside each chain publisher, which no longer knows a publish was failover-originated; if we want a landed-only signal I'd add it publisher-side separately. The metric is now multichain_sign_deliberator_publish.

integration-tests/tests/cases/mpc_with_stream.rs
yvonneanne: Renamed to deliberator_publish_fixture

integration-tests/tests/cases/mpc_with_stream.rs
```diff
@@ -49,6 +49,111 @@ async fn test_sign() {
     );
 }
```
volovyks, line 51: Do we want to test bidirectional cases? Or do they use the same flows?

chain-signatures/node/src/metrics/requests.rs
```diff
@@ -104,6 +104,26 @@ pub(crate) static SIGN_REQUEST_DELAYED: LazyLock<CounterVec> = LazyLock::new(||
...
+pub(crate) static SIGN_DELIBERATOR_PUBLISH: LazyLock<CounterVec> = LazyLock::new(|| {
+    try_create_counter_vec_with_node_and_version(
+        "multichain_sign_deliberator_publish",
+        "Deliberator publish attempts by a non-proposer participant whose deadline passed without observing the proposer's response, by chain. Counts attempts, not landed transactions: at most one per entry while it stays in the backlog.",
+        &["chain"],
+    )
+    .unwrap()
+});
+
+pub(crate) static SIGN_DELIBERATOR_WAITING: LazyLock<prometheus::IntGaugeVec> = LazyLock::new(
```
volovyks: What is the reason for having this metric?

chain-signatures/node/src/metrics/requests.rs
```diff
@@ -104,6 +104,26 @@ pub(crate) static SIGN_REQUEST_DELAYED: LazyLock<CounterVec> = LazyLock::new(||
     .unwrap()
 });
 
+pub(crate) static SIGN_DELIBERATOR_PUBLISH: LazyLock<CounterVec> = LazyLock::new(|| {
+    try_create_counter_vec_with_node_and_version(
+        "multichain_sign_deliberator_publish",
```
volovyks: I agree that this is nice to have such a metric, but our Grafana already has 100+ panels :)
WDYT about emitting a warning each time the deliberator needs to publish?
Logs should be our go-to place for figuring out what is wrong. Grafana is mostly for measuring our success rates, etc.

volovyks (commented):
The failover idea is ~90 lines, but the PR is ~986, and most of the difference looks like it follows from two placement choices rather than from the problem:

1. The sweep lives in a standalone task — sitting outside the stream is what forces caught_up onto the Backlog as shared atomic state, even though stream/ops.rs already has resume_pending_publish_requests doing this exact scan and just discards the !is_proposer branch; calling it from the Block arm instead of a 1s timer keeps it inside the stream and makes the gate structural ("observed through block N" is exactly the precondition for "the proposer's response didn't land") rather than a flag that can drift.

2. The schedule lives in memory — stamping publishing_since on PublishState would replace the tracked map with a pure function of the entry, and fix the clock restarting on every node restart.

yvonneanne:
> The failover idea is ~90 lines, but the PR is ~986, and most of the difference looks like it follows from two placement choices rather than from the problem:
> 
> 1. The sweep lives in a standalone task — sitting outside the stream is what forces caught_up onto the Backlog as shared atomic state, even though stream/ops.rs already has resume_pending_publish_requests doing this exact scan and just discards the !is_proposer branch; calling it from the Block arm instead of a 1s timer keeps it inside the stream and makes the gate structural ("observed through block N" is exactly the precondition for "the proposer's response didn't land") rather than a flag that can drift.
> 2. The schedule lives in memory — stamping publishing_since on PublishState would replace the tracked map with a pure function of the entry, and fix the clock restarting on every node restart.

Agreed on both, will work on them now. Thanks a lot for the suggestions!

chain-signatures/node/src/metrics/requests.rs
yvonneanne: It existed because the standalone task computed a per-tick liveness count; going with your suggestion the stream sweep has no such number to publish, so I'll drop it.

chain-signatures/node/src/metrics/requests.rs
yvonneanne: The warning already exists, I'll drop this metric too.

yvonneanne:
> > The failover idea is ~90 lines, but the PR is ~986, and most of the difference looks like it follows from two placement choices rather than from the problem:
> > 
> > 1. The sweep lives in a standalone task — sitting outside the stream is what forces caught_up onto the Backlog as shared atomic state, even though stream/ops.rs already has resume_pending_publish_requests doing this exact scan and just discards the !is_proposer branch; calling it from the Block arm instead of a 1s timer keeps it inside the stream and makes the gate structural ("observed through block N" is exactly the precondition for "the proposer's response didn't land") rather than a flag that can drift.
> > 2. The schedule lives in memory — stamping publishing_since on PublishState would replace the tracked map with a pure function of the entry, and fix the clock restarting on every node restart.
> 
> Agreed on both, will work on them now. Thanks a lot for the suggestions!

Both your suggestions are now in: the sweep runs from the stream's Block arm, and the deadline is anchored on a publishing_since stamp, which also stops the clock restarting on every stream restart. I went further than the split and cut the publish stand-down gate entirely, since failover doesn't need it, which takes rpc/mod.rs from 390 lines to 40 and restores the retry_rpc! jitter. Both metrics are gone, the bidirectional case now has a test. 

The PR is now 156 lines of logic, 89 comments, 411 tests.

doc/protocol_properties.md
```diff
@@ -174,12 +174,15 @@ Generation is skipped entirely while \< t nodes are in the active set, so the fa
...
+
+The draw is anchored on a stamp stored with the entry, so it survives a restart rather than restarting the clock, and the sweep runs on block arrival, so a node that is not observing its chain holds fire instead of publishing blind. One participant with a healthy stream is enough.
+
+Two gaps remain. The delay follows chain finality, so a silent proposer costs a first failover of about `1.5·L`, tens of minutes on Ethereum. And entries checkpointed before the stamp existed carry no anchor and never fail over, which leaves the old behaviour for that set until it drains.
```
yvonneanne:
```suggestion
```

integration-tests/tests/cases/mpc_with_stream.rs
yvonneanne: Same flow apart from the dedup key: both legs share a sign id, so the fired-once and in-flight sets are keyed by PublishKind too, or a fired leg 1 would suppress leg 2 for good. `publish_failover_fires_once_per_leg` in ops_tests.rs covers that, and nothing else in the sweep branches on kind, so a fixture-level bidirectional case would only re-cover the plain path.

Does that suffice?

chain-signatures/node/src/stream/ops.rs
```diff
@@ -84,18 +86,72 @@ pub(crate) async fn requeue_pending_sign_requests(
...
+pub(crate) async fn publish_failover_due(ctx: &mut StreamContext, chain: Chain) {
+    if !ctx.caught_up {
+        return;
+    }
+
+    let publishable = ctx.backlog.publishable_requests(chain).await;
+    if publishable.is_empty() {
+        ctx.published.clear();
+        return;
+    }
```
ChaoticTempest: I'm not so sure if putting the published metadata in `StreamContext` is best here since now we'd have to track two separate areas just for whether we've published or not. We can possibly put this metadata in `BacklogEntry` and have it skip serialization on it

ChaoticTempest (approved):
looks good to me besides the one minor thing I mentioned

chain-signatures/node/src/stream/ops.rs
yvonneanne: Thanks for the suggestion, it is better to track it there and skip serialization. Please have a look at the implementation.

integration-tests/tests/cases/mpc_with_stream.rs
volovyks: yeah, it shoud

volovyks (approved):
Looks good, no blockers from me

PR #1176, opened 2026-08-27
feat: reset checkpoints by height

ChaoticTempest:
based on the work done @yvonneanne.

This resets the contract and nodes' checkpoint back to empty, by triggering a regression based on an empty digest overwritten to the near governance contract.

Previous work was done at https://github.com/sig-net/mpc/pull/1173 but that was more complex introducing a contract serialization change with a new `CheckpointDirective` enum. We avoid that here in this PR with what I mentioned above with just using an empty digest. This is safe to do since we follow the same code path for regression, so doing a reset here is basically a regression. A regression today already clears out local and pending checkpoints.

### DIFF
--- a/chain-signatures/contract/src/lib.rs
+++ b/chain-signatures/contract/src/lib.rs
@@ -1180,10 +1180,6 @@ impl MpcContract {
         &self.latest_checkpoints
     }
 
-    fn latest_checkpoints_mut(&mut self) -> &mut IterableMap<Chain, ConsensusCheckpointDigest> {
-        &mut self.latest_checkpoints
-    }
-
     fn insert_checkpoint(&mut self, chain: Chain, checkpoint: ConsensusCheckpointDigest) {
         self.latest_checkpoints.insert(chain, checkpoint);
     }
--- a/chain-signatures/contract/src/lib.rs
+++ b/chain-signatures/contract/src/lib.rs
@@ -1252,14 +1252,14 @@ impl VersionedMpcContract {
     }
 
     /// Settle each chain's *canonical reset checkpoint*: the one asserting an
-    /// empty backlog at `resume_after`.
+    /// empty backlog just before `height`.
     ///
-    /// Its digest is derived from `(chain, resume_after)` alone, so every node
+    /// Its digest is derived from `(chain, height - 1)` alone, so every node
     /// can rebuild it locally and regress onto it through the ordinary
     /// divergence path. Votes for the chain are dropped, since they describe
     /// the abandoned history.
     ///
-    /// `resume_after` is exclusive: indexing resumes at `resume_after + 1`.
+    /// `height` is inclusive: indexing resumes at `height` itself.
     ///
     /// The resulting checkpoint is deliberately indistinguishable from one the
     /// network settled with an empty backlog at the same height, because it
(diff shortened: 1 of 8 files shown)


chain-signatures/contract/src/lib.rs
```diff
@@ -1380,10 +1455,106 @@ mod tests {
...
+        contract.reset_checkpoints(vec![CheckpointReset {
+            chain: Chain::Solana,
+            height: 6,
+        }]);
+
+        // `height` is inclusive: indexing resumes at 6 itself, whose empty
+        // checkpoint lives at 5.
+        // The reset is an ordinary settled checkpoint whose digest is derived
+        // from the pair, which is what every node re-derives locally.
+        assert_eq!(
```
volovyks, line 1467: Not a blocker for this PR, but it would be great to make a step towards the genesis block design and save the starting block as well. Plus, have a contract API to get it.

chain-signatures/contract/src/lib.rs
```diff
@@ -1380,10 +1455,106 @@ mod tests {
...
+            .to_string()
+            .contains(&CheckpointError::CheckpointBehind.to_string()));
+        assert!(contract.checkpoint_votes(Chain::Solana).is_empty());
+
+        // Re-submitting the reset checkpoint itself is an idempotent no-op,
+        // which is what lets nodes confirm it rather than fight over it.
+        let reset = ConsensusCheckpointDigest::new(
+            Chain::Solana,
+            100,
+            mpc_primitives::reset_checkpoint_digest(Chain::Solana, 100),
```
volovyks: 101 -> 100 logic is a bit unclear here, would be nice to add a constant

chain-signatures/contract/src/lib.rs
volovyks: It is important to know it, since we can then fetch this info and start over from the same block. Our indexing capabilities will catch up eventually.

chain-signatures/node/src/backlog/checkpoints.rs
```diff
@@ -47,27 +47,34 @@ impl fmt::Debug for Checkpoint {
...
         Self {
             chain,
-            block_height: 0,
+            block_height,
             pending_requests: Vec::new(),
             cumulative_digest: Self::empty_cumulative_digest(),
         }
     }
 
     pub fn empty_cumulative_digest() -> [u8; 32] {
```
volovyks, line 65: Do we need this wrapper?

integration-tests/tests/cases/checkpoint.rs
```diff
@@ -277,3 +281,135 @@ async fn test_consensus_alignment_consensus_changes_while_fetching() {
...
+        let mut digests = HashSet::new();
+        for node in &network.nodes {
+            digests.insert(node.backlog.checkpoint(chain).await.unwrap().digest());
+        }
+        digests
+    };
+    assert_eq!(
+        digests_before.len(),
+        network.nodes.len(),
+        "test setup should leave every node on a different checkpoint"
```
volovyks: This comment does not represent the check.

volovyks (approved):
Nice work @ChaoticTempest and @yvonneanne! I like the simplicity.

chain-signatures/contract/src/lib.rs
ChaoticTempest: as discussed, this is a separate feature entirely and will be done later

chain-signatures/node/src/backlog/checkpoints.rs
ChaoticTempest: it's still a lot clearer to have it be in checkpoint

chain-signatures/contract/src/lib.rs
ChaoticTempest: made it clearer

integration-tests/tests/cases/checkpoint.rs
ChaoticTempest: rewrote the comment

PR #1182, opened 2026-08-28
Remove `pause_proposing_until`

volovyks:
Removing `pause_proposing_until` (Added in [#832](https://github.com/sig-net/mpc/pull/832), built for [#793](https://github.com/sig-net/mpc/issues/793)).

It predates presignature reservations, which fixed the real problem — a rejected proposer now just returns its reservation.

This flow is effectively dormant (~44 hits/day on devnet vs ~90k reorganizes), while the pause itself can be a hidden node-local mode: a paused elected proposer silently sits out for a full generation_timeout.

Worst case without it is one cheap re-propose that gets rejected within a round trip. Reject reasons are no longer acted on, but now get logged in the reorganize reason instead of being dropped.

### DIFF
--- a/chain-signatures/node/src/protocol/request/task.rs
+++ b/chain-signatures/node/src/protocol/request/task.rs
@@ -53,9 +53,6 @@ impl GeneratingPhase {
         state: &mut SignState,
         mailbox: &PositMailbox,
     ) -> SignPhase {
-        // We successfully committed to generating; future rounds should be unrestricted.
-        state.pause_proposing_until = None;
-
         let sign_id = ctx.sign_id;
 
         tracing::info!(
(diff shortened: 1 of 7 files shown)


chain-signatures/node/src/protocol/request/task.rs
yvonneanne: Not introduced by your PR, but should we use save the round in L125 in case it's

yvonneanne:
```suggestion
            round,
            my_round = state.round(),
            ?reason,
```

yvonneanne (approved):
LGTM, thanks! 

While reviewing I noticed that we don't use stale round correctly when already generating, fix here https://github.com/sig-net/mpc/pull/1185

PR #1185, opened 2026-08-31
fix(mpc-node): exchange rounds on a late-Propose reject

yvonneanne:
`reject_late_propose` is the only phase that neither tells a behind peer its round nor records an ahead peer's. [#1082](https://github.com/sig-net/mpc/pull/1082) added `PositMessage::stale_round` for exactly this and routed the two posit-phase sites through `record_peer_round`, but left this one at `None`. Checked the other 12 `stale_round: None` sites, all correct.

Now a behind peer gets `StaleRound` carrying our round, which the existing receiver logic already picks up, and an ahead peer's round is recorded. Nothing changes outside this function, and rounds still only go up.

Two tests, one per direction, both fail against the previous unconditional `AlreadyGenerating`/`None`. They reuse the harness already in `posit.rs`, which is the only reason that file changes.

### DIFF
--- a/chain-signatures/node/src/protocol/request/posit.rs
+++ b/chain-signatures/node/src/protocol/request/posit.rs
@@ -453,7 +453,7 @@ impl PositPhase {
 }
 
 #[cfg(test)]
-mod tests {
+pub(crate) mod tests {
     use super::*;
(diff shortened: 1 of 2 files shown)


volovyks (approved):
Nice catch!

PR #1186, opened 2026-08-31
fix(bidirectional): reject invalid requests at admission, quarantine stuck ones

yvonneanne:
A bidirectional request whose deterministic derivations fail (unknown target chain, underivable epsilon, undecodable transaction) is admitted today, signed, and half-answered: every node's respond processing then fails at the same spot forever, and the entry sits in pending-publish for good. Admission now runs the same derivations respond processing needs and rejects such requests before they enter the backlog.

Requests can also re-enter the backlog through checkpoint recovery, skipping admission. Those get the same treatment one step later: the first time such a request provably cannot be completed, it is dropped. Every node reaches that verdict at the same point from the same data, so the network stays in agreement about which requests are pending.

Note for applications: a request with an undecodable `serialized_transaction` now gets no response at all instead of a leg-1 response and a permanently stuck leg 2. The request was unfulfillable either way; it now fails fast and identically on every node.

The decode check makes `sign_and_hash_transaction` reachable from any observed request event, so the legacy `35 + 2 * chain_id` computation switches to checked arithmetic instead of panicking on a crafted chain id.

The remaining commits are groundwork for the #1175 rework, which builds on this branch:

* `PublishAction.public_key` is removed: no chain publisher reads it, publishers sign with their own payer keys, and `PublishAction::new` keeps taking the key it actually needs for signature reconstruction.
* The fixture protocol and its chain streams now share one `Backlog`. They each had their own, so `mark_publishing` failed with `NotFound` in every fixture test and the backlog half of the sign path was silently skipped.
* `publishable_requests` takes the read lock; it never mutates.
* The redis container helper retries a lagging port mapping instead of unwrapping, which flaked under parallel container starts.

### DIFF
--- a/chain-signatures/node/src/protocol/test_setup.rs
+++ b/chain-signatures/node/src/protocol/test_setup.rs
@@ -24,11 +24,15 @@ pub struct TestProtocolChannels {
 }
 
 impl MpcSignProtocol {
+    /// `backlog` is the one the caller's chain streams write to. Handing the sign
+    /// side its own would leave every request `NotFound` when it marks publishing,
+    /// which silently skips the backlog half of the sign path in every fixture.
     pub async fn new_test(
         my_account_id: AccountId,
         storage: TestProtocolStorage,
         channels: TestProtocolChannels,
         contract: ContractStateWatcher,
+        backlog: Backlog,
     ) -> Self {
         let generating = channels.msg_channel.subscribe_generation().await;
         let resharing = channels.msg_channel.subscribe_resharing().await;
@@ -42,7 +46,7 @@ impl MpcSignProtocol {
             channels.mesh_state.clone(),
             channels.msg_channel.clone(),
             channels.rpc_channel.clone(),
-            Backlog::new(),
+            backlog,
         );
         Self {
             my_account_id,
--- a/chain-signatures/node/src/protocol/test_setup.rs
+++ b/chain-signatures/node/src/protocol/test_setup.rs
@@ -24,9 +24,6 @@ pub struct TestProtocolChannels {
 }
 
 impl MpcSignProtocol {
-    /// `backlog` is the one the caller's chain streams write to. Handing the sign
-    /// side its own would leave every request `NotFound` when it marks publishing,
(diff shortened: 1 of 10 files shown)


chain-signatures/node/src/protocol/test_setup.rs
```diff
@@ -24,11 +24,15 @@ pub struct TestProtocolChannels {
 }
 
 impl MpcSignProtocol {
+    /// `backlog` is the one the caller's chain streams write to. Handing the sign
+    /// side its own would leave every request `NotFound` when it marks publishing,
+    /// which silently skips the backlog half of the sign path in every fixture.
```
yvonneanne:
```suggestion
```

chain-signatures/node/src/stream/ops.rs
isSerge, line 1: Not a blocker, but seems like we have validation scattered across multiple places: `ops.rs`, `sign_bidirectional.rs`. We probably should centralize it in `stream/validation.rs` and keep `ops.rs` as a pure orchestration layer. Can be separate PR. What do you think?

yvonneanne: Agreed on ops.rs being orchestration only. I went for the types rather than a stream/validation.rs: validate_bidirectional_event is now SignBidirectionalEventExt::validate, sitting next to the target_chain/epsilon/RLP derivations it composes, and verify_entry_signature is BacklogEntry::verify_signature. So the split you spotted between ops.rs and sign_bidirectional.rs is gone, all of it is in one place per type now.

I skipped the shared module because the two checks are about different types, and a validation.rs holding both would mostly be a bag. Happy to add one if you'd rather have a single file to read.

Wdyt?

isSerge: yeah, moving it to types looks cleaner

volovyks (approved):

volovyks:
Looks good overall, but build and tests need fixing.
Let's separate fixes and changes in the future; it is a bit hard to jump contexts.

PR #1187, opened 2026-08-31
refactor(ethereum): Unify RPC stacks and share 429 backoff gate

isSerge:
Currently we have two Ethereum clients: for indexer and for publisher. Each using own HTTP stack with divergent retry, timeout and error policies for the same RPC endpoint - if indexer gets rate-limited publisher will be unaware of cooldown and keep retrying. This PR unifies stack for both clients.

### Changes

- Refactor Indexer client (`indexer_eth_direct_rpc.rs`) to use Alloy transports - becomes more than 200 lines shorter
- Update publisher to use `retry_rpc_gated!` macro, same `SharedBackoff` is re-used for both publisher and indexer


volovyks (approved):
Nice!

PR #1189, opened 2026-08-31
feat(midnight): serialize responses with FAB

Pessina:
## Summary

- serialize Midnight bidirectional responses with FAB-compatible Compact layout while preserving the existing ABI and Borsh paths for other chains
- validate Midnight response schemas and match fakenet capacity, padding, type, and size rules using TypeScript-generated golden vectors
- route Midnight outcomes through FAB and cover the local Midnight-to-Ethereum-to-Midnight stack flow

## Verification

- affected Rust package suites, clippy, and formatting checks
- TypeScript unit, type, lint, formatting, and deterministic-vector checks
- local Midnight-to-Ethereum-to-Midnight full-stack integration test


chain-signatures/chain-ethereum/src/respond_bidirectional/midnight.rs
isSerge, line 1: Not sure `chain-ethereum` should contain Midnight knowledge and have Midnight-specific dependencies. Maybe we should move it to `chain-midnight/src/respond/fab.rs` - `chain-midnight` already depends on `alloy` and `hex`.

isSerge: We can also move `Output/AbiField` to `chain-integration-core` to make these shared across various chain integrations

chain-signatures/chain-ethereum/src/respond_bidirectional.rs
```diff
@@ -51,6 +57,7 @@ impl Output {
         match format {
             SerDeserFormat::Abi => encode_abi(data, &schema),
             SerDeserFormat::Borsh => encode_borsh(data, &schema),
+            SerDeserFormat::Fab => unreachable!(),
```
isSerge, line 62: `if format == Fab` + `unreachable!()` combo looks weird. Maybe we can do routing on the node side `node/src/respond_bidirectional.rs` since it already depends on specific chain integration crates:

```rust
match format {
       SerDeserFormat::Abi  => mpc_chain_ethereum::respond_bidirectional::encode_abi(&output, schema),
       SerDeserFormat::Borsh => mpc_chain_ethereum::respond_bidirectional::encode_borsh(&output, schema),
       SerDeserFormat::Fab  => mpc_chain_midnight::respond::serialize_fab(&output, schema),
   }?;
```

isSerge (commented):
Left few comments about the structure: I think we should keep `chain-ethereum` focused on EVM, generic encoding, etc. and keep Midnight-specific stuff within `chain-midnight`.

chain-signatures/chain-ethereum/src/respond_bidirectional/midnight.rs
Pessina: I understand ur concern, the issue it's that the AbiField do not carry the extra data Midnight needs the `max_bytes`, `max_items` and the EVM types are not 1:1 with Midnight types, so we need a validation/coercion layer to go from ABI -> FAB.

The rest of the file follow the existing pattern 

1 - de-serialize EVM Return Value
2 - re-serialize as FAB for Midnight (or Borsh for Solana, etc.)

The second issue it's that this layer return bytes, we lose the original value + schema so moving it to the node it's a large refactor. 

I would propose defer this changes when we need introduce another chain that requires us to extract this logic out of EVM, for example when Midnight can call Solana then we will need

1 - de-serialize Borsh from Solana Return Value
2 - re-serialize in Fab for Midnight

That's when the common Midnight module makes sense. 

I'm afraid to do all this refactor now due to the code freeze, it will touch all chains. The current changes are self-contained in Midnight 

But I fully agree with you.

Pessina: I'm happy to add a TODO/Issue and tackle it later. What do you think?

isSerge: yes, that would be great

Pessina: Created: https://github.com/sig-net/mpc/issues/1196

chain-signatures/node/src/respond_bidirectional.rs
```diff
@@ -67,6 +67,10 @@ impl CompletedTx {
                 output.extend_from_slice(&borsh_data);
                 Bytes::from(output).into()
             }
+            SerDeserFormat::Fab => {
```
BRBussy, line 70: Not at this location, but the hashing used on the midnight side has changed to no longer use keccak256 (it will not be available in next next compact release that we are targeting). So the hashing used in `calculate_respond_bidirectional_hash_message` needs to change based on chain and use poseidon to align with transient hash usage.

isSerge (approved):

PR #1192, opened 2026-08-31
chore(mpc-node): stop logging two normal outcomes as warnings

yvonneanne:
Two of the three most frequent warnings on devnet describe normal protocol outcomes, not failures. Both are phrased as errors, and together they account for thousands of WARN lines per hour under load, which makes the warning stream hard to read and buries real problems. The third commit replaces the visibility they were providing with a counter.

### `inbox: failed to decrypt/verify messages`

Every one of 500 sampled instances in a one-hour window carried `err=Idempotent`, which means the batch signature was already in the dedup LRU: a peer's outbox retried a send that had in fact landed. Nothing failed to decrypt, and no signature failed to verify. The message says otherwise.

It now gets its own arm at `debug` with a message that names what happened. Real decrypt/verify errors keep the WARN, and `UnknownParticipant` keeps its retry path (now an explicit arm rather than a `matches!` guard inside the error branch; the trailing `continue` was the last statement of the loop body, so dropping it does not change control flow).

### `presignature required triples are not known`

Fires when a peer proposes a presignature over a triple pair we do not hold, so we reply `RejectWithReason(MissingArtifact)` and the proposer retries with another pair.

Demoted to `debug` and reworded to state the decision. devnet runs `RUST_LOG=mpc_node=debug`, so the line is still there when someone reads the debug stream; it just leaves the WARNING severity bucket where alerting and Cloud Logging triage happen.

### `multichain_posit_rejects_total{protocol, reason}`

A per-line WARN was the wrong instrument for "how often do we refuse a posit" anyway. The new counter is incremented where the triple and presignature managers send their reply, so the rate is visible per reason (`missing_artifact`, `already_generating`, `stale_round`, `invalid_request`, `unknown`) without a log line per event, and it now covers reasons that were never logged at all. The signature posit path has several scattered send sites and is left for a follow-up; it slots in under the same `protocol` label without breaking anything built on this metric.

### Context, and what this does not fix

The rate is load-driven, not steady: on devnet it was >=10k/hour cluster-wide during load-test bursts and 0/hour when idle. The rejects are a symptom. In the 17:00-18:00 UTC hour on 2026-08-31 no posit was aborted for too many rejections, while 5934 presignature generations hit the 45s `generation_timeout` against 6729 completions, with container CPU limit utilization averaging 68-85% across all 12 pods and peaking over 100%. Each timeout burns a taken triple pair, which is what makes holder sets diverge in the first place. Issue to follow for that and for the reject-vs-wait asymmetry between `process_posit` and `PendingTriples::fetch`.

### Not touched

`presignature already generated` / `already generating` in the same handler are the same shape (a reject reply logged at WARN), but a duplicate proposal for a presignature we already have is arguably worth seeing, so I left the logs. They are counted by the new metric either way.

### Testing

`cargo check -p mpc-node --lib`, `cargo clippy -p mpc-node --lib` and `cargo fmt` pass. No behaviour outside logging and metrics changes.

### DIFF
--- a/chain-signatures/node/src/protocol/presignature.rs
+++ b/chain-signatures/node/src/protocol/presignature.rs
@@ -438,10 +438,6 @@ impl PresignatureSpawner {
             self.triples.contains_reserved(id.pair_id).await
                 || self.triples.contains(id.pair_id).await
         } {
-            // We do not hold the proposed triple pair: it has not reached us
-            // yet, or we already spent it. Rejecting is the designed answer
-            // here and the proposer retries with another pair, so this is a
-            // normal outcome rather than a fault.
             tracing::debug!(
                 ?id,
                 ?from,
--- a/chain-signatures/node/src/protocol/presignature.rs
+++ b/chain-signatures/node/src/protocol/presignature.rs
@@ -438,11 +438,15 @@ impl PresignatureSpawner {
             self.triples.contains_reserved(id.pair_id).await
                 || self.triples.contains(id.pair_id).await
         } {
-            tracing::warn!(
+            // We do not hold the proposed triple pair: it has not reached us
+            // yet, or we already spent it. Rejecting is the designed answer
+            // here and the proposer retries with another pair, so this is a
+            // normal outcome rather than a fault.
+            tracing::debug!(
                 ?id,
                 ?from,
                 ?action,
-                "presignature required triples are not known"
+                "rejecting presignature posit: triple pair not held locally"
             );
             PositInternalAction::Reply(PositAction::RejectWithReason(
                 PositRejectReason::MissingArtifact,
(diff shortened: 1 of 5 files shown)


chain-signatures/node/src/protocol/presignature.rs
```diff
@@ -438,11 +438,15 @@ impl PresignatureSpawner {
             self.triples.contains_reserved(id.pair_id).await
                 || self.triples.contains(id.pair_id).await
         } {
-            tracing::warn!(
+            // We do not hold the proposed triple pair: it has not reached us
+            // yet, or we already spent it. Rejecting is the designed answer
+            // here and the proposer retries with another pair, so this is a
+            // normal outcome rather than a fault.
```
yvonneanne:
```suggestion

yvonneanne:
Closing in favour of a narrower change: only the inbox Idempotent log fix will be reproposed. The presignature reject demotion and the posit-reject counter are dropped for now.

PR #1195, opened 2026-09-01
chore(mpc-node): quiet duplicate inbox batch log

yvonneanne:
### `inbox: failed to decrypt/verify messages`

Every one of 500 sampled warning instances with this message in a one-hour window carried `err=Idempotent`, which means the batch signature was already in the dedup LRU: a peer's outbox retried a send that had in fact landed. Nothing failed to decrypt, and no signature failed to verify. The message says otherwise.

It now gets its own arm at `debug` with a message that names what happened. Real decrypt/verify errors keep the WARN, and `UnknownParticipant` keeps its retry path (now an explicit arm rather than a `matches!` guard inside the error branch; the trailing `continue` was the last statement of the loop body, so dropping it does not change control flow).

### Testing

`cargo check -p mpc-node --lib` passes. No behaviour outside logging changes.

### DIFF
--- a/chain-signatures/node/src/protocol/message/inbox.rs
+++ b/chain-signatures/node/src/protocol/message/inbox.rs
@@ -230,7 +230,7 @@ impl MessageInbox {
                 // retry once the map catches up.
                 Err(MessageError::UnknownParticipant(_)) => retry.push((encrypted, timestamp)),
                 // A batch we have already ingested, resent because the peer's
-                // outbox retried a send that had in fact landed. 
+                // outbox retried a send that had in fact landed.
                 Err(MessageError::Idempotent) => {
                     tracing::debug!("inbox: dropped duplicate message batch");
                 }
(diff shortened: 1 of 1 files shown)


chain-signatures/node/src/protocol/message/inbox.rs
```diff
@@ -226,14 +226,16 @@ impl MessageInbox {
...
-                    } else {
-                        tracing::warn!(?err, "inbox: failed to decrypt/verify messages");
-                    }
-                    continue;
+                // Sender is not in our participant map yet; keep the batch and
+                // retry once the map catches up.
+                Err(MessageError::UnknownParticipant(_)) => retry.push((encrypted, timestamp)),
+                // A batch we have already ingested, resent because the peer's
+                // outbox retried a send that had in fact landed. Dropping it is
+                // the dedup working, not a decryption or signature failure.
```
yvonneanne:
```suggestion
                // outbox retried a send that had in fact landed. 
```

volovyks (approved):

PR #1197, opened 2026-09-01
Remove unused docker tests

volovyks:
I wanted to remove it like a year ago. So far, I'm not aware of any use cases and do not think anybody uses it. The setup itself is a lot of code duplication.


yvonneanne (approved):
Less code 🎉 

Looks complete.

ChaoticTempest (approved):

PR #1198, opened 2026-09-02
fix(posit): resync peers that reject proposal as missing holder

yvonneanne:
**Problem:** After a proposal was rejected because the presignature holder list is stale will be returned to the pool and re-used. Only state sync repairs holder lists, and it runs on peer reconnect, so with a stable peer set nothing corrects them: on devnet one presignature has been re-proposed more than 1145 times in for 23 hours, costing every affected sign request a full posit timeout.

**Solution:** A MissingArtifact reject is the holder saying it does not have the share, which is proof our holder list is stale. A proposer can then report that peer to the mesh as desynced, and the existing sync path prunes it, dropping any artifact whose remaining holders fall below threshold. This reuses the repair path we already have and fixes everything that peer is missing, triples included, not only the artifact that happened to be proposed.

**Implementation:** mpsc channel from the sign path to the mesh, mirroring the existing synced-peer channel: report_node_desynced in mesh/connection.rs, a receiver arm in the mesh loop, and the sender threaded from cli through SignatureSpawner into SignTask. The transition is Active -> Syncing only, so repeated rejects are no-ops. 

Notet that this transition takes the peer out of the active set until the sync succeeds (about 200ms for a full burst on devnet), and wait_for_active_participants has no fallback if active drops below threshold.


yvonneanne:
replaced by #1199

PR #1199, opened 2026-09-02
fix(posit): resync peers that reject as non-holder

yvonneanne:
**Problem**: After a proposal was rejected because the presignature holder list is stale will be returned to the pool and re-used. Only state sync repairs holder lists, and it runs on peer reconnect, so with a stable peer set nothing corrects them: on devnet one presignature has been re-proposed more than 1145 times in for 23 hours, costing every affected sign request a full posit timeout.

**Solution**: A MissingArtifact reject is the holder saying it does not have the share, which is proof our holder list is stale. A proposer can then report that peer to the mesh as desynced, and the existing sync path prunes it, dropping any artifact whose remaining holders fall below threshold. This reuses the repair path we already have and fixes everything that peer is missing, triples included, not only the artifact that happened to be proposed.

**Implementation:** sign path reports to the mesh over the existing synced-peer channel, now carrying a SyncKind { Synced, Desynced } tag: report_node_desynced in mesh/connection.rs, one receiver arm dispatching on the tag, and the sender threaded from cli through SignatureSpawner into SignTask. One FIFO also means a Synced and a Desynced report for the same peer are applied in emission order rather than in whichever order select! happens to pick. The transition is Active -> Syncing only, so repeated rejects are no-ops.

Notet that this transition takes the peer out of the active set until the sync succeeds (about 200ms for a full burst on devnet), and wait_for_active_participants has no fallback if active drops below threshold.


chain-signatures/node/src/cli/mod.rs
```diff
@@ -637,6 +639,7 @@ struct MeshHandles {
     contract_watcher: ContractStateWatcher,
     contract_state_tx: watch::Sender<Option<ProtocolState>>,
     synced_peer_tx: mpsc::Sender<Participant>,
+    desynced_peer_tx: mpsc::Sender<Participant>,
```
ChaoticTempest: couldn't we just use the same channel for report sync or desynced? Would require one less thing to manage by the mesh and less code too. We can have a new enum `SyncKind { Synced, Desynced }` passed along with the participant to say what their new status is.

chain-signatures/node/src/protocol/request/posit.rs
```diff
@@ -348,6 +348,24 @@ impl PositPhase {
...
                         if counter.enough_rejects(ctx.governance.threshold) {
+                            // A MissingArtifact reject means that holder never stored the
+                            // presignature, so our holder list for it is stale. State sync
+                            // owns holder lists, so put the peer back through it.
+                            for (peer, _) in counter
+                                .rejects
+                                .iter()
+                                .filter(|(_, reason)| {
+                                    matches!(reason, PositRejectReason::MissingArtifact)
+                                })
```
ChaoticTempest: nit: probably worth making a new method for this in PositCounter itself

ChaoticTempest (commented):
Looks good for the most part

chain-signatures/node/src/cli/mod.rs
yvonneanne: yes, that's better, I changed the implementation, please have another look

chain-signatures/node/src/protocol/request/posit.rs
yvonneanne: done

ChaoticTempest (approved):

volovyks (approved):
Looks good!
Let's see how frequent the syncs would be after the merge.

PR #1202, opened 2026-09-03
chore:  clean up & refactor checkpoint lifecycle, remove in memory checkpoints

ChaoticTempest:
Previously, `Checkpoints` maintained an in-memory mirror in parallel with `CheckpointStorage`. This PR eliminates the in-memory duality, establishing `CheckpointStorage` as the single source of truth for both pending and confirmed checkpoints. It also streamlines the checkpoint recovery/regression lifecycle and hardens startup error handling.

## Changes
* Removed the in-memory mirror from `Checkpoints`. `CheckpointStorage` now directly handles pending checkpoints
* New atomic pending counter per chain, used by `has_slot(chain)`.
* Made `PendingRequests::from_checkpoint` and `Backlog::recover_by_checkpoint` infallible: Checkpoint deserialization and backlog recovery are purely in-memory operations, they no longer return `Result` or require `.unwrap()` at call sites.
* refactored to have cleaner api surfaces such as `Checkpoints::hydrate`, `Checkpoints::latest`, and `Backlog::hydrate`.
* supervisor will now loop on backlog failing to hydrate itself (just loading the counter for pending checkpoints).


chain-signatures/node/src/storage/checkpoint_storage.rs
```diff
@@ -187,117 +197,103 @@ impl CheckpointStorage {
...
+                        redis.call('SET', KEYS[3], pending)
+                        local num = tonumber(height)
+                        local index = redis.call('HGETALL', KEYS[2])
+                        for i = 1, #index, 2 do
+                            if tonumber(index[i+1]) <= num then
+                                redis.call('HDEL', KEYS[1], index[i+1])
+                                redis.call('HDEL', KEYS[2], index[i])
+                            end
                         end
+                        return { true, redis.call('HLEN', KEYS[1]) }
```
yvonneanne: The success branch returns two elements instead of a three-tuple. This is what cases::store::test_pending_checkpoint_persistence is failing on in CI: Array response of wrong dimension (response was array([int(1), int(1)])).

https://github.com/sig-net/mpc/actions/runs/33729188460/job/100565008559?pr=1202#step:20:72

chain-signatures/node/src/stream/supervisor.rs
```diff
@@ -152,7 +166,16 @@ async fn run_supervised_with_watchdog<I: ChainIndexer, T: ChainTelemetry>(
...
         )
-        .await;
+        .await
+        {
+            tracing::error!(
+                %chain,
+                %err,
+                "failed to recover backlog; retrying in {ERROR_RESTART_DELAY:?}"
+            );
+            tokio::time::sleep(ERROR_RESTART_DELAY).await;
```
yvonneanne: Iiuc, if Redis is unavailable or the checkpoint keys are unreadable, recover_backlog fails immediately rather than after doing work, so each chain's supervisor spins at 1 Hz for the whole outage: a storage round trip and an error log per chain per second, for as long as it lasts.

Is that on purpose?

yvonneanne (commented):
Pushing promote-and-trim into a single Lua script is a good direction.

Afaics, some in-memory duality is left, though: `pending_counts` mirrors `storage.pending_count()` with nothing spanning the two. confirm commits in Redis and then writes the counter, and since `detect_regression` runs inside a `select`! branch it can be dropped in between, leaving the counter high so `has_slot` stalls dispatch early. That window predates this PR; what's new is that the only unconditional re-derivation, Checkpoints::hydrate, runs once per process because `recover_backlog` gates it on `load_local`. After that the counter is corrected only as a side effect of the next uncancelled confirm.

Since `pending_count` is an `HLEN`, I'd move `backlog.checkpoints().hydrate(chain)` out and leave `recover_by_checkpoint` inside it. Drift then self-heals on every restart, rather than needing every path that touches the counter to be cancel-safe.

chain-signatures/node/src/storage/checkpoint_storage.rs
ChaoticTempest: nice catch, fixed

chain-signatures/node/src/stream/supervisor.rs
ChaoticTempest: hmm, yeah probably once a second is too fast. Added expontential backoff.

yvonneanne (approved):

PR #1207, opened 2026-09-04
chore: refactor node local signing progress

ChaoticTempest:
Separate out SignStatus to be able to cleanly see that we have node local vs on chain sign progression.

```
pub enum SignStatus {
    PendingGeneration,
    PendingPublish { .. },
    PendingExecution { .. },
    PendingGenerationBidirectional,
    PendingPublishBidirectional { .. },
}
```
is now purely represented for on chain sign tracking:
```
pub enum SignStatus {
    Sign(SignProgress),
    Bidirectional(BidirectionalProgress),
}
```
where we have the following node local progress tracking:
```
pub enum SignProgress {
    Generating,
    Publishing(Arc<PublishState>),
}

pub enum BidirectionalProgress {
    Initial(SignProgress),
    Executing(Arc<BidirectionalTx>),
    Final {
        respond_request: Arc<IndexedSignRequest>,
        progress: SignProgress,
    },
}
```


volovyks (approved):
I like the new state machine. Makes much more sense.
What is the plan for migration? Would we be able to remove Legacy code after that?

P.S. Such changes are pretty big in the state of code freeze. Please merge only when DEV is 100% green and observe how the PR affects the network. Revert in case of any issues.

ChaoticTempest:
@volovyks migration is automatic. We can remove migration code after this lands on testnet/devnet

yvonneanne (commented):
I like the direction, keeping the original request instead of overwriting it at phase 2 is a real improvement. And I agree with Serge that this is a rather large change considering we have a freeze. 

Afaics, there's no down-migration and no version bump, so a revert hits an unknown SignStatus variant in the same v14 namespace, recover_backlog fails, and supervisor.rs:163 retries forever without starting the indexer. Same for a mixed-version window: an un-upgraded node fetching a checkpoint from an upgraded peer can't decode the body, gets None from every peer, and loops in find_consensus_checkpoint...
Regardless of how you fixed it, please put the revert procedure in the PR body.

yvonneanne:
Instead of initial and final, which are both not very descriptive maybe we could use other terms?

What do you think of the following?
```
SigningTransaction(SignProgress)
AwaitingExecution(Arc<BidirectionalTx>)
SigningResponse { respond_request, progress }
```

ChaoticTempest:
@yvonneanne `SigningTransaction` and `SigningResponse` makes it too ambiguous what state we are in in bidirectional requests. There's actually an `initial` signing transaction and then a `final` signing transaction. That's why I had it as initial, then executing, then final.  

For reverts to the PR and serialization state, we wouldn't be able to do a revert on the PR/branch itself, but rather would need the other side of doing `from = NewestSignEntry` on the serde deserialization. Adding revert code will just bloat the PR itself, so I would leave it out. Any sort of serialization change on revert I'd assume would require additional steps to revert. Even our bumping of versions don't give us a clear cut way to do a pure revert on the PR/branch itself since our checkpoints have advanced. So for serialization change reverts, we'd need to undo the changes and also apply a migration code change on top.

But good news is that I already tested the migration here on whatever checkpoint data we have on devnet/testnet and it works.

yvonneanne (approved):

PR #1211, opened 2026-09-04
fix: require --sign-sk instead of falling back to account_sk

volovyks:
Technically a breaking change, but it is not safe to reuse the key, and I hope we are not using it in practice.
cc @auto-mausx


ChaoticTempest:
this looks good to me pending review from @auto-mausx to see if there are any breaking changes with it

auto-mausx:
This is the default setup, to use the sk, however it is possible to miss it, so I cleaned it up in this PR https://github.com/sig-net/mpc-infrastructure/pull/23

This PR won't break anything, but to cover our bases, the one I linked will make it safer for future partners.

auto-mausx (approved):

PR #1215, opened 2026-09-04
fix(solana): accept versioned transactions in block fetches

isSerge:
This should fix current problem with Solana:
```
failed to fetch Solana block; retrying in 10.3s
   err: RPC response error -32015: Transaction version (1) is not supported by the
        requesting client. Please try the request again with:
        "maxSupportedTransactionVersion": 1
   attempt=32 maxAttempts=u64::MAX
   ```
   
`SolanaClient::block_fetch_config()` declared `max_supported_transaction_version: Some(0)` (client parses legacy and v0 transactions only). Error is deterministic for a given block - the retry loop (`retry_until_ok` unbounded attempts) spun on it indefinitely

### DIFF
--- a/chain-signatures/chain-solana/src/client.rs
+++ b/chain-signatures/chain-solana/src/client.rs
@@ -182,7 +182,7 @@ impl SolanaClient {
             transaction_details: Some(TransactionDetails::Full),
             rewards: Some(false),
             commitment: Some(CommitmentConfig::finalized()),
-            max_supported_transaction_version: Some(0),
+            max_supported_transaction_version: Some(u8::MAX),
         }
     }
 
@@ -612,7 +612,7 @@ mod tests {
             config.commitment.map(|c| c.commitment),
             Some(solana_sdk::commitment_config::CommitmentLevel::Finalized)
         );
-        assert_eq!(config.max_supported_transaction_version, Some(0));
+        assert_eq!(config.max_supported_transaction_version, Some(u8::MAX));
     }
 
     #[tokio::test]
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: chain-signatures/chain-solana/src/client.rs [hunk 1]
```diff
@@ -179,7 +179,7 @@ impl SolanaClient {
             transaction_details: Some(TransactionDetails::Full),
             rewards: Some(false),
             commitment: Some(CommitmentConfig::finalized()),
-            max_supported_transaction_version: Some(0),
+            max_supported_transaction_version: Some(u8::MAX),
```
volovyks: Does not look as a good practice. What if future version brings something we can not parse?

Also, do we need to limit retries instead?

isSerge: this error should not be retried since it's deterministic - this change was planned as follow up PR after current issue is resolved

>  What if future version brings something we can not parse?

Not sure pinning specific version will prevent this (network will move past it eventually)

ChaoticTempest: the issue is even if we make it max, when we eventually hit something we cannot parse, we'd be stuck at the parsing area anyways or should be. I think for now, probably making targeted towards the version we have should be correct. Otherwise, we could potentially skip the block entirely. So if we make it max, we should definitely make our parsing logic for solana does not skip the block, and simply halts or retries

isSerge: Updated to 1

volovyks (approved):

PR #1217, opened 2026-09-04
feat(loadtests): k6 driver for the bidirectional round trip

ppca:
Adds a k6 driver for the Solana → Ethereum bidirectional round trip, and a manual workflow to run it.

**Why a new script.** `/ping` is synchronous, so `http_req_duration` measures the sign request. `POST /sign_bidirectional` returns 202 with a job id in milliseconds and settles 30+ mins later — the existing thresholds would pass while every round trip failed. One iteration submits, then polls the job to a terminal state.

Phase timings (lease, signature, confirmation, respond, total) come from the job's own record, not wall-clock around polling. Failures carry the service's `failureReason` as a tag; rejections are counted separately and split by cause, so a saturated service doesn't read as a broken one. `setup()` aborts if any derived address is underfunded.

**Workflow is `workflow_dispatch` only.** Every job spends gas, and a cadence is also a claim about respond timings that no green run has established yet. `k6 run` is the last command in its step, so thresholds decide the verdict.

**Also fixes** `fail` never being imported in `k6-load-test.js`; it threw `ReferenceError` on every failed validation, visible in current scheduled runs.

Validated against dev. The MPC was down, so the failure path is exercised and the success path is not.


.github/workflows/k6-bidirectional-loadtest.yml
```diff
@@ -0,0 +1,129 @@
...
+# also a claim about respond-leg timings that no green run has established yet;
+# measure first, then decide the cadence and the thresholds together.
+on:
+  workflow_dispatch:
+    inputs:
+      lt_strategy:
+        description: Arrival rate (per minute — the service caps at 10)
+        type: choice
+        options:
+          - rate_1_min
```
volovyks: Can we follow the same naming we already have? (rpc_*)

.github/workflows/k6-bidirectional-loadtest.yml
```diff
@@ -0,0 +1,129 @@
...
+      environment:
+        description: MPC network
+        type: choice
+        options:
+          - dev
+          - testnet
+        default: dev
+        required: true
+      lt_mode:
+        description: Transaction mode
```
volovyks: Why do we need this mode in the context of a Load test?

.github/workflows/k6-bidirectional-loadtest.yml
```diff
@@ -0,0 +1,129 @@
...
+        required: true
+
+permissions:
+  contents: read
+
+# The derived addresses are shared state: two runs would lease the same ones and
+# collide on their nonces. Queued rather than cancelled, since cancelling
+# mid-flight abandons jobs whose Ethereum transaction is already paid for.
+#
+# This only serialises runs of this workflow. The same addresses are also
```
volovyks: I believe most of the comments in this PR can be 1 or 2 lines

volovyks (commented):
There is no output to Grafana. Is that in the plans?

We may consider moving Grafana reporting to the pinger. But that needs investigation.
CI becomes a lightweight traffic generator.

.github/workflows/k6-bidirectional-loadtest.yml
ppca: changed to rpm

.github/workflows/k6-bidirectional-loadtest.yml
ppca: removed, as for a load test this is not necessary

.github/workflows/k6-bidirectional-loadtest.yml
ppca: trimmed

ppca:
> There is no output to Grafana. Is that in the plans?
> 
> We may consider moving Grafana reporting to the pinger. But that needs investigation. CI becomes a lightweight traffic generator.

planning to do output to grafana as a followup. there might be some issues using Grafana Cloud k6 for such long running test. I'll investigate and also investigate the option of having pinger report to pinger.

And seems like our grafana dashboard for k6 load test `<grafana_link>` is not showing anything for the past runs. will look into that too.

volovyks (approved):

PR #1220, opened 2026-09-05
feat(midnight): cache execution outputs in GCS

Pessina:
Midnight's Compact circuits require byte-payload sizes at compile time. Serialized execution outcomes vary in size across requests, so the shared response contract cannot know their size when compiled. Add optional GCS storage for the exact attested bytes off-chain before publishing the execution signature on-chain.

- Keep storage configuration scoped to Midnight and verify the full flow with a local Docker emulator.
- The bucket name and output-serving URL should include `cache`.

Validated: full local Midnight → Ethereum → Midnight flow, 581 Rust library tests, 41 TypeScript tests, build, type checks, lint, and formatting.

### DIFF
--- a/chain-signatures/chain-midnight/src/config.rs
+++ b/chain-signatures/chain-midnight/src/config.rs
@@ -57,6 +57,13 @@ impl Default for IndexerConfig {
 /// Runs the out-of-process intent builder, the piece that stays TypeScript for its proving stack.
 #[derive(Clone, PartialEq)]
 pub struct PublisherConfig {
+    /// Optional GCS bucket; when absent, final responses only publish on-chain.
+    pub output_storage_bucket: Option<String>,
+    /// Object namespace; use a distinct prefix for deployments that reset chain state.
+    pub output_storage_prefix: String,
+    pub output_storage_timeout: Duration,
+    #[cfg(feature = "sandbox")]
+    pub output_storage_emulator_endpoint: Option<String>,
     /// argv of the builder, program first: a list so no operator path is word-split.
     pub intent_gen_command: Vec<String>,
     /// Funds respond transactions.
@@ -77,6 +84,11 @@ pub struct PublisherConfig {
 impl Default for PublisherConfig {
     fn default() -> Self {
         Self {
+            output_storage_bucket: None,
+            output_storage_prefix: "v1".to_string(),
+            output_storage_timeout: Duration::from_secs(30),
+            #[cfg(feature = "sandbox")]
+            output_storage_emulator_endpoint: None,
             // The `bin` name the TypeScript package installs, resolved on the fixed child PATH.
             intent_gen_command: vec!["midnight-publisher".to_string()],
             funding_seed: String::new(),
@@ -102,6 +114,9 @@ impl Default for PublisherConfig {
 impl fmt::Debug for PublisherConfig {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("PublisherConfig")
+            .field("output_storage_bucket", &self.output_storage_bucket)
+            .field("output_storage_prefix", &self.output_storage_prefix)
+            .field("output_storage_timeout", &self.output_storage_timeout)
             .field("intent_gen_command", &self.intent_gen_command)
             .field("funding_seed", &"<redacted>")
             .field("proof_server_url", &self.proof_server_url)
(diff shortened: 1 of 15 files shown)


chain-signatures/chain-midnight/src/output_storage.rs
```diff
@@ -0,0 +1,315 @@
...
+    timeout: Duration,
+}
+
+impl GcsOutputStore {
+    pub(crate) async fn connect(
+        config: &PublisherConfig,
+        network_id: &str,
+        central_address: MidnightAddress,
+    ) -> anyhow::Result<Option<Self>> {
+        config.validate_output_storage()?;
```
isSerge: Can we include `validate_output_storage` into `MidnightConfig::validate()`?

chain-signatures/chain-midnight/src/config.rs
```diff
@@ -102,6 +114,9 @@ impl Default for PublisherConfig {
 impl fmt::Debug for PublisherConfig {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
         f.debug_struct("PublisherConfig")
+            .field("output_storage_bucket", &self.output_storage_bucket)
+            .field("output_storage_prefix", &self.output_storage_prefix)
+            .field("output_storage_timeout", &self.output_storage_timeout)
```
isSerge: should `Debug` also include `output_storage_emulator_endpoint`?

integration-tests/src/gcs.rs
```diff
@@ -0,0 +1,91 @@
...
+            .start()
+            .await
+            .context("starting the GCS emulator")?;
+        let mut emulator = Self {
+            endpoint: String::new(),
+            bucket: format!("mpc-test-{}", uuid::Uuid::new_v4()),
+            container,
+            client: Client::builder().timeout(Duration::from_secs(5)).build()?,
+        };
+        emulator.refresh_endpoint().await?;
```
isSerge: can we set `endpoint` right away without calling `refresh_endpoint`?

isSerge (approved):
Nothing blocking, just left a few questions

integration-tests/src/gcs.rs
Pessina: Addressed here: https://github.com/sig-net/mpc/pull/1220/commits/1041c1f8ee5fd460a6806e43b51cee1d47611a14

chain-signatures/chain-midnight/src/output_storage.rs
Pessina: Addressed here: https://github.com/sig-net/mpc/pull/1220/commits/1041c1f8ee5fd460a6806e43b51cee1d47611a14

chain-signatures/chain-midnight/src/config.rs
Pessina: Addressed here: https://github.com/sig-net/mpc/pull/1220/commits/1041c1f8ee5fd460a6806e43b51cee1d47611a14

chain-signatures/chain-midnight/src/config.rs
```diff
@@ -57,6 +57,13 @@ impl Default for IndexerConfig {
 /// Runs the out-of-process intent builder, the piece that stays TypeScript for its proving stack.
 #[derive(Clone, PartialEq)]
 pub struct PublisherConfig {
+    /// Optional GCS bucket; when absent, final responses only publish on-chain.
+    pub output_storage_bucket: Option<String>,
+    /// Object namespace; use a distinct prefix for deployments that reset chain state.
+    pub output_storage_prefix: String,
+    pub output_storage_timeout: Duration,
+    #[cfg(feature = "sandbox")]
+    pub output_storage_emulator_endpoint: Option<String>,
```
ppca: nit:is it better to wrap all output_ params in one object naming sth like  output_storage_config, shape like
struct OutputStorageConfig {
..
}
where each field inside is required, and output_storage_config field is optional?

ppca (approved):
looks good! just want to check: how do we plan to have consumers who receive RespondBidirectional event to get the output bytes? Are they getting it from the GCS store too?

Pessina:
> looks good! just want to check: how do we plan to have consumers who receive RespondBidirectional event to get the output bytes? Are they getting it from the GCS store too?

Yes, that's correct. We will expose a URL and they fetch directly from GCS. 

It's ok to do so because this value can be fetched from other sources, such as the client runs `debug_traceTransaction` on their own and get the value. So we publish on GCS to make it easy for them, but that's not mandatory from our side. 

cc: @ppca

chain-signatures/chain-midnight/src/config.rs
Pessina: Fixed: https://github.com/sig-net/mpc/pull/1220/commits/8cf62cb16d8d19a7ba46eafbe086dd1d3cf98fd0

Issue #992, opened 2026-07-13
canton: publish failure due to 404 contract not found


ppca:
[log](https://console.cloud.google.com/logs/query;query=SEARCH%2528%22canton%22%2529%0Aseverity%3D%22ERROR%22%0AjsonPayload.message%3D%22canton:%20failed%20to%20publish%20signature%22;lfeCustomFields=jsonPayload%252Ftarget;cursorTimestamp=2026-07-13T07:22:42.653534239Z;duration=P7D?project=near-cs-dev)

<img width="1022" height="260" alt="Image" src="https://github.com/user-attachments/assets/fdef1482-7446-444f-ab11-9703a1b80e07" />

ppca:
for example for this request, none of the publish were successful(log `<gcp_link>`)
```
jsonPayload.signId="SignId(\"ecb6b3b4573d054b1e0fed8c87c33719fa61db38173dea3a3abbadc48ff69059\")"
```

PR #1221, opened 2026-09-05
fix(canton): stop retrying accepted duplicate commands

Pessina:
Stop signature submission retries when Canton returns HTTP 409 with structured DUPLICATE_COMMAND after accepting an earlier submission. This prevents unnecessary retries from later surfacing as CONTRACT_NOT_FOUND.

- Keep other errors and verified ledger-event completion unchanged.
- Cover both response phases, timeout-to-duplicate retries, and publication metrics.

Related to #992.

### DIFF
--- a/chain-signatures/chain-canton/src/client.rs
+++ b/chain-signatures/chain-canton/src/client.rs
@@ -4,7 +4,7 @@ use crate::{
     daml::{CantonSignature, EcdsaSigData},
     ledger_api::{
         ActiveContractEntry, Command, CumulativeFilter, EventFormat, GetActiveContractsRequest,
-        IdentifierFilter, JsCommands, LedgerEndResponse, PartyFilter,
+        IdentifierFilter, JsCantonError, JsCommands, LedgerEndResponse, PartyFilter,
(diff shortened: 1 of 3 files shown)


chain-signatures/chain-canton/src/client.rs
```diff
@@ -229,8 +247,19 @@ impl CantonClient {
...
+                    if status == reqwest::StatusCode::CONFLICT
+                        && serde_json::from_str::<JsCantonError>(&body)
+                            .is_ok_and(|error| error.code == "DUPLICATE_COMMAND")
+                    {
+                        // Acceptance ends submission retries; verified ledger events still settle the request.
+                        return Ok(SubmissionOutcome::AlreadyAccepted);
+                    }
+                    anyhow::bail!("{context} failed: {status} {body}");
+                }
+                Ok(SubmissionOutcome::Transaction(resp.json().await?))
```
isSerge, line 262: I need to enhance retry RPC macros so we can provide chain-specific non-retryable errors
https://github.com/sig-net/mpc/issues/1222

isSerge (approved):

Issue #1165, opened 2026-08-24
Github workflow: Add `pull_request` `path` filters to skip redundant runs


isSerge:
Currently every PR has to run all workflows no matter what is changed and this is going to get worse as we add more chains. We could skip redundant runs by adjusting filters for each workflow

volovyks:
Agree. I was thinking what is best - nightly tests ot this. I'm totally fine with choosing the `path` approach.
Note: we need to make sure that any change to the interface triggers all the tests.

volovyks:
Appeared to be hard. We may get back to this once Chain integrations are 100% separated.

PR #1225, opened 2026-09-07
feat(ci): enhance PR path filters for Canton and Midnight workflows

isSerge:
Problem is described in #1165 

This PR targets Canton and Midnight, because their setup is expensive (might do the same for Solana and Ethereum next, need to split them between separate jobs).

### Changes

Added `pull_request.paths` filters to the three chain workflows:

- `canton.yml`
- `midnight.yml`
- `midnight-publisher-ts.yml`

Each one is now triggered if chain specific or shared core paths affected (previously run on every PR). Shared core includes:

- `chain-integration-core`
- `primitives`
- `crypto`
- `keys`
- `utils`
- `compact-hashing`
- `node`
- `signet-crypto`
- `signet-primitives`
- `Cargo.toml/Cargo.lock`
- `justfile`
- `.config/nextest.toml`

Note: `merge_group` does not support filtering, so it will always run everything (safety net)

### DIFF
--- a/.github/workflows/canton.yml
+++ b/.github/workflows/canton.yml
@@ -7,7 +7,40 @@ on:
     paths:
       - integration-tests/**
       - chain-signatures/**
+  # PRs run only on Canton or shared-core changes; merge_group stays
+  # unfiltered as the safety net. Keep core in sync with other chain workflows.
   pull_request:
+    paths:
+      - ".github/workflows/canton.yml"
+      # Shared core: changes here trigger every chain workflow.
+      - "chain-signatures/chain-integration-core/**"
+      - "chain-signatures/primitives/**"
+      - "chain-signatures/crypto/**"
+      - "chain-signatures/keys/**"
+      - "chain-signatures/utils/**"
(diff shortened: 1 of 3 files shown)


volovyks (commented):
Nice change

We may avoid setup duplication using something like `dorny/paths-filter` (not tested).

We need to update the PR required jobs list (strangely, all the jobs have now been executed; the filters probably do not work as expected).

The list of paths may be extended. For example, with changes in integration tests themselves.

isSerge:
@volovyks I was planning to use `dorny/paths-filter` for a single gate job in the following PRs.

All jobs were triggered because all three paths are modified in this PR. I believe this is how it should work - we should validate workflow if we change it.

volovyks:
@isSerge, what about paths like `integration-tests/src/canton.rs`?

isSerge:
it's covered here https://github.com/sig-net/mpc/pull/1225/changes#diff-2541d7c7784711aaee860040e5e96df5b5b14a725a622c5eee4b5aeb5a176239R31

volovyks (approved):

PR #1232, opened 2026-09-07
refactor: move publish from generation

volovyks:
`SignGenerator::run` now returns the `FullSignature`; the backlog and RPC hand-off moves to `GeneratingPhase::hand_off` in `request/task.rs`. The generator no longer knows about `RpcChannel` or `Backlog`.

- `GenerateCtx` loses `rpc` and `backlog`.
- Signature is reconstructed once and passed to both the backlog and `rpc.publish_signature`; the duplicate check in `RpcChannel::publish` is gone with the method (no other callers).
- Docs on `SignPhase::Complete` and `hand_off` state that `Ok` means handed off, not on chain.

No behaviour change.

### DIFF
--- a/chain-signatures/node/src/protocol/request/task.rs
+++ b/chain-signatures/node/src/protocol/request/task.rs
@@ -30,7 +30,10 @@ pub enum SignPhase {
     Posit(PositPhase),
     /// Commit the reserved presignature and run the signing protocol to completion.
     Generating(GeneratingPhase),
-    /// Terminal: the request finished (`Ok`) or aborted (`Err`).
+    /// Terminal. `Ok` means generation finished and the signature was handed
+    /// off to the backlog and the RPC queue; it does not mean a response is on
+    /// chain, which the backlog state and the chain streams track. `Err` means
+    /// the request aborted.
     Complete(Result<(), SignError>),
 }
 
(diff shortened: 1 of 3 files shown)


chain-signatures/node/src/protocol/request/task.rs
```diff
@@ -26,7 +30,10 @@ pub enum SignPhase {
     Posit(PositPhase),
     /// Commit the reserved presignature and run the signing protocol to completion.
     Generating(GeneratingPhase),
-    /// Terminal: the request finished (`Ok`) or aborted (`Err`).
+    /// Terminal. `Ok` means generation finished and the signature was handed
+    /// off to the backlog and the RPC queue; it does not mean a response is on
+    /// chain, which the backlog state and the chain streams track. `Err` means
+    /// the request aborted.
```
yvonneanne, line 36: The doc says Ok means handed off to the backlog **and** the RPC queue, but hand_off returns () and swallows both failure paths: the early return on reconstruction failure, and the warn!-only mark_publishing error. Reconstruction failure currently reports success, so handle_task_exit logs "completed successfully" and the duration metric records a success for a request that produced nothing. Pre-existing, but this PR is what writes the invariant down. Rephrase (and create an issue to fix it properly or fix it in this PR?).

chain-signatures/node/src/protocol/request/task.rs
```diff
@@ -112,11 +119,65 @@ impl GeneratingPhase {
         };
 
         match result {
-            Ok(()) => SignPhase::Complete(Ok(())),
+            Ok(output) => {
+                self.hand_off(ctx, state, output).await;
+                SignPhase::Complete(Ok(()))
+            }
```
yvonneanne, line 125: Minor: hand_off is awaited outside the select!. Before, mark_publishing ran inside generator.run on the &mut generation branch, so a late Propose still got a Reject from mailbox.recv(). Now that peer waits out its posit timeout on a request we've already published. mark_publishing's write lock queues behind publishable_requests, which scans the whole map under a read lock on every block, so the window scales with backlog size.

yvonneanne (approved):
I like it! 
Dropping the second reconstruction is a good catch. 

Nit: chain-integration-core/src/publish.rs:32
PublishAction::new now has no production caller anymore, so the validate-before-publish gate no longer sits at the publish boundary. Fine today, since both producers trace to the same validated reconstruction. Delete it with its tests, or make it the only constructor.

PR #1233, opened 2026-09-07
refactor: resolve the presignature in GeneratingPhase

volovyks:
SignGenerator::new takes a PresignatureTaken. The proposer/deliberator
split (commit our reservation, or wait for our copy in storage) lives in
GeneratingPhase::take_presignature; PendingPresignature is removed.

### DIFF
--- a/chain-signatures/node/src/protocol/request/task.rs
+++ b/chain-signatures/node/src/protocol/request/task.rs
@@ -7,10 +7,14 @@ use super::posit::PositPhase;
 use super::state::SignState;
 use super::*;
 
+use crate::storage::presignature_storage::PresignatureTaken;
+
 /// Generating phase — see [`SignPhase::Generating`].
 pub struct GeneratingPhase {
     pub proposer: Participant,
     pub presignature_id: PresignatureId,
+    /// Our reservation when we are the proposer; `None` for a deliberator,
+    /// whose copy is taken from storage when generation starts.
     pub presignature: Option<PresignatureReservation>,
     pub accepted_participants: Vec<Participant>,
 }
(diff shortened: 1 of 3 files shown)


chain-signatures/node/src/protocol/request/task.rs
```diff
@@ -117,6 +111,35 @@ impl GeneratingPhase {
...
+                interval.tick().await;
+                if let Some(taken) = ctx.presignatures.take(id, self.proposer).await {
+                    break taken;
+                }
+            }
+        })
+        .await
+        .map_err(|_| format!("timeout ({timeout:?}) waiting for presignature {id} to be available"))
+    }
+
```
yvonneanne, line 142: The justification for the 45s wait doesn't match the accept gate.

/// we accepted the Propose knowing only that the presignature exists or is still being generated

The check is ctx.presignatures.contains(presignature_id) (posit.rs:123), whic is true only once the presignature is fully stored. contains_generating exists but has no caller on the posit path. A deliberator therefore always accepted having seen the presignature stored, and "still being generated" can't be why it's missing later. The cases that actually reach the poll are: GC'd/expired in between, taken concurrently, or Redis unreachable (take returns None for all three, indistinguishably). None of those get better by waiting, so this is 45s of latency before an inevitable reorganize, while the proposer burns its own 45s in parallel. The "in storage or generating" phrasing at posit.rs:122 is stale too. Could we either shorten the wait for those cases or reword to say what it's really covering?

yvonneanne (approved):
Less I/O, I like it.

Two comments:

1. The tasks.rs docstring's justification for the 45s wait doesn't match the code (details inline).

2. take() isn't cancel-safe.  using.insert(id, mine) happens before the invoke_async await and is only removed on the error arms, so an abort (mod.rs:299/367/434) that lands mid-round-trip leaks the entry. Because take starts with if using.insert(...).is_some() { return None }, that id then fails every future take on this node. If the abort landed before the script ran server side, the presignature is still in Redis, still passes contains, still gets accepted in a later posit, and the deliberator polls 45s logging "taking artifact that is already in use" every 250ms before reorganizing. (pre-existing, not introduced here, fixable with a guard)

volovyks:
Good findings @yvonneanne ! This PR is a pure refactor—no logic changes. I will merge it and address the issues separately.

PR #1235, opened 2026-09-07
refactor: Sign spawner cleanups

volovyks:


chain-signatures/node/src/protocol/request/mod.rs
```diff
@@ -158,6 +158,35 @@ pub struct SignatureSpawner {
...
+    fn new(
+        node_account_id: near_account_id::AccountId,
+        contract: ContractStateWatcher,
+        presignatures: PresignatureStorage,
+        mesh_state: watch::Receiver<MeshState>,
+        msg: MessageChannel,
+        rpc: RpcChannel,
+        backlog: Backlog,
+        sync_report_tx: SyncReportSender,
+    ) -> Self {
```
ChaoticTempest, line 171: If the reason to add a `new` function is to do a clean up then IMO, we should have it be cleaned up to the point where we don't have `clippy::too_many_arguments`

ChaoticTempest (approved):
looks good to me besides one nit

PR #1237, opened 2026-09-07
refactor(posit): one round gate for sign posits

volovyks:
Do not review yet

Future-round messages wait in PositMailbox (recv_up_to) instead of a
second buffer on SignState, and the mailbox tracks the highest round
peers have shown us. SignState::gate classifies a message against our
round; recv_current answers stale ones with StaleRound. SignTask::reject
builds every posit reject. Removes the three copies of this logic from
wait_for_propose, the posit loop and reject_late_propose.


volovyks:
Closing for now.

PR #1250, opened 2026-09-08
ci: drop develop push

volovyks:
Test workflows ran three times per change: on the PR, in the merge queue, and again on push to develop. The third run is redundant: develop only advances through the queue (rulesets block direct pushes), and the queue fast-forwards develop to the exact commit it tested, so its checks are already attached to every develop commit. The push run only duplicated them and produced flaky reds (8 of the last 60 Integration runs on develop).

What was changed:
1. Drop the push-to-develop trigger from Unit, Integration, Canton, Midnight and Midnight Publisher TS. Deploy and load-test workflows keep theirs.
2. Save Rust caches from `merge_group` runs instead of develop pushes, so caches don't go stale. `nightly.yml` keeps its develop condition.
3. Document the policy in the `_ci-changes.yml` comment.

### DIFF
--- a/.github/workflows/unit.yml
+++ b/.github/workflows/unit.yml
@@ -1,20 +1,5 @@
 name: Unit
 on:
-  push:
-    branches:
-      - main
-      - develop
-    paths:
-      - ".github/workflows/unit.yml"
-      - "chain-signatures/**/*.rs"
-      - "chain-signatures/Cargo.*"
-      - "integration-tests/**/*.rs"
-      - "chain-signatures/contract-eth/**/*.sol"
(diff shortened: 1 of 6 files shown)


.github/workflows/unit.yml
```diff
@@ -1,20 +1,5 @@
 name: Unit
 on:
-  push:
-    branches:
-      - main
```
isSerge, line 5: is dropping `main` intended?

volovyks: yes, merge_queue is enabled for main and develop, so we will run tests anyway

isSerge (approved):

PR #1254, opened 2026-09-08
refactor(node): invert the signing delta once, and don't panic on zero

volovyks:
Computes the signing delta's inverse once instead of twice per signature, and rejects a zero delta as an InitializationError rather than unwrapping mid-signing.


ChaoticTempest (approved):

ChaoticTempest:
we might want to guarantee this never fails at the typelevel or actually just use `invert_vartime` since delta is a public parameter

volovyks:
This is a cryptographically impossible situation. But it is better to have some error handling. Should be enough.

PR #1256, opened 2026-09-08
fix(solana): propagate failed respond-event sends

volovyks:
Makes the two respond-event sends in `emit_events` propagate with `?` like the sign arm above them, so a closed channel surfaces instead of the indexer silently dropping respond events.


isSerge (approved):

ppca (approved):

ChaoticTempest (commented):
it's a good step that we're propagating the error, but we should properly handle the case actually. The `SendError` here is a shutdown signal, so we need to be able shutdown on each run_supervised loop iteration.

It looks like this will constantly be retried until successful which could be wasteful

PR #1258, opened 2026-09-08
refactor(node): drop the constant latest_block_height from /state

volovyks:
Removes `latest_block_height` from the `/state` response: it has been hardcoded to 0 since #594 and nothing reads it.


ppca (approved):

ChaoticTempest (commented):
this is going to be a breaking change. Old nodes are going to error out on querying `/state` on new nodes, but new nodes querying old nodes will just ignore the field

volovyks:
@ChaoticTempest nodes are not using this endpoint. You are probably thinking about `/status`, and this is `/state`. Only used for debugging.

PR #1263, opened 2026-09-09
feat(backlog): request lifecycle transitions via typestate

### LINKED ISSUE #779: Backlog should make it easier to transition to new sign request states
Currently, backlog has advance which purely does one state transition from AwaitingResponse to PendingExecution which does not cover the whole entire transition tree we need to represent persistently.

So, I propose we do something like a typestate where we represent in code a set of rust types that allow us to transition from one state to another, but in the background, we're just calling Backlog::set_state. The typestate would enforce the transition graph for us, so there's no complication underneath where things are persisted.

ChaoticTempest:
## Overview
Part 2 of the typestate refactoring stack (closes #779).

This PR introduces the **typestate pattern** (`SignEntry<State>`) to govern MPC request lifecycles at compile-time. Instead of mutating backlog entries via ad-hoc status checks and runtime error paths, `SignEntry` is now the single source of truth for all request state transitions and mutations across protocol signing and RPC publishing.

Stacked on top of #1261 (`phuong/refactor/backlog-decompose`).

### 1. Typestate (`backlog/request.rs`, `backlog/mod.rs`)
- Introduced `SignEntry<State>` with strongly-typed lifecycle stages:
  $$\text{Generating} \longrightarrow \text{Publishing} \longrightarrow \text{Executing} \longrightarrow \text{Final}\langle\text{Generating}\rangle \longrightarrow \text{Final}\langle\text{Publishing}\rangle$$
- State transitions are encoded directly on `SignEntry` via linear consumption:
  - `entry.advance(...)`
  - `entry.complete().await`
  - `entry.remove().await`
- `Backlog::insert`, `Backlog::get`, `requeueable_requests`, and `publishable_requests` now return typed `SignEntry` handles.
- Eliminated dead `PublishState` enum; published state fields are tracked directly in `Publishing`.
- Removed obsolete runtime `BacklogError` variants made impossible by the type system.

### 2. Protocol Signing & RPC Pipeline Integration
- `types::SignCommand::Request` now wraps `SignEntry<Generating>`.
- Protocol signature generation spawner (`protocol/signature.rs`) strictly accepts `SignEntry<Generating>` and advances directly to `SignEntry<Publishing>`.
- RPC publishing queue consumes `SignEntry<Publishing>` upon on-chain confirmation.

### 3. Test Fixtures & Backlog Tests Modernization (`backlog/mock.rs`, `backlog/tests.rs`)
- Added `BacklogTestExt` trait providing `insert_mock_sign`, `insert_mock_bidirectional`, `insert_mock_executing`, and `insert_mock_final`.
- Updated `backlog/tests.rs` to validate all state transitions via the typed API.


chain-signatures/node/src/backlog/request.rs
```diff
@@ -0,0 +1,795 @@
...
+            })?;
+        match &mut entry.status {
+            SignStatus::Sign(progress)
+            | SignStatus::Bidirectional(BidirectionalProgress::Initial(progress))
+            | SignStatus::Bidirectional(BidirectionalProgress::Final { progress, .. }) => {
+                *progress = SignProgress::Publishing(publish);
+            }
+            SignStatus::Bidirectional(BidirectionalProgress::Executing(_)) => {}
+        }
+        Ok(())
```
yvonneanne: this can overwrite an existing Publishing (resetting publishing_since and publish_dispatched, which pushes out the failover deadline) and returns Ok without writing anything when the entry is Executing, so advance hands back a Publishing handle no backlog entry holds.

```suggestion
let progress = match &mut entry.status {
    SignStatus::Sign(p)
    | SignStatus::Bidirectional(
        BidirectionalProgress::Initial(p) | BidirectionalProgress::Final { progress: p, .. },
    ) => p,
    SignStatus::Bidirectional(BidirectionalProgress::Executing(_)) => {
        return Err(BacklogError::InvalidPublishingTransition)
    }
};
if !matches!(progress, SignProgress::Generating) {
    return Err(BacklogError::InvalidPublishingTransition);
}
*progress = SignProgress::Publishing(publish);
Ok(())
```

chain-signatures/node/src/backlog/mod.rs
```diff
@@ -726,34 +582,15 @@ impl StateManager for Backlog {
...
-    #[error("transaction not found")]
-    TransactionNotFound,
-    #[error("cannot advance sign request: status must be pending generation or publishing")]
-    InvalidAdvanceTransition,
-    #[error("cannot mark publishing: status must be pending generation")]
-    InvalidPublishingTransition,
-    #[error("cannot transition to bidirectional response: id must match and request must be RespondBidirectional")]
-    InvalidBidirectionalResponseTransition,
+    #[error("failed to reconstruct signature")]
+    InvalidSignature,
```
yvonneanne, line 586:
```suggestion
    InvalidSignature,
    #[error("invalid transition to publishing")]
    InvalidPublishingTransition,
```
error variant needed to prevent duplicate publishing was deleted

chain-signatures/node/src/types.rs
```diff
@@ -24,6 +25,44 @@ pub type Epoch = u64;
...
+        tokio::spawn(async move {
+            while let Some(cmd) = near_sign_rx.recv().await {
+                match cmd {
+                    mpc_chain_near::SignCommand::Request(req) => {
+                        let (entry, _) = backlog.insert(req).await;
+                        if sign_tx.send(Self::Request(entry)).await.is_err() {
+                            break;
+                        }
+                    }
+                    mpc_chain_near::SignCommand::Completion(id) => {
```
yvonneanne, line 54:
```suggestion
                    mpc_chain_near::SignCommand::Completion(id) => {
                        backlog.remove(Chain::NEAR, &id).await;
```
or are we removing it from the backlog else where?

yvonneanne (commented):
Nice! Typestate split makes a lot of sense.

Two comments in-line.

chain-signatures/node/src/types.rs
ChaoticTempest: this is a forwarder, since near uses a completely different `SignCommand`. This part has nothing to do with the backlog

chain-signatures/node/src/backlog/request.rs
ChaoticTempest: good catch, actually this should be within the typestate compilation itself. `record_publishing` will be moved out to only `SignProgress == Generating`

chain-signatures/node/src/types.rs
ChaoticTempest: but then again, near actually does insertion in the `SignCommand` here. Doesn't hurt to add `backlog.remove`

chain-signatures/node/src/backlog/request.rs
yvonneanne: The body is unchanged, so this doesn't add the check. <Generating> is what the handle was at creation, not what the entry is now: SignEntry::generating and Backlog::insert both don't consult storage, and it's then kept in SignatureSpawner::requests and cloned into every task incarnation and retry round.

Reachable case: a peer's respond event moves the entry from Generating to Executing (process_respond_event casts Initial<AnyProgress>), and no Completion is enqueued on that transition, so our leg-1 task keeps running. When it finishes, record_publishing silently does nothing and advance returns a SignEntry<Publishing> no entry backs, so publishable_requests never sees it and mark_publish_dispatched flips an atomic nobody holds.

Publishing behaviour is the same as develop. But develop warned here via InvalidPublishingTransition, and your Err case already breaks without publishing, so restoring the variant is a small win.

ChaoticTempest: ahh I see what you're saying. Changed it to your version

yvonneanne (approved):
Both comments addressed, approving.

Follow-up suggestion (pre-existing problem, probably better done in a separate PR):

advance_bidirectional_to_execution is the only respond-event path that
doesn't enqueue SignCommand::Completion (the plain-Sign path and
process_respond_bidirectional_event both do). So when a node sees the leg-1
respond event, its backlog entry goes to Executing but its own leg-1 task
stays in SignatureSpawner::requests, and since the task reorganizes into a
fresh round on failure rather than exiting, it can sit there indefinitely.

Leg 2 reuses the same SignId, so if that task is still in `requests` when
the execution confirmation arrives, handle_sign drops leg 2 as a duplicate
with no retry and the entry sits in Final(Generating) until a restart
requeues it.

PR #1264, opened 2026-09-09
refactor(stream): simplify execution watching and ops with typestate

### LINKED ISSUE #779: Backlog should make it easier to transition to new sign request states

ChaoticTempest:
## Overview
Part 3 of the typestate refactoring stack (closes #779).

Leverages the `SignEntry` typestate introduced in PR 2 to streamline `stream/ops.rs`, clean up bidirectional execution watching, and eliminate obsolete defensive runtime checks and boilerplate test harnesses across stream integration tests (-623 net lines).

Stacked on top of `phuong/feat/typestate-core`.

### 1. Stream Operations Simplification (`stream/ops.rs`)
- **`advance_bidirectional_to_execution`**: Replaced manual parameter extraction and transaction hashing with `event.to_bidirectional_tx(...)`, reducing the function to a linear ~25-line operation.
- **`unwatch_execution`**: Directly returns `Option<SignEntry<Bidirectional<Executing>>>` in a single operation, eliminating the secondary `backlog.get_by` lookup.
- **`process_execution_confirmed`**: Dropped redundant `sign_id` and `source_chain` arguments. Since the entry in the backlog is the authoritative source of truth, defensive mismatch checks (`unwatched_sign_id != sign_id` and `source_chain != pending_tx.source_chain`) have been completely removed.
- **`process_respond_bidirectional_event`**: Removed redundant completion warning check.

### 2. Execution Watcher Cleanup (`chain-ethereum/src/execution_watcher.rs`)
- Removed the private wrapper `struct ExecutionWatcher`; `ExecutionWatchers` now directly maps `BidirectionalTxId -> Arc<BidirectionalTx>`.

### 3. Test Suite Pruning (-942 lines removed)
- Deleted tests verifying runtime error handling for states and mismatches that are now strictly impossible at compile-time (e.g. `process_execution_confirmed_warns_but_still_uses_watcher_sign_id`).
- Converted `ops_tests.rs`, `stream_tests.rs`, `ethereum_stream.rs`, and `solana_stream.rs` to use `insert_mock_executing` and typed transitions.

### DIFF
--- a/chain-signatures/chain-ethereum/src/execution_watcher.rs
+++ b/chain-signatures/chain-ethereum/src/execution_watcher.rs
@@ -655,6 +655,32 @@ mod tests {
     use serde_json::json;
     use std::sync::Arc;
 
+    fn test_watcher_tx(
+        tx_hash: alloy::primitives::B256,
+        from_address: alloy::primitives::Address,
+        nonce: u64,
+    ) -> Arc<BidirectionalTx> {
+        Arc::new(BidirectionalTx {
+            id: BidirectionalTxId(tx_hash.0),
+            sender: [0u8; 32],
+            serialized_transaction: vec![],
+            source_chain: Chain::Solana,
+            target_chain: Chain::Ethereum,
+            caip2_id: "eip155:31337".to_string(),
+            key_version: LATEST_MPC_KEY_VERSION,
+            deposit: 0,
+            path: "m/44'/60'/0'/0/0".to_string(),
+            algo: "secp256k1".to_string(),
+            dest: Chain::Ethereum.to_string(),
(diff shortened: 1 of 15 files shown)


isSerge (approved):

yvonneanne (approved):
the refactor is good, the three redundancy removals look correct

PR #1267, opened 2026-09-09
feat(solana): add benchmarks

isSerge:
Replicates the `chain-ethereum` benchmark setup for Solana: a standalone bench binary driving the real indexer over a fixed historical slot range against a live RPC (Solana does not have eRPC proxy solution), with feature-gated timing/RPC instrumentation and a single summary report.

### Changes

- `src/bench.rs` - feature-gated module with per-method RPC counters, timing buckets and other setup
- `examples/bench_catchup.rs` +` examples/helpers/mod.rs` - standalone binary to run benchmarks, drives `SolanaIndexer` for specified historical range against real RPC endpoint, emits `Catchup Benchmark Report` on completion
- `README.md` - include crate description and baseline benchmark results for reference


volovyks (approved):

chain-signatures/chain-solana/examples/helpers/mod.rs
```diff
@@ -0,0 +1,106 @@
...
+
+    tracing::info!(start, end, "{label}: starting catchup");
+
+    let drain = tokio::spawn(async move {
+        while let Some(ev) = events_rx.recv().await {
+            tracing::debug!(?ev, "{label} drained event");
+        }
+    });
+
+    let blocks_stream = indexer.catchup_blocks(end, start).await?;
```
yvonneanne: Calling catchup_blocks directly instead of drain_range skips emit_block_markers_for_drained_inactive_slots and thus excludes backpressure. Is it like that on purpose?

yvonneanne (commented):
I like the timing decomposition!

chain-signatures/chain-solana/examples/helpers/mod.rs
isSerge: sharp point, it should use `drain_range` to match production

isSerge: updated, also added two marker-related counters

PR #1279, opened 2026-09-09
feat: indexer parser versioning for clearing outdated pending checkpoints

ChaoticTempest:
This adds in a new `INDEXER_PARSER_VERSION` const to the `ChainIndexer` interface. It is required to invalidate the pending checkpoints on making a parser change. So we need to bump this version going forward for any code changes to parsing on each chain's indexer.

This clearing out is managed by the supervisor and on seeing a new version, clears out the pending. Currently, since it is missing, it will not clear out any pending


volovyks (commented):
IMO, this PR patches a problem that needs a more fundamental solution.

chain-signatures/chain-integration-core/src/indexer.rs
```diff
@@ -11,6 +11,9 @@ use tokio_util::sync::CancellationToken;
 #[async_trait::async_trait]
 pub trait ChainIndexer: Send + Sync + 'static {
     const CHAIN: Chain;
+    /// Bump when this chain's indexer parsing changes and unconfirmed
+    /// checkpoints produced by the old parser must be dropped.
+    const INDEXER_PARSER_VERSION: u64 = 0;
```
yvonneanne, line 16: The trait default `= 0` plus an explicit `= 0` in all five chain impls: why both? A default means a new chain crate compiles without anyone thinking about this. Dropping it costs a one-liner in the four test indexers.

Separately, did you consider keying pending on something that changes automatically per build (binary version or git sha) instead of a hand-maintained const? Nothing enforces the bump, and a forgotten bump is silent. The cost is re-indexing from the confirmed checkpoint on every deploy. If that is cheap enough on devnet today, it removes the "remember to bump" burden entirely.

chain-signatures/node/src/backlog/checkpoints.rs
```diff
@@ -204,6 +204,34 @@ impl Checkpoints {
         Ok(count)
     }
 
+    /// Align durable pending checkpoints with the binary's indexer parser version.
+    /// Returns true when pending was evicted.
+    pub(super) async fn apply_parser_version(
```
yvonneanne, line 209: Can the following happen in a rolling deploy? The majority of old nodes confirms further old-parser checkpoints while the first upgraded node has already cleared its pending, which then regresses onto them via align_backlog_with_consensus.

yvonneanne (dismissed):
Mechanism looks correct and the tests cover the storage paths well.

chain-signatures/node/src/backlog/checkpoints.rs
yvonneanne: To say why I am asking: indexing resumes from the confirmed `processed_block_height` and never re-reads below it, so if those heights settle under the old parser they keep that parse permanently. Nothing later revisits them. That is what I want to be sure about, since the value of clearing pending is redoing blocks that were already indexed.

chain-signatures/node/src/stream/recovery.rs
```diff
@@ -24,6 +25,23 @@ pub(crate) async fn recover_backlog(
...
+    if load_local {
+        match backlog.apply_parser_version(chain, parser_version).await {
+            Ok(true) => {
+                tracing::info!(
+                    ?chain,
+                    "dropped unconfirmed pending checkpoints for re-parse after parser bump"
+                );
+            }
+            Ok(false) => {}
+            Err(err) => {
```
yvonneanne, line 39: Why swallow this when `hydrate` two lines below propagates a storage error with `?` against the same Redis? The case this arm actually catches is a partial failure, version loaded but `clear_pending` failed, and then the node runs with stale pending until someone restarts it again, which is the state this PR exists to avoid.

Suggest returning `Result<bool, CheckpointError>` from `apply_parser_version`, mapping storage errors the way `hydrate` and `regress` already do, and using `?` here. The supervisor then retries with its existing backoff, and `load_local` stays true so the check re-runs on the retry.

yvonneanne:
@volovyks I think you are right, and I hit a concrete instance of it while reviewing.

The const clears only unconfirmed pending, so it re-parses blocks above the last confirmed checkpoint and nothing below. In a rolling deploy the nodes still on the old binary can keep confirming blocks while the first upgraded node has already cleared its pending, and since indexing never re-reads below the confirmed height, those blocks keep the old parse for good. So it works when all nodes restart together and quietly does less than it looks like otherwise.

PR #1281, opened 2026-09-10
fix(protocol): announce posit START only after securing locally

yvonneanne:
**Problem**: All three posit paths broadcast START before the work that can still fail, so a proposer that drops out after announcing leaves peers acting on a round it has abandoned. Presignature and signature lose the artifact on every peer but the owner.

**Fix**: Do the local work first, announce START on success. A failure now ends the round before peers hear about it, leaving them on the deliberator expiry path they already handle.

This makes the waste one-time instead of recurring, it does not remove it: the artifact is now orphaned on the holders rather than destroyed there while the proposer keeps re-serving it. Nothing prunes an artifact whose owner no longer holds it.

### DIFF
--- a/doc/posit_state_machine.md
+++ b/doc/posit_state_machine.md
@@ -117,12 +117,13 @@ stateDiagram-v2
     state "Organizing" as OrgIn
     state "<b>Propose sent</b>
     1. tally ACCEPT and REJECT
-    2. once enough ACCEPTs send START to accepters" as ProposeSent
+    2. once enough ACCEPTs, commit the presignature
+    3. send START to accepters" as ProposeSent
     state "Generating" as GenOut
 
     OrgIn --> ProposeSent: proposer, PROPOSE sent
     ProposeSent --> GenOut: START sent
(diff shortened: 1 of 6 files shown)


doc/posit_state_machine.md
```diff
@@ -117,12 +117,13 @@ stateDiagram-v2
...
     1. tally ACCEPT and REJECT
-    2. once enough ACCEPTs send START to accepters" as ProposeSent
+    2. once enough ACCEPTs, commit the presignature
+    3. send START to accepters" as ProposeSent
     state "Generating" as GenOut
 
     OrgIn --> ProposeSent: proposer, PROPOSE sent
     ProposeSent --> GenOut: START sent
-    ProposeSent --> OrgIn: too many REJECTs, or timeout
+    ProposeSent --> OrgIn: too many REJECTs, timeout, or commit failed
```
volovyks: I thought `commit failed` is not possible, since the presignature can not dissapier wile "taken".

PR #1282, opened 2026-09-10
fix(solana): require solana validator funding and live slots before readiness

isSerge:
This should fix flaky `test_solana_stream_resumes_gap_free_after_outage`. 

Problem was that `wait_for_validator_ready` was resolving once RPC endpoints answered, `funded` was warn-only and there was no confirmation chain is advancing. This resulted in `transaction expiration and insufficient fee-payer funds` after restart.

Fix is to harden readiness path: `funded` is a requirement + add `wait_for_block_production` method

### DIFF
--- a/integration-tests/src/containers.rs
+++ b/integration-tests/src/containers.rs
@@ -1,7 +1,7 @@
 use std::collections::HashMap;
 use std::env;
 use std::path::PathBuf;
-use std::time::Duration;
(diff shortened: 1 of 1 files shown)


yvonneanne (approved):
Nice! Gating on funded plus proving slots advance is a much stronger readiness signal than "RPC answers"

PR #1285, opened 2026-09-10
fix(posits): take the deliberator presignature once instead of polling

volovyks:
Deliberators saw their presignature share stored at Accept, so a miss in Generating is final; take once and reorganize instead of polling for 45s. Storage issues should be solved at the storage layer.


yvonneanne (approved):
Good call to do this separately and fix the comments!

PR #1287, opened 2026-09-11
chore: fix reachable advisories, drop the `google-apis-rs`, migrate to `cargo-deny`

isSerge:
This PR fixes every reachable advisory that had an available patch, removes the unmaintained `google-apis-rs` dependency chain from the node, and replaces `cargo-audit` with feature-aware `cargo-deny` in CI


chain-signatures/node/src/protocol/message/inbox.rs
```diff
@@ -209,7 +209,7 @@ impl MessageInbox {
         while let Some((encrypted, timestamp)) = self.pending_decrypt.pop_front() {
             let decrypted: Result<(Participant, Vec<Message>), _> =
                 SignedMessage::decrypt_with(&encrypted, cipher_sk, participants, |sig| {
-                    if self.idempotent.put(sig.clone(), ()).is_some() {
+                    if self.idempotent.push(sig.clone(), ()).is_some() {
```
volovyks: [inbox.rs:212](vscode-webview://0djktbu94qo63qp3922bpiojkt7vdclturpfabdecfv329a3tr3l/chain-signatures/node/src/protocol/message/inbox.rs#L212) treats Some from push as a replay, but push also returns Some on capacity eviction. After 65,536 signatures the node rejects every new message until restart. lru 0.18 still has put, so revert all four call sites.

deny.toml
```diff
@@ -0,0 +1,42 @@
...
+    { id = "RUSTSEC-2026-0098", reason = "rustls-webpki 0.101.7 URI name constraints accepted; pinned by rustls 0.21 via helios' jsonrpsee-http-client 0.19 (hyper-rustls 0.24)" },
+    { id = "RUSTSEC-2026-0099", reason = "rustls-webpki 0.101.7 wildcard name constraints accepted; same chain as RUSTSEC-2026-0098" },
+    { id = "RUSTSEC-2026-0104", reason = "rustls-webpki 0.101.7 CRL parsing panic; same chain as RUSTSEC-2026-0098; we do not use CRLs" },
+    { id = "RUSTSEC-2026-0118", reason = "hickory-proto NSEC3 unbounded loop on cross-zone responses; no fixed upgrade available; via reqwest 0.12 consumers (alloy/helios/near)" },
+    { id = "RUSTSEC-2026-0119", reason = "hickory-proto O(n²) name compression CPU exhaustion; fix needs hickory 0.26+/reqwest 0.13, blocked by alloy/helios/near pins" },
+    { id = "RUSTSEC-2026-0222", reason = "wasmtime 45 type-index mixup; only reachable via near-sdk unit-testing dev-only feature (near-vm-runner pins wasmtime ^45, no patched release)" },
+    { id = "RUSTSEC-2026-0269", reason = "wasmtime 45 filesystem sandbox escape; same chain as RUSTSEC-2026-0222; dev-only, no WASI filesystem exposed to guest code" },
+]
+
+[licenses]
```
volovyks, line 31: The licenses, bans, and sources sections are never run by CI. Running them locally fails with over 1,300 license rejections. We need to delete them or make them pass and run them.

volovyks (commented):
Looks good
Nice move with `deny`

deny.toml
isSerge: CI now runs them

volovyks (approved):

PR #1288, opened 2026-09-11
ci(load test): Replace post-merge k6 run with tiered Solana load schedule

volovyks:
Removes the k6 runs triggered on push to develop and after merged PRs. They tested the previous build behind a fixed one-hour sleep and always reported green because of `--exit-on-running`.

Load against Solana dev now comes from three staggered schedules through the pinger:

- **Low**: 0.1 rps for 1h every hour
- **Mid**: 1 rps for 20m at 09:15 and 21:15 UTC
- **High**: a 12-minute ramp through 1, 2, 5, and 10 rps at 03:30 and 15:30 UTC

Thresholds now fail the GitHub job, and each run writes a step summary with tier, rate, and result. The ramp also fails on dropped iterations, so a run that silently sent less than it was asked to cannot pass.

All tiers together spend about 0.18 SOL/day from the pinger key, roughly 50 days on the current balance. Ethereum load stays with the Grafana synthetic checks.

### DIFF
--- a/loadtests/k6-load-test.js
+++ b/loadtests/k6-load-test.js
@@ -65,6 +65,33 @@ const strategies = {
     },
   },
 
+  // Steps up through the constant-rate tiers in one run so a single test shows
+  // where latency breaks instead of a pass/fail at one rate. ~3,240 requests.
+  "ramp_1_10": {
+    scenarios: {
+      ramp: {
+        executor: 'ramping-arrival-rate',
+        startRate: 1,
+        timeUnit: '1s',
+        preAllocatedVUs: 40,
+        maxVUs: 300,
+        stages: [
+          { target: 1, duration: '3m' },
+          { target: 2, duration: '3m' },
+          { target: 5, duration: '3m' },
+          { target: 10, duration: '3m' },
+        ],
+      },
+    },
+    thresholds: {
+      http_req_failed: ['rate<0.03'],
+      http_req_duration: ['p(95)<10000'],
+      // k6 skips iterations when every VU is still waiting on a signature. A
+      // run that silently sent less than it was asked to is not a pass.
+      dropped_iterations: ['count<10'],
+    },
+  },
+
 };
 
 export const options = (() => {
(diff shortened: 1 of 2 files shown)


volovyks:
PS: Our pinger is not ready to handle such a load. Additional work is required there.

ppca (approved):

loadtests/k6-load-test.js
```diff
@@ -65,6 +65,33 @@ const strategies = {
...
+          { target: 10, duration: '3m' },
+        ],
+      },
+    },
+    thresholds: {
+      http_req_failed: ['rate<0.03'],
+      http_req_duration: ['p(95)<10000'],
+      // k6 skips iterations when every VU is still waiting on a signature. A
+      // run that silently sent less than it was asked to is not a pass.
+      dropped_iterations: ['count<10'],
```
yvonneanne: should `dropped_iterations: ['count<10'],` be added to rps_1 (L32) and rps_0_1 (L17) thresholds block as well?

loadtests/k6-load-test.js
```diff
@@ -65,6 +65,33 @@ const strategies = {
...
+          { target: 5, duration: '3m' },
+          { target: 10, duration: '3m' },
+        ],
+      },
+    },
+    thresholds: {
+      http_req_failed: ['rate<0.03'],
+      http_req_duration: ['p(95)<10000'],
+      // k6 skips iterations when every VU is still waiting on a signature. A
+      // run that silently sent less than it was asked to is not a pass.
```
yvonneanne: This gates the whole ramp including the 10rps stage you expect to break capacity. Would it make sense to scope thresholds per-stage/tag or can we find out else where what rps we can currently sustain?

yvonneanne (approved):
Nice one!

Combining  fixing --exit-on-running bug and sensible staggering of tiers!

I assume this will lead to alerts on mpc-alerts-dev as well?

loadtests/k6-load-test.js
volovyks: Not sure if that is possible. Yes, I expect it to be read, but this is not part of our CI. Our job would be to make it green.

If I'm missing something important here, I'll address it separately. Please elaborate.

PR #1298, opened 2026-09-14
fix(ci): ignore RUSTSEC-2026-0285, which has no reachable fix

ppca:
- cargo-deny is failing on every branch, not just one: `rustls 0.23.43` accepts TLS 1.3 handshake messages across an encryption-level boundary.
- No upgrade path. The fix is rustls 0.23.45, which needs `aws-lc-rs ^1.18`; `near-crypto` pins `=1.16.2` on 0.37.4 as well as our 0.37.2.
- The transcript stays authenticated, so a handshake cannot be altered or completed by an attacker. Ignored on the same terms as the ten entries above it.
- The yanked `spin` / `chacha20` / `core2` crates in the same report are warnings and do not fail the build; left alone.

Verified with `cargo deny check advisories -W unmaintained -W unsound` → `advisories ok`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01XVpsCWQQcEBg3NoxpjBW3E


ppca:
Superseded by #1296, which carries the same RUSTSEC-2026-0285 ignore with a more complete reason — it also names `google-cloud-storage =1.17.0` forcing rustls' aws-lc provider, a second blocker this PR missed. Closing in favour of that one.

PR #1301, opened 2026-09-15
docs: add bidirectional calls design doc

yvonneanne:
PR for design doc discussion on bidirectional calls. It includes entities, API, guarantees, per-entity pseudocode with the properties each achieves, proof sketches, open points.


doc/bidirectional_calls.md
```diff
@@ -0,0 +1,503 @@
...
+  the same transaction rather than a new one.
+* *Request*: what a call asks for, the tuple (tx, dest, key, schemas): the
+  transaction, its destination chain, the key parameters to sign it with,
+  and the schemas for decoding its output and encoding the response. Written
+  `req` in the pseudocode.
+* *Request ID* (rid): a collision-resistant hash over (contract, tx, dest,
+  key) in a length-committing encoding, so within one source chain a rid
+  names one execution and nothing else. I.e.,  rid(a) = rid(b) <=> a.tx = b.tx. The schemas are outside it, so that
+  two calls for one transaction cannot both be outstanding. The key
+  parameters must be canonical, or two rids could name one execution.
```
yvonneanne: @DavidM-D 

I followed your suggestion to add "Request ID(a) = Request ID(b) <=> Transaction ID(a) = Transaction ID(b)", adapting it to this doc's notation.

> Do Request IDs need to be globally unique, or unique per chain? EVMs often have duplicate transaction IDs which are executed on different chains?

Afaics, it's enough if they are unique per chain, if the MPC ensures it doesn't mix them.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,503 @@
...
+* *Outcome*: a pair (kind, data). kind is Executed or Failed.
+  * *Executed* means the transaction was finalised, succeeded, and its
+    return data decoded against the contract's schema; data is that decoded
+    return data.
+  * *Failed* means the transaction was finalised and reverted; data is a
+    bounded prefix of the revert reason. (Appendix A adds two more reasons
+    as extensions.)
```
yvonneanne: I tried to keep it simple, we should discuss how much the contracts could benefit from more reasons now.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,503 @@
...
+  * A transaction that succeeded but whose return data does not decode has
+    no outcome: the MPC reports nothing.
```
yvonneanne, line 57: the contract probably wouldn't be able to handle this well, would it?

see also [https://sig-tar2858.slack.com/archives/C07SH5FRZU2/p1788478782628319](https://www.google.com/url?q=https://sig-tar2858.slack.com/archives/C07SH5FRZU2/p1788478782628319&sa=D&source=docs&ust=1789460656424133&usg=AOvVaw10ttjl4ZWUlLty1QOPNf1W)

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,503 @@
...
+    transaction: SerializedTransaction,
+    destination: ChainId,
+    key:         (DerivationPath, KeyVersion, SigningScheme),
+    schemas:     (OutputDeserialization, ResponseSerialization),
+) -> RequestId | Refused
+// the four arguments together are the request, written req below
+
+// implemented by the application contract, called by the library
+on_response(rid: RequestId, outcome: (Executed | Failed, data))
+```
```
yvonneanne, line 116: Wdyt of this API? @DavidM-D

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,503 @@
...
+* G5 Delivery: if tx(c) is finalised under a signature the MPC issued for
+  call(c), and its return data, if any, decodes against the contract's
+  schema, a resp(c, o) is eventually accepted.
+
+### 3.3 Assumptions
+
+* Chains  
+  * Finality: no re-orgs on source or destination chain below the finality criterion the MPC uses to index requests and attest outcomes.  
+  * Payload-level replay protection: once a transaction has executed on the destination chain under a key, no transaction with the same bytes executes again under that key, whatever signature it bears (EVM account nonce, spent UTXO, Solana durable nonce). Solana transactions using a recent blockhash do not qualify; see open points.  
+  *  Source and destination chains, eventually make progress: outages only delay delivery
```
yvonneanne:
```suggestion
  *  Source and destination chains eventually make progress: outages only delay delivery
```

yvonneanne: @DavidM-D 
> Are you saying doesn't go down, or doesn't go down forever?

Forever, is it more clear like that?

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,503 @@
...
+        drop
+    if height <= e.known:                               // C3c
+        drop
+    last_seen[e.dest] = max(last_seen[e.dest], height)  // C3d
+    delete outstanding[rid]                             // C4
+    self.on_response(rid, (kind, data))
+```
+
+The entry keeps `dest` because the rid is a hash and cannot yield it, and
+C3d needs it to pick the last_seen to raise.
```
yvonneanne: @DavidM-D, added to address your question here https://docs.google.com/document/d/12ugwE4t9StC-8ZJvFaaxdoqwzGSedp3CMA7_H8w-ziw/edit?disco=AAACHKr7TQc

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,503 @@
...
+
+When the library starts tracking a destination, last_seen[dest] starts at
+dest's finalised height at that moment. At deployment and on the first call
+to a new destination 0 is equivalent, since nothing has executed under the
+contract's key there yet. At an upgrade from a version without last_seen a
+lower start would let a response to a call answered before the upgrade be
+accepted again.
+
```
yvonneanne, line 253: YAP: If entries already outstanding at the upgrade start at zero, then a request ID completed and re-issued before the upgrade can accept one replayed old attestation. Acceptable, or do you want a stricter start height and the lost late responses?


David: Starting at zero is fine I think.
Conceptually it's equivalent to having done your last request a very long time ago, and have had no transactions hit the contract since. If that's insecure, then our model is generally insecure.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,503 @@
...
+Properties:
+
+* C1 A call is refused while an entry with the same rid is outstanding.
+* C2 Every outstanding entry records last_seen[dest] at creation.
+* C3 A response is accepted only if it verifies, an entry for its rid is
+  outstanding, and its height is strictly above that entry's recorded
+  height. Acceptance raises last_seen to at least that height.
+* C4 Entries are removed by acceptance only. No timers.
+
+### 4.2 Library on Midnight: inbox and processing
```
yvonneanne, line 263: > David: I'm going to gut a few implementation details out of this pseudo code. It's just a FIFO queue that is low contention.

Great, much appreciated!

doc/bidirectional_calls.md
yvonneanne: @DavidM-D For the first deployment with last_seen 0 works great. At a contract upgrade the contract has already accepted executions at some heights, and last_seen should be at least those or a replayed old Response for a re-issued request passes C3c (rid, attestation and key are all the same, only the height check stands in the way)....

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,503 @@
...
+It holds no per-application state, verifies nothing, and anyone may call
+it. It emits three events:
+
+* `SignRequest { contract, rid, req }`, when a contract asks for a
+  signature.
+* `Signature { rid, signature }`, when a signature is published, for the
+  broadcaster.
+* `Response { contract, rid, att, sig }`, when an attestation is published.
+  Where the chain allows it, the contract's `response` handler is called in
+  the same transaction.
```
yvonneanne: > @DavidM-D : These two don't actually have to be per source chain. They just have to be made available somewhere. We may want them on the same chain so we don't have spooky downtime at a distance when another system goes down, but it might make our integrations cheaper if we don't.

if we want them to be on the same chain, then we'd have to encode the source chain into the rid, won't we?

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,503 @@
...
+  receipt. So a response reporting exec(c) passes C3 and is accepted.
+
+## 6. Open design points
+
+* Which replay protection to assume. This draft assumes once a transaction
+  has executed, the same bytes never execute again under that key. EVM
+  nonces, spent UTXOs and Solana durable nonces give this. Using a recent
+  blockhash can be made to work if the MPC remembers completed rids for as
+  long as a differently signed copy could still execute, about a minute on
+  Solana.
```
yvonneanne, line 521: >> YAP: What's the roadmap for Solana? 

> @DavidM-D Probably fix the contract so it detects staleness on it's responses, and don't let other chains call it until we have EdDSA. And then think really hard about how we can tie down the entropy in our EdDSA implementation, so that a Sign_Respond request will only land once on Solana.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,493 @@
+# Bidirectional calls: entities, API, properties
+
+Status: draft, to be moved to github
```
yvonneanne:
```suggestion
```

doc/bidirectional_calls.md
DavidM-D: No it wouldn't, but if this is an issue on our end, we can report the error, then get the MPC to return the correctly decoded value at our leisure.

doc/bidirectional_calls.md
DavidM-D: It looks good to me, one of the outstanding questions I have is whether we actually need to send the Serialized transaction to the MPC.

The upside of it is that we make it much easier for people to transmit their transactions to other chains, and we force people to guarantee availability for their transaction, so someone can always move things along if a transaction is blocking a key.

The other side is it's not strictly necessary, a contract can always force availability through events, and it probably restricts the transactions users are able to send.

WDYT?

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,491 @@
...
+## Appendix: the failures the MPC does not report
+
+Two things leave a call unanswered: a transaction that never executes,
+because nobody broadcast it or because it can no longer be included, and a
+request the MPC refuses to process. An earlier draft answered both with
+`Failed`. Both were dropped, for the same reason.
+
+Every attestation carries the height of the block it describes (M2), and the
+library accepts it only if that height is above what it has already seen
+(C3c). A `Failed` of either kind has no block to name, since nothing
```
DavidM-D: I think this is incorrect, I think a blockheight should be attached to a failure, to mean, at this Blockheight, this transaction became permanently unviable. For example if, within that block caused a nonce to be used.

This gives us replay protection and causality on failures.

There are issues with this, sometimes the block that made this transaction unviable happened a long time ago, in that situation I think the correct response is to respond with nothing, but I'm not 100% on this, WDYT?

doc/bidirectional_calls.md
DavidM-D: This is fine, provided that RID hashes the chain ID that it's calling

yvonneanne: Great, we're on the same page then.

doc/bidirectional_calls.md
yvonneanne: The tx_id is derived from the bytes+signature and the MPC needs to know the tx_id...

yvonneanne: Whether the bytes arrive as calldata or in an event is a cost question the design does not care about

doc/bidirectional_calls.md
yvonneanne: You're right, I oversimplified it here and it's incorrect as stated. 

If we accept that nonce spent before the watcher has been armed may lead to unanswered calls, then we could do it report Unviable.

I'll adapt it accordingly after thinking about it some more, if you have a better idea, speak up.

doc/bidirectional_calls.md
yvonneanne: What do you mean by report the error? An attestation confirming the problem?

The node cannot tell our bug from genuinely returned bytes not matching a valid schema.

So if we attest to it, the question is if/where the library or contract should handle it. If the request is removed from the outstanding set, then we cannot later fix it. If the request remains in the outstanding set the on reponse logic becomes more complex.

doc/bidirectional_calls.md
DavidM-D: Good point, since request ID often doesn't = transaction ID. I guess on chains where it does, this wouldn't need to be there.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,477 @@
...
+To avoid this, ordering and processing are separated:
+
+First we put the call, or validated response into the inbox/outbox, as a compact call.
+
+We then issue Impact ops which stamp with and update the last seen on these calls/responses.
+
+```
+case message of
+    Call(rid, dest)
+        outstanding[rid].last_seen <- copy last_seen[req.dest]
```
yvonneanne:
```suggestion
        outstanding[rid].known <- copy last_seen[dest]
```
to make it consistent @DavidM-D

doc/bidirectional_calls.md
DavidM-D: Oh you mean if the MPC understands the response, but the contract doesn't, yeah the contract is SoL I think.

yvonneanne: Yep.

Afaics, the options are:

1. Report nothing and **drop** the request. This is the doc right now, and [#1302](https://github.com/sig-net/mpc/issues/1302). The call hangs forever and the contract cannot distinguish it from one still in flight.
2. Report nothing and **keep** the request pending, so a later decoder fix delivers the real value. Your original suggestion. It recovers our bugs, but when the failure is genuine the entry lives forever with nothing that can ever clear it.
3. Attest Undecodable and **remove** the request. The contract learns the transaction executed, but no later bugfix can recover our the serialization problem retrospectively.
4. Report with a separate event that a decoding problem happened but **keep** the request in the pending set so a bug fix potentially fixes it.

yvonneanne: From a complexity point of view I like 3 best

yvonneanne: Decision from a call with David: report nothing, drop the request and log why. Later we'll build a node explorer to go through the logs.

doc/bidirectional_calls.md
yvonneanne: Done (including an mpc pseudcode change to let the watcher start right after seeing a request)

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+* *Signet contract*: our contract on the source chain, through which
+  requests, signatures and responses are published. It holds no per-
+  application state.
+* *MPC network*: the nodes that sign and attest.
+* *Broadcaster*: whoever submits the signed transaction to the destination
+  chain. Untrusted; the design does not depend on who it is.
+
+Happy path:
+
+1. The application contract calls the library, which records the call and
```
volovyks, line 19: This is correct, but the way we store requests on the contract can change without the MPC team/repo knowing it. Since this is an MPC doc, I would drop app-level implementation details.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+3. Any entity can broadcast the signed transaction to the destination chain.
+4. When the transaction is finalised on the destination chain, the MPC
+   attests the outcome and publishes the attestation to the signet contract.
+5. Once delivered to the application contract, the library accepts or drops
+   it, and on acceptance runs the contract's response handler.
+
+## 2. Vocabulary
+
+* *Transaction*: the bytes of an unsigned destination-chain transaction.
+* *Transaction ID*: the identifier under which the destination chain records
```
volovyks, line 32: What if we generate another signature for the same transaction? Whould that be a problem? The TransactionId would be different.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+  rid(a) = rid(b) exactly when a and b agree on all four. One rid names one
+  execution, and by key derivation (section 3.3) one execution
+  names one rid. The schemas are outside it, so that two calls for one
+  transaction cannot both be outstanding. The key parameters must be
+  canonical, or two rids could name one execution.
+* *Outcome*: a pair (kind, data), with three kinds.
+  * *Executed*: the transaction was finalised, succeeded, and its return
+    data decoded against the contract's schema; data is that decoded return
+    data.
+  * *Failed*: the transaction was finalised and reverted; data is empty.
```
volovyks, line 50: I would clarify if it means that we will never get any other outcome (is this a final state?).

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+  * *Unviable*: a finalised transaction carrying other bytes took tx's
+    nonce, so tx can never be included (section 3.3); data is empty. The
```
volovyks, line 52: What do you mean by "data is empty"?

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+    nonce is whatever replay protection tx consumes (account nonce, spent
+    output). Unlike the other two, this outcome is best-effort: section 4.4
+    says when it is missed.
+  * A transaction that succeeded but whose return data does not decode has
+    no outcome: the MPC reports nothing.
```
volovyks, line 57: And MPC clears it from its backlog?

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+  execution, and by key derivation (section 3.3) one execution
+  names one rid. The schemas are outside it, so that two calls for one
+  transaction cannot both be outstanding. The key parameters must be
+  canonical, or two rids could name one execution.
+* *Outcome*: a pair (kind, data), with three kinds.
+  * *Executed*: the transaction was finalised, succeeded, and its return
+    data decoded against the contract's schema; data is that decoded return
+    data.
+  * *Failed*: the transaction was finalised and reverted; data is empty.
+  * *Unviable*: a finalised transaction carrying other bytes took tx's
```
volovyks, line 51: Can also be `Superseded`, but no strong opinion here.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+    nonce, so tx can never be included (section 3.3); data is empty. The
+    nonce is whatever replay protection tx consumes (account nonce, spent
+    output). Unlike the other two, this outcome is best-effort: section 4.4
+    says when it is missed.
+  * A transaction that succeeded but whose return data does not decode has
+    no outcome: the MPC reports nothing.
+* *Attestation key*: a signing key the MPC derives from its root key, the
+  source chain and the contract, used for nothing but attestations to that
```
volovyks: and key version

volovyks: It is actually derived from more parameters. Mostly taken from the original request.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+  contract.
+* *Attestation*: a statement (rid, version, height, outcome) signed with
+  rid's contract's attestation key at that version, each field
+  length-committed. height is
```
volovyks, line 63: redundant new line

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+  on destination chain dest(c), naming it rid(c). Making the same request
+  again is a new call with the same rid.
+* exec(c): tx(c) is included in a finalised destination block.
+  height(exec(c)) is that block's height.
+* resp(c, o): a response for rid(c) carrying outcome o.
+
+How a call ends, as seen by the application contract:
+
+* *Refused*: the library rejects call(c), and the caller learns it at once.
+* *Accepted*: the contract takes a resp(c, o) as the answer to call(c) and
```
volovyks, line 82: This is a 100% successful flow. What about the one where the contract rejects the outcome?

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+  runs its response handler (section 3.1).
+* *Unanswered*: no response is ever accepted. A retry must be a new
+  transaction: the same bytes give the same rid, which C1 refuses.
```
volovyks, line 85: C1 is not defined yet.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+    key:         (DerivationPath, KeyVersion, SigningScheme),
+    schemas:     (OutputDeserialization, ResponseSerialization),
+) -> RequestId | Refused
+// the four arguments together are the request, written req below
+
+// implemented by the application contract, called by the library
+on_response(rid: RequestId, outcome: (kind, data))
+```
+
+The application configures the library with its attestation public key
```
volovyks, line 118: Is it so? Should it be the MPC Root key? Since it depends on the request parameters.

doc/bidirectional_calls.md
volovyks: Ok, now I understand that it would be hard to write the MPC-only spec and then the Applications spec.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+    and migrations. A contract that keeps its key and loses this state can be
+    replayed against everything it ever executed.
+  * Every published response is delivered, and again if the library dropped
+    it: on Midnight someone enqueues it and runs `process` (section 4.2),
+    elsewhere the signet contract or anyone calls `response` (section 4.3).
+    A handler that runs inside `response` may fail, since that reverts the
+    acceptance; one that runs outside must not, since the entry is already
+    gone.
+* MPC
+  * at most f of the n nodes are faulty, a signature or attestation needs a
```
volovyks, line 177: This can be skipped. Not a part of the bidirectional spec. More like the MPC spec.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+  * Honest nodes observe the same finalised destination state, receipts and
+    return data included, and compute the attestation content as the same pure
+    function of what they read and the schemas, which an upgrade does not
+    change for requests already made; otherwise nodes split and no
+    attestation reaches the threshold. The same holds for `authentic` and
+    `processable`: a request admitted by too few nodes is never attested.
+  * A change to `authentic`, `processable` or the attestation function
+    applies only to requests made at or after a source height the upgrade
+    names; applied to requests in flight it leaves them unanswered.
+  * Key derivation is collision-resistant: distinct (source chain, contract,
```
volovyks, line 190: And this is a key derivation spec.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+### 4.1 Library (inside the application contract)
+
+```
+state (per application contract):
+    attestation_key: KeyVersion -> PublicKey   // section 3.1
+    last_seen:   ChainId -> Height       // 0 for every chain, see below
+    outstanding: RequestId -> Entry
+    Entry = { dest: ChainId, known: Height }
+
+on sign_bidirectional(req) from the application logic:
```
volovyks, line 214: Should we check the `>last_seen` condition?

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+  broadcaster and for nodes that were not in the signing round.
+* `Response { contract, rid, att, sig }`, when an attestation is published.
+  Where the chain allows it, the contract's `response` handler is called in
+  the same transaction, with a bounded gas allowance, since it runs on the
+  MPC's gas, and without letting its failure suppress the event. The
+  library verifies what it receives (C3), so anyone may call `response`
+  directly, which is how a response is delivered where the signet contract
+  cannot call it, and again after a failed handler.
+
+### 4.4 MPC network
```
volovyks, line 321: Again, more like MPC spec

volovyks (commented):
Great! Very good doc. It was hard to review, and I'm sure it was much harder to write.

Not approving; let's wait for people from Applications to do that.

doc/bidirectional_calls.md
yvonneanne: Right, section 2 listed three of the five inputs. The attestation key is derived from the root key, the source chain, the contract, a reserved path and the key version, with the key version taken from the request. Fixed, and M1 had the same omission.

On the root key: the library holds the public key rather than the root key so the contract does not have to run the derivation on chain. The derivation is per contract, so one contract has one attestation key per key version, which is the `KeyVersion -> PublicKey` map in 4.1.

yvonneanne: Covered above.

doc/bidirectional_calls.md
yvonneanne: Covered above.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+* *Transaction ID*: the identifier under which the destination chain records
+  a submitted transaction, computable from the transaction and the signature
+  in the encoding that chain accepts.
+* *Request*: what a call asks for, the tuple (tx, dest, key, schemas): the
+  transaction, its destination chain, the key parameters to sign it with,
+  and the schemas for decoding its output and encoding the response. Written
+  `req` in the pseudocode.
+* *Request ID* (rid): a collision-resistant hash over (contract, tx, dest,
+  key) in a length-committing encoding, so within one source chain
+  rid(a) = rid(b) exactly when a and b agree on all four. One rid names one
```
scafuro, line 41: This is a  really important property, I think we satisfy it, but just to make sure:  @BRBussy @esaminu  can you confirm that we have a length-committing encoding so that  rid(a) = rid(b) exactly when a and b agree on all four? (and point me to the code:)) 

The classical edge case where two "distinct" calls hash to the same value are cases like this: 
   a=0011 |11| 00   b= 001 | 111 | 00
here
a = b  but  if the lengths are not fixed, they can get serialized in two different requests.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+  execution, and by key derivation (section 3.3) one execution
+  names one rid. The schemas are outside it, so that two calls for one
```
scafuro, line 43: the key-derivation function is independent of transaction and  rid. 

what property of key derivation are we trying to use here?

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+  transaction cannot both be outstanding. The key parameters must be
+  canonical, or two rids could name one execution.
```
scafuro, line 45: I am not sure what property  are we trying to capture:
"The schemas are outside it, so that two calls for one transaction cannot both be outstanding. The key parameters must be canonical, or two rids could name one execution.

doc/bidirectional_calls.md
scafuro, line 51: The definition of Unviable is not super clear to me as I am not clear on what finalized means for us.

How is finalized in fail different from a finalized in unviable?

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+  * *Executed*: the transaction was finalised, succeeded, and its return
+    data decoded against the contract's schema; data is that decoded return
+    data.
+  * *Failed*: the transaction was finalised and reverted; data is empty.
+  * *Unviable*: a finalised transaction carrying other bytes took tx's
+    nonce, so tx can never be included (section 3.3); data is empty. The
+    nonce is whatever replay protection tx consumes (account nonce, spent
+    output). Unlike the other two, this outcome is best-effort: section 4.4
+    says when it is missed.
+  * A transaction that succeeded but whose return data does not decode has
```
scafuro, line 56: Can we clarify what "does not decode" mean?

And why it is ok to do nothing given that the transaction succeeded?
 would assume that a contract calling the MPC to execute a transaction will want to know that the transaction is succeeded with "data that does not decode".

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+  length-committed. height is
+  the height of the block that includes the transaction the outcome
+  describes, attested once that block is final, in the destination chain's
+  own numbering (a slot on Solana): the execution's block, or for Unviable
+  the block that took the nonce.
+* *Response*: an attestation delivered to its contract.
+
+Events for a call c made by the application contract:
+
+* call(c): the contract asks for transaction tx(c) to be signed and executed
```
scafuro, line 72: what is the difference between a call "c" and a request? and if there is no difference, what is the reason to keep them both?

Does tx(c)  mean "serialization of c into a transaction accepted on destination" ?

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+  on destination chain dest(c), naming it rid(c). Making the same request
+  again is a new call with the same rid.
```
scafuro, line 74: "Making the same request again is a new call with the same rid."

Here there might be the answer, we think of a call as an action  over a request.

Which makes totally sense, but  then what I think trips me off, is that we use the notation 
call(c) -- as if we are calling c  --  so  c now is not an action but is a subject.

In my view,   we want to use notation call(req) -- instead of call(c) -- when call () means the action of calling to execute (req).
Using this notation,  then the sentence  "Making the same request again is a new call with the same rid."
becomes obvious because  if I run  call(req)  and then call(req) again, of course, I except the exact same behavior.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+* exec(c): tx(c) is included in a finalised destination block.
+  height(exec(c)) is that block's height.
+* resp(c, o): a response for rid(c) carrying outcome o.
+
+How a call ends, as seen by the application contract:
+
+* *Refused*: the library rejects call(c), and the caller learns it at once.
+* *Accepted*: the contract takes a resp(c, o) as the answer to call(c) and
+  runs its response handler (section 3.1).
+* *Unanswered*: no response is ever accepted. A retry must be a new
```
scafuro, line 84: "no response is ever accepted"
Can we expand/clarify what we mean here?

Here was my thought process and confusion:

The word _unanswered_  to me suggests that we mean that the MPC did not answer yet (probably,   because the MPC does not know yet what was the faith of the transaction).   with this interpretation,  unanswered it's more like a pending transaction.

But the language "no response is **ever** accepted"  could be interpreted  in a  couple of ways:
1. the contract will never accept any response on that call  (even if the MPC provided one)
2. the MPC knows that they will never accept to provide a response on that call.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+  transaction: the same bytes give the same rid, which C1 refuses.
+
+A call is *outstanding* from the moment it is made until a response to it is
+accepted; the library records it as an entry in `outstanding` (section 4.1).
+An unanswered call is outstanding forever, and from inside the contract this
```
scafuro, line 89: Ok, this part provides some clarification about unanswered = outstanding, but  I think we should change the explanation of unanswered.

I also wonder if  *outstanding* should be another item of the list instead of being treated differently.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+### 3.2 Guarantees
+
+G1 to G3 are safety, G4 is liveness. G2 and G4 together give "exactly one
+accepted response per call that meets G4's premise".
+
+* G1 Causal order: an accepted resp(c, o) reports destination state that
+  does not happen-before call(c).
+* G2 At-most-once: for each execution, at most one response reporting it
+  as the outcome of its own transaction is ever accepted, however many calls
+  named that transaction and however many times it is attested.
```
scafuro, line 135: We defined response as the attestation delivered to the contract. 
In G2 we are implicitly saying that the delivery process has a mechanism whereby  it recognizes that many - potentially distinct - attestations, are mapped to the same execution.

Was this the intention ? and is it worth it saying it explicitly somewhere?

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+* G3 Integrity and finality: an accepted resp(c, o) carries the true, final
+  outcome of tx(c), never of another transaction.
+* G4 Delivery: if tx(c) is finalised under a signature the MPC issued and
```
scafuro, line 138: we said earlier that a refused transaction is a transaction that is finalized & reverted. 


Do we want to say : "if tx(c) is accepted"  to avoid any ambiguity or backslash?

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+  published for call(c), and its return data, if any, decodes against the
+  contract's schema, a resp(c, o) is eventually accepted.
+
+After the upgrade that introduces last_seen, G1 and G2 are suspended per
```
scafuro, line 142: I would move this paragraph somewhere else.

In this section we are providing definitions, so I would don't talk about solutions.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+Nothing is promised for a transaction that never executes, for an execution
+whose return data does not decode (M5), or for a request the MPC refuses to
+process. Reporting any of them needs machinery the rest of this design does
+without.
+
+### 3.3 Assumptions
+
+* Chains
+  * Finality: no re-orgs on source or destination chain below the finality
+    criterion the MPC uses to index requests and attest outcomes.
```
scafuro, line 155: I am not sure what we mean here. Are we assuming something about the chains or something about the Mac criterion?

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,537 @@
...
+    every other that names it. Solana transactions using a recent blockhash
+    do not qualify; see open points.
+  * Source and destination chains eventually make progress: an outage only
+    delays delivery
+  * Chain identifiers are injective: two ChainIds the MPC accepts never name
+    the same chain.
+* Contracts and library
+  * Durable contract state: the library's state (section 4.1) survives upgrades
+    and migrations. A contract that keeps its key and loses this state can be
+    replayed against everything it ever executed.
```
scafuro, line 169: I agree to assume that the state must survive.

I am not sure I know what we mean here by replayed " _A contract that keeps its key and loses this state can be replayed against everything it ever executed."_

Also,  I am reading this sentence "" _A contract that keeps its key and loses this state can be replayed against everything it ever executed."_"  as: this is an attack that happens  if the assumption about the state fails. Is this true?
If so, I suggest to reframe it so that it's more obvious?

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,543 @@
...
+  application state.
+* *MPC network*: the nodes that sign and attest.
+* *Broadcaster*: whoever submits the signed transaction to the destination
+  chain. Untrusted; the design does not depend on who it is.
+
+Happy path:
+
+1. The application contract calls the library, which records the call and
+   asks the signet contract to emit a sign request.
+2. The MPC signs the transaction, publishes the signature on the source
```
ppca, line 21: do we need to define or provide context for why source chain and target chain is mentioned?

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,543 @@
...
+
+* *Transaction*: the bytes of an unsigned destination-chain transaction.
+* *Transaction ID*: the identifier under which the destination chain records
+  a submitted transaction, computable from the transaction and the signature
+  in the encoding that chain accepts.
+* *Request*: what a call asks for, the tuple (tx, dest, key, schemas): the
+  transaction, its destination chain, the key parameters to sign it with,
+  and the schemas for decoding its output and encoding the response. Written
+  `req` in the pseudocode.
+* *Request ID* (rid): a collision-resistant hash over (contract, tx, dest,
```
ppca, line 39: should key_versioon matter for request id?

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,543 @@
...
+  source chain, the contract, a reserved path and the request's key version,
+  used for nothing but attestations to that contract.
+* *Attestation*: a statement (rid, key version, height, outcome) signed
+  with rid's contract's attestation key at that key version, each field
+  length-committed. height is
+  the height of the block that includes the transaction the outcome
+  describes, attested once that block is final, in the destination chain's
+  own numbering (a slot on Solana): the execution's block, or for Unviable
+  the block that took the nonce.
+* *Response*: an attestation delivered to its contract.
```
ppca, line 68: what does its contract mean?

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,543 @@
...
+* exec(c): tx(c) is included in a finalised destination block.
+  height(exec(c)) is that block's height.
+* resp(c, o): a response for rid(c) carrying outcome o.
+
+How a call ends, as seen by the application contract:
+
+* *Refused*: the library rejects call(c), and the caller learns it at once.
+* *Accepted*: the contract takes a resp(c, o) as the answer to call(c) and
+  runs its response handler (section 3.1).
+* *Unanswered*: no response is ever accepted. A retry must be a new
```
ppca, line 84: what does accept mean? does it mean received and approved?

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,543 @@
...
+  transaction: the same bytes give the same rid, which C1 refuses.
+
+A call is *outstanding* from the moment it is made until a response to it is
+accepted; the library records it as an entry in `outstanding` (section 4.1).
+An unanswered call is outstanding forever, and from inside the contract this
+is indistinguishable from a response that has not arrived yet.
+
+*Happens-before*, for one application contract: the transitive closure of
```
ppca, line 92: I'm a bit lost on this paragraph 😅 are we trying to define *Happens-before*? then I guess we need a newline after `for one application contract:` so that it's clear it is the same structure as above.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,543 @@
...
+two kinds of edge. On one chain, an earlier block, transaction or step
+within a transaction happens-before a later one. Across chains, a
+destination block happens-before the source transaction in which this
+contract accepts a response attesting it. Nothing else is an edge, not a
+cross-chain call, not a dropped response, not another contract's
```
ppca, line 97: what does it mean nothing else is an edge? an edge between mpc and contract?

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,543 @@
...
+* G2 At-most-once: for each execution, at most one response reporting it
+  as the outcome of its own transaction is ever accepted, however many calls
+  named that transaction and however many times it is attested.
+* G3 Integrity and finality: an accepted resp(c, o) carries the true, final
+  outcome of tx(c), never of another transaction.
+* G4 Delivery: if tx(c) is finalised under a signature the MPC issued and
+  published for call(c), and its return data, if any, decodes against the
+  contract's schema, a resp(c, o) is eventually accepted.
+
+After the upgrade that introduces last_seen, G1 and G2 are suspended per
```
ppca, line 142: probably helpful to link to the last_seen upgrade?

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,543 @@
...
+    key derivation is valid: parameters canonical, key derivable for the
+      source chain, path not the attestation key's
+    req.tx is non-empty, parses as an unsigned transaction in dest's
+      format, commits to dest (EVM: carries its chain id), and attaching
+      any signature yields a well-formed signed transaction
+    req.schemas are valid for dest and source chain respectively
+
+state:
+    pending: RequestId -> Entry
+    Entry = { req, contract, signatures: Set<Signature>, attestation? }
```
ppca, line 342: why Set<Signature> instead of 1 signature?

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,543 @@
...
+    e = pending[rid] if rid in pending
+    if e and signature verifies over e.req.tx
+      under derived_key(e.contract, e.req.key):
+        e.signatures.add(signature)
+
+on destination block at height h finalised on chain dest:
+    for (rid, e) in pending for dest and no e.attestation:
+        ours = { txid(s, e.req.tx) for s in e.signatures }
+        account = derived_address(e.contract, e.req.key)
+        if some id in ours has a receipt r in a block at height h' that is
```
ppca, line 364: I think our current implementation also looks up the id in past blocks, not just this block. this backfill mechanism is quite important especially if destination chain has already moved on a lot while source chain was doing a catchup.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,543 @@
...
+need. Whether a node has admitted the request by then depends on how far
+its destination indexing runs ahead of its source indexing, so a nonce
+taken soon after the request is made may be seen by fewer nodes than the
+threshold: Unviable is best-effort.
+
+Restarts. A node keeps `pending` and its position on every chain durably,
+advancing the position only once the event's effects are persisted, and
+resumes indexing from that position, so what it admitted survives, what it
+dropped stays dropped (M4), no destination block goes unprocessed, and
+each entry resumes at the step it is missing: signing, lookup or
```
ppca, line 414: what does lookup mean here?

ppca (commented):
will continue on the rest next week :)

doc/bidirectional_calls.md
BRBussy: This bit: `contract calls the library which records the call` could be reworded something like: `contract uses the library to construct protocol structures`. Written that way would be more accurate as the library doesn't perform any `recording` which sounds as though it writes state. The libraries tend to provide only helpers or utility functions to the application contract.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,543 @@
...
+Happy path:
+
+1. The application contract calls the library, which records the call and
+   asks the signet contract to emit a sign request.
+2. The MPC signs the transaction, publishes the signature on the source
+   chain, and starts looking for the transaction on the destination chain.
+3. Any entity can broadcast the signed transaction to the destination chain.
+4. When the transaction is finalised on the destination chain, the MPC
+   attests the outcome and publishes the attestation to the signet contract.
+5. Once delivered to the application contract, the library accepts or drops
```
BRBussy: - Perhaps expand a bit what what specifically it is that is delivered back to the application contract, i.e. the attestation and an untrusted destination chain outcome/result.
- It would be more accurate to say that the application contract `uses helper functions/utilties from the library to verify the provided output against the attestation`, the library itself does not drop or accept.

doc/bidirectional_calls.md
BRBussy: Currently the schemas are also includded within the data that is hashed, however, I'm not sure how that relates directly to `two calls for one transaction cannot both be outstanding`? Since the request Ids may still differ for requests on transactions with the same schemas.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,543 @@
...
+* *Request*: what a call asks for, the tuple (tx, dest, key, schemas): the
+  transaction, its destination chain, the key parameters to sign it with,
+  and the schemas for decoding its output and encoding the response. Written
+  `req` in the pseudocode.
+* *Request ID* (rid): a collision-resistant hash over (contract, tx, dest,
+  key) in a length-committing encoding, so within one source chain
+  rid(a) = rid(b) exactly when a and b agree on all four. One rid names one
+  execution, and by key derivation (section 3.3) one execution
+  names one rid. The schemas are outside it, so that two calls for one
+  transaction cannot both be outstanding. The key parameters must be
```
BRBussy: In the midnight implementation 2 requestIds can currently name 1 transaction due to the presence of a `requestNonce`. Outstanding work in this [issue](https://github.com/sig-net/midnight-integration/issues/120).

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,543 @@
...
+  requests, signatures and responses are published. It holds no per-
+  application state.
+* *MPC network*: the nodes that sign and attest.
+* *Broadcaster*: whoever submits the signed transaction to the destination
+  chain. Untrusted; the design does not depend on who it is.
+
+Happy path:
+
+1. The application contract calls the library, which records the call and
+   asks the signet contract to emit a sign request.
```
BRBussy: This bit: `contract calls the library which records the call` could be reworded something like: `contract uses the library to construct protocol structures`. Written that way would be more accurate as the library doesn't perform any `recording` which sounds as though it writes state. The libraries tend to provide only helpers or utility functions to the application contract.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,543 @@
...
+  in the encoding that chain accepts.
+* *Request*: what a call asks for, the tuple (tx, dest, key, schemas): the
+  transaction, its destination chain, the key parameters to sign it with,
+  and the schemas for decoding its output and encoding the response. Written
+  `req` in the pseudocode.
+* *Request ID* (rid): a collision-resistant hash over (contract, tx, dest,
+  key) in a length-committing encoding, so within one source chain
+  rid(a) = rid(b) exactly when a and b agree on all four. One rid names one
+  execution, and by key derivation (section 3.3) one execution
+  names one rid. The schemas are outside it, so that two calls for one
```
BRBussy: Currently the schemas are also included in the data that is hashed, however, I'm not sure how that relates directly to `two calls for one transaction cannot both be outstanding`? Since the request Ids may still differ for requests on transactions with the same schemas.

doc/bidirectional_calls.md
BRBussy: I do believe that the hashing that we perform to generate our requestId for the midnight implementation is length-committing. See the [hashing helper here](https://github.com/sig-net/midnight-integration/blob/04bf076846941628f9542582e78735ff2bd50b6a/packages/signet-midnight/src/Signet.compact#L187). I believe it is length committing as the hashing function is statically configured by request type and size at compile time.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,543 @@
...
+  transaction cannot both be outstanding. The key parameters must be
+  canonical, or two rids could name one execution.
+* *Outcome*: a pair (kind, data), with three kinds.
+  * *Executed*: the transaction was finalised, succeeded, and its return
+    data decoded against the contract's schema; data is that decoded return
+    data.
+  * *Failed*: the transaction was finalised and reverted; data is empty.
+  * *Unviable*: a finalised transaction carrying other bytes took tx's
```
BRBussy: Perhaps the description of unviable being `a finalised transaction took tx's nonce` should be described as an example instead of the rule/definition of this outcome as some chains have ways to render a transaction invalid without use of nonces - e.g. block height or timebounds etc.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,543 @@
...
+    nonce, so tx can never be included (section 3.3); data is empty. The
+    nonce is whatever replay protection tx consumes (account nonce, spent
+    output). Unlike the other two, this outcome is best-effort: section 4.4
+    says when it is missed.
+  * A transaction that succeeded but whose return data does not decode has
```
BRBussy: Perhaps worth being explicit about the `MPC reporting nothing` meaning that the request _has been dropped_ and the MPC reports nothing.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,543 @@
...
+    no outcome: the MPC reports nothing.
+* *Attestation key*: a signing key the MPC derives from its root key, the
+  source chain, the contract, a reserved path and the request's key version,
```
BRBussy: Worth noting that the path of the attestation key is not strictly reserved at the moment in MPC code. i.e. if an application contract chooses to construct a request with the path equal to what the MPC uses to derive the response key the path is allowed. i.e. the request is not dropped.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,543 @@
...
+  used for nothing but attestations to that contract.
+* *Attestation*: a statement (rid, key version, height, outcome) signed
+  with rid's contract's attestation key at that key version, each field
+  length-committed. height is
+  the height of the block that includes the transaction the outcome
+  describes, attested once that block is final, in the destination chain's
+  own numbering (a slot on Solana): the execution's block, or for Unviable
```
BRBussy: similar to earlier comment: unviability may be the result of some other criterion than nonce unavailability, as such it would be more accurate here to describe the `nonce-taken` case as an example rather than the definition.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,543 @@
...
+* call(c): the contract asks for transaction tx(c) to be signed and executed
+  on destination chain dest(c), naming it rid(c). Making the same request
+  again is a new call with the same rid.
+* exec(c): tx(c) is included in a finalised destination block.
+  height(exec(c)) is that block's height.
+* resp(c, o): a response for rid(c) carrying outcome o.
+
+How a call ends, as seen by the application contract:
+
+* *Refused*: the library rejects call(c), and the caller learns it at once.
```
BRBussy: The library typically just provides helper functions or utilities for validation, the application contract is what actually rejects or accepts (perhaps this is implementation detail).

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,543 @@
...
+* *Accepted*: the contract takes a resp(c, o) as the answer to call(c) and
+  runs its response handler (section 3.1).
+* *Unanswered*: no response is ever accepted. A retry must be a new
+  transaction: the same bytes give the same rid, which C1 refuses.
+
+A call is *outstanding* from the moment it is made until a response to it is
+accepted; the library records it as an entry in `outstanding` (section 4.1).
```
BRBussy: Similar comment about what the library does/doesn't do: The application contract marks the entry as outstanding. The library may provide some helpers for that, but it is the responsibility of the application contract.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,543 @@
...
+
+### 3.1 API offered to the application contract
+
+```
+sign_bidirectional(
+    transaction: SerializedTransaction,
+    destination: ChainId,
+    key:         (DerivationPath, KeyVersion, SigningScheme),
+    schemas:     (OutputDeserialization, ResponseSerialization),
+) -> RequestId | Refused
```
BRBussy: In current implementations a request is never refused at the Signet contract API. The application contract may store multiple `pending` copies of the same request and request signature multiple times for the same requestId.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,543 @@
...
+// the four arguments together are the request, written req below
+
+// implemented by the application contract, called by the library
+on_response(rid: RequestId, outcome: (kind, data))
```
BRBussy: Perhaps implementation detail, but the library does not call an `onResponse` handler. The library provides an attestation validation helper function. The application contract verifies the given output with the attestation and then responds accordingly.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,543 @@
...
+not settle.
+
+### 3.2 Guarantees
+
+G1 to G3 are safety, G4 is liveness. G2 and G4 together give "exactly one
+accepted response per call that meets G4's premise".
+
+* G1 Causal order: an accepted resp(c, o) reports destination state that
+  does not happen-before call(c).
+* G2 At-most-once: for each execution, at most one response reporting it
```
BRBussy: From whose perspective is G2 written? `at most one response reporting it as the outcome of its own transaction is ever accepted` doesn't clarify who does this acceptance. Does this mean: `at most one response is ever reported (and attested) as being the outcome of a particular transaction`? An application contract may choose to process the same attestation+output pair more than once for some application specific reason.

BRBussy (approved):

doc/bidirectional_calls.md
yvonneanne: I added the following at the top, please re-open the comment if this is not sufficient.

"## TL;DR

A contract on the source chain makes an asynchronous call to a destination
chain, and gets at most one response. The response is final when it comes:
the call executed, it reverted, or it can never execute. Until then nothing
arrives, and a call that never executes is never answered. The rest of the
document is what the MPC, the signet contract and the library have to do
for that to hold."

doc/bidirectional_calls.md
yvonneanne: the statement here is about transactions on the destination chain in general, not specific to our approach. 
dealing with more signatures for the same transaction is considered later in the doc. 
I'll resolve the comment, please re-open if you want me to change something here.

doc/bidirectional_calls.md
yvonneanne: Yes, otherwise our properties don't hold. 
It already is in there: key in the hash is the API's (DerivationPath, KeyVersion, SigningScheme). That was only visible in the section 3.1 signature, so I've named the three fields where the request tuple is defined in section 2.

The reason the version has to be in the rid is the G2 sketch in section 5.

doc/bidirectional_calls.md
yvonneanne: please unresolve if there is something you want me to change in the doc here.

doc/bidirectional_calls.md
yvonneanne: You're right that it doesn't follow as written. Two calls with the same transaction bytes can still get different rids through dest or the key parameters, and that's intended: those are different executions. I'll rephrase.

Excluding the schemas is the necessary half, not the sufficient one. They are the only part of a request that doesn't change the execution, so they are the only way two calls naming one execution could get two rids. Put them in the hash and both stay outstanding, one execution matches both, and both responses are accepted (section 4.1), which G2 forbids.

On "currently they are included": correct, and inconsistently. Midnight binds the full schema buffers into the request id, with a test pinning it. This must be changed to adhere to this fix.

doc/bidirectional_calls.md
yvonneanne: Only injectivity, in the direction execution → rid. Derivation is a function of (source chain, contract, key parameters) alone, and it has to be: the address must exist before there is a transaction or a signature, which is how the MPC finds an execution in M3, derived_address(contract, key), see section 5.

I rephrased it to make it more clear, tell me if more explanations are needed here.

doc/bidirectional_calls.md
yvonneanne: Great! I take this comment as FYI for me only and resolve it. Tell me if I'm missing something.

doc/bidirectional_calls.md
yvonneanne: @DavidM-D @BRBussy is it an option for a contract to reject the outcome? the case where the handler does not complete (e.g., out of gas) is treated in a later section.

yvonneanne:
Thank you all for the questions and comments, they were exactly what the doc needed. I restructured it to take your suggestions into account, it now has a TL;DR and reader guide, one noun (request) instead of call and request, the outcomes as a table with Unviable defined generally, the assumptions reduced to the chains, with the library's needs in 4.1 and the MPC properties at the start of Section 5, and the upgrade exception moved to 4.1. Replies to each thread below say what changed for it. Section 4.2 (Midnight) is unchanged and David's to edit.

doc/bidirectional_calls.md
yvonneanne: Thanks, this made me add a reader guide at the top that says which section is normative for whom: 4.1 and 4.2 for the library, 4.3 for the signet contract, 4.4 for the MPC. The library parts have to stay, since the guarantees only hold with both halves, but now it is explicit which half is ours.

doc/bidirectional_calls.md
yvonneanne: Source and destination chain are now also the first two entities in Section 1.

doc/bidirectional_calls.md
yvonneanne: 4.4 is what this repo implements for bidirectional requests, so it stays. The generic MPC properties it depends on moved to Section 5 with pointers, and the reader guide says which section is normative for whom.

doc/bidirectional_calls.md
yvonneanne: Yes, every outcome is final: the block is final, and once the transaction has executed nothing later changes that. Now said right under the outcome table.

doc/bidirectional_calls.md
yvonneanne: A zero-length data field, the kind alone is the information. Now said explicitly.

doc/bidirectional_calls.md
yvonneanne: Yes: it drops the request and logs why (M5). The library's entry stays outstanding, which is the stranded-rid point in Section 6. Added to the vocabulary text.

doc/bidirectional_calls.md
yvonneanne: I thought about it and kept Unviable, as the one kind for every way a transaction becomes impossible to include. Two reasons. The application does the same thing whatever the cause, retry with a new transaction, so the cause does not belong in the API. And the kind is a field of the attestation that every library decodes: one kind means adding expiry detection later is a node-only change, while a second kind is a versioned change on every source chain. Superseded would also be wrong when the nonce went to an unrelated transaction, say a plain sign under the same key. The definition is now the general one, can never be included, with the used-up nonce as what the MPC detects today; a reason can go into data later if anyone needs it.

doc/bidirectional_calls.md
yvonneanne: Rewritten. The point is that rids are one-to-one with executions, and two things could break that. Schemas in the hash: two requests for one transaction with different schemas would have different rids, both outstanding and signed, and the single execution attested twice. Non-canonical key parameters: two encodings of one key would be two rids for one execution.

doc/bidirectional_calls.md
yvonneanne: Both mean the block is final under the finality assumption. Failed: req.tx is in a final block and reverted. Unviable: a different transaction from the same key is in a final block and used up req.tx's replay protection, so req.tx can never be included. Now a table, which I hope makes the three kinds easier to compare.

doc/bidirectional_calls.md
yvonneanne: Does not decode means the return bytes do not parse under the output schema the request named, either because the schema is wrong or because the destination contract changed its return type. Nothing is reported because the node cannot tell which, and the contract could not act on bytes it cannot decode. The decision in the earlier thread with David was: report nothing, drop the request and log it. Now spelled out under the outcome table.

doc/bidirectional_calls.md
yvonneanne: You were right, the pair was not earning its keep. I dropped call as a defined object. There is one noun now, the request req with fields req.tx, req.dest, req.key and req.schemas, and it is made, it executes, it is answered.

doc/bidirectional_calls.md
yvonneanne: Same fix as above: no more call(c). Guarantees and sketches are stated over req.

doc/bidirectional_calls.md
yvonneanne: Rephrased: the rid definition now spells out both halves, schemas out of the hash and canonical key parameters, each with the failure it prevents. The Midnight change is item 6 of the code gaps note, which will get its own PR.

doc/bidirectional_calls.md
yvonneanne: Agreed, and thanks for the timebound examples. Unviable is now defined as can never be included; the used-up nonce is what the MPC detects today, and expiry by height or timebound is an open point in Section 6.

doc/bidirectional_calls.md
yvonneanne: Same fix: the height is now the block holding the transaction the outcome describes, for Unviable the one that used up the replay protection.

doc/bidirectional_calls.md
yvonneanne: The contract that made rid's request. Reworded.

doc/bidirectional_calls.md
yvonneanne: Right, it was a forward reference. It now points to Section 4.1 and the label is introduced there.

doc/bidirectional_calls.md
yvonneanne: Your reading 1, thanks for spelling out both. It is the contract's view: no response is ever accepted, because none is ever published or because every published one is dropped. Rewritten.

doc/bidirectional_calls.md
yvonneanne: Outstanding is the state between made and accepted, not an ending, so it stays its own paragraph. The paragraph now says that, and that unanswered is unobservable from inside the contract.

doc/bidirectional_calls.md
yvonneanne: Passes the library's checks (C3) and the handler runs. Rewritten under Accepted.

doc/bidirectional_calls.md
yvonneanne: Thanks, reformatted as a definition with the two edge kinds as bullets.

doc/bidirectional_calls.md
yvonneanne: It means the relation has only those two edge kinds. In particular a request being made is not an edge to its execution, and a dropped response adds nothing. Rewritten with those examples.

doc/bidirectional_calls.md
yvonneanne: Yes, intended, and worth saying. All attestations of one execution carry one rid, because the rid is one-to-one with executions (Section 2). G2 now says so itself.

doc/bidirectional_calls.md
yvonneanne: Good catch. G4 now says: included in a final block, whether it succeeded or reverted.

doc/bidirectional_calls.md
yvonneanne: Agreed, moved to the last_seen paragraph in 4.1, with a one-line pointer from 3.2.

doc/bidirectional_calls.md
yvonneanne: Moved next to the last_seen paragraph in 4.1 and linked from 3.2.

doc/bidirectional_calls.md
yvonneanne: Fixed, thanks.

doc/bidirectional_calls.md
yvonneanne: Agreed, it is the MPC's property and not this design's. Section 3.3 now holds only the assumptions about the chains. The threshold, agreement and key derivation properties moved to the start of Section 5, where the sketches use them, with a pointer to protocol_properties.md.

doc/bidirectional_calls.md
yvonneanne: Same treatment: moved to Section 5 with a pointer to ACCOUNT_DERIVATION.md. The sketch only needs that different (chain, contract, key parameters) derive different addresses.

doc/bidirectional_calls.md
yvonneanne: It is about the chains: a block the MPC treats as final stays in the chain, and the MPC reads only from such blocks. Rewritten, and 3.3 is now only about the chains.

doc/bidirectional_calls.md
yvonneanne: Yes, exactly, it describes what happens when the assumption fails. Moved to 4.1 and rephrased: a contract that keeps its key and loses outstanding and last_seen accepts, for a rid made again, a response to a request it already answered.

doc/bidirectional_calls.md
yvonneanne: Not here. C2 records last_seen so that the response can be checked against it later (C3c). At request time there is nothing to compare yet, the execution has not happened. I added a comment in the pseudocode, since the question is a natural one.

doc/bidirectional_calls.md
yvonneanne: Two signing rounds for one request can each produce a signature, for instance when a node restarts and joins a second run, and each gives a different transaction ID, so the MPC has to watch all of them. Added a comment in the pseudocode.

doc/bidirectional_calls.md
yvonneanne: Agreed, and that is what the pseudocode meant: a receipt in a final block at height h' is any height. It was too easy to misread, so it now says this block or any earlier one, and the prose says the lookup by transaction ID works however late a node starts.

doc/bidirectional_calls.md
yvonneanne: The search for the request's execution on the destination chain, the block handler in 4.4. Now named as such.

doc/bidirectional_calls.md
BRBussy: looks good now :)

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,736 @@
...
+    outstanding[rid] = { req.dest, known: last_seen[req.dest],   // C2
+                         req.key.key_version }
+    signet.sign_bidirectional(rid, req)
+    return rid
+
+on response(rid, att = (kind, height, data), sig):
+    if rid not in outstanding:                          // C3a
+        drop
+    e = outstanding[rid]
+    if not verify(sig, H(rid || att),                   // C3b
```
scafuro: In the final implementation we decided that we also sign the length of the outcome (since it can be of variable length, we want to remove the attack vector of an adversary re-playing the same signature with an outcome that can be interpreted in two ways depending on the length). 


So we sign: H(rid || block-height || length-outcome || outcome)


We can change this in other parts of the document.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,736 @@
...
+
+* Threshold (Section 2 of protocol_properties.md). Fewer than t nodes cannot
+  produce a signature or an attestation, honest nodes alone can, and honest
+  nodes eventually publish what they produce.
+* Agreement. Honest nodes admit the same requests and compute the same
+  attestation for a rid, as a function of final destination state and the
+  request's schemas only.
+* Distinct keys (ACCOUNT_DERIVATION.md). The derivation path contains the
+  source chain and the requesting contract, so different (source chain,
+  contract, key parameters) derive different keys, and the sender of an
```
scafuro: by key parameters here we mean "key path" ?

In this paragraph   we are saying: there are two type of derived keys: the keys that sign transactions ( derived from the  paths contained in the request from contract C) and the single key to sign attestation, fixed for contract C.

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,736 @@
...
+
+* *Refused*: the library rejects it inside `sign_bidirectional`, and the
+  caller learns it from the return value.
+* *Accepted*: the library takes a response as the answer to the request,
+  removes the request and runs the contract's response handler (Section
+  3.1). A response the library does not accept is dropped without a trace:
+  there is no rejected ending, and the contract is not told.
+* *Unanswered*: no response is ever accepted, because none is ever
+  published (the transaction never executes, its output does not decode,
+  or the MPC did not admit the request) or because every published one is
```
scafuro: we haven't talked about "admission" anywhere. It's clear what it means, but I was wondering if it would be better to say earlier that there is a concept of admission before signing ?

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,768 @@
...
+  directly, which is how a response is delivered where the signet contract
+  cannot call it, and again after a failed handler.
+
+On Midnight, `SignRequest` is split in two. The application contract keeps
+the request in its own ledger, and the signet contract emits only a
+notification with the rid, the caller's address and where in the caller's
+ledger the request is stored; the MPC reads the request from there
+(Section 7). Midnight's signet contract also cannot see which contract
+called it, so there `authentic` rests on the MPC checking that the caller
+named in the notification is the contract that made the call.
```
yvonneanne: @Pessina I added this paragraph to represent what we've just merged. Speak up if you disagree

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,768 @@
...
+Vocabulary shared across versions carries no suffix: `RequestId`,
+`OutputKind`, `Signature`, the enums (`MPCSignatureAlgorithm`,
+`MPCDestination`, `TxParamType`) and the transaction parameter structs, which
+`TxParamType` tags. Two more carry no suffix: the signet contract's circuits,
+since a deployed singleton never changes, and the notification (below),
+which carries its own version byte.
+
+`OutputKind` is hashed by position, so variants may only be appended.
+Reordering them would make a digest signed under the old order attest a
+different kind.
```
yvonneanne: @BRBussy new content, please revise if necessary

doc/bidirectional_calls.md
```diff
@@ -0,0 +1,768 @@
...
+
+### Notification
+
+The application contract stores its `SignBidirectionalEventV1` in its own
+ledger, and the signet contract emits only a notification per request: a
+version byte, the request id, and a 128-byte payload. In version 1 the
+payload is the caller's address (32 bytes), the depth of the ledger path
+to the caller's request index (1 to 4), that path (4 bytes, only the
+first depth bytes used), and zero padding. The MPC reads the request from
+that index under the request id.
```
yvonneanne: @BRBussy new content, please revise if necessary (if you think it's too much detail in the wrong place we can omit it).

PR #1302, opened 2026-09-15
fix(node): don't attest failure on decode error

yvonneanne:
A terminal output-extraction failure resolved the execution as
`ExecutionOutcome::Failed`, which signs the deadbeef failure attestation.
That attestation is indistinguishable from a real revert, so a vault refunds
on it while the destination transaction actually executed.

The trigger is caller-reachable, e.g., a contract passing one ERC20 schema for a
token the user chooses hits it whenever the token's return shape differs, as
USDT's does.

`ExecutionOutcome::ExtractionFailed` now resolves such a request without
signing anything: the watcher is dropped, the entry removed (deterministic
across nodes, so checkpoints stay aligned) and the sign id retired.
`Failed` keeps its three legitimate producers, revert, replaced sibling and
consumed nonce, where attesting failure is correct. 

Deliberately not covered:

- funds stay committed on the source chain and the dropped request is not
  recoverable; 
- a schema able to say "returns bool or nothing" would keep the ERC20 case
  from reaching the classifier at all

Rollout: while some nodes are un-upgraded they will still propose signing
the failure attestation and it is possible they will or will not reach the threshold.


chain-signatures/node/src/stream/ops.rs
```diff
@@ -259,14 +259,23 @@ pub async fn process_execution_confirmed(
...
             format!(
                 "failed to transition pending tx to final response for sign id {sign_id:?}, tx_id {tx_id:?}, source_chain {source_chain}"
             )
-        })?;
+        })?
+    else {
+        // Retire the id: a leg-1 task on a node that lagged the publish would
+        // outlive its request.
+        ctx.sign_tx
+            .send(SignCommand::Completion(sign_id))
```
yvonneanne, line 274: I used a direct send here instead of  try_enqueue, as on ops.rs:144 and ops.rs:291. Was this the right choice because the other's are in catch up?

ppca: Looked at the code, and I honestly think we shouldn't use try_enqueue for any Completion events. Reason is, if catchup is true, then we'd drop the Completion event, but this event is not stored in backlog as Request events do, so basically we'd always regenerate the signatures altho the request's corresponding respond() event has already been observed...

ppca (approved):

esaminu (approved):

chain-signatures/node/src/stream/ops.rs
yvonneanne: @isSerge ppca suggests Completion should never go through try_enqueue. Wdyt, can you confirm?

isSerge: can't say I am very familiar with the mechanism, I would trust ppca

yvonneanne: Thanks both. Keeping the direct send here. The other Completion paths going through try_enqueue (dropped during catchup, then re-signed because Completion isn't in the backlog) is a separate bug, so I'll handle it in a follow-up rather than widen this PR.

PR #1304, opened 2026-09-15
chore(solana): raise the default checkpoint interval to 1200 slots

ppca:
Currently we checkpoint solana much more often than other chains (most checkpoint every 6-10 mins).

Raises the default Solana checkpoint interval 10x, from 120 slots (~48s) to 1200 slots (~8 min at ~400ms/slot). `CHECKPOINT_INTERVAL_SOLANA` still overrides it; the test-environment value (`5`) is unchanged.

The default is read only by `Backlog::set_processed_block`. Tests that mention 120 pass the interval explicitly or scale with the configured value.


ChaoticTempest:
isn't this a bit too high now? I remember us moving to finalized blocks for solana, which is about 12.8 seconds per slot. Make it about 256mins per checkpoint for every 1200 slots.

ChaoticTempest:
actually, finalized blocks are 400ms a part from each other. The 12.8s is time from initial block to the block being finalized. Subsequent finalized blocks are not 12.8s a part

auto-mausx (approved):

PR #1307, opened 2026-09-16
fix(ethereum): replace `helios` meta-crate with `helios-ethereum` and evict `libp2p 0.51` from graph

isSerge:
Using `helios` meta crate was forcing us to pull `helios-opstack` unconditionally, which drags `libp2p/discv5`. This PR updates imports to only use `helios-ethereum`.

### Changes

- `chain-ethereum/Cargo.toml`: `helios` meta-crate -> `helios-ethereum` direct git dep (same rev, same code), feature renamed to `helios = ["dep:helios-ethereum"]`
- update imports 
- `deny.toml`: remove RUSTSEC-2024-0344


volovyks (approved):
Nice

PR #1309, opened 2026-09-17
fix: malformed solana signature

ChaoticTempest:
a non MPC node can publish a signature to our solana program. We should just ignore if we cannot parse. Will need to hot patch this into testnet.


auto-mausx (approved):

ChaoticTempest:
@volovyks @ppca can you guys hot patch this to testnet when you get the chance? Or we can wait on the next release, since I skipped the block for now

volovyks:
@ChaoticTempest, since it is not a burning issue, we can wait. But the testnet release is not blocked; we can do it anytime.

PR #1311, opened 2026-09-17
feat(midnight): index fallible singleton calls

Pessina:
Index singleton emissions from guaranteed and fallible phases of fully applied Midnight transactions, preserving ledger order and checking transaction hashes. Add captured-transaction and production-reader regressions, and update integration guidance.


ppca (approved):

chain-signatures/chain-midnight/src/emissions.rs
```diff
@@ -47,51 +48,59 @@ pub struct Emission {
...
 fn log_items<P: ProofKind<DefaultDB>>(
     call: &ContractCall<P, DefaultDB>,
+    transcript: &Transcript<DefaultDB>,
 ) -> anyhow::Result<Vec<VersionedLogItem<DefaultDB>>> {
+    let program = Vec::from(&transcript.program);
+    // A leading checkpoint marks a phase boundary and only charges gas in the VM.
+    // The remaining singleton program emits literal logs without reading state or
+    // call context. Reject other programs: empty-state replay is valid only here.
+    let log_program = match program.as_slice() {
+        [Op::Ckpt, rest @ ..] => rest,
```
ppca: this is stricter than before.
Two ledger-valid programs fail this:
  - A single Noop anywhere. ContractCall::well_formed (ledger verify.rs:1885) only rejects adjacent Noop pairs, so one Noop is
    well-formed. It makes the op count odd, so remainder isn't empty, or it shifts the pairing so a pair isn't [Push, Log].
  - More than one Ckpt. split_at consumes only n checkpoints, so a circuit with several leaves an interior Ckpt in the
    guaranteed program. The match arm strips at most one leading Ckpt.

  Why a rejection is worse than before — rpc.rs:394:

  ```
let calls = emissions_in(&tx, singleton).map_err(|err| {
      BlockHold::new(
          "emission-schema-hold",
          height,
          ...
```
  Every error from emissions_in is now a BlockHold. On develop today, one class of error was survivable (rpc.rs:350):

 ``` 
if let Some(unsupported) = err.downcast_ref::<UnsupportedFallibleCall>() {
      tracing::warn!(
          reason = "unsupported-fallible-singleton-call", height, ...
          "midnight transaction skipped: fallible singleton calls are unsupported"
      );
      continue;                       // skip the tx, keep indexing
  }
```

  
And BlockHold doesn't retry — indexer.rs:352-370, process_block_retrying: "Permanent scanner failures escape immediately;
  transport failures retry here." A BlockHold is logged at error level and escapes the retry loop, so the chain stops at that
  height until someone intervenes.

chain-signatures/chain-midnight/src/rpc.rs
```diff
@@ -331,45 +377,32 @@ impl MidnightRpc {
...
+                        err.context(format!(
+                            "candidate extrinsic {} ledger transaction",
+                            found.details.index()
+                        )),
+                    )
+                },
+            )?;
+            // TxApplied covers both logical phases. Partial results cannot identify
+            // committed fallible segments through this node's status events.
+            let calls = emissions_in(&tx, singleton).map_err(|err| {
```
ppca, line 399: UnsupportedFallibleCall will now halt the chain as well. is that intended?

ppca (commented):
just wondering if these two are what we want

chain-signatures/chain-midnight/src/rpc.rs
```diff
@@ -129,6 +129,57 @@ impl subxt::events::StaticEvent for TxApplied {
...
+struct TxPartialSuccess(TxAppliedDetails);
+
+impl subxt::events::StaticEvent for TxPartialSuccess {
+    const PALLET: &'static str = "Midnight";
+    const EVENT: &'static str = "TxPartialSuccess";
+}
+
+/// Accept only one unambiguous full-success status from this extrinsic.
+/// These are RPC-reported events; this check does not verify their inclusion.
+fn fully_applied_hash(
```
isSerge, line 143: Should this also handle `TxDiscarded` for completeness?

chain-signatures/chain-midnight/src/rpc.rs
Pessina: It's because now we support it.

Before, if the event was emitted in a fallible segment (even if it succeeded), we would skip it.

Today we branch on:

* TxApplied -> Can include fallible segments, but all must succeed; then we index.
* TxPartialSuccess -> Can include fallible segments, but at least one failed; then we skip. (I will work on it later because it can include valid events, but it's significantly more work, so we will document it for integrators and skip it for now.)

We halt the indexer if the tx is undecodable. It means there is a drift between the MPC and Chain versions, and we need to intervene.

chain-signatures/chain-midnight/src/emissions.rs
Pessina: Good point. I was too restrictive regarding Noop and Ckpt. It's safe to loosen the indexer and accept Noop/Ckpt.

Fixed on: https://github.com/sig-net/mpc/pull/1311/commits/5a874d9f18104e9ffe293b9928351235ebfff72e

chain-signatures/chain-midnight/src/rpc.rs
Pessina: U are asking because of the README changes, right?

My bad, there is no `TxDiscarded` in the event enum: https://github.com/midnightntwrk/midnight-node/blob/651e043b61ed445bf7a5066c60c87ea7bd606073/pallets/midnight/src/lib.rs#L241-L258

I’ve corrected the README: https://github.com/sig-net/mpc/pull/1311/commits/5a874d9f18104e9ffe293b9928351235ebfff72e

For indexing, we only require `TxApplied`. We log `TxPartialSuccess` for information, but skip it.

isSerge: I was confused why it was in readme, but not handled here, now all clear

isSerge (approved):

ppca (approved):

PR #1312, opened 2026-09-17
refactor(logging): enhance logging and minimize noise

isSerge:
This PR contains few adjustments to minimize log noise from routine operations based on 24h log audit and proposes logging policy. 

### Changes

- Add `doc/logging.md` describing log level policies and some guidelines. I kept it minimal - let me know if we should include anything.

- Downgrade log level for routine operations:
  - `chain-ethereum/event_parsing.rs`: downgrade "No valid Ethereum sign requests found in logs" from `warn` to `info`. It was firing on every log batch with no requests, ~200 entries/day
  - `chain-ethereum/execution_watcher.rs`:  "transaction replaced by sibling tx mined in this block" and "transaction replaced or dropped (nonce consumed by another tx)", ~380 entries/day, both are routine outcomes that watcher handles on it's own.

- Add `utils/throttle.rs` 60s-window rate limiter and apply it to two posit warns ("returning presignature to pool due to REJECTs" and "…proposer timeout"), which repeat per reorganization round. Throttling preserves the signal (occurrence count) at 1 entry/min/key instead of one per event

- Flatten `anyhow` chains in `backlog/consensus.rs` and `stream/supervisor.rs`: `Caused by:` lines ingest as separate log entries, splitting one event across N lines:

**Before** (4 log entries, cause chain is detached from the event):
```
   2026-09-17T02:04:20.734Z WARN mpc_node::stream::supervisor: chain run() failed; restarting chain=ethereum result=Err(
       error sending request for url (https://rpc…/)
       Caused by:
           connection closed before message completed
   )
```

**After** (renders the chain inline, one event - one entry, queries match complete story):

```
   2026-09-17T02:04:20.734Z WARN mpc_node::stream::supervisor: chain run() failed; restarting chain=ethereum error=error sending request for url (https://rpc…/): connection closed before message
 completed
```


chain-signatures/chain-ethereum/src/event_parsing.rs
```diff
@@ -31,7 +31,7 @@ pub fn parse_filtered_logs(logs: Vec<Log>) -> Vec<IndexedSignRequest> {
         }
     }
     if indexed_requests.is_empty() {
-        tracing::warn!("No valid Ethereum sign requests found in logs");
+        tracing::info!("No valid Ethereum sign requests found in logs");
     }
```
yvonneanne: I think the condition should go rather than change level.

parse_filtered_logs is reached for every block, not just ones with contract logs.
client.rs:125 turns a bloom-negative block into CatchupItem::BatchBlock with an empty
logs vec, and nothing between process_range and here filters on emptiness. So any
block without a SignatureRequested log hits line 33 with indexed_requests empty and
prints "No valid Ethereum sign requests found in logs", when there were no requests to
parse in the first place. Do we want to read about that?

When `logs` is non-empty every failure already warned individually on line 30, so the
aggregate adds nothing there either. I would delete lines 33-35. If you want to keep
a signal for the malformed case, gate it on `!logs.is_empty()` and leave it at warn,
which is the case worth reading.

chain-signatures/utils/src/throttle.rs
```diff
@@ -0,0 +1,118 @@
...
+    }
+
+    /// Records an occurrence for `key`. Returns the total occurrence count for
+    /// the elapsed window when the entry should be logged, `None` while
+    /// suppressed. The first occurrence of a key logs immediately.
+    pub fn check(&self, key: &str) -> Option<u64> {
+        self.check_at(key, Instant::now())
+    }
+
+    pub fn check_at(&self, key: &str, now: Instant) -> Option<u64> {
```
yvonneanne: Nits

- `.lock().unwrap()` panics on poison inside a logging helper. `unwrap_or_else(
  PoisonError::into_inner)` costs nothing here since the critical section can't leave
  broken state.
- The key cap on line 39 returns Some(1) for every new key once full, and nothing is
  ever evicted, so saturation permanently turns the throttle off. Two static keys
  can't reach it, so take `&'static str` to keep it structurally true and drop the
  `to_owned()`.

yvonneanne (approved):
Thanks!

doc/logging.md
```diff
@@ -0,0 +1,20 @@
+# Log Level Policy
+
+**error** means a human needs to look at it: the node will not recover on its own, or correctness may
+be affected. **warn** means the node is degraded but expected to self-heal through its normal
+mechanisms (retry, reorganization, resync), escalating to error if it doesn't. Everything else —
+routine operation, expected transitions, diagnostics — is **info** or **debug**. The test for any warn
+or error: if on-call would gain nothing from reading it, it does not belong at that level. 
+
+## Choosing a level
```
volovyks: I would say this section repeats the first one ~50%

doc/logging.md
```diff
@@ -0,0 +1,20 @@
+# Log Level Policy
```
volovyks, line 1: I would extend the rules with this (present as you think is best):
```rust
// good
tracing::info!(%chain, height, "processed block");
tracing::warn!(%chain, error = %format_args!("{err:#}"), "chain run() failed; restarting");

// bad: data baked into the message
tracing::info!("Processing block number {} with hash {:?}", number, hash);

// bad: Debug of an anyhow error
tracing::warn!(?result, %chain, "chain run() failed; restarting");
```

volovyks, line 1: I wanted to have a security section in logs doc, but I realized it would be a duplicate effort with the doc we already have. But a ling to it may be nice.

Somethig like:
```
Secrets are covered by doc/SECRET_MANAGEMENT.md; that includes URLs carrying API keys, which must be logged as host only.
```

volovyks, line 1: We already have logging docs in integration_tests/README.md
What do you think about merging them into a single .md file? (This one).

chain-signatures/chain-ethereum/src/event_parsing.rs
isSerge: I agree, let's delete it, unless someone has a strong opinion

doc/logging.md
isSerge: Let me merge and update `integration_tests/README.md` to have reference

isSerge: added an item in guidelines section with a reference

isSerge: added examples section

doc/logging.md
isSerge: updated

chain-signatures/utils/src/throttle.rs
isSerge: good points, updated

PR #1314, opened 2026-09-17
fix(node): ignore a repeated sign request

yvonneanne:
`Backlog::insert` overwrote any entry with the same id with a fresh Generating
one, and `process_sign_request` enqueued a signing task whether or not the id
was already present. Request ids carry no nonce (`compute_request_id` hashes
sender, payload, path, key version, chain id, algo, dest and params), so a
caller that submits the same request twice emits the same id twice. The second
one drops a signature that was waiting to publish, or resets an entry that is
already executing, whose execution confirmation then finds nothing to complete.

`insert` now returns early without touching the entry when the id is present,
and the caller returns before enqueueing. The `(entry, bool)` signature stays.
The only other caller is the NEAR bridge, which is left alone: that indexer only
runs under `env == "integration-tests"` and forwards each request once.

`test_stream_requeues_replaced_ethereum_recovery_entry_after_catchup` asserted
the old behaviour, that a replay replaces a recovered entry and its index
timestamp. It is renamed and now asserts the recovered entry survives with the
time it was first indexed at; the requeue after catchup is unchanged, since
`requeueable_requests` still picks the entry up. The new test replays a request
whose entry has advanced past Generating, which the existing idempotence test
does not cover: reverting either production hunk makes it fail.


chain-signatures/node/src/backlog/mod.rs
```diff
@@ -189,24 +189,23 @@ impl Backlog {
...
-        let entry = BacklogEntry::new(Arc::clone(&request));
-        let (prev, len) = {
+        let len = {
             let mut pending = self.pending(&chain).write().await;
-            let p = pending.insert(id, entry);
-            (p, pending.len())
+            if pending.get(&id).is_some() {
+                return (SignEntry::generating(request, self), false);
+            }
+            pending.insert(id, BacklogEntry::new(Arc::clone(&request)));
```
ChaoticTempest, line 202: I guess we need the `PendingRequests::entry` api similar to `HashMap::entry` api at some point, but this is fine for now

ChaoticTempest (approved):

chain-signatures/node/src/backlog/mod.rs
yvonneanne: Fully agree, we can add it once there is another call site.

PR #1317, opened 2026-09-18
refactor(logging): adjust logging and observability for reorganize

isSerge:
Currently `reorganizing sign request` warn often dominates the warn stream, this PR implements round-aware log level for reorganize instead of per round warn

### Changes

- `request/state.rs`: `warn` on round 0 and every 10th round, everything in between is `info`
- `protocol/signature.rs`: downgrade advancement-failure to `info` - event always funnels into `reorganize` anyway and `multichain_signature_generator_failures` already counts it
- `metrics/protocols.rs`: added `multichain_sign_reorganizes_total` metric to measure rotation pressure so we can adjust alert if necessary (previously only visible through logs)


yvonneanne (commented):
Good change!

One problem I see: the error -> info downgrade in signature.rs leaves the two recv aborts (inbox closed, timeout) at warn, so a plain timeout now outranks a poke() failure where the protocol itself broke. Moreover, SignError::Aborted is a unit variant, so the reason at task.rs:111 renders as "signature generation failed: Aborted", and the round-0 warn carries no err and no awaited. Giving Aborted a String payload fixes both, and then the downgrade costs nothing. 

Nits:
* bump_round is max(round + 1, highest_seen_round), so a request dragged forward by StaleRound can skip every multiple of 10 and go quiet. round >= last_logged + 10 holds the rate without needing to land on a multiple.
* self.round == 0 || is redundant, 0.is_multiple_of(10) is already true.
* An unlabeled total can't separate 500 requests rotating once from 5 rotating a hundred times. A histogram over the round at reorganize would.

chain-signatures/node/src/protocol/request/state.rs
```diff
@@ -64,12 +69,28 @@ impl SignState {
...
-            reason,
-            "reorganizing sign request"
-        );
+        protocols::SIGN_REORGANIZES.inc();
+        protocols::SIGN_REORGANIZE_ROUND.observe(self.round as f64);
+
+        // Wedged requests rotate forever; a watermark on the last warned round
+        // keeps the warn rate at ~1 per 10 rounds even when StaleRound jumps
+        // `bump_round` past whole decades.
+        if self.round == 0 || self.round >= self.last_warned_round + 10 {
```
volovyks, line 74: Why aren't we reusing the existing mechanics for frequent logs?

chain-signatures/node/src/protocol/request/state.rs
```diff
@@ -64,12 +69,28 @@ impl SignState {
...
+            tracing::warn!(
+                sign_id = ?self.entry.sign_id(),
+                round = self.round,
+                reason,
+                "reorganizing sign request"
+            );
+            self.last_warned_round = self.round;
+        } else {
+            tracing::info!(
```
volovyks, line 83: Looks like we can remove this code duplication by moving this log outside of the `if`.

chain-signatures/node/src/metrics/protocols.rs
```diff
@@ -299,6 +299,27 @@ pub(crate) static SIGNATURE_GENERATOR_FAILURES: LazyLock<Counter> = LazyLock::ne
     .with_label_values(&[] as &[&str])
 });
 
+pub(crate) static SIGN_REORGANIZES: LazyLock<Counter> = LazyLock::new(|| {
```
volovyks: Would it be enough to observe these 2 things in logs? Grafana dashboard is already bloated with data. And we already have POSIT latency.

chain-signatures/node/src/protocol/request/state.rs
isSerge: Where do they exist?

chain-signatures/node/src/protocol/request/state.rs
isSerge: Not sure I understand - log level is different based on the condition

chain-signatures/node/src/metrics/protocols.rs
isSerge: agree, removed

chain-signatures/node/src/protocol/request/state.rs
volovyks: I have not noticed that. Have not noticed that.

chain-signatures/node/src/protocol/request/state.rs
volovyks: throttle.rs
Not sure we want to handle this case differently.

isSerge: These are different cases: throttle is a wall-clock and this one is a watermark (round progression). We can extract it later if there is a need for a second watermark elsewhere

volovyks (approved):

PR #1321, opened 2026-09-18
fix(node): reject the reserved attestation path

yvonneanne:
The attestation key a node signs responses with is derived by the same function as
request keys, with the requesting contract as sender and a fixed path per source
chain (node/src/respond_bidirectional.rs), and nothing rejected a request naming that
path. `validate()` runs only for `SignKind::SignBidirectional`, while plain `sign`
takes a free-text path and an arbitrary 32-byte payload, so a contract asking for its
own chain's attestation path with payload keccak(request_id || output) gets a valid
attestation for any request id and outcome, and forges a response to itself.
Reachable today on Solana and Hydration, the two chains with both an attestation key
and a plain `sign` entrypoint.

`process_sign_request` now rejects a request whose path is the one its source chain's
attestation key is derived under, after the per-kind match so it covers every sign
kind. The network's own attestations are unaffected: they are built internally and
handed to the entry by `SignEntry::advance`, never reaching `process_sign_request`.
The rejection propagates to the supervisor, which logs and continues, so it drops the
request without stalling indexing, the same shape as the existing `validate()` bail.

The test asserts the rejection and that a Solana request carrying another chain's
reserved string is still admitted; that control is the only case in the file that
admits a request on a chain which has a reserved path, so a predicate that rejected
everything on the four bidirectional chains would fail it. Two limits: checkpoint
recovery restores entries without passing admission, so the fix is not retroactive
across a restart, and `is_respond_bidirectional_path` is false for a chain
`respond_bidirectional_path` does not cover, so a chain that gains an attestation key
has to be added there.


ChaoticTempest (approved):
only issue is that prior checkpoints would run into this issue again since this code is only ran on indexing the request

yvonneanne:
> only issue is that prior checkpoints would run into this issue again since this code is only ran on indexing the request

you're absolutely right, classical TOCTOU problem, I added the missing check so it works even in that case, thanks for noticing!

yvonneanne:
> > only issue is that prior checkpoints would run into this issue again since this code is only ran on indexing the request
> 
> you're absolutely right, classical TOCTOU problem, I added the missing check so it works even in that case, thanks for noticing!


Fixing it changed the code considerably, so I'd appreciate another look

chain-signatures/node/src/protocol/request/mod.rs
```diff
@@ -453,6 +454,18 @@ impl SignatureSpawner {
...
 
+                // Every route into signing passes here, checkpoint recovery included (it skips
+                // admission). After the duplicate guard, so a refusal can't retire a tracked request.
+                if claims_attestation_key(entry.request()) {
+                    tracing::error!(
+                        ?sign_id,
+                        chain = %entry.chain(),
+                        "refusing to sign on the reserved attestation path"
+                    );
+                    self.retire_task(sign_id, "reserved attestation path");
```
ChaoticTempest: why do we need `retire_task` here? It should not have created the sign task yet since this is a new sign task entry? Or am I missing something?

yvonneanne: for the mailbox and the dead tag, not the task

Let's do the following instead to avoid any confusion
 
```suggestion
    // Drop posits peers sent ahead of the request, and any that arrive later.
    self.mark_dead(sign_id);
    self.posit_mailboxes.remove(&sign_id);
```

Issue #1324, opened 2026-09-18
leg2 of bidirectional has zero success after sign traffic burst


ppca:
<img width="1051" height="372" alt="Image" src="https://github.com/user-attachments/assets/ca325f7b-f0ae-433c-957c-82c04b687358" />

[grafana ](https://signetwork.grafana.net/d/a8258407-c08f-4796-9d3e-31caacde8653/chain-signatures-prod?from=2026-09-17T18:37:00.000Z&to=2026-09-18T18:59:00.000Z&timezone=utc&var-node_account_id=$__all&var-environment=dev&refresh=5s&viewPanel=panel-172)

I looked into one of the request, and see that the 2nd sign request of the bidirectional was skipped as a duplicate([log)](https://console.cloud.google.com/logs/query;query=SEARCH%2528%22cc346dc97421a577e0f63217f70e8c485cc0a975d03ffdaf80a35b7bb003d36e%22%2529%0A%0AjsonPayload.signId%3D%22SignId%2528%5C%22cc346dc97421a577e0f63217f70e8c485cc0a975d03ffdaf80a35b7bb003d36e%5C%22%2529%22%0Alabels.%22k8s-pod%2Fapp%22%3D%22multichain-9%22;lfeCustomFields=jsonPayload%252Ftarget;cursorTimestamp=2026-09-18T16:52:54.498730684Z;duration=PT3H?project=near-cs-dev). AI confirms this for all the requests after burst.

ppca:
this is what AI says:
Yes — it's systematic, and the before/after comparison is stark.

  dev-11, today (since 09:00 UTC)

<img width="684" height="156" alt="Image" src="https://github.com/user-attachments/assets/63480217-b9cd-4cd5-b2d7-e1351cee9900" />

  Every single one was skipped in the same second as its transition, with the same message: skipping duplicate sign request.

  dev-11, pre-burst control (09-17, 19:00–23:00)

<img width="312" height="155" alt="Image" src="https://github.com/user-attachments/assets/f4092157-da1a-471e-a0fd-09a07410bdf7" />

  So the duplicate skip isn't new — it already happened to about a quarter of requests before the burst — but back then most
  leg 2s still completed. The difference is that leg-1 tasks used to finish and retire themselves, so by the time leg 2
  arrived the sign id was usually free. Under saturation, leg-1 tasks never finish, the id is always still occupied, and the
  skip rate goes to 100%.

[...]

PR #1325, opened 2026-09-18
fix(node): stop a bidirectional second leg from being dropped as a duplicate

ppca:
part of https://github.com/sig-net/mpc/issues/1324
## Summary
A second leg reuses its first leg's sign id, so while that leg's task was still tracked the duplicate guard dropped it and the round trip never finished.

- **Retire the responded leg:** advancing to execution enqueues `SignCommand::LegCompleted`, dropping that task without marking the sign id dead.
- **Kind-aware duplicate guard:** same kind is still a duplicate; a change of kind is the next leg, which replaces the leg it supersedes.
- **Completions name their leg:** the command carries the finished leg's kind, so one arriving late can't retire a second leg that already started.
- **Stop events skip the catchup gate:** nothing replays them, so the completion goes straight to the sign queue, as `AbortChain` does. Overtaking its own request is a no-op: nothing is tracked, and the entry is already `Executing`, which catchup does not replay.

## Impact on dev
Today **248 of 248** second legs were skipped as duplicates, none completed. Pre-burst it was 27% (73 of 270); load takes it to 100%.

Out of scope: proposer-slot starvation, and late task reaping (#1326, #1327).

## Testing
Five tests cover retirement, supersede, and late or untracked completions; each fails without its fix.


chain-signatures/node/src/protocol/request/mod.rs
```diff
@@ -396,9 +431,24 @@ impl SignatureSpawner {
...
+                // A bidirectional request's second leg carries its first leg's
+                // sign id, so a change of kind is the next leg superseding the
+                // tracked one, not a duplicate.
+                if let Some(tracked) = self.requests.get(&sign_id) {
+                    let tracked_kind = tracked.entry.request().request_kind();
+                    let incoming_kind = entry.request().request_kind();
+                    if tracked_kind == incoming_kind {
+                        tracing::info!(?sign_id, "skipping duplicate sign request");
+                        return;
+                    }
```
yvonneanne: The rule is symmetric but the semantics are directional, so a leg-1 request arriving while leg 2 runs replaces the leg that already responded. Reaching it needs a stream restart whose re-scan window covers leg 1's slot, since Backlog::insert resets the entry to Initial holding the leg-1 request and the catchup requeue then feeds it back.

```suggestion
                    let is_next_leg = tracked_kind == RequestKind::SignBidirectional
                        && incoming_kind == RequestKind::RespondBidirectional;
                    if !is_next_leg {
                        tracing::info!(?sign_id, "skipping duplicate sign request");
                        return;
                    }
```

chain-signatures/node/src/protocol/request/mod.rs
```diff
@@ -672,6 +722,301 @@ mod tests {
...
+                sign_id,
+                kind: mpc_primitives::RequestKind::Sign,
+            },
+            &cfg,
+        );
+        assert!(!spawner.test_requests_contains(&sign_id));
+        assert!(!spawner.test_tasks_contains(sign_id));
+        assert!(
+            !spawner.test_dead_ids_contains(&sign_id),
+            "the second leg reuses this sign id, so it must not be marked dead"
```
yvonneanne:
```suggestion
            "the id stays live to buffer posits until leg 2 is admitted, not to admit it"
```

yvonneanne (approved):
Fixes the reported bug and the tests are solid. One suggestion on the supersede direction, and one thing worth writing down about posit routing.

chain-signatures/node/src/protocol/request/mod.rs
ppca: Taken. Added `test_a_replayed_first_leg_does_not_supersede_the_second` for the direction you describe — it fails against the symmetric check (leg 2 loses its task and the tracked kind reverts to SignBidirectional).

chain-signatures/node/src/protocol/request/mod.rs
ppca: Taken verbatim — the posit-routing reason is the one worth keeping.

